PR #79, opened 2016-03-09
Button disabled state on init. Typo in dialog example

gorskygriff:
The situation when the 'disabled' status is not set.
`<button pButton type="submit" [disabled]="!loginForm.valid">Sign In</button>`

And some typo in dialog example


cagataycivici:
Thanks, dialog typo is fixed now and we'll redo button as native component in 0.7 so not merging that part. Thank you.

cagataycivici:
See #81 please.

PR #146, opened 2016-03-25
fix(Editor): rename next() to emit()

ericmartinezr:
This is a small change, but it's a fix since `next()` was deprecated long ago for EventEmitter, instead we should be using `emit()`. See https://github.com/angular/angular/pull/5302.

PS : I found out about this component through twitter and by digging in the code I saw this. I hope this it's not a problem.

PS2 : Sorry if I missed some guidelines, but I didn't see any contribution guide in your repo, I hope also not being doing anything wrong.

Cheers!

### DIFF
--- a/components/editor/editor.ts
+++ b/components/editor/editor.ts
@@ -169,14 +169,14 @@ export class Editor implements AfterViewInit,OnDestroy {
                 htmlValue = null;
             }
             
-            this.onTextChange.next({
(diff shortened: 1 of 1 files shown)


cagataycivici:
Welcome, we use next for all components until now and will migrate to emit in next release. We'll do a find-replace to replace all so not merging the pr now. Thank you for the feedback.

PR #272, opened 2016-04-27
Autocomplete improvements

jonbmul:
* If the user enters a value that corresponds with an object found in the suggestion list, and then hit TAB or Enter, that object should be selected.

* Prevent multiple identical selections to be made when autocomplete is in multiple-mode

* Closing the autocomplete suggestion on the blur event in stead of using a global event listener attached to the body. This makes the suggestions go away when tabbing to the next field

* When revisiting an autocomplete field that already has suggestions, those should be shown when field receives focus again

### DIFF
--- a/components/autocomplete/autocomplete.ts
+++ b/components/autocomplete/autocomplete.ts
@@ -232,24 +232,37 @@ export class AutoComplete implements AfterViewInit,DoCheck,AfterViewChecked,Cont
     }
     
     selectItem(item: any) {
-        let itemIndex = this.domHandler.index(item);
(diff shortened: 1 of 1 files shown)


cagataycivici:
Thank you, I'm not sure some of them like displaying suggestions when revisiting, for other we'll make some changes based on your feedback. Thank you.

jonbmul:
You're welcome. FYI, there are other autocomplete components out there that automatically displays suggestions when receiving focus, for example https://material.angularjs.org/latest/demo/autocomplete

Issue #226, opened 2016-04-19
schedule component.


the-r3dshirt:
Can't get the example to work locally.

I think the problem is within schedule.js the lifecyle hook is called for ngDoCheck on line 159 this.schedule hasn't been set yet. It isn't set until ngAfterViewInit is called which happens after ngDoCheck.

I worked around the problem by inserting a check/set for this.schedule on line 161:
if (!this.schedule) {
    this.schedule = jQuery(this.el.nativeElement.children[0]);
}

PR #286, opened 2016-04-29
Schedule component: Fix uninitialized object

adrienverge:
It seems that in some cases, `ngDoCheck()` is called before `this.schedule` is initialized with `ngAfterViewInit()`.

This commit adds a check to prevent such failures.

Fixes: #226


cagataycivici:
Thank you.

Issue #293, opened 2016-05-02
DataTable - expandableRows with some hidden columns


gatapia:
A data table with some hidden columns does not correctly show expanded content. i.e. expanded content sits in a cell with colspan ignoring hidden columns.

Example:
```
<p-datatable [value]="list" expandableRows="true">
  <p-column expander="true"></p-column>
  <p-column header="shown col"></p-column>
  <p-column [hidden]="true" header="hidden col"></p-column>
</p-datatable>
```
Expanding rows in this table will look strange, this is because the expaned row content has wrong colspan (3 instead of 2).

cagataycivici:
Will check for beta.6.

gatapia:
Still happening for scrollable tables, fix here:
https://github.com/primefaces/primeng/pull/383

PR #327, opened 2016-05-07
fix for bug 293

gatapia:
Simple fix for this issue: https://github.com/primefaces/primeng/issues/293

"DataTable - expandableRows with some hidden columns"


cagataycivici:
Thanks

PR #331, opened 2016-05-11
Update schedule.ts

nickraphael:
Hi guys,

I've found an error in your schedule.ts code.

eventDrop: (event, delta, revertFunc, jsEvent, ui, view) => {
this.onEventDragStop.emit({
'event': event,
'delta': delta,
'revertFunc': revertFunc,
'jsEvent': jsEvent,
'view': view
});
},

This method incorrectly emits onEventDragStop. It should emit onEventDrop.


cagataycivici:
Thank you.

PR #339, opened 2016-05-11
Add TagInput

FallenRiteMonk:
A input field for tags like in stackoverflow.

This is based on a module I found https://www.npmjs.com/package/angular2-tag-input, but has some minor changes/improvements.


cagataycivici:
Thank you, due to company policy we prefer to implement new components ourselves based on our development process so unable to accept the PR, however this feature is coming very soon. Here is the ticket;

https://github.com/primefaces/primeng/issues/1149

PR #342, opened 2016-05-11
Fix overlay panel out of viewport if targeted at element on right side of viewport

mamohr:
Overlay panel will now open to left side if there is not enough room on right side


Mrtcndkn:
This commit is not related to this request sorry for complication.

cagataycivici:
We've also fixed this, thanks for the PR.

PR #344, opened 2016-05-11
Add (imageClicked) event. Outputs {image: image, imageIndex: i, origi…

brtnshrdr:
I thought it would be useful to have a click handler on the image, so I created one. 
It outputs the image object clicked, the index from the images array, as well as the original click event


cagataycivici:
Thank you.

PR #367, opened 2016-05-16
modal should be boolean

sdmoralesma:
* In component's documentation the type of the modal attribute is boolean, then, should be changed to "true" to be consistent in examples.
* The description of the attribute "modal" is not correct


cagataycivici:
modal="false" will still be modal, correct way is [modal]="false", as this is confusing we keep booleans as modal="modal" like html5 boolean attributes e.g disabled="disabled". I've corrected the doc.

Issue #364, opened 2016-05-15
custom list of objects to dropdown 


lilling:
dropdown gets only array of objects that has two properties:
`{label: 'Audi', value: 'Audi'}` as strings. otherwise either you get exception `Cannot read property 'label' of undefined` if has'nt got label property, or the ngmodel wont change if the value property is an object

It would be great if the developer will be able to fill the dropdown with custom objects and a property to show to the user. something like value in data table and field in column.

brtnshrdr:
https://github.com/primefaces/primeng/pull/370 should give you that feature if and when it's merged :)

lilling:
@brtnshrdr thx!

do you know if they're going to do this feature?

brtnshrdr:
No idea! But I made a pull request so we can only wait and see!

monoflash:
The same problem:

Template:
```
<p-dropdown [options]="item.rates">
            <template let-element>
              <li style="position: relative">
                <div>111111</div>
              </li>
            </template>
</p-dropdown>
```

Error:
![Error](http://p.jiq.ru/on42+)

cagataycivici:
Fixed for beta10.

victorcarvalhosp:
Can you give some working example? I updated to beta.11 and still got this error...

christoferw:
"fieldLabel" is not in the current master branch available, as far as i could see there have been a merge conflict which moved that fix out again?
@cagataycivici am i right?

PR #370, opened 2016-05-16
#364 - Ability to use custom objects in dropdown.

brtnshrdr:
Uses `<p-dropdown [options]="cities" field="city" fieldLabel="cityLabel"></p-dropdown>` syntax

(If object was as so)
```
{
  city: 'New York',
  cityLabel: 'New York, New York'
}
```


lilling:
@cagataycivici. When its going to be reviewed?

cagataycivici:
I couldn't get this working at all, I think we have another problem. [attr.data-value] does not work when we try to get the selected item from the li element. I'm thinking about alternatives.

cagataycivici:
Trying to find a way without adding new options as listbox works with objects.

PR #379, opened 2016-05-19
local publishing + datatable changedetection fix

joe380:


cagataycivici:
Will merge the necessary part only in datatable.ts

PR #389, opened 2016-05-23
DataTable: row custom style or css class

ijsgaus:


cagataycivici:
Thank you, this seems a bit more complicated than I expected so not merging it. We'll try to find a simpler solution and if not reopen. Please create a feature request at our tracker with your requirements.

PR #394, opened 2016-05-23
Added viewRender-event

PhilippRoessner:
- Added viewRender-event from Fullcalendar to schedule-component
- Added viewRender-documentation to showcase


cagataycivici:
Shouldn't you emit viewRender event instead of eventResize? Why do you emit this.onEventResize? Is it a typo?

PhilippRoessner:
sorry, yes, it's a typo :/ you're right

PR #421, opened 2016-05-26
DataTable: Add ability to specify class to use on the table tag, and the ability to specify the class to use if the table is responsive.

shahmirn:
Add ability to specify class to use on the table tag, and the ability to specify the class to use if the table is responsive.


cagataycivici:
Why not just [class]="tableStyleClass" ?

shahmirn:
@cagataycivici Changed code to use tableStyleClass and added documentation.

cagataycivici:
Thank you, we'll provide tableStyleClass but rather keep the responsive class as fixed for now.

cagataycivici:
See #1155

PR #449, opened 2016-06-02
PanelMenuSub child item url link fix

daotoad:
Child items of a PanelMenuSub were linking to their parent's url attribute.  Now link to child url attribute.


cagataycivici:
Fixes #525

PR #458, opened 2016-06-07
OnSelectioChange event

Toktik:


cagataycivici:
I think we should just pass delta, oldDelta and source to the event without processing them.

PR #471, opened 2016-06-09
Add firstDay as option to schedule

MathijsHoogland:
So you can define a different first day in the calendar (1 for monday, 2 for tuesday etc)


cagataycivici:
firstDay belongs to the locale config so better avoid duplicating it with a separate attribute. You can customize firstDay by binding an object that defines firstDay.

PR #493, opened 2016-06-16
allowHover for tooltip

PhilippRoessner:
I needed to keep the tooltip open when the mouse is over the tooltip


cagataycivici:
There are too many character changes, it is hard to find what has changed for review, please update the PR and we'll reopen.

PR #502, opened 2016-06-21
Datatable: Better multisort

feltech:
This change allows multisort to work on multiple columns without having to hold down the CTRL key (undocumented).  This is necessary for mobile, and is more standard with how other tables work.

The `sortOrder` for a column cycles between -1, 0 and 1 as the column is clicked.  0 is handled as a special case, where that column is removed from the `multiSortMeta` list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ typings
 
 # misc 
 *.log
+
+# ide
+.idea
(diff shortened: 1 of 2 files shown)


cagataycivici:
We'll wait for user feedback, last time we've removed the metaKey, many users demanded it back.

PR #503, opened 2016-06-21
Datatable: Expand rows by ID

feltech:
This follows on from PR #502.  Can be cherry picked if #502 is not accepted.

A new optional `@Input() idField` is added, which makes `expandedRows` use a field of the row data, rather than the entire row object itself, as the tracking key.  This means when data is updated dynamically, selected rows will still be expanded, assuming a row(s) with matching `idField` exists after the update.  

If `idField` is not used (i.e. the current behaviour) then the memory pointers to row objects change after an update, even if the row data itself is identical, so the check for equality in `expandedRows` fail and the previously expanded row is collapsed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ typings
 
 # misc 
 *.log
+
+# ide
+.idea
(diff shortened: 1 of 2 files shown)


cagataycivici:
Not sure if the value added justifies the amount of changes, I don't see a problem of using expandedRows.

PR #543, opened 2016-07-01
Handle "null value options" and resize on options change

zpul:


cagataycivici:
I think this is obselete now after Dropdown reimpl.

PR #550, opened 2016-07-04
Different notation for styles

ndsmyter:
The notation to configure the styles is different from the one that was mentioned in the documentation. On the website, the correct notation is used in the "source" tab, but the wrong is used in the "documentation" tab.


cagataycivici:
Resolve, thanks.

PR #563, opened 2016-07-05
 Adding management of the first line number

tellurS:
I would like to have @input first.
To use a component in applications such as routing parameter and etc


cagataycivici:
Better way would be using a setter  I think to listen changes on first. I'll do this for beta 12.

PR #577, opened 2016-07-07
To control for the expanded rows

tellurS:
The application needed to manage open records, such as through the URL-routing


cagataycivici:
Merged manually thank you.

PR #579, opened 2016-07-07
External control filters

tellurS:
In the application it is necessary to control the filter value, for example through the URL-routing


cagataycivici:
Merged manually

PR #603, opened 2016-07-14
Added a "Limited Range" option for p-slider

ryanwawr:
An application I've been working on uses a group of related sliders to assign a fixed quantity of something among a number of sliders. In this scenario, the maximum value for each slider should be the total quantity, since any slider could have the entire quantity assigned to it, but the range of each slider should be capped dynamically such that it cannot be increased past the point where the sum of all slider values is equal to the total quantity. (The payment slider at https://www.humblebundle.com is similar to what I'm trying to develop, though theirs readjusts the other sliders automatically).

To allow for this, I added "limitedMin" and "limitedMax" attributes as inputs to p-slider and changed the logic in the slide function to prevent the value of the slider from exceeding these values in either direction. The effective range of the slider can be adjusted dynamically simply by changing these inputs. I added an example with this functionality to the demo page (link to a screen recording below), but did not go so far as to change any of the documentation.

http://screencast.com/t/BinHh0lavO1

### DIFF
--- a/components/slider/slider.ts
+++ b/components/slider/slider.ts
@@ -22,6 +22,10 @@ export class Slider implements AfterViewInit,OnDestroy,OnChanges,ControlValueAcc
     @Input() min: number;
 
     @Input() max: number;
+	
+	@Input() limitedMin: number;
+	
+	@Input() limitedMax: number;
 
     @Input() orientation: string;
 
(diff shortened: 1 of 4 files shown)


components/slider/slider.ts
```diff
@@ -66,6 +70,15 @@ export class Slider implements AfterViewInit,OnDestroy,OnChanges,ControlValueAcc
                     this.onChange.emit({originalEvent: event, values: ui.values});
                 }
                 else {
+					if(ui.value > this.limitedMax){
```
ryanwawr: If value of slide exceeds limitedMin or limitedMax, set ui value and slider handle value to limitedMin/Max and prevent slide from occurring past these points

components/slider/slider.ts
```diff
@@ -66,6 +70,15 @@ export class Slider implements AfterViewInit,OnDestroy,OnChanges,ControlValueAcc
                     this.onChange.emit({originalEvent: event, values: ui.values});
                 }
                 else {
```
ryanwawr: If value of slide exceeds limitedMin or limitedMax, set ui value and slider handle value to limitedMin/Max and prevent slide from occurring past these points

PR #604, opened 2016-07-14
Hide calendar on select

ryanwawr:
I added a "hideOnSelect" input to p-calendar which, when set to true, will cause the datepicker to close when the user hits the Enter key in the associated input. 

This is the default behavior of the jquery-ui datepicker, but does not currently happen in the primeng calendar input because jquery skips the hide command if a custom onSelect callback has been defined in the calling code.


cagataycivici:
So this is only to hide it when enter key is pressed?

ryanwawr:
Yes, just an option to hide it when the enter key is hit. That seems to be
the default functionality of the jQuery datepicker, but the line that
normally hides it is skipped, since the onSelect callback is implemented in
calendar.ts.

PR #605, opened 2016-07-14
Fix: A couple fixes for sorting of data

shahmirn:
Fix: Sort data when the data in the grid changes, ignoring stopSortPropagation

Fix: Use custom sort function if one exists for a column, and the sortField specified on the table matches that column


cagataycivici:
I could not understand what the PR fixes, stopSortPropagation should not be ignored as it causes UI flicker and may constantly sort the even if you move mouse. That was added to fix it. Please create separate issues first each explaining the cause with a reproducible case and we'll review.

PR #611, opened 2016-07-15
Fix: Implement ondestroy and add unsubscribe

shahmirn:


cagataycivici:
Had conflicts so merging manually

PR #615, opened 2016-07-16
Checkbox: Fix for boolean model using 'name' attribute

cghislai:
New angular forms require a name attribute to be set when using template-driven forms components inside a `<form>` element:

    If ngModel is used within a form tag, either the name attribute must be set
        or the form control must be defined as 'standalone' in ngModelOptions.

                      Example 1: <input [(ngModel)]="person.firstName" name="first">
                      Example 2: <input [(ngModel)]="person.firstName" [ngModelOptions]="{standalone: true}">


cagataycivici:
We've decided to add binary attribute instead at #636

PR #616, opened 2016-07-16
Checkbox: Add clickable labels

cghislai:
Add a label property to display clickable label next to the checkbox


cagataycivici:
There are conflicts so merging manually with slight changes. Thank you. Please see #649 649

Issue #587, opened 2016-07-11
InputMask with ngModel not updating first character


andrevcf:
`<p-inputMask mask="(99) 9999[9]-9999" [unmask]="true" name="cel" [(ngModel)]="cel" required></p-inputMask>`

With the example above,  when i type 2 digits the model is updated with 1 digit only.
For example: I type "4", the model is blank. When I type "34" the model is now 3. When I type "342" then the model is 34. 
The backspace is working fine and updating correctly

KonbOgonb:
Current implementation onKeyDown event as time moment when to fetch masked or unmasked value. This approach is wrong because when onKeyDown is fired value does not contain newly added character. So if we perform validation on the fly with each change of text, we are going to receive text without last character. Due to it we have this bug, when first character is not validated when it is entered. Possible solution could be to used key up event of <input> to flash text to value.

ryanwawr:
Having the same problem. Would definitely like to see this fixed soon, since it's a fairly significant issue when trying to use p-inputMask with a search element.

PR #638, opened 2016-07-21
Fix for issue #587 - Input mask with ngModel lags by one character

ryanwawr:
This is related to issue #587, where a value bound to p-inputMask with ngModel was lagging behind the input by one keypress, i.e. when the input field displayed "345", the bound value was "34".

I implemented the onBeforeWrite function in inputmask.ts in order to have access to the current value of the input, which was not immediately available in onKeyDown. I left the implementation of onKeyDown as is and added a check to make sure event.target != null in onBeforeWrite in order to maintain correct behavior during paste and undo actions. 

Tested all this by adding inputs to the inputmask demo page in my local copy of the repo and verifying that the value bound by ngModel stayed synchronized with the value displayed in the inputmask field.

### DIFF
--- a/components/calendar/calendar.ts
+++ b/components/calendar/calendar.ts
@@ -107,8 +107,6 @@ export class Calendar implements AfterViewInit,OnChanges,OnDestroy,ControlValueA
     @Input() locale: any;
     
     @Input() icon: string = 'fa-calendar';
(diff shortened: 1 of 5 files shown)


cagataycivici:
Thank you.

PR #646, opened 2016-07-22
fix(dropdown): ngModel set before options

MarkPerryBV:
When the ngModel is set before the [options] binding is populated then
when the options do arrive (possibly from an Observable) then the
selectedValue is not highlighted in the ui.


digitalcraftco:
Awe thanks! I was battling this for a few days. Couldn't get the label and UI Highlight working between different views with ngModel binding.

cagataycivici:
Merged with some changes.

Issue #284, opened 2016-04-29
DataTables multiple selection without CTRL key


RicardoVaranda:
Is there currently a way to enable multiple selection on a data table without the use of a meta key or ctrl key?

If not can someone please point out where the prototype for the row selection is?

cagataycivici:
No, as it is based on file system behavior where you need a key to multi select.

RicardoVaranda:
I got that part but where would I have to go to change  this behavior? I've been looking at the node_modules files for primeng and can't seem to find anything related to selections.

RicardoVaranda:
I've created a new pull request to address this issue, please take a look at it here:

[https://github.com/primefaces/primeng/pull/287/commits](url)

cagataycivici:
Thank you for the pr after discussion, we've decided to wait for more user requests of this same requirement to proceed.

rahuja910:
Will this feature be out with Beta-9?

PR #647, opened 2016-07-22
Updated documentation to match change from #284.

mrmateo:
The commit for #284 changed the functionality of unselecting a row in single mode and dropped the requirement for having the metakey selected during the click. The documentation still referenced needing the meta key pressed. This PR updates the documentation.

If the intention was to still have the metakey selected to unselect a row in single selection mode (The original issue only made reference to multi selection mode) let me know and I can create a ticket and PR with the fix. Otherwise, this update to the docs should be good.

Thanks!


cagataycivici:
Thank you.

PR #686, opened 2016-08-02
PickList: Added Event Emitters to emit the array of selected items either transferred left or right so they can be accessed later on

ZeroCoolX:
Two Event Emitters have been added so that someone implementing picklist can access the element(s) that are actually being transferred.


jyoti-nayyar:
Hi , I have used picklist in my project and used npm to install the primeNG package. Please let me know when will the npm repository be updated with the above events as I need to use them.

cagataycivici:
This is fixed in a similar way today. Thank you.

PR #690, opened 2016-08-03
Fixed 2-way binding ngModel

lthh89vt:
Editor Content is not being updated right after ngModel changed.
So removed selfChange variable

Explanation (Hope I explain it well enough):

First time
-------------------
ngModel is updated -> writeValue -> Editor pick up on-text change and set selfChange

2nd time
------------------------
ngModel is updated 
-> Because self change = true -> set self change = false and STOP update editor content.

3rd will be same as first as selfChange is now false unless type in anything then it' same as 2nd time.

Cheers,

### DIFF
--- a/components/editor/editor.ts
+++ b/components/editor/editor.ts
@@ -86,8 +86,6 @@ export class Editor implements AfterViewInit,ControlValueAccessor {
     
     onModelTouched: Function = () => {};
     
-    selfChange: boolean;
-
     quill: any;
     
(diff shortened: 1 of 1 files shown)


lthh89vt:
Not sure what that selfChange is for but it seems to come from the old code where you have the loop in writeValue()

cagataycivici:
It was to avoid writing the model value back to model as setText or pasteHTML calls text-change, but removing it should not cost much. Will merge manually, thank you.

PR #699, opened 2016-08-04
Add Base64 Encoded Image Support

Kaffiend:
Setting of type property of "encoded" allows for showing of base64 encoded images using the same object properties as type "image" adding the requirement of setting the format property to the image type for decoding target ie. format = "png"


Kaffiend:
Closed, found issues on image next prev

PR #703, opened 2016-08-06
Fixes Carousel Page Links. Resolves #554.

### LINKED ISSUE #554: Carousel page links redirect to #
In showcase, carousel page links are not working. Currently they redirect to http://www.primefaces.org/primeng/# Instead of changing carousel page. I tested showcase http://www.primefaces.org/primeng/#/carousel with Firefox and Chrome and both browsers suffered from the same issue.

mko:
Fixes issue #554.


cagataycivici:
I'll keep # and do prevent default instead, thanks for the tip.

juanlet:
This is not working, the next and prev do nothing

PR #753, opened 2016-08-18
Fixing default selection on datatables.

jeffrysteven:
Fixing default selection on datatables based on [(selection)] attr.


cagataycivici:
Thanks but I'm not sure about string comparison, we have some object equality check method that I'll try.

PR #795, opened 2016-08-26
Fixed InputMask placeholder

kryzys69:


cagataycivici:
slotChart is for mask placeholder and placeholder attribute itself is used as the html5 input placeholder.

PR #804, opened 2016-08-28
fixes #802 unmasking

### LINKED ISSUE #802: InputMask - Unmask is not working
When I'm working with a currency-InputMask and changing the radix-point to ",", then unmask is not working:

`<p-inputMask alias="currency" [unmask]="true" [options]="{prefix:'',radixPoint:',',autoGroup:false}" styleClass="form-control" [(ngModel)]="tarif.preis"></p-inputMask>`

Maybe it's related to jquery.inputmask?

PhilippRoessner:
decimal and currency with changed radixPoint is now unmasked correctly


cagataycivici:
Thank you.

PR #832, opened 2016-09-01
handle cyclic objects in domhandler.equals

dsoltesz:


cagataycivici:
Assigned to beta.17 so we won't forget it this time.

PR #867, opened 2016-09-06
Open fileupload request before onBeforeUpload for header modification

fensminger:
Hello,

I try your fileupload component with the beta14 and I can not add a header (authentication token) to the upload command.

After this modification, the upload works.

Could you include this proposition in your next release ?

Thanks,
Frederic.

### DIFF
--- a/components/fileupload/fileupload.ts
+++ b/components/fileupload/fileupload.ts
@@ -193,12 +193,13 @@ export class FileUpload implements OnInit,AfterContentInit {
             }
         };
         
+        xhr.open('POST', this.url, true);
+
         this.onBeforeUpload.emit({
             'xhr': xhr,
             'formData': formData 
         });
         
-        xhr.open('POST', this.url, true);
         xhr.send(formData);
     }
 


cagataycivici:
Done, thank you.

predatorkill:
Can you please show the code in order to add an authentication token into the header while uploading? 

Thanks!

kuba-010:
Hi, not sure if I should submit this as a seperate bug. The documentation refers to OnBeforeUpload to set headers for example. However this seems to fail with the xhr must be open error. I've had a peak at the code and it looks like instead of OnBeforeUpload the emit goes to OnBeforeSend - if I then set the headers there as below it works OK. OnBeforeSend  isn't mentioned anywhere in the docs so seems like a mix up somewhere?

This works:
```
// (onBeforeSend)="onBeforeSend($event)"
onBeforeSend(event) {
        event.xhr.setRequestHeader("Authorization", this.token);
}
```

This doesn't:
```
// (onBeforeUpload)="onBeforeUpload($event)"
onBeforeUpload(event) {
        event.xhr.setRequestHeader("Authorization", this.token);
}
```

PR #887, opened 2016-09-12
add possibility to use dropdown as filter in column

nadirowski:
There are some cases when it would be nice to use a dropdown as a filter instead of textbox.
Here is my proposition for this


michalig:
Looks great to me. Looking forward to see it in the next release :)

lbamburski:
I would be nice to have an option to attach some service when there are many records. Also an option for multiselection.

cagataycivici:
We'll do customizable filter elements so you'll add whatever component instead of predefined elements.

PR #896, opened 2016-09-14
Make the severity flexible

s3kob:
With this changes you can set your own severity by adding CSS styles like this:
.ui-growl-message-whatever .ui-growl-item {
    background: green none repeat scroll 0 0;
}

The displayed icon should work the same way.


cagataycivici:
There is a conflict, please resolve the conflicts.

Issue #849, opened 2016-09-02
Enhancement request: Expose panelVisible field and show method in p-dropdown


ryanwawr:
It would be nice if the panelVisible property in the Dropdown class was public (or at least had a public getter) so that I could easily determine the state of the dropdown from code. 

Would also be nice if the show() method, which is already public, did not require two private fields as input. There really doesn't seem to be any reason for it to require any input arguments at all, since panel and container are already accessible from within the method anyway. Allowing show to be called without arguments would make it much easier for users to customize the shortcut for opening the dropdown, which would be a nice alternative to only allowing Alt+DownArrow.

ryanwawr:
Looks like this isn't an issue anymore since these fields were recently made public.

mmubasher:
@ryanwawr on upgrading from primeng 6 to version 8, panelVisible has been removed. Is there any other feature added for this?

PR #911, opened 2016-09-16
Fix issue #849: Added dropdownOpen property and refactored show() to be called with no parameters

ryanwawr:
As described in issue #849, it would be nice to be able to determine whether the dropdown panel is open or not and to be able to hide/show it from a component implementing p-dropdown. Among other things, this would make it much easier for users to customize hotkeys for showing the dropdown panel.

I added a public property called dropdownOpen that returns the value of panelVisible, since, ideally, a user should be able to read this field but should only be setting it by means of the hide/show functions. 

The show function was already public, but was invoked with two private fields, essentially rendering it private. I refactored the method to be called with no parameters, since the parameters being passed in are just fields on the dropdown component that can be referenced from within show without being explicitly passed in.

### DIFF
--- a/components/dropdown/dropdown.ts
+++ b/components/dropdown/dropdown.ts
@@ -97,8 +97,12 @@ export class Dropdown implements OnInit,AfterViewInit,AfterViewChecked,DoCheck,O
     focus: boolean;
     
     differ: any;
+	
+	get dropdownOpen(): boolean {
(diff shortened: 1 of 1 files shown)


ryanwawr:
Looks like this isn't necessary anymore, since the fields I exposed were recently switched from protected/private to public.

Issue #868, opened 2016-09-07
Issue on Datatable with resize and reorder at same time


fedemarengo:
Datatable with resize and reorder enabled at the same time is not working properly.
If i have enabled both and i try to resize the column to the right the table takes the reorder event as well, so the grid is trying to drag the column (change order) and resize the column. Is working fine when i resize the column to the left.

cagataycivici:
This should be fixed in recent PrimeNG release, please try with them and if issue persists drop a comment and we'll reopen.

PR #920, opened 2016-09-19
Fix: Issue on Datatable with resize and reorder at same time #868

lbamburski:


cagataycivici:
The issue I see is reorder does not work but resize works, the patch had no effect when I applied it. Any ideas?

lbamburski:
It works for me. It doesn't work on IE because of the line:

event.dataTransfer.setData('a', 'b'); // Firefox requires this to make
dragging possible

should be:

event.dataTransfer.setData('text', 'b'); // Firefox requires this to make
dragging possible

as someone noticed.

Best regards


Pozdrawiam
Łukasz Bamburski

2016-11-01 9:17 GMT+01:00 Çağatay Çivici <notifications@github.com>:

PR #922, opened 2016-09-19
MultiSelect can show selected label as tokens instead of comma-separa…

### LINKED ISSUE #907: Replace plain string label in MultiSelect with <ul> list like in autocomplete widget
Is it possible to replace plain comma-separated label with "ul>li list the same way  autocomplete component does? The idea is that I need to have onClick events on individual selected items.

Dema:
…ted list. Fixes #907 
Not sure about "close" button on each token. Currently commented it out.


cagataycivici:
We've decided to keep it as it is and will reconsider if more users demand it.

mohan1304:
Hi,
The width of multiselect drop down is gradually increasing if we select multiple items but I don't want this to happen because of this feature my layout is looking odd. Could you please let me know how can I limit the width of multi select

PR #932, opened 2016-09-20
Please add this to toggle icon display

Syrbox:
I need to show the toggle icon only when I have data. This should show it only if expander="true"


Syrbox:
I made the change on the web, haven't tested on local

Issue #917, opened 2016-09-19
DataTable with scroll and few data causes headers and table cells to misalign


sigad:
In a datatable with scroll, if the height of the data is smaller than the specified scroll height, the headers and the table cells misalign (there is some gap for the scroll bar in the header, even if no scrollbar is displayed).

Here is the definition of the datable: 

```
<p-dataTable [value]="data"
                         selectionMode="multiple"
                         [(selection)]="selectedItems"
                         [totalRecords]="totalElements"
                         [rows]="size"
                         [lazy]="true"
                         (onLazyLoad)="loadLazy($event)"
                         [responsive]="true"
                         [resizableColumns]="false"
                         scrollable="true" scrollHeight="400px">
```

I am using 1.0.0-beta.15.

billdwhite:
Probably a duplicate of [487](https://github.com/primefaces/primeng/issues/487)

GopiAnnan:
@cagataycivici ,

I have tried with 2.0.2 , When Scrollable is true and it has not many records to scroll. still have the issue. 
<div class="ui-datatable-scrollable-header-box"  style="margin-right: 17px;">
This creating issue.

Header has inline style, margin-right: 17px; . this has to be 0 px when there is no Scrollable content.  

Please let me know if I am missing something.

PrajwalGonsalves07:
https://github.com/primefaces/primeng/issues/487#issuecomment-277251152

Check the above link. There is a work-around for this.

Issue #859, opened 2016-09-05
Datatable scrollable="true" breaks column template header


mcfin:
Hi PrimeNG Team,

I've found that when setting `scrolling="true"`, the datatable does not properly apply the template using `pTemplate type="header"`. 
The template using `pTemplate type="body"` does work though.

erlopezh:
Same problem here.

kethan:
Please how to resolve this... i am using it in production...

idanmalka:
am i missing something? i cant find a way to fix the header allignment too

ctrl-brk:
There's kind of hack to fix it. See this post:
http://forum.primefaces.org/viewtopic.php?f=35&t=48781

Issue #487, opened 2016-06-15
DataTable with explicit widths causes headers and table cells to misalign


gatapia:
The following table:

```
<p-dataTable [value]="entities" 
      [lazy]="true" [totalRecords]="totrows" (onLazyLoad)="refresh($event, table)"
      selectionMode="single" (onRowSelect)="edit($event)" 
      sortField="ContactName" [sortOrder]="1"
      [globalFilter]="globalfilter"
      scrollable="true" [scrollHeight]="winheight" #table>    
    <p-column [style]="{'width':'300px'}" field="Company" header="Company" [sortable]="true" [filter]="true"></p-column>
    <p-column field="ContactName" header="Name" [sortable]="true" [filter]="true"></p-column>
    <p-column [style]="{'width':'250px'}" field="Email" header="Email" [sortable]="true" [filter]="true"></p-column>    
    <p-column [style]="{'width':'80px'}" field="IsKeyContact" header="Key Contact" [sortable]="true" [filter]="true">
      <template let-entity="rowData">{{entity.IsKeyContact ? "Yes" : "No"}}</template>
    </p-column>
    <p-column [style]="{'width':'80px'}" field="DateNextFollowUp" header="Follow Up" [sortable]="true" [filter]="true"></p-column>
  </p-dataTable>  
```

(And any other table that uses explicit widths) causes the following layout issues:

![image](https://cloud.githubusercontent.com/assets/135965/16096412/308ec0ec-338c-11e6-81e5-3c9145384d7b.png)

I am using the Modena theme, so not sure if it could be this.  There is no CSS interfearing with the table (I assume given encapsulation and my brief inspection appears to support this).

cagataycivici:
Can you also add width to ContactName column as well.

gatapia:
Setting the width on ContactName (which I dont want to do as I want it to take remaining space) yields slightly less misaligned results but still not great:

![image](https://cloud.githubusercontent.com/assets/135965/16180671/ac8b6164-36cf-11e6-9c57-bd3d7c4b15b2.png)

gatapia:
Just to clarify further, this only happens with scrollable=true tables as the header is a separate element from the body.

billdwhite:
+1 on this; same problem here

sylvainpey:
I have the same problem with relative width (in %). I precise that it appears only if the scrollbar is not displayed (not enought lines). Actually, the scrollbar width is always "reserved" on the right of the header but not in the lines part.

[...]

Issue #465, opened 2016-06-08
dataTable when scrollable and scrollheight is set - style issue


Floppynator:
a dataTable style is broken when scrollable and scrollHeight attribute is set.

rahuja910:
any update on this?

sittingbool:
Same here. IfI set a percentage value into the scrollHeight it sets the max-height of the body to 0px

sittingbool:
The error happens Here: https://github.com/primefaces/primeng/blob/master/components/datatable/datatable.ts#L1620.

My Setup: I have the table inside a tab-panel. The first tab has a selection which changes the list shown in the table, the second has the table. After I change tabs, select and come back to the table this line will set 0px and never be called again.

cagataycivici:
Unable to replicate with newer versions, please try with 2.0.2.

BartoszBaczek:
This issue still affects me - both on 2.0.2 and 2.0.3

cagataycivici:
Can you provide a plunkr please?

BartoszBaczek:
Yup, I'm working on it right now.

cagataycivici:
Thanks, we'll do 4.0.RC1 on wednesday, I can ask my colleague to check this until then.

BartoszBaczek:
There is everything fine with with scrollWidth. The scrollHeight is the one, that seems to have problems:

http://plnkr.co/edit/lwsjtNJCWW5qEBtOOXiY?p=preview

ericrovtar:
Was there ever a fix for this? I'm still finding the issue in the latest npm package.

PR #936, opened 2016-09-21
Scrollable works with resizableColumns, reorderableColumns,

lbamburski:
sortable, columnResizeMode, scrollHeight, scrollWidth, header,footer, filter, paginator, selectionMode,
  expandableRows with inner table and column toggler.

  Fixed footer width and partially pagination width when scrollWidth is used

Fixes issues with scrollable:
-DataTable with scroll and few data causes headers and table cells to misalign #917
-dataTable when scrollable and scrollheight is set - style issue #465
-Column Reorder and Resize support for Scrollable DataTable #238
-Maybe: DataTable with explicit widths causes headers and table cells to misalign #487
-Maybe: Datatable scrollable="true" breaks column template header #859


lemuellumapas:
This fixes a lot of my problems. When will this be merged?

cagataycivici:
There is a conflict and issue with new line character, we'll do this for 1.0 final. Thank you.

PR #938, opened 2016-09-21
Fix issues 859 and 860

mcfin1:
I have created these issues:
http://forum.primefaces.org/viewtopic.php?f=35&t=47021
http://forum.primefaces.org/viewtopic.php?f=35&t=47020

It fixes those to.


cagataycivici:
We've fixed this today, thanks for the PR. Also PR misses the styles for alignment.

mcfin1:
You are right. Only did the template part.

Issue #304, opened 2016-05-03
TableState for Table


gatapia:
It would be good to be able to remember the state of the table to be able to give the user continuity as they use an application.  This state would include:
- sorts
- filters
- column widths (if resizeable)
- page number
- expanded state?
- toggled columns?

A nice clean API to get and set this would be great:

```
# template
<p-datatable (stateChanged)="onStateChanged($event)" 
    [initialState]="loadInitialState()"...

# component
onStateChanged(event: any) {
  localStorage.set('table-state', event.data);
}

loadInitialState() {
  return localStorage.get('table-state');
}
```

Or perhaps even a totally automatic API, like:

```
<p-datatable persistState="true"...
```

GraemeSMiller:
Would be a great feature. Automatically is a nice to have. Best for me if it exposed functions so can hookin and save to remote storage etc

gatapia:
Just looking at source now and it looks like accessing the sort/filter state is simple enough, just attach to onFilter/onSort events.  There appears to be now way to set these values back.

Somewhat related, onSort fires the onFilter event even if filter has not changed, i.e. filter data (get filter event), sort data (get filter event then sort event).  I think that filter event is redundant.  minor issue though.

vijosoco:
Is there any update on this issue.  I am using the LazyLoad event to store the sort and filter information.  I am using local storage (which isn't ideal - but it works).  Now when i navigate away from the page with my datatable i can persist those values.  Once i return to that page (onInit) i can get those values back from localstorage and use them in very next query (lazy load).  HOWEVER, there seems to be no way to visually update the data table UI to show that those filters are being used or that the sort order on a column has changed.

Is there any way to update the datatable header fields on the UI?

lbamburski:
This pull request solves this. It uses localStorage for lazy load. Everything is updated visually. For integration with MultiSelect for toggle columns, event onPersistentStateApplied should be used similar to this:

onPersistentStateApplied(event): void {
        if (!event.state || !event.state.visibleColumnsState) {
            return;
        }

        let eventColumns = event.state.visibleColumnsState;

        this.columns.splice(0, this.columns.length);

        for (let i = 0;i < eventColumns.length;i++) {
            let eventColumn = eventColumns[i];
            this.columns.push({
                field: eventColumn.field,
                header: eventColumn.header,
                style: eventColumn.style
            });
        }

        let columnsOptionsByLabel = {};
        for(let i = 0; i < this.columnOptions.length; i++) {
[...]

PR #948, opened 2016-09-23
Table settings persit

lbamburski:
Allows DataTable to remember it's state of pagination, sorting, columns style, toogled columns,  filters.
Fixes: [New Feature] DataTable get/set state (for remembering sorts, filters, etc) #304


cagataycivici:
Thanks for the PRO but changes are a bit too much, significant new features should be developed by our team following our processes as a company policy. We'll add this feature in near future though.

PR #968, opened 2016-09-27
Fixed #966 and #971

### LINKED ISSUE #966: Angular 2.0.1 breaks p-dataTable expandableRows
The following change breaks how p-dataTable queries for the expandableRow template:
core: ContentChild descendants should be queried by default ([0dc15eb](https://github.com/angular/angular/commit/0dc15eb)), closes [#11645](https://github.com/angular/angular/issues/11645).

The issue in my case was that an other deeper located template (used for a column body) was taken for the row-expansion, but the let-[variable] was not set to the `$implicit` value (actually was set to `"rowData"` which is `undefined` in this case).

Please introduce a selector directive (e.g. `pRowExpansion`, like `pTemplate` for p-column) for row-expansion, and use as selector for `@ContentChild(PrimeRowExpansion)`.

ribizli:


ribizli:
Sorry seems I've messed up with two commits...

cagataycivici:
We have pTemplate to identify templates, I'll use it instead. Thank you.

ribizli:
Careful with that, deep querying can still find the column templates
instead of the expansion template...

Issue #876, opened 2016-09-09
Enhancement request: ability to add button(s) to the tabs row in the TabView component


eddbog:
Hello guys, and thanks a lot for your amazing components!

It would be nice to have ability to show some custom button(s) on the tabs row of a TabView component - right next to the other tabs. Such button could generate an event which can be used by client code to perform different actions, for example adding a new tab, closing all tabs except active, etc.

eddbog:
Closed in favor of Donriver ticket

PR #977, opened 2016-09-28
Implemented #876. Added canClose guard callout.

eddbog:
This change implements the following new features:
1. Ability to have customizable buttons right next to the tabs. This can be useful if we want to create a new tab dynamically and then instantiate or route some other component into it.
2. Ability to cancel tab closing process. This is useful when the tab cannot be closed due to business logic restrictions, for example when there are some unsaved changes in the child component.


eddbog:
Closed due to re-implementing in Donriver fork.

PR #980, opened 2016-09-28
Fixed calculations of x coordinate of resize helper.

Dema:
Fixes issues with incorrect helper positioning and wrong delta width calculations for columns
when the datatable is not the leftmost one.


cagataycivici:
Dup of #1065

PR #986, opened 2016-09-29
Filtered rows are not refreshed if [value] is changed

Dema:
Calling this.filter() in ngDoCheck()


cagataycivici:
Reverting due to #1170

PR #991, opened 2016-09-30
Lightbox : Display image title and close button

raphayol:
Simple change in order to display the image title on lightboxes.
If a title is present this will display the close button too.
If there is no title the comportment is the same as before : lightbox is display without title and close button.


cagataycivici:
Thank you

PR #992, opened 2016-09-30
Spinner: added readonly

kryzys69:
I've added "readonly" setting to spinner component. I'm not sure about a naming convention. because in calendar it is called "readonlyInput"


cagataycivici:
Don't you need to bind it to input and buttons, also disable functionality when readonly is on?

kryzys69:
It is meant to disable only input, so maybe it should be called "readonlyInput" (like in calendar). In enables using spinner on the phone without showing keyboard.

cagataycivici:
It seems we've removed readonly binding so PR is invalid, I'll add it.

PR #1019, opened 2016-10-04
IE11 rendering bug with <p-calendar></p-calendar> CalendarModule --- Update calendar.ts

MisaelSabb:
Two days ago I downloaed "primeng" with this cmd comand:	npm install primeng
and I imported correctly the "CalendarModule".

There is a problem with <p-calendar></p-calendar> on IE 11  (Version 11.187.14393.0)

The problem is related to the (mouseenter) and (mouseleave) events.

Once the Calendar popup is shown, on mouseOver the input field changes abnormally his size and color.

You can check this issiue on your web site (using IE 11): http://www.primefaces.org/primeng/#/calendar
If you change the used "themes" using the drop-down menu it works correctly.

I solved the issiue removing the following code inside the template (line 18 of calendar.ts file) :
(mouseenter)="hovered=true" (mouseleave)="hovered=false"

Maybe it's not the best solution but just a simple workaround.
Thank for your attention :)


cagataycivici:
This should be fixed with new calendar.

Issue #1027, opened 2016-10-05
TriStateCheckbox not taking initial values for false


JiroDan:
Does not appear to accept initial values correctly, true seems to work, but false and null will both show up as if null.

Looking at the javascript, 
this.value = value || null;

How can I set the initial value to true or false or null? 

Behavior sets true to true but both false and null to null (from the toggle state).

PR #1031, opened 2016-10-05
Fixed TriStateCheckbox to allow all three states to be preset.

JiroDan:
Fix for  #1027


cagataycivici:
Fixed with a slightly different approach, also can't accept the PR due the differences in space/tab.

PR #1059, opened 2016-10-12
Added timezone as option. Fixed locale to reflect changes in fc 3.0

sebastianhaas:
I added `timezone` as additional option. 

https://fullcalendar.io/docs/timezone/timezone/

I also fixed `locale` to reflect changes in fc 3.0, where `locale` is s string containing a locale rather than an object.

https://fullcalendar.io/docs/text/locale/


cagataycivici:
Has conflicts so merging manually, thanks.

PR #1067, opened 2016-10-14
Propagating the blur event for InputMask and AutoComplete controls

odedolive:
We needed the blur (focus lost) event on some of the controls to be able to hook in custom validation logic. I added the propagation of this even to the masked textbox and the auto complete textbox


cagataycivici:
What is the reason for changing || with && at inputmask? How is it related to focus and blur?

odedolive:
I'm sorry, these issues are not related.
Changing || to && is a fix to another issue with the "autoClear" property on the input mask control.

I checked it into my local branch and it updated the pull request.

So in fact, there are 3 different issues in this pull request - 2 commits that implement the missing blur event on the autocomplete and input mask controls, and one commit that fixed an issue with the autoClear implementation on the input mask control.

Thank you.

PR #1071, opened 2016-10-17
TabView: Implement ability to cancel close tab action

donriver:
This is useful when the tab cannot be closed due to business logic restrictions, for example when there are some unsaved changes in the child component.


cagataycivici:
Resolved via #1075 instead.

donriver:
Closed as obsolete in favor of [#1075](https://github.com/primefaces/primeng/issues/1075)

PR #1072, opened 2016-10-17
TabView: Implement ability to have customizable buttons right next to the tabs

donriver:
This can be useful if we want to create a new tab dynamically and then instantiate or route some other component into it.


donriver:
Closed due to existing workaround

PR #1106, opened 2016-10-22
Calendar: perf. optimization and model sync

moshensky:
Perf optimization reducing isSelected and isToday callback calls.
Sync calendar's input with model value.
Update showcase docs with example.
After component view init if value is defined than input will be synced. If value will be changed from model than input will be synced again.


cagataycivici:
We'll keep as it is and consider this if performance issues are reported. I'd rather not keep state at date metadata when creating the calendar to avoid maintaining it since we leave it to angular expression. This was actually discussed when we were doing the calendar, still if performance issues are reported, we'll reconsider. Thank you.

moshensky:
Will you add ui sync to model's value?

PR #1117, opened 2016-10-24
When a checkbox is disabled, I was still able to tab into it and change it's state using the keyboard

odedolive:


cagataycivici:
PR contains unrelated code such as inputmask, please do a new PR with only related code. Thank you.

cagataycivici:
Fixed via #1121

PR #1119, opened 2016-10-24
Allow resetting of the DataScroller

mroskamp:
It'd be nice to be able to use the DataScroller in such a way that it can reused for different data sets. For example, an infinite scrolling list that's searchable by the user.

When the user first arrives on the page, a set of search results appear. The user can then change the search criteria to provider better results for their specific need. The onLazyLoad handler is aware of the search criteria and returns the data accordingly. In order for this to work though, the scroller needs to be reset, so the loaded data starts again at the beginning.

### DIFF
--- a/components/datascroller/datascroller.ts
+++ b/components/datascroller/datascroller.ts
@@ -116,6 +116,7 @@ export class DataScroller implements AfterViewInit,DoCheck,OnDestroy {
     reset() {
         this.first = 0;
         this.dataToRender = [];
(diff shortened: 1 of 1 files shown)


cagataycivici:
Don't we need to call load() after reset() as you clear the dataToRender(). reset just clears the data I from what I see?

mroskamp:
Yes, you're right. I realized that last night, I've made the change to call load().

PR #1120, opened 2016-10-25
DomHandler equals method wasn't working properly

eugeniobac:
DomHandler equals method was returning false even when the object was equal


cagataycivici:
Thank you.

Issue #1136, opened 2016-10-26
DataTable lazy loading


bogdanvaduva:
I created a component based on DataTable.
DataTable has filters and pagination.
First when I load the data, everything is fine. Then I enter something in one filter column, at this point everything goes crazy because onFilterKeyup -> filter -> onLazyLoad (which loads the data) -> ngDoCheck (which also does a filter) -> filter and stays here. I solved the problem by adding a property called filtered and changing the following in datatable.ts

DataTable.prototype.onFilterKeyup = function (value, field, matchMode) {
var _this = this;
this.filtered = false;
if (this.filterTimeout) {
clearTimeout(this.filterTimeout);
}
this.filterTimeout = setTimeout(function () {
_this.filters[field] = { value: value, matchMode: matchMode };
_this.filter();
_this.filterTimeout = null;
_this.filtered = true;
}, this.filterDelay);
};

and

DataTable.prototype.filter = function () {
if (!this.filtered)
this.first = 0;
if (this.lazy) {
if (!this.filtered)
this.onLazyLoad.emit(this.createLazyLoadMetadata());
}
....

Now it seems to work fine. Could someone tell me if I did somenthing wrong in my component or there is an issue in datatable.ts?

Regards.

mwaltersva:
I'm also experiencing this issue. It works fine in beta.16, but once you enter a filter term in 17+, it repeatedly calls the lazyLoad function.

ElaAlex:
Same issue here. Using Angular 2.1 together with primeng beta 20. Lazy loading works but as soon as I add filter capability and enter a value into one (contains-) filter a constant reloading circle starts.

bogdanvaduva:
Hi @mwaltersva, @ElaAlex if you make the changes within the datatable.ts it should work fine, but I would like to have these changes into the main branch.
Regards.

cagataycivici:
Please do a PR and we'll review.

bogdanvaduva:
Thanks.
Regards

PR #1145, opened 2016-10-27
DataTable filter loop with lazy loading

bogdanvaduva:
I've added the filtered property which should solve the issue with filter() loop.

### DIFF
--- a/components/autocomplete/autocomplete.ts
+++ b/components/autocomplete/autocomplete.ts
@@ -60,7 +60,7 @@ export class AutoComplete implements AfterViewInit,DoCheck,AfterViewChecked,Cont
     
     @Input() placeholder: string;
     
-    @Input() readonly: number;
+    @Input() readonly: boolean;
         
     @Input() disabled: boolean;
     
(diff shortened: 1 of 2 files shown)


cagataycivici:
What is the original issue with loop?

bogdanvaduva:
It stays in a loop if you use lazy loading and filter.

maleet:
For me looping occurs because of  ngDoCheck. I get new data so there are changes with diff checking. And if there are filters also then it's in loop... new data from backend, ngDoCheck with changes and filter method call.

mwaltersva:
#1136 Just tagging the original issue

cagataycivici:
This turned out to be a regression of #986 which has better solutions like resetting table when actual data changes, so reverting that to fix this. Logged at #1170

bogdanvaduva:
Thanks.
I think there is also a problem with hasFilter() because returns true even if the filters are empty (I means when you enter something in a filter and then you delete it).

dan-madaket:
@cagataycivici There is another regression on this issue.  I'm having this problem in primeng 4.0.0-rc.2.

PR #1152, opened 2016-10-28
Add events for focus and blur

mroskamp:
We want to be able to drive some logic off of the focus/blur state of the dropdown.


cagataycivici:
Fixed via #1156

aykutyararbas:
multiselect onBlur is triggered when a multiselect option is checked.

Steps to produce: 
1-Click multiselect drop down to see choices
2-Click one of the choices.
3-See onBlur  triggered.

Issue #553, opened 2016-07-04
DataTable Column custom sorting by default


Coffee-Tea:
When I use default sorting my custom sorting function does not work.
On click it works perfect, but does not when table just loaded.

I thin that I should run .sort() function in dataTable after load table, but I don know how to do it?
Maybe I have to do something else?

Thank you!

shahmirn:
I ran into the same issue, so I've created a pull request https://github.com/primefaces/primeng/pull/605 that should fix this

ryanwawr:
Would be nice to see that fix pulled in sometime soon. I'm having the same issue with the initial custom sort not working. Worked around it for now by calling my custom sort function manually once the datatable is populated, but would prefer not to have to resort to such a messy solution.

Samisama:
Hello, is there a way to fix that problem ? I'm also using sortable="custom" on one of my columns but the default sort on this column doesn't call my custom sort function.
Thank you

cagataycivici:
This should be fixed in recent PrimeNG release, please try with them and if issue persists drop a comment and we'll reopen.

PR #1167, opened 2016-10-31
Custom sort function on load is not invoked

ryanwawr:
Added to fix issue #553 . If the sort field for a table is a column with a custom sort function, the 'customSort' event will not be emitted when the table is first loaded, since 'sortColumn' is not set. 

I added a line to the condition in sortSingle() that will also emit the 'customSort' event if
1) 'SortField' is set
2) The column whose 'field' property matches 'sortField' has its 'sortable' field set to 'custom'


cagataycivici:
But sortColumn still won't be set so how could the sortFunction be emitted?

cagataycivici:
Please review  https://github.com/primefaces/primeng/commit/583e89160e64b4db0f33ba9d61d3f4360a4d7a60

It sets it before sorting if sortColumns is not available on load.

ryanwawr:
Ah, you're right. I missed setting sortColumn, but it looks like your commit should resolve the issue.  Thanks!

JiroDan:
Doing this in onDoCheck means that you must check to see if the columns are bound first;
634:          if(!this.sortColumn && this.columns) {

Otherwise it throws, Cannot read property 'find' of undefined: 635

PR #1175, opened 2016-11-01
Use contains for dropdown filter

mauriciovigolo:
Changed to search instead of startsWith.
It's easier to find the available options in the dropdown.


cagataycivici:
Shouldn't it be faster to use indexOf?

mauriciovigolo:
Hi @cagataycivici. Sure you are right! Tks!!

PR #1182, opened 2016-11-01
Field undefined extended sub-property protection

saverett:
Use ng2 elvis-operator style protection for data field sub-properties that could be undefined to help prevent `Cannot read property 'X' of undefined` exceptions.

**For example:**
In the following datatable column declaration, primeng datatables would throw a `Cannot read property 'c' of undefined` exception if `b` was `undefined`. It is often the case that extended field sub-properties are present for some collections/rows, and not present for others. This PR solves this problem, while still allowing primeng datatables' built-in features that rely on the fully-qualified `field` property (such as sorting and filtering) to work correctly:

`<p-column field="a.b.c" header="C"></p-column>`


saverett:
Any updates on whether this PR will be merged? This would greatly help my primeng datatables implementation.

cagataycivici:
Thank you.

PR #1196, opened 2016-11-03
Calendar localization

kryzys69:
My change allows to dynamically apply localization to calendar picker

### DIFF
--- a/components/calendar/calendar.ts
+++ b/components/calendar/calendar.ts
@@ -40,7 +40,7 @@ export interface LocaleSettings {
                         <span class="fa fa-angle-right"></span>
                     </a>
                     <div class="ui-datepicker-title">
-                        <span class="ui-datepicker-month" *ngIf="!monthNavigator">{{currentMonthText}}</span>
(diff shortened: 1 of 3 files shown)


components/calendar/calendar.ts
```diff
@@ -231,14 +236,8 @@ export class Calendar implements AfterViewInit,OnInit,OnDestroy,ControlValueAcce
     constructor(public el: ElementRef, public domHandler: DomHandler,public renderer: Renderer) {}
 
     ngOnInit() {
-        let today = new Date();
```
cagataycivici: Why remove these parts?

kryzys69: Because they are not used anymore, both `weekDays` and `currentMonthText` are pulled directly from `locale`. `locale` has setter that recreates component with new locale.
I hope I haven't missed anything... in my case and in modified calendar demo it works well.

mlofjard:
I hope this gets merged soon. :)

cagataycivici:
I suspect this may fail if first day of week is sunday or monday, wdyt? I'll add it with some changes.

cagataycivici:
See;

https://github.com/primefaces/primeng/issues/2361

PR #1208, opened 2016-11-04
Input textarea placeholder

ryanwawr:
Added a few lines to host the placeholder attribute for textarea in the InputTextarea directive and allow it to be set as an input.


ryanwawr:
Just realized that this is entirely unnecessary, since inputTextarea is a directive. Whoops.

Issue #654, opened 2016-07-24
Datatable column sorting does not work with global filter


kurigohan:
When a global filter used on the datatable, single column sorting stops working.

Goldflow:
same issue here

Shawson:
+1 with 1.0.0-beta.17

mariahlynne:
Any chance we can get this fix included in RC5?

anugoyal12:
Hello everyone,

any idea which release this fixed issue is available?

PR #1210, opened 2016-11-04
Fix issue #654: using a global filter prohibits sorting

AngularPadawan:
This fixes issue #654. Sorting now works in a data table with an enabled global filter. I fixed this by making hasFilter() aware of the global filter.


anugoyal12:
Which version is this fix available. I am working on a similar story and if this is fixed, will fix my issue. Please let me know, will appreciate.

cagataycivici:
Merged manually, thanks.

PR #1227, opened 2016-11-07
Disables click on disabled accordion tab

andreipreda:
Clicking on a disabled accordion tab was following the anchor tag hash.
Not sure that was the desired behaviour. Just in case it helps, a small fix to return false when toggled & disabled.


cagataycivici:
Thanks

Issue #1192, opened 2016-11-02
Calendar 'onSelect' not fired on time change


RomeoMikeFox:
It seems that the `onSelect` event is not fired up if one only changes the **time** in a `p-calendar` with date and time. I didn't test with a time only calendar.

Is it possible to trigger it when time is also modified?

PR #1233, opened 2016-11-08
Fix issue #1192 : Add onSelect event on time change

RomeoMikeFox:
To fix the issue #1192 I raised, I added the 'onSelect' event triggering when time is changed, as it is done for the date modifications.

I also added an example in the showcase dealing with both onBlur and onSelect events.

Hope it helps.


cagataycivici:
Thank you, we'll add the onselect part.

PR #1245, opened 2016-11-09
Fixed #1243

### LINKED ISSUE #1243: site.css font-size overwritten
In the [site.css - line 205](https://github.com/primefaces/primeng/blob/master/showcase/resources/css/site.css#L205)

The font-size is set to 90%. This makes the showcase looking different to someone who is using primeng in an application. 
```
/* override */
.ui-widget {
    font-size: 90%;
}
```

kartheininger:


cagataycivici:
Fixed and also applied 14px font-size to body as ui-widget is defined in em usually 1em in community themes. Community themes get their font-size from the base font-size that should be defined in body.

PR #1253, opened 2016-11-09
fix of left percent offset for (range) slider

berc:
fix of left percent offset for (range) slider

### DIFF
--- a/components/slider/slider.ts
+++ b/components/slider/slider.ts
@@ -201,16 +201,16 @@ export class Slider implements AfterViewInit,OnDestroy,ControlValueAccessor {
     
     updateHandleValue(): void {
         if(this.range) {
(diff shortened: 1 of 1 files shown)


cagataycivici:
What is the actual issue here?

berc:
slider position is moved to the right side by min value, because bad percentage calculation. Percentage for "left" css translate begins not from 0, but from percentage of min value from zero and also slider button position on right side is out of the 100 % and out of the slider component.

cagataycivici:
How can we replicate the issue? Could you please post a simple test code for it?

berc:
set min to 500 and max to 1000 and you should see that slider is going out of the component.

cagataycivici:
Ok, I'll try it tomorrow and review the PR. Thank you.

cagataycivici:
The PR has not fixed anything so reverting, we'll fix this on our side, thank you.

PR #1259, opened 2016-11-10
Real readonly attribute to p-calendar

victorschinaider:
I was using the p-calendar, but in a new work, i needed to not allow changes in p-calendar value. But the 'readonlyInput' is just to not allow keyboard digits, but we can select in calendar. I think 'readonlyInput' is not the best name for this behavior, 'readonly' is just to read, not to change.

So i changed the attribute 'readonlyInput' to 'noTypeful', and i created an attribute 'readonly' to not allow changes, the real behavior.

I updated the code, and the docs.
It needs review.
Thanks.

Ps.: package.json changes is not part of the commit

### DIFF
--- a/components/calendar/calendar.ts
+++ b/components/calendar/calendar.ts
@@ -25,7 +25,7 @@ export interface LocaleSettings {
     template:  `
         <span [ngClass]="{'ui-calendar':true,'ui-calendar-w-btn':showIcon}" [ngStyle]="style" [class]="styleClass">
             <input type="text" pInputText *ngIf="!inline" (focus)="onInputFocus($event)" (keydown)="onInputKeydown($event)" (click)="closeOverlay=false" (blur)="onInputBlur($event)"
-                    [readonly]="readonlyInput" (input)="onInput($event)" [ngStyle]="inputStyle" [class]="inputStyleClass" [placeholder]="placeholder||''" [disabled]="disabled"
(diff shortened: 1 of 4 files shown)


cagataycivici:
noTypeful isn't very descriptive for me, why don't you use disabled?

victorschinaider:
Yes, you're correct, maybe "noTypeful" isn't the best descriptive name. 

But, i continue with the ideia that "readonlyInput" isn't the best name to the behavior definied.

If we look to html docs https://www.w3.org/TR/html5/forms.html#the-readonly-attribute.

> The readonly attribute is a boolean attribute that controls whether or not the user can edit the form control

So, when i choose "readonlyInput", usually, a developer will think that the element will have the behavior above. So the value of the element will not change, but it not occurs.

> The difference between disabled and readonly is that read-only controls are still focusable, so the user can still select the text and interact with it, whereas disabled controls are entirely non-interactive.

I don't use "disabled" because of the definition above. I want user to interact with the control, disabled element shouldn't allow this.

Any ideias?

cagataycivici:
Not sure about this, readonlyInput is derived from PrimeFaces actually, it is used to set the input as readonly and only allow selecting date with the overlay without input. readonly also should not show the overlay maybe, again I'm not sure, we need more feedback on this so not taking an action at the moment.

PR #1268, opened 2016-11-10
Make Dropdown work with any kind of data without following SelectItem…

maniacneron:
DropDown component only works with data that following SelectItem Interface. It is hard to transfrom rest results to this format only for using in dropdown component. So I added optionLabel and optionValue attributes to provide label and value keys. Default values are "label" and "value" so it does not affect if options are following SelectItem interface.

### DIFF
--- a/components/dropdown/dropdown.ts
+++ b/components/dropdown/dropdown.ts
@@ -19,7 +19,7 @@ export const DROPDOWN_VALUE_ACCESSOR: any = {
             (mouseenter)="onMouseenter($event)" (mouseleave)="onMouseleave($event)" (click)="onMouseclick($event,in)" [ngStyle]="style" [class]="styleClass">
             <div class="ui-helper-hidden-accessible">
                 <select [required]="required" tabindex="-1">
-                    <option *ngFor="let option of options" [value]="option.value" [selected]="selectedOption == option">{{option.label}}</option>
+                    <option *ngFor="let option of options" [value]="option[optionValue]" [selected]="selectedOption == option">{{option[optionLabel]}}</option>
                 </select>
             </div>
             <div class="ui-helper-hidden-accessible">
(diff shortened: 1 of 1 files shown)


cagataycivici:
The idea is good but we haven't received any negative feedback until now regarding SelectItem api, we prefer to have this API to represent the option tag and to follow a certain interface. We'll reconsider if we receive more feedback in future. Also if we do something like this, it should also do the same for all input component that use SelectItem API.

chpasha:
This obligation to implement some interface is indeed a pain, since the incoming data has to be converted to SelectedItem just to be shown or has to implement it (which is also not very good, if some dto object has to know about SelectedItem)

cagataycivici:
I see your point but we need an interface to represent option and optgroup elements, for example to disabled an item in the list and also to group them with optgroup.

chpasha:
can't you convert any items to any interface internally? for instance we tell the dropdown that the label field name is "name", "enabled" = "active" and so on. In the options setter you create wrapper for every incoming element converting it to SelectItem-Implementation which has for instance
get label(): string {
  //I skip checking if prop exists 
  return element[labelField];
}
this is actually what we have to do in the client code - convert incoming options to array of SelectItem-Wrappers

cagataycivici:
It would be better if you wrap in at app side rather than we do it. But again, if we get more feedback, that is fine to do the change.

digitalcraftco:
To add more feedback. We hit this problem also as soon as we started implementing p-dropdown with our rest api. None of the data models contained "label" or "value". Changing or adding columns in DB was going to be a huge pain. Mapping results client side is not optimal either.

We ended up extending SelectItem in our model and then deserialize the data. Which is nice because sometimes you'll want to manipulate data like below example for label. So now it's nice to have the SelectItem Interface stay consistent for elements requiring option tags.

`export class VehicleGroup extends SelectItem implements SerializableDataModel<VehicleGroup> {
	public id: number;
	public value: number;
	public isDepositRequired: boolean;
	public description: string;
	public label: string;

	public methodName() {
		return 'vehiclegroup';
	}

	public deserialize(input) {
		if (input) {
			this.id = input['id'];
			this.isDepositRequired = input['isDepositRequired'];
			this.description = input['description'];
			this.label = this.id + ' - ' + this.description;
			this.value = this.id;

			return this;
		}
	}
}`

sanjarbek:
I also need custom SelectItem interface, so I extended it with additional data field which contains data object itself. In template I used data field to show custom data. But now I have problems on filter. It would be nice to have at least additional data field on SelectItem, so there is not to extend it.

ImproSnake:
We could need the Requested Function too. 

Do you really have to change or delete the interface ? Cant we just define the value and label proberty via in the directive like :

`<p-dropdown [options]="data" [filter]="true" [(ngModel)]="selected" labelName ="XXXX" dataName = "XXX">
</p-dropdown>`

or would this be a problem ? 

Greetings

Issue #1269, opened 2016-11-10
DataTable doesn't update filtered results when source data changes


ats17:
Normally, DataTable updates its view any time its value input changes; however, this doesn't happen when a filter is applied. 

To reproduce:
1. Create a DataTable with filtering enabled on at least one column and add sample data
2. Update the data in code and see that the DataTable view updates to match
3. Enter some filter text that passes at least one row
4. Update the data in code (for one of the displayed rows) and see that the DataTable view **does not update to match**

ats17:
I've added pull request #1270 as one way to fix this.

ats17:
Closed pull request #1270 because I buggered the repo. Replaced with PR #1279.

cagataycivici:
Tracking at #1295

Issue #1170, opened 2016-11-01
bug: datatable filter function with lazy


lsa-shanghai:
datatable.ts 
line: 524
if(this.hasFilter()) {
      this.filter();
}
When LazyLoadEvent fired, it will update the datatable's value and it will fire LazyLoadEvent in a circle.

cagataycivici:
Same as #1145

erkanerol:
I use     "primeng": "4.0.0-rc.2" and "@angular/core": "4.0.1". The bug still exists. The lazy load event is fired again and again when lazy=true. Should I check it in my code or set another parameter?

AlexanderPruss:
+1. The bug still seems to be there in 4.0.0-rc.2.

PR #1270, opened 2016-11-10
Fix for issue #1269 - DataTable doesn't update filtered results when source data changes

ats17:


cagataycivici:
This is likely to cause an infinite loop, see;

#1170

Please try this with a lazy loading datatable first.

ats17:
Good point, I missed that and the other related issues. I've added a check to skip the filter call when lazy loading since in that case the data source would be handling the filtering.

My original commit checked filteredValue, but that would miss cases where nothing passed the filter before the data was updated. Now I'm checking hasFilter(). 

To avoid unneeded filtering, I've rewritten hasFilter() to only return true when at least one filter contains text. This looks to be compatible with all current callers.

components/datatable/datatable.ts
```diff
@@ -1065,15 +1065,12 @@ export class DataTable implements AfterViewChecked,AfterViewInit,AfterContentIni
     }
 
     hasFilter() {
-        let empty = true;
-        for(let prop in this.filters) {
-            if(this.filters.hasOwnProperty(prop)) {
-                empty = false;
```
ats17: To minimize changes, I probably should have simply added 
`if (this.filters[prop].value.length) `
in front of
`empty = false;`
in the original code.

ats17:
I've made a mess out of this repo so I'm going to close this pull request and open a clean one

cagataycivici:
Thanks, no need for a new PR, we'll do it at 1295

PR #1276, opened 2016-11-11
Input parameter 'unit' added to ProgressBar component

ladislav-bozek:
It would be nice to let user to define '%' as input parameter. It allows user to hide '%' or change it to some other unit e.g. 'MB' . With default '%' the backward compatibility should be preserved.


cagataycivici:
Great, thanks!

PR #1279, opened 2016-11-11
Fix for #1269

ats17:
Avoids infinite loop with lazy loading by not filtering when lazy, since the data source is responsible for lazy filtering anyway.
Tightens hasFilter() to only be true if filter text exists. This avoids unnecessary runs through the filter.


cagataycivici:
Thank you, we'll do this ourselves for RC5 at;

https://github.com/primefaces/primeng/issues/1295

Issue #725, opened 2016-08-10
Selected values in multiselect label are (null) when list of options loaded *after* bound selectedValue


tbertran:
Title is a bit convoluted but this [plunker ](http://plnkr.co/edit/ds1QX7) illustrates the problem.
Note that if `selectedCars` is populated **before** `cars`, it works fine.

FYI, I'm having this issue because, in my case, those two come from two different GET requests.

michalig:
I'm not sure if this is the same issue but I have similar problem when using multiselect together with ng-translate. If I use translated string as a label, after refreshing the page (with F5) it is not displayed at all. I've tried to fix that somehow but I guess this is because ng-translate loads translation asynchronously and multiselect is rendered before it finishes.

@tbertran did you have a similar problem with multiselect?

tbertran:
I did not use it with `ng-translate` so I don't know.

mozdynjr:
I've also had this problem with **@angular/*=2.4.7** and **primeng=^2.0.1**.
But I guess it works as suppose to. If you want to assign/reference selected values to/from a collection, you first need to fetch this collection.

My case was in general like this:
There was a component with standard binding:
`<p-multiSelect [options]="brandOptions" defaultLabel="All Brands" [(ngModel)]="preselectedBrands" (onChange)="onCarBrandChange($event)"></p-multiSelect>`

and I was fetching options by service like so:
```
private prepareBrandOptions() {
        this.carService.getBrands()
            .then(brands => this.convertToOptions(this.brandOptions, brands));
    }
```
and at the same time doing `this.restoreBrandFilter();` which actually populated `this.preselectedBrands` from local storage.
`this.preselectedBrands` was always faster than REST call from `this.carService.getBrands()` and as an end result I'd always got nulls as labels (but the selections in drop-down were ok).

So I've moved `this.restoreBrandFilter();` simply like so
```
private prepareBrandOptions() {
        this.carService.getBrands()
            .then(brands => {
                this.convertToOptions(this.brandOptions, brands);
                this.restoreBrandFilter();
            });
    }
```
Probably could be done more intelligently, but was sufficient at that time.
Hope it helps, because the issue is not so obvious at the first sight.
It would be nice to have some warning or reminder in the docs...

giscafer:
i got the same issue

broweratcognitecdotcom:
This problem exists in version 4.1.0. We tried 5.0 also. Still a problem.

PR #1280, opened 2016-11-11
Added input property compareField for comparing property between valu…

petrbalat:
…es in multiselect

### DIFF
--- a/components/autocomplete/autocomplete.ts
+++ b/components/autocomplete/autocomplete.ts
@@ -245,7 +245,7 @@ export class AutoComplete implements AfterViewInit,DoCheck,AfterViewChecked,Cont
     }
     
     show() {
-        if(!this.panelVisible) {
+        if(!this.panelVisible && this.focus) {
             this.panelVisible = true;
             this.panel.style.zIndex = ++DomHandler.zindex;
             this.domHandler.fadeIn(this.panel, 200);
--- a/components/autocomplete/autocomplete.ts
+++ b/components/autocomplete/autocomplete.ts
@@ -26,7 +26,7 @@ export const AUTOCOMPLETE_VALUE_ACCESSOR: any = {
                     <span class="ui-autocomplete-token-label">{{field ? val[field] : val}}</span>
                 </li>
                 <li class="ui-autocomplete-input-token">
-                    <input #multiIn type="text" pInputText (input)="onInput($event)" (keydown)="onKeydown($event)" (focus)="onFocus()" (blur)="onBlur()" autocomplete="off">
+                    <input #multiIn type="text" pInputText [attr.placeholder]="placeholder" (input)="onInput($event)" (keydown)="onKeydown($event)" (focus)="onFocus()" (blur)="onBlur()" autocomplete="off">
                 </li>
             </ul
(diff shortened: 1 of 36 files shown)


cagataycivici:
What problem does this solve?

petrbalat:
If you have value as object in SelectItem then compare with operator == is a problem because objects in options can be different instance but with the same values. With this commit you can compare by id for example.

petrbalat:
select2 has a id property for this func. So maybe it will be better rename the property compareField

cagataycivici:
What is the case where you have to compare different select item instances.

petrbalat:
If you have different api which load options e.g. we have components for clients e.g. <clients-select> which always loads all record from /api/clients ... but in other edit form e.g. <edit-bill> clients is property from some dto object. Like I said select2 has id property, kendo has valueField http://www.telerik.com/kendo-angular-ui/components/dropdowns/api/DropDownListComponent/#valuefield-string  ...

cagataycivici:
I see but we need more details, we should not add it just because other libs have it. When you load the options, you bind them to the multi select. When you load them again to change them, same instances are bound as well. Can you post some code where you need compareField? It is not clear.

petrbalat:
e.g. you have a cmp
@Component({
    selector: 'multiselect-clients-cmp',
    template: `
<p-multiSelect [options]="allClients | toSelectItem:'name'" [ngModel]="clients" compareField="id"
defaultLabel="" (onChange)="valueChange($event.value)"></p-multiSelect>
`
})
export class MultiSelectClientsCmp extends AbstractCmp implements OnInit {

    @Input('clients')
    clients:any[] = [];
    @Output('clientsChange')
    clientsChange = new EventEmitter<Array<any>>();

    allClients:any[] = [];

    constructor(http:Http, notificationService:NotificationService) {
        super(http, notificationService, null, null);
    }

    ngOnInit() {
        //clienti do comboboxu
        this.http.get(`/api/admin/clients`).map(res => res.json()).subscribe((res:Array<any>) => {
            this.allClients = res;
        });
    }

    valueChange(items:Array<any>) {
        this.clientsChange.emit(items);
    }
}


but in other cmp

<form>
            <div class="form-group">
                <label>label...</label>
                <multiselect-clients-cmp [(clients)]="dto.clients"></multiselect-clients-cmp>
            </div>
</form>

//delay because of bug  #725  .. https://github.com/primefaces/primeng/issues/725
this.http.get(`/api/admin/user/${this.userId}`).map(res => res.json()).delay(200).subscribe((res:any) => {
            this.dto = res; // clients in dto are diferent instance...but with the same values and unique id
});

stephanecot:
I need the same thing, but maybe I have a better exemple:

This Multiselect has a problem when used in a form in 'modification'.
I have a get request to populate the [**options**] property and then I populate **myForm** ==> this.**myForm**.setValue(**data**). (**data** is another get request)
[**options**] is a list of object: {id: xxx, label: yyy} and my form has a property **zzz** whitch is a list of the same object (but not not the same instance!)

The problem is that the objects in **options** field can't be matched with the list of objects  of **myForm.zzz**. It's not the same instance. They re not way to indicate it's the same object (no method equals() like in java :) ).

cagataycivici:
This PR has too many unrelated commit, please create a issue ticket so we can discuss first.

PR #1286, opened 2016-11-12
spinner crashes angular in dev mode

mmikeyy:
This solves a problem when angular is run in dev mode under the following conditions:
- spinner is in a reactive form;
- spinner's initial value is not null or empty string (some numeric value)
- spinner control's inclusion in template is conditional (within element with *ngIf attribute)
Problem: angular crashes whenever the control is hidden, then displayed. 
Reason: angular reports that a value was changed after it was checked. 

The timeout proposed eliminates the error with no side effect that I could detect.


cagataycivici:
Thank you but timeout is not the best solution, we'll do this at 1287 with binding.

PR #1343, opened 2016-11-17
Listbox's events should pass the current value

Ov3rlo4d:
Currently listbox's events don't behave according to their documentation: they just pass the original event instead of also passing the currently selected value.


cagataycivici:
Thank you

PR #1355, opened 2016-11-18
added function to change to JSON format when style option is acquired.

mikosu3:
because [ngStyle] need JSON args.

I tried the following 2 patterns with RC5, but both errors occurred.
```
<p-inputMask [(ngModel)]="val" mask="99-9999" style="background-color: #bde9ba; text-align: right"></p-inputMask>
<p-inputMask [(ngModel)]="val" mask="99-9999" style='{"background-color": "#bde9ba", "text-align": "right"}'></p-inputMask>
```


cagataycivici:
Please use forum first;

Should have been;

[style]='{"background-color": "#bde9ba", "text-align": "right"}'

mikosu3:
thank you for anser.
i will use the forum from the next

Issue #705, opened 2016-08-07
Empty tooltips are still displayed


mmikeyy:
I have conditional tooltips (some elements generated by a *ngFor loop need them, others do not). 

I tried assigning empty text or null value to the tooltip for items not requiring one. The arrow and a small empty container are still rendered.

I just added these few lines to the tooltip.js code and the problem is solved:

``` javascript
 Tooltip.prototype.show = function () {
        if (!this.text){              // start
            return
        }                             //end
        this.create();
```

This would be a minor but welcome enhancement to the component.

dzhavat:
Fixed in  #1361

PR #1361, opened 2016-11-18
[Bugfix] Do not show the tooltip if the text is an empty string

### LINKED ISSUE #1190: Empty tooltip
Currently when the value is not there the tooltip is still being displayed. There is no use from empty tooltip. Tooltip should not be shown at all, when the text is null or empty.

dzhavat:
Fixes #1190 and #705


cagataycivici:
Will merge manually as it does not suit coding convention.

PR #1373, opened 2016-11-20
Patch 2

mmikeyy:
regarding getHeight function, used in dialog resizing, there's no point in getting offsetHeight, then subtract padding and border to obtain the original height.

If we're going to write the corrected height to style.height in the end, we should get style.height as original height value.

Why is this important? Because style.height is the inner or outer height depending on what box-sizing is set to.

If box-sizing is set to 'box-content', style.height is outer height. And if we adjust the inner height and then write the result to style.height, we get a dialog that shrinks uncontrollably when resized.

### DIFF
--- a/components/dom/domhandler.ts
+++ b/components/dom/domhandler.ts
@@ -265,10 +265,8 @@ export class DomHandler {
     }
 
     public getHeight(el): number {
-        let height = el.offsetHeight;
         let style = getComputedStyle(el);
-        height -= parseFloat(style.paddingTop) + parseFloat(style.paddingBottom) + parseFloat(style.borderTopWidth) + parseFloat(style.borderBottomWidth);
-        return height;
+        return parseFloat(style.height);
     }
 
     public getWidth(el): number {
--- a/components/dom/domhandler.ts
+++ b/components/dom/domhandler.ts
@@ -221,7 +221,7 @@ export class DomHandler {
 
         if (margin) {
             let style = getComputedStyle(el);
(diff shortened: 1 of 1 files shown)


cagataycivici:
Merged floats, getHeight stays as it is, there is already getOuterHeight to calculate total height, getHeight is the inner height without padding. Of course they are same in border-box.

mmikeyy:
Hi,

Thanks for merging the floats.

But the other change you're rejecting leaves a problem unsolved.

If you leave things as they are, the resizing logic for the height is as
follows:

1) get content height;
2) add deltaY
3) adjust height.

Note that in 3, we adjust 'height', and in 1 we obtained 'content height'.
They are the same only if box-sizing is content-box.

If box-sizing happens to be set to 'border-box', this logic makes no sense.
You're reading a parameter (content height), correcting for deltaY, and
writing the result to another parameter (outer height).

I'm not insisting because I'm some crazy lunatic! I just happened to use a
library with a css rule that set box-sizing to border-box and found that
the primeng dialogs could no longer be resized.

The library (primeng) should not go haywire just because the rule
"box-sizing: border-box" happens to be present in another library or in the
user's CSS...

Another way to circumvent this problem is to set the style "box-sizing:
content-box!important" on div.dialog-content in the dialog template. With
!important otherwise it could be overridden by an 'important' CSS rule. But
I found this less elegant.



Le lun. 21 nov. 2016 à 11:20, Çağatay Çivici <notifications@github.com> a
écrit :

PR #1382, opened 2016-11-21
Show active state on menuBar items that use routerLinks

mischkl:
Simple change to Menubar to make items that use routerLinks show the active state (the "ui-state-active" class is added) whenever their routes are activated.

### DIFF
--- a/components/menubar/menubar.ts
+++ b/components/menubar/menubar.ts
@@ -13,7 +13,7 @@ import {Router, RouterModule} from '@angular/router';
             <template ngFor let-child [ngForOf]="(root ? item : item.items)">
                 <li #item [ngClass]="{'ui-menuitem ui-widget ui-corner-all':true,'ui-menu-parent':child.items,'ui-menuitem-active':item==activeItem}"
                     (mouseenter)="onItemMouseEnter($event,item,child)" (mouseleave)="onItemMouseLeave($event,item)">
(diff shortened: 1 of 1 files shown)


cagataycivici:
Do you know why so many lines changed? Is there an issue with new line char?

mischkl:
Ah, sorry I think that was caused by the autoformatting in IntelliJ that strips the whitespace...

mischkl:
If you want I can re-submit without that

mischkl:
I reverting everything but the essentials, should be good to go.

mischkl:
Any news on this? Actually I am thinking it would make sense for a bunch of other menu components. This is actually a feature that is very much needed for our application, so it would be great it it could be added shortly.

Some background info: I tried adding the same functionality without modifying the components but found it pretty much impossible, since there is no simple way of getting at the menu item elements and adding classes conditionally from outside of the component.

I did notice that this is something that seems to be working in the Ultima theme Live Preview, which was one of the reasons why we opted for using PrimeNG in the first place. @cagataycivici could you speak to whether there is any way of getting this working, outside of submitting pull requests for every MenuItem-based component?

cagataycivici:
How about adding styleClass property to menuitem api so that you can provide it from the model? I guess in that case, you need a way to programmatically figure out if link is active.

cagataycivici:
Fixed via;

https://github.com/primefaces/primeng/issues/2393

PR #1392, opened 2016-11-22
markup weekend on calendar

mikosu3:
day of week property added to mark up saturday and sunday on datepicker.
for now, i display blue on saturday and red on sunday.

### DIFF
--- a/components/calendar/calendar.css
+++ b/components/calendar/calendar.css
@@ -243,4 +243,11 @@
 
 .ui-timepicker > div span {
     font-size: 16px;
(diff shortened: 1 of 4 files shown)


cagataycivici:
Can't really add colors to core.css, all colors must go in theme.css. So unfortunately can't accept in its current stage although idea is good.

mikosu3:
If i add all colors go in theme.css can you accept it?

Also, I would like to display the holidays in red too.
That is not too difficult.

PR #1398, opened 2016-11-22
Don't render upload and cancel buttons on file upload if auto="true"

maniacneron:
Don't render upload and cancel buttons on file upload if auto="true"


pmeems:
I agree this is unlogical to show the upload and cancel buttons when auto="true" or auto="auto".

+1

cagataycivici:
Merged manually.

PR #1412, opened 2016-11-23
Header checkbox may not always get checked

### LINKED ISSUE #1206: DataTable: Select all rows in a table
When I set `selectionMode="multiple"` on a column I get a checkbox in the table's header. It is used for selecting all rows. I noticed that if the table is paginated and there is more than one page, selecting all rows only selects the ones on the current page. Is there a way to select the total number of rows?

saverett:
Fixes #1206
Allows paged datatable header checkbox to behave correctly when programmatically selecting entire dataset


veplechuc:
@saverett Is this fix already out ?

PR #1414, opened 2016-11-24
Support HTML in growl message

ooboqoo:
Before change, some string like "first line. <br> and  here goes the second line." do not work as expected.
This revision brings the flexibility without security risks. (eg. 'first line <br> second line.<script>alert("lalala")</script>' comes with no alert dialogue.)


pkwiatko:
dadasd
adasd
adadad

PR #1418, opened 2016-11-24
Check if nested object is not null before try get value.

upmauro:
DATA

```
[
   {
      "id":"1",
      "relation":null
   }
]
```
PROBLEM

`<p-column field="item.relation.id"></p-column>`


cagataycivici:
Duplicate of #1412?

PR #1424, opened 2016-11-25
Fired the onChange event when spinner buttons pressed

Droxx:
the onChange event would emit on the input being changed. But not when the spinner buttons were pressed


cagataycivici:
Fixed manually, thank you.

PR #1431, opened 2016-11-26
Implemented New feature 1430

jpheintz:
Added new feature of displaying seconds in the calendar date time picker.


cagataycivici:
Thank you but we only accept defect fixes for PRs at the moment, we do the new features by ourselves. We'll consider your feedback when adding second support in near future, thank you.

PR #1440, opened 2016-11-28
getOffsetFunction doesn't account for scrolling

tritibo:
getOffsetFunction does not seem to take account of the vertical scroll while iterating parent elements.

Step to reproduce (in Chrome):
go to http://www.primefaces.org/primeng/#/tooltip
give an height to a parent tooltip element from console:
`document.getElementById('CONTENTSIDEindent').setAttribute('style', 'height: 500px; overflow: auto')`

scroll the page and try the tooltip

### DIFF
--- a/components/dom/domhandler.ts
+++ b/components/dom/domhandler.ts
@@ -298,8 +298,8 @@ export class DomHandler {
         let y = el.offsetTop;
 
         while (el = el.offsetParent) {
-            x += el.offsetLeft;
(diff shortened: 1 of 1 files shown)


cagataycivici:
This PR breaks existing functionality, so even without making the parent scrollable or changing the height, showing after scrolling the page breaks it. We need alternatives to handle these cases such as supplying the overflown container to the tooltip so we can do math based on it.

Issue #1451, opened 2016-11-29
Datatable with dynamic columns crashes with a "Cannot read property 'xxx' of undefined" for nested objects. 


bogandy99:
**I'm submitting a 
```
[x ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Datatable with dynamic columns crashes with a "Cannot read property 'xxx' of undefined" for nested objects. 
The following example crashes if the fieldvalue for the column is land.landName as land in the second json object is missing (undefined). 
Sample json:
[
{
      "id": 48,
      "ZIP": "555",
      "city": "city1",
      "land": {
        "id": -1908176136,
        "landName": "land-name"
      },
    },
   {
      "city": "city2",
      "ZIP": "111",
      "id": 49
    }
]

**Expected behavior**
Not to crash when setting the fieldvalue as land.landName.

**Minimal reproduction of the problem with instructions**
*STEPS TO REPRODUCE* 
A simple datatable:
 <p-dataTable
      [value]="zipList"
      scrollable="true" scrollHeight="70%" scrollWidth="75%"
      selectionMode="single" [(selection)]="selectedZip" (onRowDblclick)="rowDblClick($event)"
      [contextMenu]="cm">
      <p-column *ngFor="let col of cols" [field]="col.field" [header]="col.header" [sortable]="col.sortable"></p-column>
        <footer><p-tabMenu [model]="items" [activeItem]="activeItem"></p-tabMenu></footer>
    </p-dataTable>
... 
    this.cols = [{field: 'city', header: 'City', sortable: true},
        {field: 'land.landName', header: 'Land', sortable: true},
        {field: 'ZIP', header: 'ZIP', sortable: true}];

**What is the motivation / use case for changing the behavior?**
Improve primeng datatable component so that it supports nested objects. 

**How to fix**
It seems that the issue is in the datatable.ts file, method resolveFieldData:
 resolveFieldData(data: any, field: string): any {
        if(data && field) {
            if(field.indexOf('.') == -1) {
                return data[field];
            }
            else {
                let fields: string[] = field.split('.');
                let value = data;
                for(var i = 0, len = fields.length; i < len; ++i) {
                    value = value[fields[i]];
                }
                return value;
            }
        }
        else {
[...]

PR #1452, opened 2016-11-29
fix for #1451

thrallala:
A proposed fix for issue [https://github.com/primefaces/primeng/issues/1451](url)


cagataycivici:
Isn't this duplicate of #1182?

PR #1459, opened 2016-11-30
Fixed #1310 (multi line support confirm dialog)

### LINKED ISSUE #1310: Confirm Dialog message with multiple lines
Apparently `\n` or `<br>` are not permitted with message area of confirm dialog.

I believe that would be a nice addition to be able to define line endings for `message` text.

upmauro:
Add `<pre></pre>` tag on message to allow multiple line using ` \n`.


cagataycivici:
From;

https://css-tricks.com/snippets/css/make-pre-text-wrap/

So how does this work? pre doesn't seem to wrap by default?

cagataycivici:
Closing due to no response.

Issue #1434, opened 2016-11-26
Paginated, Filtered, Lazy Loaded DataTable bug & fix


davisruk:
**I'm submitting a ...**  (check one with "x")
```
[ X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**

DataTable with Pagination, LazyLoading and Filter enabled spawns infinite onLazyLoad callbacks when a filter has a value, there is more than one page and the user clicks a page number. Callback initially requests the correct page of data but all other callbacks request page 1.
The reason for this is that the DataTable ngDoCheck function is incorrectly setting lazyFilteredByUser to false when lazyloading is enabled and the filter has a value. In lazy loading mode the application is responsible for filtering and retrieving the correct data for the requested page. Current code is below; further posts on the PrimeNG forum:
[http://forum.primefaces.org/viewtopic.php?f=35&t=48297&sid=bae2acdf7ca3c456a85c8fc5bccc7ca6](url)
`   DataTable.prototype.ngDoCheck = function () {
        var changes = this.differ.diff(this.value);
        if (changes) {
            this.dataChanged = true;
            if (this.paginator) {
                this.updatePaginator();
            }
            if (this.hasFilter()) {
                if (this.lazy) {
                    //prevent loop
                    if (this.lazyFilteredByUser)
                      this.lazyFilteredByUser = false;
                    else
                      this.filter();
                }
                else {
                    this.filter();
                }
            }
            if (this.stopSortPropagation) {
                this.stopSortPropagation = false;
            }
            else if (!this.lazy && (this.sortField || this.multiSortMeta)) {
                if (this.sortMode == 'single')
                    this.sortSingle();
                else if (this.sortMode == 'multiple')
                    this.sortMultiple();
            }
            this.updateDataToRender(this.filteredValue || this.value);
        }
    };
`
**Expected behavior**

The filtering should be the responsibility of the application for lazy loading. Changing NgDoCheck to the following produces the expected behaviour:
`   DataTable.prototype.ngDoCheck = function () {
        var changes = this.differ.diff(this.value);
        if (changes) {
            this.dataChanged = true;
            if (this.paginator) {
                this.updatePaginator();
            }
            if (this.hasFilter()) {
                if (this.lazy) {
[...]

PR #1466, opened 2016-11-30
Fix lazy loading/filter loop #1434

mwaltersva:
Fix for #1434 as described by @davisruk


cagataycivici:
With this filteredByUser is always true, I think we need an alternate fix.

PR #1471, opened 2016-12-01
Fix bug: PanelMenu not opening external links

gilhanan:
PanelMenu allways fire event.preventDefault(), so anchor tag can't href to external tag


cagataycivici:
Thank you.

Issue #1454, opened 2016-11-29
Seconds support for Calendar


cagataycivici:
Add seconds support for calendar time picker.

kiran3807:
Hi, Any update regarding this issue ?

cagataycivici:
Planned for 2.0.

PR #1477, opened 2016-12-01
Added the support for seconds in calender-time-picker

kiran3807:
Hi,
This is in response to : https://github.com/primefaces/primeng/issues/1454

As mentioned now we can set seconds in the calender.
Have tested for the following cases =>

1. Wether the input text is changing when we click on increment/decrement buttons
2. When input text is being changed wether the change is being reflected in the seconds area near increment/decrement buttons

3. Correct time is being emmited for onSelect output.

time format for days months and years is same as it was before. However for hours and minutes it will be => hours : minutes : seconds. If not in this format existing exception ("Invalid time") is thrown

Please suggest changes and improvements if any :)

Thanks

### DIFF
--- a/components/calendar/calendar.css
+++ b/components/calendar/calendar.css
@@ -226,6 +226,10 @@
     margin-left: 0px;
 }
 
+.ui-timepicker > .ui-second-picker {
+    margin-left: 0px;
+}
+
 .ui-timepicker > .ui-separator {
     margin-left: 0px;
     min-width: 12px;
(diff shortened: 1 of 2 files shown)


cagataycivici:
For such cases with significant changes, our policy is to implement the features by ourselves following our own processes. Due to this reason, I'm afraid I can't accept it however we've plans to do it in 2.0.

kiran3807:
I see. Thanks for the update though

PR #1488, opened 2016-12-02
Fixed #1484 Update documentation 

### LINKED ISSUE #1484: [TreeTable] Documentation
There's a problem with the docs in `showcase/demo/treetable/treetabledemo.html`

On lines 477 and 480, where it says `let-col #node="rowData"` it should instead read `let-node="rowData"`.

If you do actually use the form currently shown, you get an error - `There is no directive with "exportAs" set to "rowData"`

dzhavat:
Fixed #1484


cagataycivici:
Merged manually

Issue #1332, opened 2016-11-17
FileUploader onBeforeUpload event is triggered after file is appended to FormData


radarsu:
Some back-end frameworks require post data to be appended before files.

So, this code in FileUpload component, in method upload():
```
upload() {
    this.msgs = [];
    let xhr = new XMLHttpRequest(),
        formData = new FormData();

    for (let i = 0; i < this.files.length; i++) {
        formData.append(this.name, this.files[i], this.files[i].name);
    }

    xhr.upload.addEventListener('progress', (e: ProgressEvent) => {
        if (e.lengthComputable) {
            this.progress = Math.round((e.loaded * 100) / e.total);
        }
    }, false);

    xhr.onreadystatechange = () => {
        if (xhr.readyState == 4) {
            this.progress = 0;

            if (xhr.status == 200)
                this.onUpload.emit({
                    xhr: xhr,
                    files: this.files
                });
            else
                this.onError.emit({
                    xhr: xhr,
                    files: this.files
                });

            this.clear();
        }
    };

    xhr.open('POST', this.url, true);

    this.onBeforeUpload.emit({
        'xhr': xhr,
        'formData': formData
    });

    xhr.send(formData);
}
```
Should be updated to this:
```
upload() {
    this.msgs = [];
    let xhr = new XMLHttpRequest(),
        formData = new FormData();

    this.onBeforeUpload.emit({
        'xhr': xhr,
        'formData': formData
    });

    for (let i = 0; i < this.files.length; i++) {
        formData.append(this.name, this.files[i], this.files[i].name);
    }

    xhr.upload.addEventListener('progress', (e: ProgressEvent) => {
        if (e.lengthComputable) {
            this.progress = Math.round((e.loaded * 100) / e.total);
        }
    }, false);

    xhr.onreadystatechange = () => {
        if (xhr.readyState == 4) {
            this.progress = 0;

            if (xhr.status == 200)
                this.onUpload.emit({
                    xhr: xhr,
                    files: this.files
                });
            else
                this.onError.emit({
                    xhr: xhr,
                    files: this.files
                });

            this.clear();
        }
    };

    xhr.open('POST', this.url, true);

    xhr.send(formData);
}
```

cagataycivici:
Please do a PR and we'll review.

svideau:
This modification seems to break #867 by opening the xhr after the onBeforeUpload(), preventing to add headers.

radarsu:
This should do the job:
```
upload() {
[...]

PR #1489, opened 2016-12-02
fileupload onBeforeUpload should be fired before appending files to FormData

radarsu:
Issue https://github.com/primefaces/primeng/issues/1332

In documentation onBeforeUpload	 event is described:
`Callback to invoke before file upload begins to customize the request such as adding header and post parameters.`

But some frameworks require post parameters to be appended to FormData before files.

When onBeforeUpload event emits xhr and FormData, there is no way to "prepend" any data to FormData object, so some backend frameworks will return errors like this:
``` Unable to expose body parameter `exampleParameter` in streaming upload!
Client tried to send a text parameter (exampleParameter) after one or more files had already been sent.
Make sure you always send text params first, then your files.
(In an HTML form, it's as easy as making sure your text inputs are listed before your file inputs.```

### DIFF
--- a/components/fileupload/fileupload.ts
+++ b/components/fileupload/fileupload.ts
@@ -170,16 +170,21 @@ export class FileUpload implements OnInit,AfterContentInit {
     onImageLoad(img: any) {
         window.URL.revokeObjectURL(img.src);
     }
-    
+
     upload() {
         this.msgs = [];
         let xhr = new XMLHttpRequest(),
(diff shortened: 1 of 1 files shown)


robkrause:
This way I can't call setRequestHeader() anymore since it has to be done after xhr.open()

MarkWilds:
uhm I think they should split this up in two events because now I can't call setRequestHeader() anymore

PR #1491, opened 2016-12-03
Fade animations fails on some locales

sittingbool:
I had some issues with fading. i found out that in some of my browsers and in electron the value of opacity will have a comma instead of a dot as float separator. thats common for germany, because we write floats with comma instead of dot. maybe the local settings of the browser cause this issue but this code version is more stable in any case.

### DIFF
--- a/components/dom/domhandler.ts
+++ b/components/dom/domhandler.ts
@@ -170,11 +170,13 @@ export class DomHandler {
         element.style.opacity = 0;
 
         let last = +new Date();
+        let opacity = 0;
         let tick = function () {
-            element.style.opacity = +element.style.opacity + (new Date().getTime() - last) / duration;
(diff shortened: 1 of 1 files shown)


sittingbool:
Would be great if you could merge this for the next version. Otherwise we can not update anymore. We had to fix this locally so far because otherwise growl would not work anymore and get stuck with a weird and too low opacity value. But sending a zip package around the dev team is not a solution ;).

cagataycivici:
Thank you.

mbaer3000:
The issue is still present for me, commit f5ad386 does not fix the Chromium issue which results in element.style.opacity returning a string with a comma instead of a dot – as in "0,15" instead of "0.15" – when using a German "de" locale. Cf. https://bugs.chromium.org/p/chromium/issues/detail?id=33263.

PR #1513, opened 2016-12-07
Fixed #1107 - Min height when label is null

### LINKED ISSUE #1107: p-dropdown blank entry compressed
When a p-dropdown entry label is blank (empty string) it appears vertically compressed:

image

It should have the same height as the rest of the entries.

Using PrimeNG 1.0.0 Beta 19, Omega Theme, and Chrome 53.0.2785.143

upmauro:
Fixed #1107 min-height when the ui-dropdown-item have a null label.

**min-height: 1em** is the best way, because "em" is "Relative to the **font-size of the element** (2em means 2 times the size of the current font)" 

Preview:

image

The same solution has applied on select2 in the past https://github.com/select2/select2/commit/4530e74e9536adda1c5b637c3e43d7bf607bb6ab

### DIFF
--- a/components/dropdown/dropdown.css
+++ b/components/dropdown/dropdown.css
@@ -72,6 +72,7 @@
     margin:1px 0;
     padding:3px 5px;
     text-align:left;
+    min-height: 1em;
 }
 
 .ui-dropdown-panel .ui-dropdown-item-group {


cagataycivici:
This looks a bit weird to me, I'll use the same approach we had for ui-dropdown-label-empty instead.

cagataycivici:
screen shot 2016-12-20 at 10 33 18

cagataycivici:
Fixed via #1626

PR #1514, opened 2016-12-07
Fixed #1505 - acceptVisible and rejectVisible not changing value

### LINKED ISSUE #1505: confirmDialog
```
```

**Plunkr Case (Bug Reports)**
it is the plunkr.
http://plnkr.co/edit/1E8iYG?p=preview

**Current behavior**

first, i click the confirm button , then i click the delete button .
the confirmDialog by the delete button  lost a button.

here is the bug.
image

**Expected behavior**

i click the confirm button  ,the confirmDialog show one button .
i click the delete button , the confirmDialog show two button

* **Angular version:** 2.2.4

* **PrimeNG version:** v1.0.1 beta 20

upmauro:
acceptVisible and rejectVisible is initialized just by declaration, so, if you change the value first time to false, the (ternary if) ever fail and the value will never change again.


cagataycivici:
Shouldn't we also check undefined?

PR #1519, opened 2016-12-07
ScrollTable does not display default filter values

Krustie101:


cagataycivici:
Merged manually, thanks.

PR #1537, opened 2016-12-08
Fix: Dropdown initial value not reflected in model / Added default value feature

spike83:
- added defaultValue parameter to dropdown
- avoid display the first item with a null value in ngModel

### DIFF
--- a/components/dropdown/dropdown.ts
+++ b/components/dropdown/dropdown.ts
@@ -74,6 +74,10 @@ export class Dropdown implements OnInit,AfterViewInit,AfterViewChecked,DoCheck,O
     
     @Input() appendTo: any;
     
+    @Input() defaultValue: any;
+
+    @Input() formControlName: string;
+
     @Output() onChange: EventEmitter<any> = new EventEmitter();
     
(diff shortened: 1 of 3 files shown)


cagataycivici:
Default value doesn't sound good, it should get it from ngModel and if not should select the first option. Also formControl name is an attribute of reactive forms. Please file an actual issue with a case to replicate and we'll review.

spike83:
@cagataycivici Thanks for your review and comment.
I opend the issue #1585. You can simply reproduce the problem in showcase by removing the null option from this.cities array. In this case the dropdown will show 'New York' and the value in model is null.
In the issue is a plunker.

So, to avoid this and other problems like not matching values coming from the model, the drop down component should override the incoming value with some kind of ‘default value’.

In my opinion the first option is not every time the best default option, so I made it possible to provide a default value.

It’s absolutely right, the `formControl ` is from reactive forms. As you can see in my pull request, I want to check if a matching option is set when `writeValue` is called but in case of Reactive Forms it is called once and when the value is provided via ngModel it it's called a second time. So, I was looking for a way to find out if the component is used with reactive forms or not.

What should be improved to become this sound good for you?

PR #1543, opened 2016-12-09
Add 'timeless' attribute to remove time from dates in calendar

amilcar-calles:
This feature will allow to select a date only with no timezone issues.

For example instead 

"2016-12-06T06:00:00.000Z"

we will get:

"2016-12-06T00:00:00.000Z" 

This will allow to send dates without timezone info to the backend, in C# for example if I send the first option and the property is dateOnly it will throw an error because the first option has "the time part" (6:00:00)

--- Maybe to be more concise, the goal should be to add an extra property named 'dateOnly' (boolean), and do that logic inside the conditional date

### DIFF
--- a/components/calendar/calendar.ts
+++ b/components/calendar/calendar.ts
@@ -464,6 +464,20 @@ export class Calendar implements AfterViewInit,OnInit,OnDestroy,ControlValueAcce
     updateModel() {
         if(this.dataType == 'date')
             this.onModelChange(this.value);
+        else if (this.dataType == 'dateOnly')
+        {
+            var value = new Date(
+                Date.UTC(
+                    this.value.getFullYear(),
+                    this.value.getMonth(),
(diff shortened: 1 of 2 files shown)


cagataycivici:
The PR misses documentation, please also include it.

amilcar-calles:
Change the first version of this feature, from [dataType]="dateOnly" to a separate attribute [timeless]="true" for clearer separation of functionality.

Added documentation to calendardemo.html (attribute list)

amilcar-calles:
This is incomplete, I will start a new PR with the new version PrimeNG and the full proposal for this feature.

The current problem is the the "timeless" feature only works on selection (by click), but if I supply a timeless date to the component, the date on the model is correct, but the date displayed on the calendar is wrong (because a UTC date gets converted to Local date).

PR #1562, opened 2016-12-12
check [null] value of the field with nested object

windwolfreal:
PR #1182 chack the [undefined] value of the field, but not for [null]. this commit fix it.


cagataycivici:
How about;

```
if (value == null){
```

Afaik, undefined == null is also true so we can both check for null and undefined with a single statement.

windwolfreal:
@cagataycivici you are right, i will close this PR.
but i found something strange. the if condition in my node_modules/primeng/datatable/datatable.js is 
[ value === undefined ], but [ value == undefined ] in the source of datatable.ts host on github. 
What are the possible causes of this result?

cagataycivici:
I'm not sure about why that happens, nevertheless I've committed the change now via #1579. Thank you.

PR #1587, opened 2016-12-13
Add input property 'cellIsEditiable' in DataTable, which verifies editability of a cell

easyping:
With the specified function of the "cellIsEditable" property, cells can
be locked for editing.
Thus, depending on the row data, it is possible to lock individual cells
for editing.


cagataycivici:
For such cases with significant changes, our policy is to implement the features by ourselves following our own processes. Due to this reason, I'm afraid I can't accept it however we've plans for improving editing in 2.0 with custom component support and features as you've mentioned.

Issue #1482, opened 2016-12-02
Chips improvements


FallenRiteMonk:
Would it be possible to extend the chips module by following functionality:
 - custom seperator keys (ex.: use space to enter a chip)
 - custom validators (ex.: a chip may not contain sybols like !"§$%&...)
 - transformations on input (ex.: make string lowercase on input)
 - prevent duplicates (same inputs are ignored)
and change/fix the following:
 - entering a blank chip results in chip height not being same as the rest of chips
 - remove the x to delete a chip if disabled is activated

wujitouch:
For the blank chips issue override the .ui-chips-token-label display value, like so:

.ui-chips-token .ui-chips-token-label {
    display: inline-block;
}

wujitouch:
https://github.com/primefaces/primeng/pull/1588

cagataycivici:
Please create a separate ticket for each one as the current ticket is too broad, better to divide into smaller pieces to manage easier.

PR #1588, opened 2016-12-13
#1482 Added validation pattern, allowInvalid, and allowDuplicate

wujitouch:
Added some features requested in #1482


wujitouch:
Added index to pTemplateWrapper to expose it in template

cagataycivici:
For such cases with significant changes, our policy is to implement the features by ourselves following our own processes. Due to this reason, I'm afraid I can't accept it however, please create separate issues as I've mentioned and we'll take care of it as the ideas are quite good to improve the component.

wujitouch:
The changes were actually very minimal. If you have a style guide or something that would allow me to follow your "own processes" I'd be more than happy to change it. I'm new to NG2 so learning your coding style would also help me :)

Thanks for the great UI framework!

PR #1601, opened 2016-12-15
Keyboard accessibility for Row Expander

digitalcraftco:
Same idea as the new expandableRowGroups functionality. Puts table rows into the tabIndex of page. The enter button now toggle rows. Add preventDefault to toggleRows() function.


cagataycivici:
This was on our todo list, thanks.

PR #1604, opened 2016-12-16
I suggest to revise AngularJS 2 to Angular 2

hryktrd:
I would like to suggest to revise AngularJS 2 to Angular 2


derehls:
I would also suggest do drop the "2" since angular will start to deliver semantic releases beginning next year beginning with v4

hryktrd:
Today when I saw README.md, that was revised to Angular 2, so I close this pull request.

PR #1610, opened 2016-12-17
Added a new event to file upload

MarkWilds:
Fileupload component didn't allow it's users to set headers anymore after 1.1.0. because of the moved onBeforeUpload event.
Added onBeforeSend event after the open method for the XMLHttpRequest to allow setting of headers once again.


cagataycivici:
Thank you!

PR #1640, opened 2016-12-20
Call FileUpload onSuccess on success response only

EddyP23:
OnUpload gets called when any success response is called (200-299)


cagataycivici:
Thanks

PR #1641, opened 2016-12-20
Implementing the blur event on the autocomplete component

odedolive:
Adding the "blur" event to the autocomplete component.
Useful when implementing custom validation logic and you need to know when focus is lost on certain controls.


cagataycivici:
Thanks, this is already available in recent releases.

PR #1642, opened 2016-12-20
Implementing the blur event on the inputmask component

odedolive:
Adding the "blur" event to the inputmask component.
Useful when implementing custom validation logic and you need to know when focus is lost on certain controls.


cagataycivici:
Rename to onBlur please due to our convention.

4kochi:
Is there a date when this will be released?

PR #1643, opened 2016-12-20
Inputmask autoclear bug fix

odedolive:
When setting the autoClear value to "false" we want to preserve the value entered by the user.

### DIFF
--- a/components/inputmask/inputmask.ts
+++ b/components/inputmask/inputmask.ts
@@ -493,7 +493,7 @@ export class InputMask implements AfterViewInit,OnDestroy,ControlValueAccessor {
         if (allow) {
             this.writeBuffer();
         } else if (lastMatch + 1 < this.partialPosition) {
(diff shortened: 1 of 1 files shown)


cagataycivici:
Had to revert because it broke the functionality where on blur incomplete value is cleared.

odedolive:
I'm probably missing something then.
We want the value to stay in the textbox if the mask is incomplete, so that we can give our users the opportunity to fix it (instead of having to re-type it).
Sometimes people forget to enter one or two characters, we don't want them to have to re-enter it all when they tab out accidentally.

Whipgit:
Am I correct in the following:
Setting autoClear="false" or autoClear="true" makes no difference?

cagataycivici:
Angular syntax kinda requires [autoClear]="false" to actually set false if default value of the property is true. Otherwise it sets autoClear as string "false" not boolean false.

Whipgit:
Thank you very much @cagataycivici

PR #1653, opened 2016-12-22
I Bracketed toggleable with []

hryktrd:
toggleable must be Bracketed with [] Otherwise toggleable="false" was not effectable.
[toggleable]="false" was effectable.


cagataycivici:
I understand but we can avoid a binding here with toggleable="true" so would like to keep it as it is. This is an Angular syntax thing, not PrimeNG Related, demos have toggleable enabled so no need for binding although I agree with your point.

PR #1654, opened 2016-12-22
Make the calendar Saturday and Sunday different colors 

mikosu3:
With the same content as Pull #1392, color was added to all theme.css.
I would like to use in my project, please accept by all means.


cagataycivici:
Due to amount of changes, we're unable to accept due to our PR policy. Thank you.

PR #1665, opened 2016-12-22
Row selection should be on a tap event instead of a touchstart

samuellaplante:
If the event is on a touchstart the handler is called as soon as the user touch the screen. This may be not the expected behaviour on mobile, because if the user is trying to scroll the content of the table, the event is fired to soon and this might not be the intention of the user.


samuellaplante:
#1666

cagataycivici:
This gives an error like hammer.js not loaded, it may cause similar problems to users.

cagataycivici:
Will reconsider if more users request it.

sgomanf:
same issue, as soon as the user touch the screen, it select the row

PR #1674, opened 2016-12-23
DatePicker : 12 hour format validation error

### LINKED ISSUE #1673: DatePIcker:  12 hourformat validation error
When hourformat set to "12" and selected date is in "PM" , change the hour to 12 in  inputfield and the date changes to AM and the date in the input filed show PM,

When the time is in "AM" and edit hour to 12 it sets as a 12"00 AM and the value is 00:00.

When the date is set to 12:00PM, change the am/pm to "AM" and date set to 12:00 am

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

**Current behavior**

**Expected behavior**

* **Angular version:** 2.0.X

* **PrimeNG version:** 2.0.X

harilal:
###Defect Fixes
#1673 
Added proper validatio in Am/PM change. In valid date will fire am error.
When the time is 12:00 PM and change to AM using the am/pm selector will change it to 00:00 AM.

###Feature Requests


cagataycivici:
There is a conflict now, what is the actual problem? I don't think 00 am is valid.

harilal:
@cagataycivici  the issue is with 12 hourformat. I will test it with new version and resolve the conflict for the new version. The main issue is some thing like if you choose an hour more than 12 in the edit input it should throw an error instead it gives a valida date with the time of 12:00. Other is if you change the date in the edit field when the time is in "PM" and you changed the hour to 12 it will switch to AM in UI and not updating the edit field. 
If the date in edit field is wrong it should throw an error instead of a valid date.

harilal:
@cagataycivici the conflict has been resolved. 
The major issue is ,the user edit the hour of the time to 12 when the time is in PM it's automatically changes to AM . Actually there is no 12:30 AM. Once its past 12 , its become 00:30. Or else it should throw an invalid date. And also when an hour changed to greater than 12 it should throw an invalid date, since its an 12 hour format.

harilal:
I will update and fix it with latest one and initiate new pull request.

Issue #1694, opened 2016-12-27
FileUpload. There is no validation for accept types


Invizible:
Currently we can set allowed types using 'accept' attribute, but component actually does not perform any validation and we can upload any file.

kt215:
Do you mean when you select "All Files" in the file dialog?

I encountered this issue too, and I validate the uploaded file contents of the server side using binary validation.

What exactly do you want primeng to do on the client side?

Invizible:
@kt215 , Yes you are right, if we select "All Files" we can upload everything.
You can take a look at my pull request. We can do file type check the same way we have validation for file size.

bhamde:
any updates on when that change is going in? quite an important bug fix

cagataycivici:
Merged

EddyP23:
@cagataycivici 

Current implementation is very confusing and actually wrong.

Say, I pass `accept='image/jpeg,image/bmp'`. Then method `isFileTypeValid` ([here](https://github.com/primefaces/primeng/blob/4.0.0-rc.3/components/fileupload/fileupload.ts))
does very strange things.

If it finds a * in accepted type (say `image/jpeg` or `image/*`), it checks whether first part of the mime-type  `image` matches the mime type of the file.
If it doesn't find a star, then it compares file extension (e.g. `.jpg`) to mime type (e.g. `image/jpeg`) which is completely wrong.

Could you please reopen and fix it?

AlissonRS:
When I click the draggable area, the browser window to select files will be pre-loaded correctly based on the `accept` input, but if I just change to "All Files" or just drag a file, the component won't prevent those files.

Currently using: `"primeng": "^7.1.3"`. As an example, I'm using it like this:

`<p-fileUpload accept=".psd,.idml" (onUpload)="onUpload($event)">`

PR #1695, opened 2016-12-27
FileUpload. Added file type validation 

Invizible:
Fix for #1694


pauly815:
Agreed, can this get merged?

cagataycivici:
PR fails in cases like accept=".html" and you can't upload an html file. Will change that part.

PR #1713, opened 2016-12-31
Update tree.ts to allow bypass propagation

ADefWebserver:
Allows bypass propagation by using propagateSelection="false" in the p-tree tag. If property is not supplied it will default to normal propagation. This was discussed in the forums at: http://forum.primefaces.org/viewtopic.php?f=35&t=48730&p=150623


cagataycivici:
Will review, thank you.

cagataycivici:
Fixed via;

https://github.com/primefaces/primeng/issues/1849

Thank you.

PR #1723, opened 2017-01-04
Fix year range in calendar not allowing dynamic ranges fixes #1125

### LINKED ISSUE #1125: Calendar's 'Relative to today's year' yearRange doesn't work
This simply prints all the numbers provided in range "-nn:+nn" in  a loop and doesn't give years relative to today's year i.e. for a year range "-0:+5" it simply prints the numbers from 0 to 5 and not years. See preview:

calendar

aaraggornn:
Allows the calendar to assign the yearRange to a dynamic range ("-nn:+nn") or ("c-nn:c+nn") as specified in the docs. Should fix #1125


cagataycivici:
Thank you but right now until more users require it, we'll only keep nnnn:nnnn format. Fixed the doc according to this. If more users require dynamic ranges, we can add this for sure. Right now we're focused on maintenance rather than new features.

armved:
Can we please include this feature?

PR #1727, opened 2017-01-04
Single FileUpload still behaves as multiple

### LINKED ISSUE #1698: File Upload Component Fails to Prevent Multiple Uploads when 'Multiple' Attribute is Omited
```
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/nRYDOJDkQsyA8ugozFDc?p=preview

**Current behavior**

The file upload component currently controls the number of files which can be uploaded through the `multiple` attribute. When the attribute is omitted, this works correctly in the file explorer, and only allows one file to be selected, but does not prevent the user from selecting more than one file if they open the explorer again.
**Expected behavior**

See plunkr above which includes instructions.

It is very common for file upload components to allow only one file upload at a time, or even a fixed number. This is not possible with the current component without intercepting the select event and manually removing files.

* **Angular version:** 2.0.X

* **PrimeNG version:** 1.0.1

* **Browser:** all 

 
* **Language:** all 

sittingbool:
In single selection mode the image should be set inside a single-item array instead of adding it to the existing images.

###Defect Fixes

###Feature Requests

### DIFF
--- a/components/fileupload/fileupload.ts
+++ b/components/fileupload/fileupload.ts
@@ -140,7 +140,10 @@ export class FileUpload implements OnInit,AfterContentInit {
                 if(this.isImage(file)) {
                     file.objectURL = this.sanitizer.bypassSecurityTrustUrl((window.URL.createObjectURL(files[i])));
                 }
-                
+                if (!this.multiple) {
+                    this.files = [ files[i] ];
+                    continue;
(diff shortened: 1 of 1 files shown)


cagataycivici:
Resolved in a different way, thank you.

PR #1728, opened 2017-01-04
Update calendar.ts

victos:
To avoid memory leaks in big applications


cagataycivici:
Thank you, fixed via https://github.com/primefaces/primeng/issues/1757

PR #1736, opened 2017-01-05
Expose `DomHandler` in library API.

memee:
`DomHandler` was a missing provider when one wanted to subclass and
modify some of primeng components. When we export it in public API
it'll be possible to subclass any component or subcomponent.


cagataycivici:
Shouldn't we export it from primeng?

```
import {DomHandler} from 'primeng/primeng';
```

memee:
@cagataycivici there's 

`export * from './components/common/api';`

in https://github.com/primefaces/primeng/blob/master/primeng.ts
which actually is doing it.

If we think of the best place it should be put in - I personally don't know. For me `common/api` was the first choice.

PR #1737, opened 2017-01-05
added keyboard and screenreader support to TreeTable

nhnb:
This PR enables TreeTables to be used via keyboard and screen readers for improved accessibility. 

It converts the expand/collapse toggle from a span-tag into a button tag. Adjusted css rules ensure that there is no visible difference. It adds two new attributes `labelExpand` and `labelCollapse` with appropriate defaults ("Expand" and "Collapse") in order to indicate the current state to screen readers.

### DIFF
--- a/components/treetable/treetable.css
+++ b/components/treetable/treetable.css
@@ -41,7 +41,8 @@
 }
 
 .ui-treetable .ui-treetable-toggler {
-    display: inline-block;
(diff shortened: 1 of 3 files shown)


cagataycivici:
Thank you, I'll merge but replace button with anchor since we use anchors mostly for accessibility.

PR #1748, opened 2017-01-06
Fixed #1655

### LINKED ISSUE #1655: Templating fails when using any *ng directives
```
[ X] bug report => Search github for a similar issue or PR before submitting
```

**Plunkr Case (Bug Reports)**
**Plunkr to demonstrate the bug with PrimeNG 1.1.1:** https://plnkr.co/edit/BQAjYy?p=preview

**Current behavior**
If `*ngIf` is used in a `<template>` inside `<p-dataList>` (and `p-orderList`, and probably other components that iterate over a template), the render of the initial data seems correct. However, adding new items to the list causes the template rendering to fail.

**Expected behavior**
Template should render correctly when `*ngIf` is used in the template.

Plunker at https://plnkr.co/edit/BQAjYy?p=preview reproduces the issue.

* **Angular version:** Plunkr uses latest from npm.

* **PrimeNG version:** 1.1.1

* **Browser:** all
 
* **Language:** all

Prontsevich:
###Defect Fixes
#1655

### DIFF
--- a/components/autocomplete/autocomplete.ts
+++ b/components/autocomplete/autocomplete.ts
@@ -92,7 +92,7 @@ export class AutoComplete implements AfterViewInit,DoCheck,AfterViewChecked,Cont
     
     @Input() multiple: boolean;
     
-    @ContentChild(TemplateRef) itemTemplate: TemplateRef<any>;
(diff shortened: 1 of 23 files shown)


cagataycivici:
We have pTemplate for such cases so can't add the PR as a result, but looking at the issue.

Issue #1646, opened 2016-12-20
Improve Growl long message handling


FallenRiteMonk:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
With a longer detail text in a growl, the ```ui-growl-message``` div gets pushed under the ```ui-growl-image fa fa-2x fa-close``` span:
![false](https://cloud.githubusercontent.com/assets/8779846/21369478/6b9e1b74-c707-11e6-82e7-c829454f0691.png)

**Expected behavior**
The div should stay in the same position it is with a short detail, just with two lines
Example short detail:
![right](https://cloud.githubusercontent.com/assets/8779846/21369508/96c90728-c707-11e6-9711-a4f5a7f5855b.png)
As it used to be before 1.1 (Using 1.0.1 it works fine)

**Minimal reproduction of the problem with instructions**
To reproduce just show a growl with a bit longer text (the exact character count is not known)
ex.: ```messageList.push({ severity: 'warn', summary: 'Error Message', detail: 'Validation failed Validation failed Validation failed' });```

* **Angular version:** 2.0.X
2.2.3

* **PrimeNG version:** 2.0.X
1.1.0

Skuriles:
Same problem here with Angular 2.4.2 and primeng 1.1.4!
Think the title and the message should be splitted and not packed in "ui-growl-message"

cagataycivici:
Growl is removed in favor of Toast.

PR #1764, opened 2017-01-09
Fix for growl messages with long detail bug #1646

cbterceiro:
Fixes growl component. Messages with longer content will not push the
title down with it

### DIFF
--- a/components/growl/growl.css
+++ b/components/growl/growl.css
@@ -31,8 +31,8 @@
 
 .ui-growl-title {
 	font-weight: bold;
-	padding: 0 0 .5em 0;
-	display: block;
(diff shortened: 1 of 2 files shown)


cagataycivici:
What is the actual problem, provide a sample to reproduce please, I'd rather not change the dom structure as it will break other things like themes, css selectors.

cbterceiro:
The problem is described on issue #1646, a growl message with long text in detail gets pushed down bringing the title with it. I believe this is not the expected behavior.

Here is a print screen from the issue described #1646.

image

cagataycivici:
I see, we can fix it with plain css then such as; 1848

Issue #1769, opened 2017-01-09
Editor Toolbar Accessibility


martine-dowden:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**

The buttons in the editor toolbar do not have any content or label making them inaccessible per the following specifications:
* WCAG success Criteria: 4.1.2 Name, Role, Value
* Section 508 Guideline 1194.22(a) Text equivalent for non-text element.

**Expected behavior**

Passes WCAG success Criteria and Section 508 Guidelines for web accessibility.

**Minimal reproduction of the problem with instructions**

  1. Load an accessibility evaluation tool such as:
  google chrome extension: Accessibility Developer Tools (https://chrome.google.com/webstore/detail/accessibility-developer-t/fpkknkljclfencbdbgkenhalefipecmb?hl=en)
  or
  google chrome extension: Axe (https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd?hl=en-US)
  2. Run the tool on a page containing the editor such as http://www.primefaces.org/primeng/#/editor.
  3. The report will show the error being generated by the buttons.

**What is the motivation / use case for changing the behavior?**

The motivations are:
- making the editor more accessible to users using assistive technologies such as screen readers
- being able to use the editor in projects requiring compliance to section 508 and/or WCAG 2.0

**Please tell us about your environment:**

Operating System: Mac
IDE: Atom

* **Angular version:** 2.0.0

* **PrimeNG version:** 1.0.0-beta.20

* **Browser:** All
 
* **Language:** TypeScript

* **Node (for AoT issues):** `node --version` =   6.0.41

PR #1770, opened 2017-01-09
added aria labels to Editor buttons (Issue #1769)

martine-dowden:


cagataycivici:
Merged manually as line breaks are incompatible. Thank you.

PR #1785, opened 2017-01-11
Improved Dropdown render performance

variable-content:
Drastically improves initial render times for dropdown component when there are a large number of dropdowns on the page (>100).

In a test rendering 100 dropdowns render speed dropped from >3seconds to render, to less than 300ms.

Fix contains two parts; only rendering the dropdown list panel to the DOM when needed, instead of eagerly  rendering it for every dropdown, and also optionally removing the hidden accessibility select list when it's not needed.

Fixes remaining issues for https://github.com/primefaces/primeng/issues/1749


variable-content:
@cagataycivici @Mrtcndkn Is is possible to get this reviewed please?

cagataycivici:
enableAccessibility is something that must always be true but I'll check the panelVisible improvement.

variable-content:
Thanks @cagataycivici, appreciate you taking a look.

Can you go into a bit more detail as to why accessibility can't be disabled via an input property? It defaults to being on, and for our use-case would be that this setting could be turned on a per user basis, and would allow the majority of users to leave it disabled and get the performance benefits on pages where we need to display a larger number of them at once.

PR #1796, opened 2017-01-11
Adding option to close dialog when the background overlay is clicked

howlettt:
(closeOnBackgroundClick)


cagataycivici:
Did some refactoring, thank you.

carlguo2016:
when i change dismissableMask to true, but it not works.

howlettt:
@carlguo2016 make sure you're using version 2.0.0 or greater. I've tested dismissableMask on 2.0.1 and it works

PR #1814, opened 2017-01-15
closed #1628 - Chips separator

### LINKED ISSUE #1628: Chips seperator
```
```

**Current behavior**
Adding chip is only possible using enter key

**Expected behavior**
Have the possibility to define one or more custom keys for inserting a chip (ex.: like space)

Gives developers the possibility to decide what is best for there use case

FallenRiteMonk:
This closes issue #1628 and gives users the possibility to specify different keys as input separators on chip input.


cagataycivici:
Many unnecessary lines seem to be changed and I can't see that actual change as a result.

Issue #1719, opened 2017-01-03
Button with Icon: the label is not updated when used with a pipe


iguissouma:
When used with a pipe such as translate (ng2-translate) the label is not correctly updated.
First display - OK
![2017-01-03_10h57_00](https://cloud.githubusercontent.com/assets/9085108/21604645/9fa4b160-d1a4-11e6-8dbb-1fb083831770.png)
After language switch - KO
![2017-01-03_11h00_18](https://cloud.githubusercontent.com/assets/9085108/21604657/b4e161c2-d1a4-11e6-8ebc-4c3522c5d982.png)

DOM

![2017-01-03_11h07_14](https://cloud.githubusercontent.com/assets/9085108/21604674/d3021bf6-d1a4-11e6-9276-efeec1dfc625.png)

"primeng": "^1.1.1",
"@angular/core": "2.2.1",
"ng2-translate": "^4.0.0",

I'm using the ultima-ng theme but I don't thik the problem is related to the theme.

donriver:
I have the similar issue, and it seems that it's related to https://github.com/primefaces/primeng/issues/1685

PR #1815, opened 2017-01-15
Fixed #1685 and #1719

### LINKED ISSUE #1685: Button label updates icon instead of label.
```
```

When label update after view initialization, the setter updates icon insead of label.

```
set label(val: string) {
        this._label = val;
        
        if(this.initialized) {
            this.domHandler.findSingle(this.el.nativeElement, '.ui-c').textContent = this._label;
        }
    }
```

The css selector should be `.ui-button-text` and not `.ui-c`;

kryzys69:
This simple change closes #1685 and #1719. I don't know what was the point of this change, but it was okay before commit b976a9d.

### DIFF
--- a/components/button/button.ts
+++ b/components/button/button.ts
@@ -64,7 +64,7 @@ export class Button implements AfterViewInit, OnDestroy {
         this._label = val;
         
         if(this.initialized) {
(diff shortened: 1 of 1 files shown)


cagataycivici:
Thank you.

PR #1823, opened 2017-01-17
Support HTML in cell

HVND:


cagataycivici:
I'd suggest using bodyTemplate instead to customize the cell content.

PR #1832, opened 2017-01-18
Unbind global listener when dropdown panel hides issue #1749

terranborn:
Current state:
1. panelVisible is used in ngClass - changing it starts angular change detection
2. n-time dropdowns will create n-time global body listeners which when clicked on body will start n-time change detection 

Fix:
1. When panel hides - unbind global listener
2. it will be initialized once again when dropdown is shown.


cagataycivici:
Good catch, thank you.

PR #1835, opened 2017-01-18
Added panelStyle and panelStyleClass attributes to dropdown component

Cselt:
**panelStyleClass**

Applying attribute `appendTo` to the dropdown element does not have any connection to the `original ui-dropdown` container element.
This means when we want to style a different look to a dropdown, we cannot apply the same style to its `ui-dropdown-panel` element since it is attached for instance to the body and they have no custom class on them.
This is why, we have to introduce `panelStyleClass` to make sure we can collaborate with the original `ui-dropdown` element.

**panelStyle**

Attribute `panelStyle` is very useful when we want to apply a custom width or any other style to `ui-dropdown-panel` element when `appendTo` is also applied.


cagataycivici:
Thank you.

PR #1841, opened 2017-01-18
[Feature] Custom chips separators

FallenRiteMonk:
Gives the user the possibility to specify which keys are used to separate chips on input.

As the last PR #1628 was closed without giving the possibility to fix, I hope this one is cleaner and will be accepted.

### DIFF
--- a/components/chips/chips.ts
+++ b/components/chips/chips.ts
@@ -50,6 +50,8 @@ export class Chips implements AfterContentInit,ControlValueAccessor {
 
     @Input() tabindex: number;
     
+    @Input() separator: number[] = [13]; // Default is enter
+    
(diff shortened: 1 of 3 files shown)


cagataycivici:
We'll consider and reopen if more users demand it. No significant value added with customizable keys as I've mentioned in previous PR as enter seems the natural choice.

BastianR:
This feature would be perfect. As an option, a separator key could be specified.

Franky238:
Yes, good feature ... and something like action `addChip()` will be better because creating chip by click on button is now impossible

nmobregon:
Dear @cagataycivici , for multiple emails, tab key is also a common choice.
I created issue [3527](https://github.com/primefaces/primeng/issues/3527) asking for this feature. 
Thanks

PR #1842, opened 2017-01-18
Remove chips input field if dissabled

FallenRiteMonk:
Removes the input field of chips if they are disabled as it is not functioning anyhow.


cagataycivici:
Reverted as it broke templating.

FallenRiteMonk:
Could you please specify a bit more precise and give me reproduce steps so tat i can try fixing it?

PR #1843, opened 2017-01-18
Chips margin bottom for multiline

FallenRiteMonk:
Adds the same margin to the bottom of a chip as is on its right, for same spacing if chips reach a second line.


cagataycivici:
We'll keep as it is thank you.

Issue #1854, opened 2017-01-19
Offset when opening up menu 


xlozinguez:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://plnkr.co/edit/NtWWnN

**Current behavior**
if an anchor tag contains another element, the menu attaches to that element

**Expected behavior**
the menu should always attach to the anchor tag

**Minimal reproduction of the problem with instructions**

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **Angular version:** 2.0.X

* **PrimeNG version:** 2.0.X

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

* **Node (for AoT issues):** `node --version` =

PR #1855, opened 2017-01-19
Adjusted target to currentTarget so that menu do not accidentally off…

xlozinguez:
…set if the a tag contains nested elements

Fixing #1854


cagataycivici:
There is a problem with line breaks, please have it fixed.

cagataycivici:
Fixed manually, thanks.

PR #1388, opened 2016-11-21
Show active state on TabMenu items that use routerLinks

mischkl:
Fixes #1386

PR #1863, opened 2017-01-19
Show active state on TabMenu items that use routerLinks

digitalcraftco:
ui-state-active has to been applied on li element. Fixes rejected PR on #1388.


cagataycivici:
Fixed via;

https://github.com/primefaces/primeng/issues/2393

Issue #1900, opened 2017-01-24
Invalid CSS selector - extra comma


brachi-wernick:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
in `tabview.css` extra comma in `.ui-tabview-title` (line 47)

**What is the motivation / use case for changing the behavior?**
when I'm doing in my less : import inline: 
`@import (inline) "../../../node_modules/primeng/resources/primeng.min.css";`
my css is broken.

PR #1901, opened 2017-01-24
Invalid css in tabview

brachi-wernick:
This is invalid CSS and breaks in the browser.

###Defect Fixes
issue: https://github.com/primefaces/primeng/issues/1900


cagataycivici:
Thanks.

Issue #1625, opened 2016-12-20
modal dialogs should prevent the background from being scrolleable


gatapia:
Using Ultima theme.

If we have a modal dialog then the background should not be scrollable when the dialog is visible.

If you could just add a 'modal-dialog-visible' class to the body this would sort the issue out.

staran:
+1, additionally the background text should not be selectable when being in the process of selecting text in the dialog and leaving its borders

iischwuerfl:
+1

hparfen:
+1

dhniels:
Did this get added in 4.1.0? I still get the page scrolling in the background when scrolling the dialog with modal set to true.

pchriste24:
To add to this issue, with modal=true you can Tab over inputs in the background and then edit their values.

brettkc:
As others have stated there still seems to be issues with this. I've got a dropdown in my modal dialog and the page in the background is still scrollable and completely messes up the rendering of the dropdown if you scroll outside of the modal.

TheAvalanche:
This feature is undocumented for some reason. But you can block `body` from scrolling by adding `[blockScroll]=true` to your `p-dialog` component.

For example: 
```
<p-dialog header="Title" [(visible)]="showDetailsModal" [modal]="true" [draggable]="false" [positionLeft]="0" [positionTop]="0" [blockScroll]="true">
...
</p-dialog>
```

ravikumarpamarthi:
Thanks @TheAvalanche

dhniels:
Using [blockScroll]="true" is anyone else having an issue where it still scrolls the body on touch devices testing on my iphone, scrolling the dialog content scrolls the body content behind the dialog. on desktop it works fine and blocks scrolling.

pedrotcm:
In IOS devices still scrolls the body, same using `[blockScroll]="true"` and `[modal]="true"`.

PR #1919, opened 2017-01-25
Adding class to body element when dialog is open

howlettt:
Adding ui-dialog-is-visible or ui-dialog-is-visible-modal class to the body element when dialog is open
#1625

### DIFF
--- a/components/dialog/dialog.ts
+++ b/components/dialog/dialog.ts
@@ -140,9 +140,12 @@ export class Dialog implements AfterViewInit,AfterViewChecked,OnDestroy {
             this.shown = true;
         } 
         
-        if(this.modal && !this._visible) {
-            this.disableModality();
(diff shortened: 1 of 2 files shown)


cagataycivici:
What is the purpose of manipulating body?

howlettt:
The main purpose of adding a class to the body is so scrolling can be disabled on the background elements while a dialog is open. See #1625

cagataycivici:
I see, instead we should make this easier by introducing a property like disableScroll="true|false".

shaoshuai314:
This issue still exists in 2.0.5. Has the fix reverted in primeng?

pedrotcm:
In IOS devices still scrolls the body, same using `[blockScroll]="true"` and `[modal]="true"`.

Issue #1929, opened 2017-01-27
DataTable rowIndex in body template is NaN


ilianiv:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/04PrXG?p=preview

**Current behavior**
Index of the row is NOT available at the template.

**Expected behavior**
Index of the row to be available at the template.

**Minimal reproduction of the problem with instructions**

Try to use let-index="rowIndex"

**What is the motivation / use case for changing the behavior?**

It does not work

* **Angular version:** 2.4.X

* **PrimeNG version:** 2.0.0-rc.1

* **Browser:** all

ilianiv:
Maybe this is the same issue https://github.com/primefaces/primeng/issues/1910

Emdee89:
can confirm this. This bug makes new version 2.0 not usable for my project

Gwiny:
Yes, I have this problem too. Would appreciate fixes.

shajeeck:
Update caused this problem! any fix ??

Emdee89:
Pull Request 1932 fixes this issue.
https://github.com/primefaces/primeng/pull/1932

PR #1932, opened 2017-01-27
DataTable rowIndex in body template is NaN

csokun:
###Defect Fixes
As describe #1929 

###Feature Requests


cagataycivici:
I've also fixed this with same solution, thanks.

PR #1944, opened 2017-01-29
Autocomplete pass index of option to itemTemplate

Petr0vi4:
In template you can access the index of option in the same way you access it in ngFor:
```
<template let-idx="index">
```


cagataycivici:
Will merge manually, thanks.

Issue #1810, opened 2017-01-13
Improve Calendar input/display formatting support


remmeier:
[ ] bug report => Search github for a similar issue or PR before submitting
[x ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35

* What is the motivation / use case for changing the behavior?

calendar has currently limited support for dateFormat="string". It should be possible to declare dateFormat and hourFormat separately for the input string and the display value. Usually backends return some standard format and it needs to be properly formatter for display to the user. As such the display and input format can differ. Potentially two new attributes inputDateFormat and inputHourFormat can be introduced.

remmeier:
or ideally it should be able to format the output with moment.js and the appropriate locales

cagataycivici:
Thank you, if more users demand it, we'll reconsider.

randdusing:
Please reconsider. Having a separate display and model format would make things much easier.

Issue #1580, opened 2016-12-13
Improve AutoComplete Object support


remmeier:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

Object support for the auto-complete component is currently rather limited. A "field" attribute can be set to choose one of the fields. A varierty of things can this way not be supported (like internationalization, combinding multiple fields, etc.).

Two options would be:

- allow to specify a template (like for the suggestions)
- allow to specify a pipe

At some point both should potentially be supported. A template is more flexible but also more work. While a pipe is simple and could be used for both the items and the suggestions.

cghislai:
There is an existing workaround that should probably be documented.
Bind the ngModel attribute to a string (using `[ngModel]`, not `[(ngModel)]`.
Use an object array for suggestions.
In the onComplete event handler, take your object instance and set the ngModel-bound  string to the value of your choice.

cagataycivici:
This should be fixed in recent PrimeNG release, please try with it and if the issue persists drop a comment with along with a test case based on plunkr below and we'll review again.

http://plnkr.co/edit/Wj39h1?p=preview

There is selectedItemTemplate now.

PR #1989, opened 2017-02-03
Improve AutoComplete Object support #1580

remmeier:
added possibilitity to specify a pipe as input to the auto-complete component. The pipe allows to transform data values before displaying them in the input control resp. as elements of the dropdown list. This is required in various scenarios like support actual objects as data, formatting of strings, etc.

please check the approach, I will then finalize and test.


odedolive:
Why use pipes when you can transform the objects using the already supported templates?

remmeier:
itemTemplates only affects the elements in the dropdown. The pipe also adjusts the contents of the input control. This means the autoComplete components becomes enabled to bind to complex objects with ngModel and to format those objects for display in the input control and dropdown. With support for this, it is not necessary to workaround this outside of the autocomplete component.

There is already rudimentary support for this with "field" and resolveFieldData. The pipe, however, is more generic and allows more powerful customizations. Looking at it again, I think the logic for the two should be merged.

remmeier:
I think it is a bit of a more general issue in primeng. With a similar pattern in Calendar you could resolve https://github.com/primefaces/primeng/issues/1810 as well.

cagataycivici:
I'd suggest templating instead for customization. We'll consider input pipe if there is demand in the future.

remmeier:
but the current template does only allow to customize the dropdown???

remmeier:
not the input field

aikrez:
I agree with remmeir. It is currently not possible to format the input which seems like an ootb feature to me.

For my use case I want to use the autocomplete for a list op people. 
Every person has a firstname and lastname so I want to display the firstname and lastname after I select a person.

Issue #1904, opened 2017-01-24
p-overlayPanel position problem


alarive:
When a big p-overlayPanel is attached to an element at the top of the page, the overlayPanel is not visible. Its style.top property gets set to a large negative number.

emilianbold:
Let me know if my pull request fixes your problem.

You could also manually tweak in node_modules/primeng/components/dom/domhandler.js the absolutePosition function and make sure you have something like

```
        if ((targetOffset.top + targetOuterHeight + elementOuterHeight > viewport.height) &&
          (targetOffset.top + windowScrollTop > elementOuterHeight))
            top = targetOffset.top + windowScrollTop - elementOuterHeight;
        else
            top = targetOuterHeight + targetOffset.top + windowScrollTop;
```

alarive:
I ended up using your PR, that fixes my issue. Thanks !

alarive:
I'm leaving the issue until your PR get accepted.

cyberrranger:
@emilianbold thx for your patch, is working
@cagataycivici can you please accept the pull request?

cagataycivici:
We first flip now and if top is less than 0, then try to fit it, this is called flipfit :)

cyberrranger:
sry is not fixed complete :/ (on split button)

![screenshot 2017-03-31 17 08 47](https://cloud.githubusercontent.com/assets/21671762/24556469/c2088568-1634-11e7-9cf4-4197b52d79ce.png)

i have on the top from the page a splitbutton with 12 entrys Jan. - Dec. is now the hight from the window > hight from all month Buttons, the overlay with the buttons open on the top

cagataycivici:
Split button doesn't seem to use this new algorithm, please create a new ticket and we'll review.

cyberrranger:
@cagataycivici please show #2439

cagataycivici:
Sure, added to 4.1.RC1.

PR #2031, opened 2017-02-09
Avoid negative offset when positioning

emilianbold:
This fixes issue #1904.

I have a large-ish OverlayPanel that constantly gets negative values for top but would be usable with this patch (and the user would have to scroll a little bit the page).

Of course, with the negative value the panel is entirely unusable as the close button is not accessible.


cagataycivici:
I've changed it a bit by trying to fit it if top is less than 0.

Issue #1963, opened 2017-01-31
Styleclass for TreeNode fails for Horizontal Tree


santoshsas:
I am not able to add styleClass to TreeNode 
I want to change style of Node ie. ui-treenode.
 I want to change entire background of TreeNode when node is selected instead of just label of node as in demo.
I see in documentation there is styleClass property for TreeNode but in common/api.d.ts there is no such property.
Thanks in advance

export interface TreeNode {
label?: string;
data?: any;
icon?: any;
expandedIcon?: any;
collapsedIcon?: any;
children?: TreeNode[];
leaf?: boolean;
expanded?: boolean;
type?: string;
parent?: TreeNode;
partialSelected?: boolean;
}
[X ] bug report => Search github for a similar issue or PR before submitting

**Current behavior**
adding styleClass property does not add class to treenode

**Expected behavior**
should add class to tree node

**What is the motivation / use case for changing the behavior?**
 I want to change entire background of TreeNode when node is selected instead of just label of node as in demo.

* **Angular version:** 2.0.2

* **PrimeNG version:** 2.0.0

cheesgno:
I'm trying to use the styleClass property for ui-treenode element on Tree (with Horizontal Layout) but the rendering of the element  doesn't contains my class.
I'm using primeng 2.0.5

PR #2046, opened 2017-02-11
Implement styleClass for ui-treenode element on Tree

nate-knight:
This adds the styleClass property for ui-treenode elements. The documentation shows 'styleClass' as a property of the TreeNode API, however the current codebase does not include this.  Issue #1963 describes this further.

### DIFF
--- a/components/common/api.ts
+++ b/components/common/api.ts
@@ -60,6 +60,7 @@ export interface TreeNode {
     type?: string;
     parent?: TreeNode;
     partialSelected?: boolean;
+    styleClass?: string;
 }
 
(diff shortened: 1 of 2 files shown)


cagataycivici:
Thank you.

nate-knight:
Thank you @cagataycivici. Do you know when 2.0.1 will be available via npm?

cheesgno:
I'm trying to use the styleClass property for ui-treenode element on Tree (with Horizontal Layout) but the rendering of the element  doesn't contains my class.
I'm using primeng 2.0.5

Issue #1860, opened 2017-01-19
Listbox: clicking on an item will unselect all other items


Dekel-Cohen:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
I tried to build a Plunkr Case for it but i encountered importing the Listbox module.
Any help with it will be much appreciated. 

**Minimal reproduction of the problem with instructions**
1. Go to the documentation of Listbox: 
    http://www.primefaces.org/primeng/#/listbox 
2. Select New York, Rome (click on the checkbox)
3. Click on London (on the title, not on the checkbox) =>New York, Rome will be unselected. 

**What is the motivation / use case for changing the behavior?**
We want to use this component in our project but this bug makes this component almost unusable.

Mrtcndkn:
Actually it's not a bug that's the way it works by design.

Dekel-Cohen:
@Mrtcndkn The user can very easily unselect all its items,
The only way he can revet the change is by closing the modal and open it again.
I don't think that the common use case is that the user wants all it's items to be unselected 
once he clicks on an item.

Will consider to change it to be less error prone?
like clear all selected items if the user *double* click on an item..

roni-frantchi:
@Mrtcndkn also note that the Multiselect dropdown component doesn't behave this way: http://www.primefaces.org/primeng/#/multiselect .
I find the behavior @Dekel-Cohen had mentioned to make more sense.
That said, maybe an `@Input()` could be added to make the Listbox multiselect behave more like the Multiselect component

Mrtcndkn:
Okey we will review this behavior i will inform you guys from here. @Dekel-Cohen @roni-frantchi

brachi-wernick:
@Mrtcndkn any progress for this?

PR #2063, opened 2017-02-13
DECISION-11756- fixing a Listbox multi select behaviour

Dekel-Cohen:
###Defect Fixes
Fix for issue https://github.com/primefaces/primeng/issues/1860
####Feature Requests
We are to change the behaviour of the compoenet only when it is configured to selected multiple values as following:
Clicking an item will not remove all other selected items, but instead it will add or remove the selected item
depending if it was already selected or not.

### DIFF
--- a/components/listbox/listbox.ts
+++ b/components/listbox/listbox.ts
@@ -160,33 +160,23 @@ export class Listbox implements AfterContentInit,ControlValueAccessor {
     }
 
     onOptionClickMultiple(event, option) {
-        let metaKey = (event.metaKey || event.ctrlKey);
         let selected = this.isSelected(option);
-        let valueChanged = false;
 
         if (selected) {
-            if (metaKey) {
-                this.value.splice(this.findIndex(option), 1);
-            }
-            else {
-                this.value = [];
-                this.value.push(option.value);
-            }
-            valueChanged = true;
+            this.value.splice(this.findIndex(option), 1);
         }
         else {
-            this.value = (metaKey) ? this.value || [] : [];
+            this.value = this.value || [];
             this.value.push(option.value);
-            valueChanged = true;
+
         }
 
-        if (valueChanged) {
-            this.onModelChange(this.value);
-            this.onChange.emit({
-                originalEvent: event,
-                value: this.value
-            });
-        }
+        this.onModelChange(this.value);
+        this.onChange.emit({
+            originalEvent: event,
+            value: this.value
+        });
+
     }
 
     isSelected(option: SelectItem) {


components/listbox/listbox.ts
```diff
@@ -160,33 +160,23 @@ export class Listbox implements AfterContentInit,ControlValueAccessor {
...
         if (selected) {
-            if (metaKey) {
-                this.value.splice(this.findIndex(option), 1);
-            }
-            else {
-                this.value = [];
-                this.value.push(option.value);
-            }
-            valueChanged = true;
+            this.value.splice(this.findIndex(option), 1);
```
Dekel-Cohen: Since it is multiple values, we do not want to clear all other items, we just want to remove the current item.

components/listbox/listbox.ts
```diff
@@ -160,33 +160,23 @@ export class Listbox implements AfterContentInit,ControlValueAccessor {
...
 
-        if (valueChanged) {
-            this.onModelChange(this.value);
-            this.onChange.emit({
-                originalEvent: event,
-                value: this.value
-            });
         }
+
+        this.onModelChange(this.value);
```
Dekel-Cohen: No real need for the ```if``` statement, since ```valueChanged``` will always be ```true```  here

components/listbox/listbox.ts
```diff
@@ -160,33 +160,23 @@ export class Listbox implements AfterContentInit,ControlValueAccessor {
...
-            else {
-                this.value = [];
-                this.value.push(option.value);
-            }
-            valueChanged = true;
+            this.value.splice(this.findIndex(option), 1);
         }
         else {
-            this.value = (metaKey) ? this.value || [] : [];
+            this.value = this.value || [];
```
Dekel-Cohen: Since it is multiple values, we do not want to clear all other items, we just want to add the current item.

cagataycivici:
Multiple selection is done with metakey, do you want to change that behavior?

Dekel-Cohen:
@cagataycivici Yes, we think that having metakeys does not make the best user experience here since 
he doesn't have any UI indication that there is support for metakeys.
Also, when defining Listbox to multiple values, we want to make the selection of multiple values
as easy as we can since it would be the common case. Holding a metakey each time you want to add/remove an item feels a bit cumbersome

cagataycivici:
I've made this customizable because we have some users who prefer to offer metaKey based selection so it is up to the page author now using metaKeySelection property. Note that in touch enabled devices it is always disabled now.

Ticket is: https://github.com/primefaces/primeng/issues/2094

I'll also apply the same solution to OrderList, PickList and DataTable now.

sarahy18:
Hi I want the same behaviour described by Dekel-Cohen for the listbox . What I have to do???
Thanks

s13571:
It still deselects all other items on click

Issue #2055, opened 2017-02-13
Datatable row deselection should optionally require the meta/ctrl key


ivanhosgood:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35


**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

Functionality evident at: http://www.primefaces.org/primeng/#/datatable/selection

**Current behavior**
<!-- Describe how the bug manifests. -->
Within the datatable component, row de-selection requires pressing the ctrl/command key while you de-select the row

**Expected behavior**
<!-- Describe what the behavior would be without the bug. -->
I would prefer if this was configurable. If the metakey is not configured, clicking on a row should toggle the selection

**Minimal reproduction of the problem with instructions**
<!--
If the current behavior is a bug or you can illustrate your feature request better with an example, 
please provide the *STEPS TO REPRODUCE* and if possible a *MINIMAL DEMO* of the problem via
https://plnkr.co or similar (you can use this template as a starting point: http://plnkr.co/edit/tpl:AvJOMERrnz94ekVua0u5).
-->
Functionality evident at: http://www.primefaces.org/primeng/#/datatable/selection


**What is the motivation / use case for changing the behavior?**
<!-- Describe the motivation or the concrete use case -->
Usability. It took a good 5 minutes to work out that row deselection had been implemented but required the use of the ctrl/command key

**Please tell us about your environment:**
<!-- Operating system, IDE, package manager, HTTP server, ... -->
Windows 10

* **Angular version:** 2.0.X
<!-- Check whether this is still an issue in the most recent Angular version -->
2.4.7

* **PrimeNG version:** 2.0.X
<!-- Check whether this is still an issue in the most recent Angular version -->
2.0.0

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
<!-- All browsers where this could be reproduced -->
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

[...]

PR #2068, opened 2017-02-13
Added DataTable selectionMeta boolean to flag whether selection is a …

ivanhosgood:
…normal toggle or toggles when the metakey is pressed

###Feature Requests
#2055 Datatable row deselection should optionally require the meta/ctrl key


cagataycivici:
Fixed via;

https://github.com/primefaces/primeng/issues/2098

PR #2103, opened 2017-02-17
Use chart.js's `update()` method when refreshing

### LINKED ISSUE #2102: Chart.js's `update()` method should be used on refresh rather than `destroy()`
```
```

**Plunkr Case (Bug Reports)**
See the demo here: http://plnkr.co/edit/zwmZKd?p=preview

**Current behavior**
On each refresh of the graph data, the graph redraws from scratch. The lines on the graph all animate from the 0 position on the y axis, rather than from their previous position.

**Expected behavior**
They should animate from their previous position.

See the demo here: http://plnkr.co/edit/zwmZKd?p=preview

Smoother animation of changes in data so that users can follow what is actually happening with the lines.

* **Angular version:** latest (used Plunkr fork)

* **PrimeNG version:** 2.0.0

* **Browser:** Tested in Safari 10 and Chrome 56
 
* **Language:** TS

jessepinho:
If you destroy and re-init the chart every time it needs to be refreshed, it will not smoothly animate between data sets. Chart.js provides an `update()` method for exactly this purpose.

### Defect Fixes
#2102

### Instead of this:
licecap

### It should be doing this (I hacked PrimeNG locally to do what I want):
licecap2

### DIFF
--- a/components/chart/chart.ts
+++ b/components/chart/chart.ts
@@ -68,8 +68,7 @@ export class UIChart implements AfterViewInit, OnDestroy {
     
     refresh() {
         if(this.chart) {
-            this.chart.destroy();
-            this.initChart();
+            this.chart.update();
         }
     }
     


jessepinho:
On second thought: One can just call the `update()` method directly using `UIChart.prototype.chart.chart.update()`. (But is that intended?)

BryantChris:
This change breaks use of the chart component in cases where the consumer sends a new dataset to the chart or changes the chart type before refresh. The destroy and init were critical for re-establishing the chart in those cases.

jessepinho:
@BryantChris Good catch. I'm not sure what the best course of action is here, but I would advise using a different name than `refresh` to indicate a complete destroy/re-init of the chart. Perhaps `reset`? `reinit`?

cagataycivici:
I'll add reinit, wdyt?

cagataycivici:
Added reinit.

Issue #2040, opened 2017-02-10
Highlight first item in AutoComplete


davladus:
**I'm submitting a usability bug

Plunkr Case (Bug Reports)
http://www.primefaces.org/primeng/#/autocomplete

Current/Expected behavior
Minimal reproduction of the problem with instructions

When I start typing letters (Advanced sample)
And selection dropdown is expanded
Then
	Observed: nothing selected
	Expected: first filtered item is selected

Please tell us about your environment:
Window 7, Chrome

Mrtcndkn:
We reviewed this and this is expected behaviour.

GFoley83:
Is there any way to programmatically default to the first item in the suggestion list, without having to click the down arrow key? 

This makes even more sense if there's only one option is the suggestion list. Currently, the user always has to use the down arrow key to select the value. 

E.g Spain should be preselected here

![image](https://cloud.githubusercontent.com/assets/838788/23009636/080c40fc-f47c-11e6-8f41-3a127dbf7e08.png)

GFoley83:
@davladus I was able to get this functionality by dispatching a down arrow key event to the autocomplete component at the end of the `completeMethod` callback. Working code:

```ts
import { AutoComplete } from 'primeng/primeng';
```
```ts
@ViewChild(AutoComplete) primeAutoCompleteComponent: AutoComplete;
```
```ts
    // (completeMethod)="search($event)"    
    search(event?: any) {
        // filter you search results array
        . . .    
        this.preSelectFirstOptionInResultsList();
    }
```
```ts
private preSelectFirstOptionInResultsList(): void {
    setTimeout(() => {

        let highlightItemIndex = this.primeAutoCompleteComponent.findOptionIndex(this.primeAutoCompleteComponent.highlightOption);

        if (highlightItemIndex < 0) {
            var arrowDownEvent = new Event('keydown', { bubbles: false, cancelable: false });
            var arrowDownEvent = document.createEvent('Event');
            (<any>arrowDownEvent).initEvent('keydown', { "bubbles": false, "cancelable": false });
            Object.defineProperty(arrowDownEvent, 'which', { 'value': 40 });

            (<Node>this.primeAutoCompleteComponent.input).dispatchEvent(arrowDownEvent);
        }

    }, 0);
}
```

haisaco:
Now, with ng4 it have issue  
Property 'input' does not exist on type 'AutoComplete'

Errabaany:
[...]

PR #2113, opened 2017-02-18
AutoComplete - added ability to optionally auto select first suggestion

GFoley83:
Currently after every search, no suggestions are highlighted in the suggestion list, even if there is only one suggestion remaining. This means that after every search, an extra step is forced on the user as they have to either click the suggestion using the mouse or by using the down arrow key. 

I've added an extra `@Input()` to the autocomplete component called `autoSelectSuggestion` which defaults to `false`. Passing `true` results in the first result of every search being auto selected in the suggestion list. This is especially useful for multi select fields when choosing many options, or as mentioned above, when there is only one suggestion remaining. This results in less keyboard/mouse interaction overall.

As part of this PR I also tidied up the logic for figuring out which option to highlight by moving it into its own function, which also tidies up the `onKeydown` function. Calling the logic to trigger the auto select happens inside the `ngDoCheck` using a simple `if` statement check on the `autoSelectSuggestion` property.

Addresses issue discussed in #2040

### DIFF
--- a/components/autocomplete/autocomplete.ts
+++ b/components/autocomplete/autocomplete.ts
@@ -95,6 +95,8 @@ export class AutoComplete implements AfterViewInit,DoCheck,AfterViewChecked,Cont
     @Input() multiple: boolean;
 
     @Input() tabindex: number;
+
+    @Input() autoSelectSuggestion: boolean;
     
     @ContentChildren(PrimeTemplate) templates: QueryList<any>;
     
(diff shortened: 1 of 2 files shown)


GFoley83:
Hi @cagataycivici, apologies for hailing directly but it would be great to get some kind of feedback on the above PR.

cagataycivici:
We will review in 2 weeks time before 4.0 final, thank you.

cagataycivici:
Will merge manually with some changes, thank you.

PR #2123, opened 2017-02-20
Schedule: enable dynamically updating all input properties

### LINKED ISSUE #2020: New method for Schedule
Hello, thanks for awesome components!

Currently is not possible to rerender the schedule when an option is updated without to refresh the all page.

Would you like to add the fullcalendar function ( like prev(), next(), prevYear(), nextYear(), gotoDate(date), incrementDate(duration), getDate(), today()) please?

Get / Set Options Dynamically https://fullcalendar.io/docs/utilities/dynamic_options/

Thank you.

yehudag:
This change will make it possible to dynamically update the editable property of the schedule using the fullcalendar methods [https://fullcalendar.io/docs/utilities/dynamic_options/](https://fullcalendar.io/docs/utilities/dynamic_options/)

Using the ngOnChanges hook, we can dynamically update most fullcalendar options.

### Defect Fixes
Closes #2020

### DIFF
--- a/components/schedule/schedule.ts
+++ b/components/schedule/schedule.ts
@@ -1,4 +1,4 @@
-import {NgModule,Component,ElementRef,OnDestroy,DoCheck,Input,Output,EventEmitter,IterableDiffers,OnInit,AfterViewChecked} from '@angular/core';
+import {NgModule,Component,ElementRef,OnDestroy,DoCheck,Input,Output,EventEmitter,IterableDiffers,OnInit,OnChanges,AfterViewChecked,SimpleChanges} from '@angular/core';
 import {CommonModule} from '@angular/common';
 
 declare var jQuery: any;
@@ -7,7 +7,7 @@ declare var jQuery: any;
     selector: 'p-schedule',
     template: '<div [ngStyle]="style" [class]="styleClass"></div>'
 })
-export class Schedule implements DoCheck,OnDestroy,OnInit,AfterViewChecked {
+export class Schedule implements DoCheck,OnDestroy,OnInit,OnChanges,AfterViewChecked {
     
     @Input() events: any[];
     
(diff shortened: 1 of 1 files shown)


yehudag:
@cagataycivici I realize that this is a substantial change, but we currently do not have a method to update the Scheduler options once it has been created. Do you have any feedback?

ElMohMoh:
Hello @yehudag,

To update the scheduler, I think they can do this:

In schedule.js add the refresh feature like this:
`Schedule.prototype.refresh = function (options){
        this.schedule.fullCalendar('option', options);
    }`

And then we can call the function like this:
`  <p-schedule #sch [events]="events"></p-schedule>
<button (click)="update(sch)">Update</button>
`
`update(sch) {
    let options : any = { weekends: false };
    sch.refresh(options);
  }`

@cagataycivici could you please add this feature in the next version?

Thank you.

yehudag:
@ElMohMoh My concern with this approach is that the component that hosts the Schedule must now be in charge of updating the Scheduler when the input properties change. I think it is cleaner to hide the fact that the schedule cannot update it's state when the input properties change.

ElMohMoh:
@yehudag,
Ok for this but your approach is for one option only. I think, it is better to find something for all options.

cagataycivici:
Will be reviewed before 2.0.2, thanks.

yehudag:
@ElMohMoh agreed

yehudag:
This should work for all options. 

```typescript
ngOnChanges(changes: SimpleChanges) {
        if (this.schedule) {
            let options = {};
            for (let change in changes) {
                options[change] = changes[change].currentValue;
            }
            this.schedule.fullCalendar('option', options);
        }
}
```

cagataycivici:
ngOnChanges is also the only way I can think of, thanks.

slickdata0:
Is it working? I'm using the timelineDay via fullcalendar-scheduler & primeNG. I'm unable to update the resources when date change. viewRender is not a good option to fetch and update the calendar.

petronics:
@yehudag Almost. It works with all the options which has a corresponding Angular _**Input()**_ property on the **Schedule** directive. 
(For example: _locale_ and _hiddenDays_)

But with the properties given inside the directive's _options_ Angular _**Input()**_ property it will fail, because they remain nested inside an options object.

primeng_scheduler

**Solution**
The solution is to walk through and extract the changes of the _changes['options'].currentValue_ object and push them to the local options object one-by-one.

yehudag:
@petronics 
Are you saying that Angular won't detect changes when properties in the `@Input() options` object? For instance: `options['whatever'] = newValue` 
If so I agree but Angular doesn't detect changes reliably when an object (or an array) is mutated.

PR #2126, opened 2017-02-21
#1973 Grid offset classes and documentation

### LINKED ISSUE #1973: Grid offset classes
```
[ x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
```

**Current behavior**
The grid system allows you to position elements on a page but only directly next to each other. Would be nice if you could offset elements so they don't sit right next to each other. the only way this can be done now is to create an empty element with a grid class.

Something like this:
```
<div class="ui-g">
  <div class="ui-g-6">...content...</div>
  <div class="ui-g-3 ui-g-offset-3">content w/ 3 grid columns of space to the left</div>
</div>
```

Instead of having to do this:
```
<div class="ui-g">
  <div class="ui-g-6">...content...</div>
  <div class="ui-g-3"></div>
  <div class="ui-g-3">content w/ 3 grid columns of space to the left</div>
</div>
```

fewer unneeded dom elements

waldo2188:
This PR add the ability to offset elements, in a basic way an responsive way.
This PR update the documentation to show how to use offsets CSS class.

###Defect Fixes
Fix #1973

### DIFF
--- a/components/grid/grid.css
+++ b/components/grid/grid.css
@@ -249,6 +249,54 @@
     width: 100%;
 }
 
+.ui-g-offset-0 {
+  margin-left: 0%;
+}
+
+.ui-g-offset-1 {
+  margin-left: 8.33333333%;
+}
+
+.ui-g-offset-2 {
+  margin-left: 16.66666667%;
+}
+
+.ui-g-offset-3 {
+  margin-left: 25%;
+}
+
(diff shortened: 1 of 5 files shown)


cagataycivici:
There are also unrelated changes like multiselect, please update the PR.

Issue #2140, opened 2017-02-23
Calendar supports Moment.JS locales


upmauro:
Changes the calendar component to try to use Moment.JS if the user imported it and set the locale options, just in case the locale input property is not informed.

cagataycivici:
Better if this is done at application side.

PR #2141, opened 2017-02-23
Ty use moment.js locale if already defined #2140

upmauro:
Feature #2140

### DIFF
--- a/components/calendar/calendar.ts
+++ b/components/calendar/calendar.ts
@@ -5,6 +5,8 @@ import {InputTextModule} from '../inputtext/inputtext';
 import {DomHandler} from '../dom/domhandler';
 import {AbstractControl, NG_VALUE_ACCESSOR, NG_VALIDATORS, ControlValueAccessor} from '@angular/forms';
 
+declare var moment;
+
 export const CALENDAR_VALUE_ACCESSOR: any = {
   provide: NG_VALUE_ACCESSOR,
   useExisting: forwardRef(() => Calendar),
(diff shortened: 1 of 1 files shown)


components/calendar/calendar.ts
```diff
@@ -221,13 +223,20 @@ export class Calendar implements AfterViewInit,OnInit,OnDestroy,ControlValueAcce
...
-    @Input() locale: LocaleSettings = {
+    @Input() locale: LocaleSettings = !moment ? {
         firstDayOfWeek: 0,
         dayNames: ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"],
         dayNamesShort: ["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"],
         dayNamesMin: ["Su","Mo","Tu","We","Th","Fr","Sa"],
         monthNames: [ "January","February","March","April","May","June","July","August","September","October","November","December" ],
         monthNamesShort: [ "Jan", "Feb", "Mar", "Apr", "May", "Jun","Jul", "Aug", "Sep", "Oct", "Nov", "Dec" ]
+    } : {
+        firstDayOfWeek: moment.localeData()._week.dow,
```
jessepinho: This looks like a private property, however...?

components/calendar/calendar.ts
```diff
@@ -221,13 +223,20 @@ export class Calendar implements AfterViewInit,OnInit,OnDestroy,ControlValueAcce
...
+        dayNames: moment.weekdays(),
+        dayNamesShort: moment.weekdaysShort(),
+        dayNamesMin: moment.weekdaysShort(),
+        monthNames: moment.months(),
+        monthNamesShort: moment.weekdaysShort()
```
jessepinho: Should be `moment.monthsShort()`

components/calendar/calendar.ts
```diff
@@ -221,13 +223,20 @@ export class Calendar implements AfterViewInit,OnInit,OnDestroy,ControlValueAcce
...
         dayNames: ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"],
         dayNamesShort: ["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"],
         dayNamesMin: ["Su","Mo","Tu","We","Th","Fr","Sa"],
         monthNames: [ "January","February","March","April","May","June","July","August","September","October","November","December" ],
         monthNamesShort: [ "Jan", "Feb", "Mar", "Apr", "May", "Jun","Jul", "Aug", "Sep", "Oct", "Nov", "Dec" ]
+    } : {
+        firstDayOfWeek: moment.localeData()._week.dow,
+        dayNames: moment.weekdays(),
+        dayNamesShort: moment.weekdaysShort(),
+        dayNamesMin: moment.weekdaysShort(),
```
jessepinho: Should be `moment.weekdaysMin()`

components/calendar/calendar.ts
jessepinho: Should be `moment.localeDate().firstDayOfWeek()`.

components/calendar/calendar.ts
```diff
@@ -221,13 +223,20 @@ export class Calendar implements AfterViewInit,OnInit,OnDestroy,ControlValueAcce
     
     @Output() onSelect: EventEmitter<any> = new EventEmitter();
     
-    @Input() locale: LocaleSettings = {
+    @Input() locale: LocaleSettings = !moment ? {
```
jessepinho: This will cause a `ReferenceError`. Do something like `typeof moment === 'undefined'` instead.

cagataycivici:
Rather not have a direct dependency to moment, thank you.

upmauro:
@cagataycivici sorry, but maybe we can try use if moment is defined, i don't know if this is a good pratice, but this method will not create a "direct" dependency.

jessepinho:
@upmauro This can also be done in your app (rather than in this library itself) by simply passing in this moment config object via the `locale` input. Does that make sense?

bhaidar:
@upmauro Great idea. Can we find all the info needed by Calendar in the moment() object? Thanks

PR #2151, opened 2017-02-24
DataTable - fixed multisorting with lazy bug

radarsu:
When using lazy loading with multiple sorting, multiSortMetadata was not being filled. Because of that multiSort information was not sent to server-side.

Also, the effect adding fa-sort-asc / fa-sort-desc arrows didn't trigger in lazy mode beacuse of this bug.


olingerc:
I would be very interested in having this merged. I do not like my workaround hacks.

cagataycivici:
Will merge manually due to line end difference.

Issue #1971, opened 2017-02-01
Dropdown does not support ChangeDetection Push strategy


cagataycivici:
Default value is not updated until component is clicked.

PR #2159, opened 2017-02-26
Enable dropdown to use OnPush change detection strategy

### LINKED ISSUE #1754: Dropdown does not hide on Push ChangeDetection
```
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/rbQWTdX8EbVHfHvqu42E

**Current behavior**
Create a component that implements OnPush change detection strategy and use the primeng dropdown component within it. Click the dropdown to open the select list panel click outside the dropdown (and don't hover the mouse back inside the panel). The panel stays open and doesn't close.

**Expected behavior**
The panel should close.

See steps above

I have components that require the performance benefits of OnPush that I'd like to also use the dropdown component within, but cannot due to this.

* **Angular version:** 2.1.X

* **PrimeNG version:** 1.0.1

* **Browser:** All
 
* **Language:** All

variable-content:
Fixes #1754 

When the dropdown is used in a large nested component tree performance noticeable degrades when opening/closing the dropdown as this triggers all components in the component tree to run through change detection.

The change in this PR allows components that use this component to use the OnPush change detection strategy to prevent the entire component tree from running change detection and therefore reducing the amount of time between the user clicking the dropdown and the panel opening. Tests in my environment show a reduction from 400ms from click to panel opening to <100ms.

### DIFF
--- a/components/dropdown/dropdown.ts
+++ b/components/dropdown/dropdown.ts
@@ -1,6 +1,8 @@
-import {NgModule,Component,ElementRef,OnInit,AfterViewInit,AfterContentInit,AfterViewChecked,DoCheck,OnDestroy,Input,Output,Renderer,EventEmitter,ContentChildren,
+import {NgModule,ChangeDetectionStrategy,Component,ElementRef,OnInit,AfterViewInit,AfterContentInit,AfterViewChecked,DoCheck,OnDestroy,Input,Output,Renderer,EventEmitter,ContentChildren,
         QueryList,ViewChild,TemplateRef,IterableDiffers,forwardRef,trigger,state,style,transition,animate,ChangeDetectorRef} from '@angular/core';
 import {CommonModule} from '@angular/common';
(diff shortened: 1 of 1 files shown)


cagataycivici:
Not sure, why we need all the RXJS dependencies, we've added simple this.cd.markForCheck(); at writeValue to fix this. See;

Let's discuss at;

https://github.com/primefaces/primeng/issues/1971

variable-content:
@cagataycivici As per your comments in #1754 this fixes a different issue. Could you please re-open? 

Regarding the usage of Observables + async pipe to me it seemed simpler to just have one line of code to update the panel visibility than having to add an additional line underneath it every time the value changes, e.g.:

this.panelVisibleSubject.next(true);

instead of:
this.panelVisible = true;
this.cd.markForCheck();

I also added the changeDetection: ChangeDetectionStrategy.OnPush line to the @Component declaration to make it obvious to developers using it than it can be safely used inside our OnPush components.

Issue #1935, opened 2017-01-27
getNextMonthAndYear return undefined as year when month is < 11


LHIOUI:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://plnkr.co/edit/mbikgiRs6oWPT8PvMJ24?p=preview

**Current behavior**

Calendar function getNextMonthAndYear return undefined year when month is less than 11 so when we use selectOtherMonths combined with showOtherMonths we get an invalid date
**Expected behavior**

The function should return the passed param year when month is less than 11 
**Minimal reproduction of the problem with instructions**

`
@ViewChild('calendar) calendar: Calendar;

  //ngOnInt lifecycle hook

  ngOnInt(){

      let nextMY = this.calendar.getNextMonthAndYear(1,2017)

      console.log('next month is '+nextMY.month + ' next year is '+nextMY.year)

      //print: next month is 2 next year is undefined 

  }
`
**What is the motivation / use case for changing the behavior?**

get the right next year 
**Please tell us about your environment:**

OS: OSX 10.12 , IDE : webstorm, package manager: npm 
* **Angular version:** 2.0.X

* **PrimeNG version:** cloned from github 

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

* **Node (for AoT issues):** `node --version` =

external-olgierd:
Created pull-request for this as part of fix for 1689.

cagataycivici:
Unable to replicate with 4.1.RC2, please provide a test case based on plunkr below and we'll review again.

http://plnkr.co/edit/Wj39h1?p=preview

Issue #1689, opened 2016-12-27
Calendar not honoring selectOtherMonths property correctly


hjalmers:
**I'm submitting a ...**  
```
[x ] bug report
```
**Current behavior**
Even when `selectOtherMonths = false` month is clickable (but not selectable) and the `ui-state-disabled` class is applied even if `selectOtherMonths = true`

**Expected behavior**
`selectOtherMonths = false` should disable the ability to click on dates in other months and the class `ui-state-disabled` should only be applied when this property is false.

**Minimal reproduction of the problem with instructions**
Use calendar and set selectOtherMonths to false

**What is the motivation / use case for changing the behavior?**
The function should work as described

* **Angular version:** 2.3.1

* **PrimeNG version:** 1.1.2

* **Browser:** all

external-olgierd:
Create a pull request with a fix for this issue and also 1689 as they are related.

hjalmers pull request is good and adds good functionality to allow specification of "disabled" dates, but I have created a pull request that fixes only the problem at hand, i.e. that selectOtherMonths option is not working properly - just to minimize the changes and potential impact.

Summary of the changes:
- applied disabled CSS class for days in other months only if the above option is not set to "true"
- year for dateMeta object is calculated properly now for next month (previously it was only working for previous months or for next months only in case if the year was also incremented)
- made sure that "selectable" property of dateMeta is set correctly for dates in other month (i.e. set to false if selectOtherMonths is not set to "true" and set to true otherwise) - this is to make sure other month dates behaves the same as other ones - if they can be selected they act as enabled ones but if they cannot be selected they behave the same as disabled dates (i.e. when click on them calendar doesnt close down)

cagataycivici:
I can't replicate with 4.0.RC1, anyone. selectOtherMonths is false by default. Note that, following enables it;

```
 selectOtherMonths="false"
```

should be;

```
 [selectOtherMonths]="false"
```

Or just leaving it as it is since it is false by default.

PR #2163, opened 2017-02-27
Fixes for #1689 and #1935

external-olgierd:
Applied fix for selectOtherMonths option as it wasnt working properly.
Correct CSS class now applied for dates from other months depending on the value of selectOtherMonths option.
Year calculated properly for date in the next month (but the same year). #1935
Selectable property set on the date from other month properly based on the value of selectOtherMonths option.


cagataycivici:

PR #2176, opened 2017-02-28
Calendar fix #1706 - handle dateFormat change

### LINKED ISSUE #1706: Date in p-calendar doesn't update when date format changes
**I'm submitting a ...** 
```
```

**Plunkr Case (Bug Reports)**

http://plnkr.co/edit/U1mypT4F5N91l0vgKCl0

**Current behavior**
Date in p-calendar doesn't update when date format changes.

**Expected behavior**
Date in input field updates when date format changes.

My app has several locations and date format depends on the selected locale. I want that date in the input field changes according to the selected locale.

* **PrimeNG version:** 1.1.1

* **Browser:** all

external-olgierd:
Date in the input field is now updated whenever dateFormat changes.


external-olgierd:
If any of the maintainers prefer to do that with a setter/getter rather than OnChange then I can modify the code.

cagataycivici:
Please update the PR due to conflict and we'll review.

PR #2177, opened 2017-02-28
Calendar fix #1036 - handle min/max date change

### LINKED ISSUE #1036: Calendar model not updated if maxDate or minDate modifies the date
jQuery UI Datepicker modifies selected date if new min/maxDate is set and current date doesn't fit.  https://github.com/jquery/jquery-ui/blob/master/ui/widgets/datepicker.js#L1931. 

In this case jQuery Datepicker does not trigger 'onSelect' event and p-Calendar doesn't update it's model but the actual input value gets changed.

**Example:***

```
date = '13.03.2013';
maxDate = '23.03.2013';
```

```
<p-calendar [(ngModel)]="date" [maxDate]="maxDate" dateFormat="dd.mm.yy"></p-calendar>
```

image

Then we change: `maxDate = '11.03.2013'` and `console.log(date)`:

```
'13.03.2013'
```

but 
image

external-olgierd:
Make sure that date value is updated correctly and onSelect event is sent out with the new date whenever minDate or maxDate is updated and currently selected date falls outside the new range.

### DIFF
--- a/components/calendar/calendar.ts
+++ b/components/calendar/calendar.ts
@@ -293,6 +293,7 @@ export class Calendar implements AfterViewInit,OnInit,OnDestroy,ControlValueAcce
     set minDate(date: Date) {
         this._minDate = date;
         this.createMonth(this.currentMonth, this.currentYear);
+        if (this.value < date) this.updateDate(date);
     }
     
     @Input() get maxDate(): Date {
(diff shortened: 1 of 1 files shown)


cagataycivici:
Please describe the issue first.

olo-b:
@cagataycivici this is as per #1036. Exact steps to reproduce:

1. Set minDate or/and maxDate on the calendar - for example 20.03.2017 and 30.03.2017 respectively
2. Select a date in range on calendar, for example 25.03.2017
3. Now change the minDate to be 26.03.2017 or the maxDate to be 24.03.2017 - the currently selected date no longer "fits" in the range.

This change makes sure that when step 3 happens, the currently selected date is updated and displayed correctly and also triggers onSelect event (as the date changes).

britvik:
I'm not sure if this is really wanted behaviour. If min/max changes then your date gets set to min/max value and user may not notice this change.

See my issue #2437 with a pull request #2438. My solution makes date invalid if min/max changes so user can correct it (him/her)self.

cagataycivici:
Please update the PR due to conflict and we'll review.

Issue #2108, opened 2017-02-17
Listbox disabled allows item selection


imfarhadk:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
none

**Current behavior**
[checkbox]="true" [multiple]="true"
when the multi-select listbox, is disabled the user can still click on the checkbox and select/unselect an item 

**Expected behavior**
user should not be allowed to select /unselect an item in disable mode

**Please tell us about your environment:**
Angular version:  2.4.5
PrimeNG version: 2.0.0-rc.2

xixi4Den:
works well when click on label. but the issue is still reproducible when click on input. v2.0.5

cagataycivici:
Which input?

xixi4Den:
checkbox. user still can select/unselect either particular or "toggle all" checkbox in disabled mode

PR #2178, opened 2017-02-28
Listbox #2108 - fix disabled property

external-olgierd:
Making sure 'disabled' option is respected in listbox component.
Cannot use '[disabled]' property with Angular2 so had to go with 'disabled' flag check before triggering relevant events.
For input text box used '[readonly]' property.


cagataycivici:
I'd prefer adding disabled check inside the method, thank you. Issue is;

https://github.com/primefaces/primeng/issues/2108

PR #2182, opened 2017-02-28
Add focus event to autocomplete #2136

upmauro:
Feature #2136


cagataycivici:
Will also do it for multiple mode which seems to be missing from the PR, thanks.

PR #2183, opened 2017-02-28
Fix DataTable header checkbox behavior #1909

### LINKED ISSUE #1909: DataTable header checkbox should not be checked when filter is applied
```
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://plnkr.co/edit/NtWWnN
_My fork_: [http://plnkr.co/edit/imms7t?p=info](http://plnkr.co/edit/imms7t?p=info)

**Current behavior**
When the datatable has both multiple selection and global filter enabled, you can check multiple items. Then if you filter out items so that all visible items are checked, the header checkbox is also checked. This allows you to deselect all (toggle header checkbox), which removes all from selection instead of only the visible items.

**Expected behavior**

To behave like primeng <=1.1.0, the header checkbox should be clear when a filter is applied so that one cannot deselect all the items in the table.

To see the issue, load [my fork](http://plnkr.co/edit/imms7t?p=preview) of the plunkr and do the following:
1. In the Filter box, type 'audi'
2. Note the header checkbox is checked
3. Uncheck the header checkbox
4. The whole selection is cleared, including the Honda item

Only the Audi items should be removed from the selection list. Or the header checkbox should not be checked. **This problem appears in the patch from PrimeNG 1.1.0 to 1.1.1.**

This bug did not exist before primeng 1.1.1, so the datatable should be modified to behave like it did before that patch.

Windows 7 Pro

* **Angular version:** 2.4.4

* **PrimeNG version:** 1.1.1

* **Browser:** all

 
* **Language:** all

tlofgren:
Instead of replacing or clearing selection, header checkbox now either adds rows not in selection already, or removes only visible, checked items from selection. This behavior is apparent once you use a combination of multiple selection and another feature that changes the dataToRender property. For example, if you use both selectionMode="multiple" and paginator="true", checking the header checkbox will add the current page of data to selection array, rather than completely replace the selection array. Unchecking the header checkbox will only remove the rows on that page from the selection array, rather than emptying the array.

###Defect Fixes
Fixes #1909, and similar defects


cagataycivici:

PR #2198, opened 2017-03-02
Change the treenode.label using [innerHTML] to be display which enables …

ttma1046:
…html content (without script tag) to be set in 'node.label'.

###Defect Fixes

###Feature Requests


ttma1046:
https://angular.io/docs/ts/latest/guide/template-syntax.html#!#property-binding

cagataycivici:
Tree node has templating which is suggested way to place custom html content instead of string based labels.

Issue #2229, opened 2017-03-07
p-calendar timeonly is not showing the correct time in input box


mitalikhundiwala:
For timepicker,

If we provide datetime in ngModel, then **this.pm** boolean value is not being calculated with provided date. It's calculated based on current time.

aclarktcc:
in formatTime this.pm needs swapped for hours > 11 

 if(this.hourFormat == '12' && hours > 11 && hours != 12) {
```

formatTime(date) {
        if(!date) {
            return '';
        }
        
        let output = '';
        let hours = date.getHours();
        let minutes = date.getMinutes();
        let seconds = date.getSeconds();
        
        if(this.hourFormat == '12' && hours > 11 && hours != 12) {
            hours-=12;
        }
        
        output += (hours < 10) ? '0' + hours : hours;
        output += ':';
        output += (minutes < 10) ? '0' + minutes : minutes;
        
        if(this.showSeconds) {
            output += ':';
            output += (seconds < 10) ? '0' + seconds : seconds;
        }
        
        if(this.hourFormat == '12') {
            output += this.pm ? ' PM' : ' AM';
        }
        
        return output;
    }
```

mitosandov:
i have exactly the same issue - i provide 23:00 and on the UI i see 23:00 AM - please release a fix for this!

mitalikhundiwala:
@aclarktcc It seems to be working fine with your provided solution. Any idea by when this fix will be available and have you submitted any pull request?

aclarktcc:
sorry been busy on a handful of projects.  I have not had a chance to build the project and do a pull request.  I was just eyeing the code and saw it while looking into the issue.  If someone has some time to contribute that would be great

SIlver--:
Has anyone been able to make a pull request yet?

LeonardoBluesMan:
Please apply this fix asap!

milosh012:
this is very ugly bug... can you @mitalikhundiwala submit a PR, and hopefully someone will merge it soon

mitalikhundiwala:
I have submitted a PR, requested to merge it soon.

cagataycivici:
This should be fixed in recent PrimeNG release, please try with it and if the issue persists drop a comment with along with a test case based on plunkr below and we'll review again.

http://plnkr.co/edit/Wj39h1?p=preview

PR #2230, opened 2017-03-07
Fix for issue #2229

mitalikhundiwala:
Setting **this.pm** boolean based on passed date value rather than current time.


mitalikhundiwala:
Updated fix as per changes done in primeng base.

cagataycivici:

PR #2251, opened 2017-03-09
Calendar date format as UTC - Fix #1113

### LINKED ISSUE #1113: Calendar issue UTC - beta19
I have a very strange issue with the new Calendar module:

Example:
- The value (date) comes from JSON API

``` typescript
import { Component, OnInit } from '@angular/core';

@Component({
  selector    : 'app-test',
  templateUrl : './test.component.html'
})
export class TestComponent implements OnInit {
  someDate: Date;

  constructor() {}

  ngOnInit() {
    // simulate API call
    setTimeout(() => this.someDate = new Date('2018-08-16T00:00:00.000Z'), 2000);
  }
}
```

``` html
<b>Value:</b> {{someDate}} <b>JSON stringify:</b> {{someDate|json}} <br>
<p-calendar [(ngModel)]="someDate"></p-calendar>
```

**This is the result (_the initial value is not set, but i think this is a known issue at the moment_):**

image

**This is what happens when you select a date:**

image

Basically the bound value is incorrect after the conversion to JSON string (it represents the day before) due to the timezone. But there is no other way to submit the value back to the API.

Is there a way for this component to work (internally) in UTC? Maybe somehow using `dateObj.getTimezoneOffset()` to compensate for the time zone difference?

This is the offending code: 
`components/calendar/calendar.ts:417` - `this.value = new Date(dateMeta.year, dateMeta.month, dateMeta.day);`

---

Edit: added comments

tplessis:
Returned date in calendar component should be in UTC format.

### DIFF
--- a/components/calendar/calendar.ts
+++ b/components/calendar/calendar.ts
@@ -490,7 +490,7 @@ export class Calendar implements AfterViewInit,OnInit,OnDestroy,ControlValueAcce
     }
     
     selectDate(dateMeta) {
-        this.value = new Date(dateMeta.year, dateMeta.month, dateMeta.day);
+        this.value = new Date(Date.UTC(dateMeta.year, dateMeta.month, dateMeta.day));
         if(this.showTime) {
             if(this.hourFormat === '12' && this.pm && this.currentHour != 12)
                 this.value.setHours(this.currentHour + 12);


cagataycivici:
Reverting due to regressions;

e.g. 2426

PR #2259, opened 2017-03-09
Hide remove option from item's when autocomplete is disabled - fix #2252

### LINKED ISSUE #2252: Disabled autocomplete values are still deletable
**I'm submitting a ...**
```
[x] bug report 
[ ] feature request
[ ] support request
```

**Plunkr Case (Bug Reports)**
Sadly your Plunkr (http://plnkr.co/edit/D1AK7i) is currently broken. So i quickly set up my own i hope thats ok.

http://embed.plnkr.co/28rbI5mtumpcCsDEww96/

**Short Description**
So the Problem is the following, if you select some values in the autocomplete component with multiple enabled and disable the input field the user is still able to delete entries from the field as well as from the model. 

**Current behavior**
Steps to reproduce:
 * select some values in the multi autocomplete 
image
 * toggle to disabled
image
 * click the "x"
image
 * see that the item was deleted
image

**Expected behavior**
Usually i would expect that the "x" shouldn't show at all if the autocomplete is disabled. e.g. 
image

* **Operating system:** Windows 10
* **IDE:** IntelliJ 2016.3.2
* **HTTP Server:** Webpack
* **Angular version:** 2.4.9
* **PrimeNG version:** 2.0.2
* **Browser:** [Chrome 56.0.2924.87] i think the browser doesnt change much there
* **Language:** [TypeScript 2.0.10]

upmauro:
Fix #2252


cagataycivici:
Thanks, I'll use ngIf instead.

PR #2278, opened 2017-03-11
Fixed #2277

### LINKED ISSUE #2277: pass null value for validation on p-calendar time only
```
```

**Current behavior**

Cant't pass null value when not required on p-calendar time only.

**Expected behavior**

pass null value when not required on p-calendar time only.

worldcreate:
When not required, pass null value for time only.


cagataycivici:
Why check for required? What difference does it make?

cagataycivici:

PR #2284, opened 2017-03-13
Add firstDay option to shedule module

vadimzham:
###Defect Fixes
Schedule module had no firstDay option. I fixed it.


cagataycivici:
firstDay is part of locale so we rather not duplicate the config again.

https://fullcalendar.io/docs/text/locale/

PR #2424, opened 2017-03-30
Feature Request: Add DataTable Checkbox Header Toggle All Pages Option

saverett:
Add `headerCheckboxToggleAllPages` DataTable property to allow the header checkbox on paged DataTables with checkbox multiple selection enabled to toggle the selection of items across all pages. Previously this was only possible via the `onHeaderCheckboxToggle` event--however, doing so produced strange results where the header checkbox would remain checked even if only the current page's items were selected. This change provides the expected results. This is an improvement over the original fix provided in #1412.
Update demo/documentation to reflect new property.

I created a new property for this so as to not alter the previous behavior of the header checkbox on paged DataTables--however, if that isn't a concern we could simplify this and remove the new property by just having the header checkbox toggle the selection of all items across all pages as its default behavior (for paged checkbox DataTables).

### DIFF
--- a/components/datatable/datatable.ts
+++ b/components/datatable/datatable.ts
@@ -487,6 +487,8 @@ export class DataTable implements AfterViewChecked,AfterViewInit,AfterContentIni
     
     @Output() onHeaderCheckboxToggle: EventEmitter<any> = new EventEmitter();
     
+    @Input() headerCheckboxToggleAllPages: boolean;
+    
     @Output() onContextMenuSelect: EventEmitter<any> = new EventEmitter();
 
     @Input() filterDelay: number = 300;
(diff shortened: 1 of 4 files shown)


SUDHAGARAN:
Hi...When we will expect this feature?..Thank you

saverett:
Thanks for your interest, @SUDHAGARAN. That will be up to @cagataycivici and the primeng team.

Duraimurugan:
can we have an ETA for this? I am desperately looking out for this feature 
@cagataycivici  can you please help on this?

veplechuc:
any news on this feature? It Is already out?

saverett:
@veplechuc, we're still waiting to hear from the primeng team. Hopefully they will merge it soon.

veplechuc:
@saverett thanks! It will be nice to have it soon!!.

adriansnelling13:
Getting this error when I add the property to my datatable:  Can't bind to 'headerCheckboxToggleAllPages' since it isn't a known property of 'p-dataTable'.

Am I missing an import?

lommez:
This feature works with lazy load?

saverett:
@lommez The table selection feature works by adding the elements for the selected rows into a `selection` array. Since not all elements are loaded at once with a lazy load table, this feature will not work with that configuration.

lommez:
Thanks @saverett.

Issue #2420, opened 2017-03-30
[BUG] RouterLink not working in p-panelMenu


igogu:
[ X] bug report => Search github for a similar issue or PR before submitting

**Plunkr Case (Bug Reports)**
It's not a working plunkr, but you can see the code
https://plnkr.co/edit/d6BSgRYbBzk7r4NB4bwy?p=info

**Current behavior**
When you set a routerlink to a child menuItem, this is not redirecting to any component. However, when a parent menuItem has a RouterLink, this can redirect to a component

**Expected behavior**
when a child of menuItem has RouterLink, it should redirect to a component.

**Minimal reproduction of the problem with instructions**
```

 this.itemsPrueba = [                      
                      {
                          label: 'Articles',
                          icon: 'fa-edit',
                          items: [
                              {label: 'Articles Gamma', icon: 'fa fa-dot-circle-o fa-fw', routerLink: ['/index/articles']},
                          ]
                      },
                      
                  ];
<p-panelMenu [model]="itemsPrueba" ></p-panelMenu>
```

For more details, go to the plunkr

**What is the motivation / use case for changing the behavior?**
In Angular 2.x.x and primeNG 2.0.5 was working right, 

**Please tell us about your environment:**
OS: Ubuntu 16.04
IDE: Eclipse Neon
package manager: npm 3.10.9
typescript 2.2.2

* **Angular version:** 2.0.X
Angular version: 4.0.0 & 4.0.1 (its not working with no one of those)

* **PrimeNG version:** 2.0.X
PrimeNG version:** 4.0.0-rc.2

* **Browser:** [all ]

 
* **Language:** [all]

snowping:
I experience probably the same issue with <p-menubar> using angular 4.0.0 and V4 rc2.  the routerLink redirection was working with angular 4.0.0 and V4 rc1.

erlopezh:
The same issue here, using angular 4.0.0 and V4 rc2 the routerLink redirection does not working

derchirurg:
Same here using angular 4.0.1 and primeng 4.0.0-rc.2.
The link is incorrect. It has href="#".

xyz765:
The same issue with MenuBar used to work fined and stopped working after updating, links shows "/#" not errors in console.

Brampage:
@xyz765 Same for me.

ImproSnake:
I got the same Issue but only on the SubMenuItems. The routerLink on a top Level MenuItem works just fine.

[...]

PR #2448, opened 2017-04-03
fixed bug #2420

mamarillo:
routerLink attribute was not being set in PanelMenu subitems. Fixed


rengare:
merge this fix please

cagataycivici:
I think this is already fixed.

PR #2452, opened 2017-04-04
add paginator to treetable

ddvkid:
I'm not sure adding a paginator to treetable will be considered as a significant change, as most of the features have already been in datatable, 

I also finished the global filter for treetable as well, see [ng-treetable](https://github.com/ddvkid/ng-treetable), if you are interested, I can submit another pr for it later on.


cagataycivici:
Thanks but it is a huge change, we'll do it after 4.0.x and will check this PR on the way. Thank you. Issue to monitor;

https://github.com/primefaces/primeng/issues/2500

PR #2459, opened 2017-04-06
Add loading property to show the loading layer

HVND:


cagataycivici:
Reverting due to conflict with the loading property already.

Issue #1623, opened 2016-12-19
Custom content inside BlockUI


papiroca-tm:
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
very short of opportunities to add icon like fa-spinner when activated blockUI

suryakand:
Yes, this is much needed.

papiroca-tm:
Of course it should be optional

mohan1304:
Yes I do need this feature. We can make it as optional

PR #2461, opened 2017-04-06
 #1623 Add icon like fa-spinner when activated blockUI

parmarques:
Feature Request  #1623

Added a optional content child and z-index input to BlockUI.
I decide to not put a icon or a label,  only add a generic content child.


cagataycivici:
Why do you have;

@ContentChild(TemplateRef) template: TemplateRef<any>;

Also we don't generally expose z-index, I'll add the support manually.

PR #2463, opened 2017-04-07
Enforce a fixed tree structure when dragging/moving nodes based on type

filipvh:
Using the array dropTypeLock one defines which type is allowed on which depth while moving/dragging.

### DIFF
--- a/components/calendar/calendar.ts
+++ b/components/calendar/calendar.ts
@@ -214,6 +214,8 @@ export class Calendar implements AfterViewInit,OnInit,OnDestroy,ControlValueAcce
 
     @Input() required: boolean;
 
+    @Input() basedOnUTC: boolean; 
+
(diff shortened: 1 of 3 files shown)


filipvh:
treetypelock

cagataycivici:
I don't understand what this PR does, what feature does it add?

filipvh:
it enforces a certain form/shape based on the string -type- to your tree
while dragging
So you can only drop a certain type on a certain level
['A','B','C'] -> on leven 0 (top level) only elements of type A, level 1
only type B

|_A
|_A
|  |_B
|  |  |_C
|  |  |_C
|  |_B
|_ A

so one cannot place a A on the 1st or 2nd level, only the 0th when moving
nodes around


2017-04-12 13:14 GMT+02:00 Çağatay Çivici <notifications@github.com>:

PR #2472, opened 2017-04-10
Next month grey date was throwing Nan/Nan/Nan

anilkhatri1305:
I have added this fix so that when you are opening the calendar and you have set the attribute [selectOtherMonths]="true", you will be able to select the next month grayed out dates. Without this fix you will get NaN/NaN/NaN.

### DIFF
--- a/components/calendar/calendar.ts
+++ b/components/calendar/calendar.ts
@@ -582,6 +582,7 @@ export class Calendar implements AfterViewInit,OnInit,OnDestroy,ControlValueAcce
         }
         else {
             m = month + 1;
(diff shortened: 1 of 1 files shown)


anilkhatri1305:
Thanks cagataycivici for merging the fix...When can we expect the RC3 release?

cagataycivici:
RC3 is due tomorrow and final is due 17th.

anilkhatri1305:
@cagataycivici Hey this fix will be there on the 2.0.7 release? When are you going to release 2.0.7?

PR #2494, opened 2017-04-12
docs(dialogdemo): add Event API of visibleChange

SangKa:
The docs of dialog lack the event api of visibleChange.


cagataycivici:
That is not a callback, it is to use [(visible)] syntax, property name is visible and documented. onShow and onHide are the official callbacks.

SangKa:
I got it!

PR #2503, opened 2017-04-12
Fix dateParse function

vazham:
###Defect Fixes
Now dateParse returns Date that not equal with Date that becomes from clicking in calendar. That fix returns equal Data parsed from input and date picked in calendar.


cagataycivici:

PR #2507, opened 2017-04-12
Add between (range) to filter constraints (datatable.ts)

aviger:
Feature Request
Add 'between' filter to the filterConstraints property of the DataTable.
this feature enables to filter according to a specific range. I chose this name because of the same operator in SQL that named 'between' (But it also make sense to call it 'range').

This filter pass values within a given range. The values can be numbers, text, or dates.
it gets also only a one range limit.

I created a new property for this so as to not alter the previous behavior of the filterConstraints or any another behavior of the DataTable.

I saw some requests for a feature like this for example:
https://forum.primefaces.org/viewtopic.php?f=35&t=50389

### DIFF
--- a/components/datatable/datatable.ts
+++ b/components/datatable/datatable.ts
@@ -1520,9 +1520,7 @@ export class DataTable implements AfterViewChecked,AfterViewInit,AfterContentIni
         },
         
         between(value: any, filter: any): boolean {
-            if (filter === undefined || filter === null ||
-                filter.start === undefined || filter.start === null ||
-                filter.end === undefined || filter.end === null) {
+            if (filter === undefined || filter === null) {
                 return true;
             }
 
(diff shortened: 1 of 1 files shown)


aviger:
**Feature Request**
Add 'between' filter to the filterConstraints property of the DataTable.
this feature enables to filter according to a specific range. I chose this name because of the same operator in SQL that named 'between' (But it also make sense to call it 'range').

This filter pass values within a given range. The values can be numbers, text, or dates.
it gets also only a one range limit.

I created a new property for this so as to not alter the previous behavior of the filterConstraints or any another behavior of the DataTable.

I saw some requests for a feature like this for example:
[https://forum.primefaces.org/viewtopic.php?f=35&t=50389](https://forum.primefaces.org/viewtopic.php?f=35&t=50389)

cagataycivici:

PR #2515, opened 2017-04-13
Support 'contains' in multi select drop down.

zkonstas:
###Defect Fixes

When filtering with 'contains' match mode with multi select the application throws an error because the filter var is an array and not a string

###Feature Requests


cagataycivici:
I can't replicate the original issue, how can filter be an array.

zkonstas:
You can replicate by having a datatable which has a column with a "multi-select" filter and set filterMatchMode="contains"

PR #2521, opened 2017-04-13
Fixed #2375

### LINKED ISSUE #2375: Autocomplete, choosing item from dropdown doesn't update ng-model
```
```

**Plunkr Case (Bug Reports)**
https://www.primefaces.org/primeng/#/autocomplete

**Current behavior**
The item doesn't get selected and ng-model is not set.

**Expected behavior**
Item should get selected and ng-model updated with the selection.

1. Make sure nothing has been selected before
2. Under Advanced click on dropdown 
3. Click on an option
4. Bug occurs and item is not selected and ng-model is not updated.

Windows 10

* **Angular version:** 2.0.X

The one used on the website

* **PrimeNG version:** 2.0.X

The one used on the website

Happens in IE11, IE10. Doesn't happen in IE9
 
Typescript

elvinmeza:
Basically the issue occurs by the way IE (10/11 so far i tested it) handles the blur event, which was causing the autocomplete panel being closed when the user clicked on the scrollbar or even the items on the list without firing the selectItem() method, hence the model was not being updated.

###Defect Fixes
Fixes [2375](https://github.com/primefaces/primeng/issues/2375)

### DIFF
--- a/components/autocomplete/autocomplete.ts
+++ b/components/autocomplete/autocomplete.ts
@@ -134,6 +134,8 @@ export class AutoComplete implements AfterViewInit,DoCheck,AfterViewChecked,Cont
     panelVisible: boolean = false;
     
     documentClickListener: any;
+
+    panelBlurListener: any;
     
     suggestionsUpdated: boolean;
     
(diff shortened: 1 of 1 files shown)


cagataycivici:
I believe there should be a fix without IE check. Will comment on issue.

Issue #2454, opened 2017-04-05
Multi filter support for Dropdown


xuanshenbo:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**

**Current behavior**

In a dropdown list with a filter, it only filters items' label, but not their value.
For example, I have:
```typescript
reasons = [];
reasons.push(label: '1910', value: {reason: 'Date not valid', severity: 'Warning'});
```

When I type 'Date ...' in the search box, no results will be returned.
**Expected behavior**

User would be able to search with both label and value that's being displayed.

**Minimal reproduction of the problem with instructions**

**What is the motivation / use case for changing the behavior?**

One item can have not only its label but may also have some of its corresponding values. Think of the Content with Filters example on PrimeNG website, imagine the logos are some ids of the car brands, together with the brand name. Users would expect to be able to search everything that is being displayed. But in fact users are only able to search the brand names, but not the ids.
**Please tell us about your environment:**

* **Angular version:** 2.0.X

* **PrimeNG version:** 2.0.X

primeng 2.0.2
* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

* **Node (for AoT issues):** `node --version` =

pcavacas:
I created pull request #2530 that implements this functionality.  It adds a new property to the dropdown component called filterOnId which is a boolean property.  If this is not present or set to false then everything works as it does today.  If this is set to true then when you filter it will filter on the label and also the stringfied version of the value property, i.e. the following test both work
this.cars.push({label: 'Audi', value: 'BMW'});
this.cars.push({label: 'Audi', value: {value: 'BMW'}});

cagataycivici:
@Mrtcndkn Introduce filterBy="field1, field2.subfield" please to customize it.

Mrtcndkn:
@cagataycivici @xuanshenbo It's in progress.

PR #2530, opened 2017-04-15
Issue #2454 Optionally filter by ID

pcavacas:
###Defect Fixes
Issue 2454

###Feature Requests
Added a new bindable property FilterOnId to allow optionally filtering on Id and then added code to implement it.


cagataycivici:
We'll do this differently with filterBy property where you can provide multiple properties from the value as well.

Issue #2533, opened 2017-04-16
PanelMenu child items should follow their own routerLink rather than their parent's


ChristopherJohnston:
**I'm submitting a ...**  (check one with "x")
```
[ X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://plnkr.co/edit/RtEBVUDiB1yQRhDpOQB7

**Current behavior**

When providing an array of MenuItem objects containing routerLink properties, the rendered panel menu uses the routerLink from the parent in the child elements, 

see lines 40 and 47 of primeng/components/panelmenu/panelmenu.ts

<a *ngIf="!item.routerLink" 
and
<a *ngIf="item.routerLink" [routerLink]="item.routerLink" 

should be
<a *ngIf="!child.routerLink" 
and
<a *ngIf="child.routerLink" [routerLink]="child.routerLink" 

**Expected behavior**

child menu items should use their own routerLink, if provided.

**Minimal reproduction of the problem with instructions**
see plunkr. provide MenuItem array to the model property of the panelmenu with routerLinks, e.g.:

{ label: 'item1', routerLink: ['/item1'], items: [ { label: 'item1.1', routerLink=['/item1/1'] }, { label: 'item1.2', routerLink:['/item1/2'] }]

The panelMenu will render but the routerlinks on the subitems will be for the parent (ie /item1) rather than their own (ie /item1/1 and /item1/2)

**What is the motivation / use case for changing the behavior?**
this is a bug. the links do not render correctly.

**Please tell us about your environment:**
windows and osx, visual studio code, npm and angular-cli

* **Angular version:** 2.0.X
4.0.0

* **PrimeNG version:** 2.0.X
4.0.0-rc.2

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
ie11, chrome, safari
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
Typescript

* **Node (for AoT issues):** `node --version` =   
[...]

PR #2534, opened 2017-04-16
Fix for #2533 - sub-items should use their own router link rather than the parent

ChristopherJohnston:
###Defect Fixes
https://github.com/primefaces/primeng/issues/2533


ChristopherJohnston:
I'd really appreciate if someone looked at this. It should be an easy integration...just 2 lines of code changed to fix an annoying bug.

Many thanks

cagataycivici:

PR #2550, opened 2017-04-19
Dialog drag sticks to mouse cursor

argnist:
If you mousedown on close button and then move mouse cursor the mouseup event is not fired and dragging is not stopping. So you shouldn't propagate mousedown event on close button.


cagataycivici:
We might need propagation for other components like dropdown to close their panels. Will fix it differently, thanks.

PR #2556, opened 2017-04-19
Add onSelect event to string parsing

vadimzham:
###Feature Requests
Now it's impossible to detect when user entered correct date from the keyboard - here is no external event to detect this.
I believe that this parsing event is equal to selecting date - factically date getting new value, UI updating.

**This PR adds onSelect event emit when parseValueFromString gettting Date object.**


cagataycivici:
Shouldn't we expose onInput or similar instead of onSelect which is confusing?

vadimzham:
Yes, probably it will be better.

cagataycivici:
See #2780

Issue #1827, opened 2017-01-17
dataTable#rowStyleClass mess up with row styles


borovsky:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://embed.plnkr.co/9eOs9R5GrAApHR0q3m8i/

**Current behavior**
rowStyleClass mess up with styles, e.g. it possible to have inconsistent styles: ui-datatable-even ui-datatable-odd in the same time, ui-state-highlight if no selectable rows

**Expected behavior**
rowStyleClass correctly generate class attribute

**Minimal reproduction of the problem with instructions**
In plunk, change one of the columns to match other column in the same row (triggers row class change). 
When row class changes, class list for row became inconsistent (even and odd class, highlight, etc)

* **Angular version:** 2.4.3

* **PrimeNG version:** 1.1.4

* **Browser:** all
 
* **Language:** all

* **Node (for AoT issues):** `node --version` =

laserus:
I have faced the same and thought that it could be due to usage of [class] and [ngClass] in the same element. Fixed this with moving rule for [class] to ngClass, and unfortunately [ngClass] cannot work as this: 
```
[ngClass]="{'my-class': condition, [dt.getRowStyle(rowData, index)]:true}"
```
Thus change the ngClass to array of classnames.

cagataycivici:
Fixed via PR

Issue #1416, opened 2016-11-24
DataTable : rowStyleClass doesn't play well with Angular architecture.


jdpearce:
I'm submitting a ...
```
[ x] bug report ?
```
This is an implementation issue, really.

**Current behavior**
The DataTable inlines the function that you use for determining the rowStyleClass. Considering this will almost certainly be coming from the component presenter class, it should at the very least bind that function to the class instance somehow.

It also doesn't play well with the style scoping. If I put my row styles in the component.scss file, Angular will scope them to be say, "oddRow[_ngcontent-smo-15]" and "evenRow[_ngcontent-smo-15]", which is determined at runtime (afaik) - the DataTable needs to modify the classes returned with this attribute selector. The issue above makes solving this second part challenging, to say the least - I haven't yet figured out a way to inject ElementRef into an inlined function.

Lastly, it seems to call the rowStyleClass function *a lot*. I was clicking a button elsewhere on a page containing a DataTable and the whole table was being rechecked on each click. I don't know if this is down to the way Angular 2 notifies regarding drawing the screen, but it seems excessive.

(Apologies for creating what might be 3 issues in one - I think the architecture of this feature needs to be revisited, so solving any one issue might well invalidate the other two)

**Please tell us about your environment:**
Windows 7, Visual Studio Code, npm, angular-cli

* **Angular version:** 2.2.1

* **PrimeNG version:** 1.0.0

* **Browser:** Chrome 54
 
* **Language:** TypeScript 2.0.3

* **Node (for AoT issues):** `node --version` = v6.9.1

jdpearce:
Just a thought, but would it be possible to invert the behaviour? i.e. make the row styling event based and we have a row style handler? You could then pass in the row data, a handle to the row element itself, perhaps with a setStyle method the handler can use which already knows what the element reference is?

Thavarajan:
actually you can use the '/deep/ '
that will almost eliminate the attribute in angular2

jdpearce:
@Thavarajan You might need to explain that one. I have absolutely no idea how your comment relates to this issue 😣

cagataycivici:
[...]

PR #2583, opened 2017-04-24
Improves usage rowStyleClass and fixes an issue with strange multiselect behavior

laserus:
###Defect Fixes
This fixes this issue: https://github.com/primefaces/primeng/issues/1827

###Feature Requests
As explained here: https://github.com/primefaces/primeng/issues/1416
rowStyleClass with a callback function and running in wrong (not-in-the-user-component) scope breaks Angular2/4 thinking. This PR introduces rowStyleMap, which can be modified by the user's component, and provided for DataTable as input. 

And fixes many requests on improvement of rowStyleClass usage.

### DIFF
--- a/components/datatable/datatable.ts
+++ b/components/datatable/datatable.ts
@@ -156,9 +156,12 @@ export class ColumnFooters {
                     </span>
                 </td>
             </tr>
-            <tr #rowElement *ngIf="!dt.expandableRowGroups||dt.isRowGroupExpanded(rowData)" [class]="dt.getRowStyleClass(rowData,rowIndex)"
+            <tr #rowElement *ngIf="!dt.expandableRowGroups||dt.isRowGroupExpanded(rowData)"
                     (click)="dt.handleRowClick($event, rowData)" (dblclick)="dt.rowDblclick($event,rowData)" (contextmenu)="dt.onRowRightClick($event,rowData)" (touchend)="dt.handleRowTouchEnd($event)"
-                    [ngClass]="{'ui-datatable-even':even&&dt.rowGroupMode!='rowspan','ui-datatable-odd':odd&&dt.rowGroupMode!='rowspan','ui-state-highlight': dt.isSelected(rowData)}">
+                    [ngClass]="[even&&dt.rowGroupMode!='rowspan'? 'ui-datatable-even':'',
+                                odd&&dt.rowGroupMode!='rowspan'?'ui-datatable-odd':'',
+                                dt.isSelected(rowData)? 'ui-state-highlight': '', 
+                                dt.getRowStyleClass(rowData,rowIndex)]">
                 <ng-template ngFor let-col [ngForOf]="columns" let-colIndex="index">
                     <td #cell *ngIf="!dt.rowGroupMode || (dt.rowGroupMode == 'subheader') ||
                         (dt.rowGroupMode=='rowspan' && ((dt.sortField==col.field && dt.rowGroupMetadata[dt.resolveFieldData(rowData,dt.sortField)].index == rowIndex) || (dt.sortField!=col.field)))"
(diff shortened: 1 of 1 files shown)


laserus:
Please, find some explanation on design in https://github.com/primefaces/primeng/issues/1827

laserus:
A little bit of explanation how I use the rowStyleMap feature:
**template.html:**
```
<p-dataTable 
            ...
            [rowStyleMap]="rowStyleMap"
            (onRowClick)="selectClicked($event.data)"
            dataKey="my_collection_id">
</p-dataTable>
```
**component.ts:**
```
	rowStyleMap = {};
	selectClicked(rowData) {
		this.selectedRow= rowData;

		this.rowStyleMap = {
			[rowData.my_collection_id] : 'table-info'
		};
	}
```

That's all, now I can tag rows as selected with bootstrap styles and do some operations on top of it.

cagataycivici:
Please provide separate PRs and also add documentation.

cagataycivici:
No need, I'll introduce rowStyleClassMap to datatable. Thanks for the idea.

cagataycivici:
How to use it though, does user need to prepopulate rowDataMap with all potential values? If so that doesn't sound good, for example how to change row color if year of a car is older than 2010? How would you create a rowStyleMap?

laserus:
@cagataycivici rowDataMap only depends on dataKey, not all possible values, if some row has a year older than 2010 and user wants to reflect it, then user put into the map that the row should have specific color. The user has all control of own data as well as many @Output events from the datatable itself, and can react in any way he wants to the changes (reflecting it into rowDataMap). 

So it adds flexibility and correct scoping compared to rowDataStyle.

laserus:
@cagataycivici what about fix for [ngClass] and [class]. When they are on the same element for me and others it produces incorrect styles (could be some angular bug???). Will you fix it or do I need a separate PR for that?

cagataycivici:
Could you please create a separate issue for ngClass and class. We'll discuss rowDataStyleMap after 4.0 since it is a bit too late for a big change.

laserus:
@cagataycivici you mean PR, not issue? Issue is already created by somebody else: https://github.com/primefaces/primeng/issues/1827

cagataycivici:
But that seems to be about dataTable#rowStyleClass again?

cagataycivici:
By the way we use [class] and [ngClass] in many places in DataTable, regardless of rowStyleClass is there any known issue for this? It seems to work fine except rowStyleClass which is subject to rewrite.

laserus:
@cagataycivici it is different. My improvement from callback function to object map is related to this: https://github.com/primefaces/primeng/issues/1416

But there is a second error, that is when rowStyleClass function is used it tries to applies [class]="dt.getRowStyleClass(rowData,rowIndex)" and that thing breaks [ngClass]. 

Yes, probably it is only because callback scoping breaks something related to angular. I do not know. But the fix that I suggest about that, fixes this issue without need to redesign rowStyleClass yet.

laserus:
I have tested locally different options, and whenever I have tried to bind to [class] this error happens... I haven't found a way to fix it othewise. No zone or changedetector helped.

laserus:
@cagataycivici after some thinking, have you ever seen binding to class like that `[class]` and not some specific `[class.my-class]` or `[ngClass]`? Previously, I have seen this direct binding to class, but not anymore. I have checked the code at angular/angular and do not find it...

E.g., like that: https://github.com/angular/angular/blob/8d300ffbfc7d46aa543bf04d0b41c26c6e2e5a9f/aio/src/app/layout/nav-item/nav-item.component.ts#L15
and  https://github.com/angular/angular/blob/8d300ffbfc7d46aa543bf04d0b41c26c6e2e5a9f/aio/src/app/layout/nav-item/nav-item.component.html#L2

Issue #2584, opened 2017-04-24
Captcha component assumes reCAPTCHA is loaded synchronously, but it's not


anabulsi:
**I'm submitting a ...**  
```
[ X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**

PrimeNG's captcha component assumes the reCAPTCHA script is already loaded. The problem is that reCAPTCHA is by default designed to be loaded asynchronously, there is no way to force it to load synchronously. The fact that captcha component is currently working is a mere coincidence, try hitting refresh repeatedly and it will fail at some point. It will fail once the angular script is executed before the reCAPTCHA script. 

To further explain, try hitting this URL in your browser: https://www.google.com/recaptcha/api.js?render=explicit  You will notice the resultant script is not the actual reCAPTCHA but a loader script. Inspect the loader script and you will notice this statement: "po.async = true". The async behavior is forced by Google and there is no way to load it synchronously. 

**Expected behavior**

reCAPTCHA offers onload callback which should be used to get notified once all required dependencies are loaded, Captcha component should listen to this callback and only then it should render.

**Minimal reproduction of the problem with instructions**

1. Open https://www.primefaces.org/primeng/#/captcha

2. keep hitting refresh and at some point it will fail

**Please tell us about your environment:**
Windows 8, VS Code, Chrome

* **Angular version:** 4.0.3

* **PrimeNG version:** 4.0.0-rc.3

* **Browser:** all 

 
* **Language:** TypeScript

* **Node (for AoT issues):** `node --version` =   6.9.0

anabulsi:
just submitted a PR to resolve this issue. Hopefully it will make it to 4.0.0 Final as captcha component is currently broken

PR #2589, opened 2017-04-24
Captcha component can now be triggered through onload parameter of reCAPTCHA

anabulsi:
captcha component now registers a callback that can be triggered through onload parameter of reCAPTCHA

###Defect Fixes
This fixes https://github.com/primefaces/primeng/issues/2584

now you can add reCAPTCHA script as follows:

`<script src="https://www.google.com/recaptcha/api.js?render=explicit&onload=recaptchaLoaded" async defer></script>`

once its loaded, it will call recaptchaLoaded which will be registered globally by captcha component and it will render the captcha.

Also, the callback name is configured through the "callback" attribute

### DIFF
--- a/components/captcha/captcha.ts
+++ b/components/captcha/captcha.ts
@@ -18,6 +18,8 @@ export class Captcha implements AfterViewInit {
     @Input() tabindex = 0;
     
     @Input() language: string = null;
+     
+    @Input() callback = "recaptchaLoaded";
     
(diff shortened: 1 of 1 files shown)


cagataycivici:
The PR does not use the callback right? Or is recaptchaLoaded a built-in name of recaptcha?

anabulsi:
you are right. In the PR I focused on resolving the issue and totally forgot about the demo page :)

PR #2600, opened 2017-04-25
Dropdown: Add `filterPlaceholder` attribute

adrienverge:
Add the ability to set a placeholder to the `<input>` used to filter entries in the list.

Example for `<p-dropdown filterPlaceholder="type here...">`:

filterplaceholder


cagataycivici:
Bettter to use attr.placeholder but I'll change it thanks.

PR #2639, opened 2017-04-28
Calendar - Adding restrictNavigation option

howlettt:
If minDate or maxDate is set, this option will prevent the user from navigating to months outside the valid months


cagataycivici:
Code has alert statements.

PR #2642, opened 2017-04-28
Fixed #1913

### LINKED ISSUE #1913: Positioning dialog on open
**I'm submitting a ...
```
```

**Current behavior**
Any new p-dialog **always** opens in the center of viewport. This behavior is defined at:
```
ngAfterViewChecked()  ->  show() -> center()
```

**Expected behavior**
If position is set in the styles (or some inputs) do not center the dialog. This behavior could be strange for large screens. Also it could be beneficial to inherit the position of dialog from it's parent position. E.g. if button clicked and the dialog is opened, it will be opened where the <p-dialog> elements are placed. 

Usability. Flexibility.

SIlver--:
Added the ability to position the dialog where indicated by the top and left style attributes.

#1913

### DIFF
--- a/components/dialog/dialog.ts
+++ b/components/dialog/dialog.ts
@@ -56,7 +56,11 @@ export class Dialog implements AfterViewInit,OnDestroy {
     @Input() width: any;
 
     @Input() height: any;
-        
+
+    @Input() openingPositionLeft: number;
+
+    @Input() openingPositionTop: number;
+
     @Input() contentStyle: any;
 
     @Input() modal: boolean;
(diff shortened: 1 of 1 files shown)


SIlver--:
Wasn't sure what the naming convention should have be for the `openingPositionTop` and `openingPositionLeft` attributes.

cagataycivici:
Will rename to positionTop and positionLeft.

ofekdeitch:
I think we should check if
``` this.openingPositionLeft != undefined && this.openingPositionTop != undefined```
instead of
``` this.openingPositionLeft && this.openingPositionTop```

If `openingPositionLeft` or `openingPositionTop` will be 0 then the statement will return false even though there is a value (!)

What do you guys think?

PR #2670, opened 2017-05-02
Tooltip: enable dynamically updating all input properties

HVND:


cagataycivici:
Not sure if I understand this PR, it just checks for disabled.

HVND:
@cagataycivici Not quite. This change will make it possible to recreate the tooltip with updated input properties

HVND:
Without (re)sets data-bound input properties:

screencaptureproject14

With (re)sets data-bound input properties:

screencaptureproject15

PR #2674, opened 2017-05-02
added several properties to different components.

p3root:
-  DataTable changes
-- added rowReadonlyField for setting rows readonly
-- added rowVisibleField to set the visibility of different rows
- ContextMenu
-- Added enabled input 
- Tree
-- added id field to tree to set the html id flag on the node
-- changed width to 100% -> this should be changed to an property as well!


cagataycivici:
We cannot except one giant PR with various changes, instead please create separate issues with detailed explanation of what the property does and we'll review. Thank you.

Issue #2703, opened 2017-05-04
Add a "Message closed" event to the Growl


DevWurm:
**I'm submitting a ...**
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**What is the motivation / use case for changing the behavior?**
Sometimes the application state has to track which messages are 'acknowledged' (or at least displayed long enough) to the user. An event when a message is closed from the Growl would make this possible. In addition an optional `id` field on the Message would make it possible to uniquely identify a closed message.

sudheerj:
The PR looks good to me.But AFAIK I'am not sure how much it is going to be useful for users.

DevWurm:
Thank you :)
In my opinion it provides an interface, which may be usefull for some users and doesn't interfere those, who don't need it.

SchneMa:
I think this could be a lighter or simpler alternative to a "confirm"dialog. So u could use it for some status messages or something like that and if closed manually mark them as read and otherwise as unread.

Something like those popup messages on Facebook.
Clicking the cross on those popups mark them as read and if you dont click them and the message closes over time its marked as unread.
Just as a little possible Use-Case example that came to mind.


I'm not in need of this till now but I think this could be an intresting feature in the future :)

PR #2704, opened 2017-05-04
Add a "Message closed" event to the growl (primefaces/primeng#2703)

DevWurm:
- Adds the `onMessageClosed` event to the growl component which fires,
if a message is removed from the view
- Add an optional `id` field to the `Message` interface, to enable growl
users to identify closed messaged
- Adapt the showcase

###Feature Requests
This is an implementation of primefaces/primeng#2703.
It is a very small feature without any braking API changes involved and it is tested on the showcase page.

### DIFF
--- a/components/common/api.ts
+++ b/components/common/api.ts
@@ -42,6 +42,7 @@ export interface Message {
     severity?: string;
     summary?: string;
     detail?: string;
+    id?: any;
 }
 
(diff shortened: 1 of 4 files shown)


DevWurm:
So what do.you think? :)

cagataycivici:
Will merge manually as no need to add it demos.

PR #2718, opened 2017-05-06
Steps disabled behvaiour

sudheerj:
Steps need to be disabled  based on menu item's disabled property


cagataycivici:
I don't understand what the PR does clearly so will implement it ourselves.

sudheerj:
We need to cover two things in this issue

        1. Disable the step component through MenuItem's disable condition(First condition)
        2. The active Index shouldn't be emitted when it disabled.So we need to put event emit statement after disable condition(Second condition)

Please let me know if it is not making sense

Issue #2666, opened 2017-05-02
VirtualScroll and selection


mcherb:
**I'm submitting a ...**  (check one with "x")
```
[] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**

**Virtualscroll mode:**
when the selection @Input change outside the datatable (whitout the row click) the dataToRender do not change

**Expected behavior**
Load (and display) the data that surround the selection (and show the selection too)

**What is the motivation / use case for changing the behavior?**
I have two DataTables that one is containing a subset (100 rows) of the other (full data, > 10000 rows). I want to get a bidirectional selection on the tables (if i select one the other is synchronized with the right selection)

**Please tell us about your environment:**
Linux, intellij idea 17, npm

* **Angular version:** 4.0.0

* **PrimeNG version:** 4.0.0

* **Browser:** all
 
* **Language:** TypeScript 2.1.0

* **Node (for AoT issues):** `node --version` =   v6.2.2

cagataycivici:
Not sure, this is an opinionated behavior, we need more feedback so if more users demand it, we'll reconsider. Thank you.

PR #2725, opened 2017-05-08
[#2666] render selection (outside the click event) when not lazy and …

mcherb:
…selectionMode='single'

###Defect Fixes
See #2666

###Feature Requests


cagataycivici:
Not sure, this is an opinionated behavior, we need more feedback so if more users demand it, we'll reconsider. Thank you.

PR #2733, opened 2017-05-09
Make it possible to make the date selection at UTC midnight. eg: If y…

filipvh:
Make it possible to make the date selection at UTC midnight. eg: If you would choose 2017-01-01T00:00:00 while in a +02:00 timezone, the UTC timestamp sent to backend would be 2016-12-31T22:00:00, which is the day before.

###Defect Fixes
Problems with timezones


cagataycivici:
Will merge manually thanks

PR #2734, opened 2017-05-09
[Feature] Add custom cell class feature for table cell

jorisvergeer:
###Feature Requests
I want to be able to set a custom background and other styling in some cells, based on some data in the row on some cells.

I added an property to the column definition where i can set a property to look for in the rowdata. In the rowdata this could be a dynamic ot statis property. 

(If there is another way I can easily set custom styling to the whole cell, i am interrested in that too)

### DIFF
--- a/components/common/shared.ts
+++ b/components/common/shared.ts
@@ -83,6 +83,7 @@ export class Column implements AfterContentInit{
     @Input() colspan: number;
     @Input() style: any;
     @Input() styleClass: string;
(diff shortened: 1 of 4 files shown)


cagataycivici:
I'd suggest using cell templating instead to have full control over cell content and styling.

jorisvergeer:
That does not work because the celltemplate is rendered inside a parent <td> which has padding. A want to keep the padding but be able to for example change the background of the whole cell.

PR #2756, opened 2017-05-11
package.json - showcase dep as devDeps, gulp build after install

gilhanan:
installing showcase dependencies as devDependencies and gulp building after installing, which make building much easier, so we can rid off from [contribution guide](https://github.com/primefaces/primeng/wiki/Building-From-Source)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,13 +3,15 @@
   "version": "4.0.1-SNAPSHOT",
   "scripts": {
     "tsc": "tsc",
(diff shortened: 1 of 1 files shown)


cagataycivici:
I know but we rather not publish any devDeps, if there is going to be a dep config, it should better be peerDeps although it doesn't install them to build PrimeNG. We'll keep as it is for now.

Issue #2565, opened 2017-04-20
Feature Request - Make SelectItem label more robust. Allow using an object with multiple props


brettkc:
Currently the `SelectItem` interface is defined as such:

`
export interface SelectItem {
  label: string;
  value: any;
}`

This is very limiting when trying to create anything beyond a very basic template. Would be great to be able to pass in an object of values that then can be used in a custom template, dropdown for example.

battmanz:
I have discovered that it's not really a limitation having to conform to the `SelectItem` interface. You can use a pipe to transform the data. For example:

```typescript
interface Person {
  name: string;
  id: number;
}

@Pipe({ name: 'toSelectItem' })
class ToSelectItemPipe implements PipeTransform {

  public transform(people: Person[]): SelectItem[] {
    if (!people) return undefined;
    return people.map(p => ({ label: p.name, value: p.id }));
  }
}

const people= [
  { name: 'Joe', id: 1 },
  { name: 'Mary', id: 2 },
  { name: 'Bob', id: 3 },
  { name: 'Suzy', id: 4 }
];
```
```html
<p-dropdown [options]="people | toSelectItem"></p-dropdown>
```

brettkc:
This is a clever way to transform data to match the interface. This still leaves you with only access to a 'label' and 'value' when constructing a template (on a dropdown for example). Sure you can concatenate some sort of big string with all the data you want on the label but you're then very limited to how you can present it via a template.

brettkc:
Perhaps this would be reasonable?

```
export interface SelectItem {
    label: string;
    value: any;
    [propName: string]: any;
}
```

battmanz:
Sorry, I've been meaning to get back to you. You can already do what you want. This works, for example:

```typescript
interface Person {
  firstName: string;
  lastName: string;
  id: number;
}

// An intersection type that combines Person and SelectItem.
type PersonSelectItem = Person & SelectItem;

@Pipe({ name: 'toSelectItem' })
class ToSelectItemPipe implements PipeTransform {

  public transform(people: Person[]): PersonSelectItem[] {
    if (!people) return undefined;

    // Create an object that has all of the Person properties AND the SelectItem properties.
    return people.map(p => Object.assign({ }, p, { label: p.firstName + ' ' + p.lastName, value: p.id });
  }
}

const people= [
  { firstName: 'Joe', lastName: 'Jones', id: 1 },
  { firstName: 'Mary', lastName: 'Robinson', id: 2 },
[...]

PR #2794, opened 2017-05-16
Add index signature to the SelectItem interface to allow for more fle…

brettkc:
…xibility when building custom templates around the interface

Related to Issue #2565


cagataycivici:
We'll reconsider if more users demand it. Thank you.

PR #2797, opened 2017-05-16
fix(DataTable): check all pages value null check

### LINKED ISSUE #2796: TypeError: Cannot read property 'length' of null for DataTable with headerCheckboxToggleAllPages true
```
```

**Current behavior**
Before the DataTable's `value` property is initialized, the `get allSelected()` function is throwing a `TypeError` when the `headerCheckboxToggleAllPages` property is set to `true`.

**Expected behavior**
The `get allSelected()` function should also check to ensure the DataTable's `value` property is also defined (not `null`) before comparing its length to the `selection` length.

Create a DataTable with paging, multiple select, and the `headerCheckboxToggleAllPages` property is set to `true`. Make sure the DataTable's value doesn't resolve right away.

The new `headerCheckboxToggleAllPages` DataTable option is missing a null check.

Windows

* **Angular version:** 4.x

* **PrimeNG version:** 4.0.1

* **Browser:** All
 
* **Language:** All

saverett:
Fixes #2796

### DIFF
--- a/components/datatable/datatable.ts
+++ b/components/datatable/datatable.ts
@@ -1313,7 +1313,7 @@ export class DataTable implements AfterViewChecked,AfterViewInit,AfterContentIni
     
     get allSelected() {
         if(this.headerCheckboxToggleAllPages) {
-            return this.selection && this.selection.length === this.value.length;
+            return this.selection && this.value && this.selection.length === this.value.length;
         }
         else {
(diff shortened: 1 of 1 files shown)


saverett:
@cagataycivici: Sorry I missed this in #2424. Please review and merge as soon as you get a chance. Thanks!

PR #2810, opened 2017-05-18
restrict ime input

hryktrd:
When IME is active, I can input multibyte charactor.
I fix it by replace other than numbers.


hryktrd:
It is solution for #2811

cagataycivici:

hryktrd:
This is fixed in other PR #3113.

PR #2818, opened 2017-05-18
Add positionRight and positionBottom properties to Dialog

### LINKED ISSUE #2816: Add positionRight and positionBottom to Dialog
```
```

To allow complete control over Dialog absolute positioning. (e.g. pin the dialog to the bottom or right of the screen)

mkennedy3000:
Fixes #2816


cagataycivici:
It is a bit uncommon as positioning is usually top, left in box model. If more users demand it, we'll reconsider. Thank you.

Sumragen:
@cagataycivici 
I'm agree with @mkennedy3000 . That PR make sense

PR #2837, opened 2017-05-20
Steps dead code with disabled menu item

sudheerj:
The disabled condition is already defined at the beginning of the method. So the specific code block is not reachable.


cagataycivici:
Fixed manually

PR #2850, opened 2017-05-22
Fixes #2846 - Growl's onClose event references wrong message when mes…

### LINKED ISSUE #2846: Growl's onClose event references wrong message when message is closed manually
```
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/0LOFUtsI2HU5ltoBozsh?p=preview

**Current behavior**
When the user manually closes a Growl message, the onClose callback receives an object containing the next message in the messages array instead of the message that was actually closed. Say we remove the second message on the array then the event is called with the third one. This does not happen when the message is automatically removed by the component.

**Expected behavior**
The callback should receive the exact message removed.

1. Go to the plunkr link provided above
2. Wait a second for the 4 messages to pop up
3. Close the 4 message in order
4. Check the console output
(It should log the four message in the correct order but instead it will log messages 2, 3, 4 and then undefined.)

It seems like unintended behaviour

* **Operating system:** Linux Mint 18.1
* **IDE:** Visual Studio Code
* **package manager:** NPM
* **Angular version:** 4.0.3
* **Angular CLI version:** 1.0.3
* **PrimeNG version:** 4.0.1
* **Browser:** Chromium 58 | Firefox 53
* **Language:** TypeScript 2.2

davisb10:
Fixes an issue where the removal of a growl message was returning the incorrect message from the onClose event.

### DIFF
--- a/components/growl/growl.ts
+++ b/components/growl/growl.ts
@@ -87,9 +87,9 @@ export class Growl implements AfterViewInit,OnDestroy {
         this.domHandler.fadeOut(msgel, 250);
         
         setTimeout(() => {
-            this.value = this.value.filter((val,i) => i!=index);
-            this.valueChange.emit(this.value);
             this.onClose.emit({message:this.value[index]});
+            this.value = this.value.filter((val,i) => i!=index);
+            this.valueChange.emit(this.value);           
         }, 250);        
     }
     


cagataycivici:
This is fixed with another PR, thanks.

https://github.com/primefaces/primeng/pull/2852

PR #2869, opened 2017-05-24
Calendar always populates AM

sudheerj:
We need to remember two points to know the reason for this change
   1. this.pm code snippet always give time zone based on default date or current time zone.
   2. this.hours is always less than 11 due to previous code lines

### DIFF
--- a/components/calendar/calendar.ts
+++ b/components/calendar/calendar.ts
@@ -1035,7 +1035,7 @@ export class Calendar implements AfterViewInit,AfterViewChecked,OnInit,OnDestroy
         }
         
         if(this.hourFormat == '12') {
-            output += hours > 11 ? ' PM' : ' AM';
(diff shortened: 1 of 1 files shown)


ohsomofo:
I apologize in advance for my ignorance about how changes make it into releases but I'm pretty new to all this. However, the same issue seems to be present in 4.0.3. Except now the initial display in the input may show PM but as soon as you start changing the time or select am/pm it will no longer change the underlying date object when changing between am/pm.

Example (using 4.0.3): 
http://plnkr.co/edit/mEctar?p=preview

sudheerj:
There are many issues raised in tracker related to calendar AM/PM(speicific to dates, timepicker etc).Can you please create a new issue with the details. By the way, I can replicate the mentioned issue.

ohsomofo:
Thank you. I created issue #3056 and did as you suggested.

PR #2870, opened 2017-05-24
MultiCheckbox requires model check

sudheerj:


cagataycivici:
We should create a new model instead if non exists.

PR #2888, opened 2017-05-26
fix multiselect label are (null) bug

giscafer:
detail see issue: https://github.com/primefaces/primeng/issues/725


cagataycivici:
Merged with a new differ

PR #2933, opened 2017-05-31
Calendar - Handle the case that minDate is in the future month of current month

darius-iko:
**Issue type**: UX issue
**Example case**: when minDate is set to any date in June, the overlay still show May month but nothing can be select from the day option.
Description: This pull request is to fix the calender into the way when minDate is set to any date in June meanwhile the current month is May, the calendar overlay will show June month instead of May month.
To test it: Rollback the calendardemo.ts to before "Original calendar demo" commits


cagataycivici:

PR #2961, opened 2017-06-01
feat(tooltip): add show/hide delay options

alexandremoore:
###Feature Requests
Add an optional show/hide delay to tooltip


cagataycivici:
This is fixed 4.1.RC1

PR #2968, opened 2017-06-02
feat(treetable): Add option to define extend button column

tomaszzmuda:
###Feature Requests
Add a property to change column of extend button on treetable


cagataycivici:
I like it!

cagataycivici:
Added doc as well.

Issue #1897, opened 2017-01-24
Editable dropdown caret jumps


m-tomczyk:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
(seen on primeng showcase) => http://www.primefaces.org/primeng/#/dropdown

**Current behavior**

1. Entering new value into editable dropdown
2. Changing cursor position in middle of value being entered
3. Typing again causes one letter to appear in desired place and next cursor jump to the end of input field

**Expected behavior**

Typing in middle of already entered value keeps cursor in place

* **Browser:** Chrome 55 | Firefox 50

koesper:
I've also run into this problem, but haven't been able to fix it yet.
We probably need to do something like this
https://gist.github.com/romuleald/f1b980d5ad01b9f525e8e0244f91711a

koesper:
With some tinkering, I've got it to work, but perhaps someone  (@cagataycivici?) will be able to clean up my solution without using a nasty setTimeout

Update [onEditableInputChange()](https://github.com/primefaces/primeng/blob/master/components/dropdown/dropdown.ts#L343)
 with this code:
```
onEditableInputChange(event) {
    // save the current cursor position
    let cursorPosition = event.target.selectionStart;
    
    this.value = event.target.value;
    this.updateSelectedOption(this.value);
    this.onModelChange(this.value);

    // set the cursor back to the original position
    setTimeout(()=>{
        event.target.selectionEnd = cursorPosition;
    }, 0);

    this.onChange.emit({
        originalEvent: event,
        value: this.value
    });
}
```
I'm not quite sure why I need the setTimeout at the moment, I expect I could be able to pass the cursorPosition to the this.onChange, but I havent found where to catch it again.

If you want, I could create a pullrequest, but I don't think my solution is good enough.

cagataycivici:
Will check for 4.1, thank you.

koesper:
@cagataycivici, does not seem to be fixed?
screencapture: 
https://app.blisk.io/screenRecords/shared/d9e5d471-9103-43da-9e29-b7594ac378cc

cagataycivici:
@koesper That is master branch right?

[...]

PR #3000, opened 2017-06-06
Rough patch for #1897 Editable dropdown input jumps

koesper:
###Defect Fixes
[https://github.com/primefaces/primeng/issues/1897#issuecomment-302093441](https://github.com/primefaces/primeng/issues/1897#issuecomment-302093441)

Remembers the editabledropdown input location, and sets it back


cagataycivici:
We need to think about a solution with a timeout.

koesper:
@cagataycivici Yes, I completely agree!
Any thoughts? This minor bug is a major problem for my application :-(

PR #3005, opened 2017-06-06
Add official editEvent method in Schedule component

TheBrockEllis:
This PR adds the official `editEvent(event)` method of FullCalendar to the Schedule component. It renames the internally used `editEvent` method to `_editEvent()` and also renames the internal `findEvent()` method as well. Updates to showcase documentation are included.

Please let me know if there is anything I can do to help get this change merged. Thanks so much for all your hard work!

### DIFF
--- a/src/app/components/schedule/schedule.ts
+++ b/src/app/components/schedule/schedule.ts
@@ -221,7 +221,7 @@ export class Schedule implements DoCheck,OnDestroy,OnInit,OnChanges,AfterViewChe
                 });
             },
             eventDrop: (event, delta, revertFunc, jsEvent, ui, view) => {
-                this.updateEvent(event);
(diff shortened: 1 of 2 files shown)


cagataycivici:
What is the use case? Thanks.

TheBrockEllis:
In my case, I'm storing additional data on the event (like `capacity` for an event that people can register for). When a user wants to register for an event, I must edit the `capacity` and redisplay the event on the calendar. According to [FullCalendar](https://fullcalendar.io/docs/event_data/updateEvent/), it needs the original event object and not a recreated one, so I would need to use  the `updateEvent()` proper to make those changes to the event without destroying the event and creating a brand new event.

Hope that makes sense. =)

ohtwadi:
+1

Issue #1257, opened 2016-11-10
Autocomplete inputStyle/inputStyleClass doesn't work for multiple mode


stalky14:
When passing in the class required for the 'inputStyleClass' attribute in autocomplete element, the class does not get applied. 
Inspector shows that the class is applied to the "class" of the input but the "ng-reflect-ng-class" returns [object Object].

I would have expected "ng-reflect-class-name" to then refer to the class.

jock-dalby:
I found on version 2.0.5 using autocomplete, inputStyleClass works fine for non-multiple inputs but not for multiples. I upgraded and tried on version 4 but still not working. Instead it just attaches the given class name to 'inputstyleclas's and' ng-reflect-input-style-class' attributes on the p-autocomplete element. Is this the intended behaviour? Is there a different approach for targeting the input element while allowing multiple selections?

PR #3016, opened 2017-06-07
TabMenu - close tab ability

rachelbr:
###Feature Requests
I'm using prime 4 for angular
Could you add same functionality as tabview has to tabmenu? For example ability to close tabs

thanks a lot


rachelbr:
I'm tring to open request 
TabMenu - close tab ability #3016
but you close it depent #1257 -there is no connection between them ,
probably i didn't open the request as expected

can you take a look on it?

Issue #2731, opened 2017-05-08
Accessibility issues on Paginator/DataTable and ConfirmationDialog


mandy6720:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://plnkr.co/edit/D1AK7i

**Current behavior**

Not really a bug, per se, but Paginator, DataTable and Modal do not meet standards for accessibility (WCAG/508 compliance).

Specifically:

-- lack of meaningful paginator alt text
When using a screen reader, the buttons for the paginator do not have meaningful text to indicate what their function is to a visually impaired user. 
(Suggestion: add aria-labels)

-- sortable columns do not have focus 
A user using keyboard/tab navigation is unable to tab to (focus) the sort arrows on a sortable column - they are unable to toggle the sort order in the DataTable. This doesn't necessarily make the table unusable for a visually impaired user, but would be a nice to have.
(Suggestion: add a tabIndex=0 to the arrow icon with an aria-label describing the sort direction)

-- focus on selectable row (shorthand selection column in DataTable)
When using the shorthand/build-in select column, a user using keyboard/tab navigation is unable to tab to/navigate to the checkboxes in that column.
(Suggestion: add a tabIndex=0 to the checkbow inputs)

-- focus on pop-up buttons (modal)
When a modal is opened, focus should go to the first interactable element on the modal window - otherwise, the user would have to tab through all of the elements on the current page before getting to the modal window.
(Suggestion: when the window is summoned, use .focus() or similar to programmatically set the focus to the upper-right close button or other interactable element)

**Expected behavior**

1. All buttons on paginator would have meaningful text for screenreaders.

2. Sort arrows for columns would be tab-able and have meaningful text.

3. The build-in selection column's checkboxes would be included in the tab navigation.

4. When a modal window is opened, focus would move to the upper righthand close button (x) or first other interactable element.

**Minimal reproduction of the problem with instructions**

[...]

PR #3024, opened 2017-06-08
Added aria-labels

dcesiel:
###Defect Fixes
This fixes adds alt text to paginator as requested in issue #2731 

"-- lack of meaningful paginator alt text
When using a screen reader, the buttons for the paginator do not have meaningful text to indicate what their function is to a visually impaired user.
(Suggestion: add aria-labels)"


cagataycivici:
But these labels are in English, we should better make them attributes or so for customization.

PR #3050, opened 2017-06-10
Dialog fires unwanted onHide event

mmikeyy:
When the dialog is created with property visible set to false, the setter for this property calls dialog#hide, which fires the event. The onHide event should only fire when a visible dialog is closed, not when it hasn't yet been visible.

The proposed change sets the property preventVisibleChangePropagation to true initially, preventing the call to dialog#hide (which fires the event) when creating the dialog. It is set back to false in dialog#AfterViewInit, which could also be done in dialog#ngOnInit if this method was added.

###Defect Fixes

###Feature Requests

### DIFF
--- a/src/app/components/dialog/dialog.ts
+++ b/src/app/components/dialog/dialog.ts
@@ -131,7 +131,7 @@ export class Dialog implements AfterViewInit,OnDestroy {
     
     preWidth: number;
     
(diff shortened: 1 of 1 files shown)


cagataycivici:
This is fixed in 4.1.RC1

Issue #3038, opened 2017-06-09
Fix: Dropdown's dimension is not synchronous when 'options' is set in lazy.


kojisaiki:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Sorry, i cannot use Plunkr by filtering.

**Current behavior**

When `options` parameter is set in lazy like using setTimeout(), Observable etc. , dropdown's dimension is not synchronous with new options.

![image](https://user-images.githubusercontent.com/10954150/26956478-fb5ac66e-4cf8-11e7-935b-1daec7c1be02.png)

its code is following:
```
        setTimeout(() => {
            this.cities2 = [];
            this.cities2.push({label:'New York', value:{id:1, name: 'New York', code: 'NY'}});
            this.cities2.push({label:'Rome', value:{id:2, name: 'Rome', code: 'RM'}});
            this.cities2.push({label:'London', value:{id:3, name: 'London', code: 'LDN'}});
            this.cities2.push({label:'Istanbul', value:{id:4, name: 'Istanbul', code: 'IST'}});
            this.cities2.push({label:'Paris', value:{id:5, name: 'Paris', code: 'PRS'}});
        }, 3000);
```

**Expected behavior**

After update `options`, in lifecycle 'AfterViewChecked', need to emit the exist function `updateDimensions()`.

![image](https://user-images.githubusercontent.com/10954150/26956604-ec6bdcfa-4cf9-11e7-9c6a-623d81690eff.png)

its code has following changes at `dropdown.ts`:
```
    set options(opts:SelectItem[]) {
        this._options = opts;
        this.optionsToDisplay = this._options;
        this.updateSelectedOption(this.value);
        this.optionsChanged = true;
        this.dimensionDiffered = true;     // new boolean field
    }
...

    ngAfterViewChecked() {

        if(this.dimensionDiffered) {
            this.updateDimensions();
            this.dimensionDiffered = false;
        }
        
        if(this.shown) {
...
```

**Minimal reproduction of the problem with instructions**

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **Angular version:** 2.0.X

[...]

PR #3059, opened 2017-06-12
Fix: Dropdown's dimension is not synchronous when 'options' is set in lazy.

kojisaiki:
From Issue: #3038


cagataycivici:

PR #3071, opened 2017-06-13
Frozen right

keyeMyria:
###Feature Requests
Add feature for freeze columns on the right.


cagataycivici:
Seems to have conflicts and issue with end of line. Also this is one of those features that we prefer to do it ourselves. Thank you.

PR #3076, opened 2017-06-13
Support for draggable HammerJs touch events (panstart, panend, panmove)

realbluehead:
###Defect Fixes

###Feature Requests
Added support for touch events pan when dragging a dialog.


cagataycivici:
Doesn't this cause a hard dependency to hammerjs?

realbluehead:
Well, I suppose it does but the code is not intrusive so if you don't use HammerJs it still works fine like before and if you use HammerJS in your project, it's a nice addition. Showing dialogs in a tablet and not being able to drag them is kind of frustrating.. Also it seems that HammerJs is the standard when developing with Angular and touch/ gesture events.

cagataycivici:
If more users demand it, we'll reconsider thank you.

PR #3086, opened 2017-06-14
(merge latest master)Fix: Dropdown's dimension is not synchronous when 'options' is set in lazy

kojisaiki:
merged latest master

same as this pr:
https://github.com/primefaces/primeng/pull/3059


cagataycivici:
Reused optionschanged callback instead, thank you.

PR #3087, opened 2017-06-14
Customized paginator position

sudheerj:


cagataycivici:
What is the issue here? It works for me as it is.

sudheerj:
It works without issues. But two things we need to consider.
1. It should work with paginatorPosition values "top","bottom" and "both" only.As of now it works with any paginatorPosition value. i.e, paginatorPosition="up" or paginatorPosition="down"
2. If we add paginatorPosition="none" to use external customized paginator then also it won't disable the existing paginator

Klinton90:
Please accept this PR. Currently there is no way to have UI pagination with custom paginator.

Current code is ambiguous:
`paginatorPosition!='bottom' || paginatorPosition =='both'"`
If value is `bottom` - top paginator is hidden, if anything else (`none`, `Bottom`, `tOp`) - it is shown. Second condition can be removed and logic stays same.

But that prohibits to use custom paginator. Whatever value is set - built-in paginator is always shown on page. Currently all pagination parameters set to `p-dataTable` (`[rows]="10" [pageLinks]="3" [rowsPerPageOptions]="[5,10,20]"`) are ignored, until `[paginator]=true` is set.

Klinton90:
Thank you!

Issue #1895, opened 2017-01-24
Calendar ehancement : support of  ISO 8601 date


rlouissidney:
Hello,

Is it possible to have an option to support ISO Date via the ISO 8601 format ?
Perhaps have a separate presentation format and input value format ?  

I use a lot of webservice with JSON data and the ISO 8601 seems to be a standard.

Thanks !

blinkxzero:
This issue is actually breaking my application. 

I need the input to be formatted as an ISO string for my database (JSON), but the p-calendar input only accepts Date types or it throws an exception.

jneveux:
Below is a test on my side (version 2.0.0). 

Date coming from the server REST end point:
`{"dateOne":"2017-03-01T10:45:09.230Z"}`
Calendar configuration:
`<p-calendar dateFormat="dd/mm/yy" [(ngModel)]="dateObjectTwo.dateOne" ></p-calendar>`
Stacktrace:
`Date type string
error_handler.js:54EXCEPTION: Uncaught (in promise): Unexpected literal at position 2ErrorHandler.handleError @ error_handler.js:54next @ application_ref.js:348schedulerFn @ async.js:93SafeSubscriber.__tryOrUnsub @ Subscriber.js:234SafeSubscriber.next @ Subscriber.js:183Subscriber._next @ Subscriber.js:125Subscriber.next @ Subscriber.js:89Subject.next @ Subject.js:55EventEmitter.emit @ async.js:79NgZone.triggerError @ ng_zone.js:333onHandleError @ ng_zone.js:294ZoneDelegate.handleError @ zone.js:334Zone.runGuarded @ zone.js:142_loop_1 @ zone.js:540drainMicroTaskQueue @ zone.js:549ZoneTask.invoke @ zone.js:420
[...]

PR #3098, opened 2017-06-14
Added ISO 8601 support to calendar

dcesiel:
###Defect Fixes
Attempts to solve #1895

The user can now set the output date type using the `@Input dataType` with either date, string or ISOString.  They can also set the input type using `@Input inputDataType` with either date, string or ISOString.


cagataycivici:
Calendar is already very complex with addition of built-in UTC support so we've decided not to add support for new formats at the moment, conversion can be done at your component instead as a workaround.

DannyTheHippo:
please add this, mongodb is one of the most used databases and it returns ISO dates.

Issue #3058, opened 2017-06-12
input is repeated on Spinner


hryktrd:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://plnkr.co/edit/VZbdZfQ7SuTV87tTkbYR?p=preview

**Current behavior**

On spinner, When I input 123 with iphone, 123123 is inputted.  And I test on chrome with IME-ON,
same probrem  occurs. (I input 123 with IME-ON, And Press F10 to change single-byte character.)

**Expected behavior**

**Minimal reproduction of the problem with instructions**

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **Angular version:** 4.0.0

* **PrimeNG version:** 4.0.0

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

* **Node (for AoT issues):** `node --version` =   6.0.0

ryanstroh:
Having this issue on Samsung s8 and samsung s5 chrome browser

arjunshankar:
This issue has been in review for sometime. Is the PR a viable fix ... ?

arjunshankar:
@hryktrd were you able to find any workarounds for this ??

hryktrd:
@arjunshankar 
Yes, I sent pull request https://github.com/primefaces/primeng/pull/3113 but that hasn't merged.

arjunshankar:
@hryktrd 
I tried using your code by extending the component. I can still reproduce the bug on Galaxy S8/S7/S6/S5 on the chrome browser while using the samsung keyboard.

The default Samsung browser seems to work fine. Also if you use any other keyboard.. like swype. that works fine as well.

What is the root cause of this issue ??
Thanks in advance.

hryktrd:
[...]

PR #3113, opened 2017-06-16
restrict that input is repeated on Spinner when IME is active.

hryktrd:
fix for #3058


cagataycivici:
Does it still happen with 4.2.1? I can't replicate on iPhone.

cagataycivici:
Can't replicate.

Issue #2773, opened 2017-05-15
appendTo property to TieredMenu component


ddyrcz:
**I'm submitting a ...**  (check one with "x")
```
[*] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

I found the same issue as here:
https://github.com/primefaces/primeng/issues/619

Where the problem occurred in the p-menu control and the solution to this issue is appendTo="body" attribute

The same problem occurs in p-tieredMenu. But there is no 'appendTo' attribute

PR #3132, opened 2017-06-18
Fix issue #2773 Popup-TieredMenu doesn't work in positioned divs

ddyrcz:
Fix issue #2773 by adding appendTo property


cagataycivici:
Merged manually, thank you.

PR #3148, opened 2017-06-20
Calendar: Add enabledDates prop to specify which dates should be selectable

jasmouse:
This property is the inverse of disabledDates. By default all dates are disabled except those passed in via enabledDates.


cagataycivici:
But there is also disabledDates, so this does the reverse?

jasmouse:
Yes this does the reverse of disabledDates. In my scenario we have a requirement to show the user which dates they can select based on whether a report was submitted on that date. These dates are random, but may only be once or twice a month. With this property it would prevent me from having to pass in all of the  disabled dates over what could be a two year period.

cagataycivici:
Sorry but there are +305 −247 line changes so we are unable to review, please update PR and we'll review again.

PR #3160, opened 2017-06-21
MultiSelect preselection may not work correctly 

am4apps:
###Defect Fixes
When trying to initialize the MultiSelect Component, if it's no the same instance of the object(options vs formControlName), fails to locate it on the options of the MultiSelect Component, and show null as label in the input.


am4apps:
taking a look to the findSelectionIndex method
the solution should be like this
     **if(option.value == val || (this.objectUtils.equals(option.value, val, this.dataKey)) {**

cagataycivici:
Seems we've fixed it already.

PR #3204, opened 2017-06-26
Feature: added infinite scroll for autocomplete

rengare:
detection of scroll being at bottom in autocomplete component


cagataycivici:
This triggers angular change detection and will make the page slow, please use NgZone to fix this.

PR #3207, opened 2017-06-26
Feature: added ability to navigate through tokens in autocomplete multiple mode.

meff34:
Add some accessibility to autocomplete component in multiple mode.

User can select some token by pressing `left arrow` or `right arrow`, and remove selected token by pressing `backspace` (with displace to left) or `delete` (with displace to right).
All the new stuff is hidden under a flag `enableControlKeys`, wich is `false` by default.


meff34:
Uh... guyz?
Someone?

cagataycivici:
There are +265 −113 line changes so I'm afraid we are unable to review what has changed.

PR #3171, opened 2017-06-21
Update inputswitch.ts

wilsonmoraes:
###Defect Fixes
Resolves ExpressionChangedAfterItHasBeenCheckedError Occurs in the angular 4.2.3 or >

PR #3209, opened 2017-06-26
Fix https://github.com/primefaces/primeng/pull/3171

HVND:
Fix https://github.com/primefaces/primeng/pull/3171


cagataycivici:
Duplicate of 3247

Issue #3151, opened 2017-06-20
radioButton: custom object comparison function for checked/unchecked state


just-jeb:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
**Current behavior**

Objects are always compared with `==` operator (`this.checked = (value == this.value)`)

**Expected behavior**

Sometimes there is a need to compare objects (for `checked` state) with a custom comparison function.
It would be nice to have it as optional property of the component.

* **Angular version:** 4.1.3

* **PrimeNG version:** 4.0.1

* **Browser:** all

 
* **Language:** all

I can create a PR for this if needed.

cagataycivici:
A custom function sounds good.

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

Issue #3149, opened 2017-06-20
dropdown: custom object comparison function


just-jeb:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
**Current behavior**

Objects are always compared with `objectUtils.equals`

**Expected behavior**

Sometimes there is a need to compare objects (for selection) with a custom comparison function.
It would be nice to have it as optional property of the component.

* **Angular version:** 4.1.3

* **PrimeNG version:** 4.0.1

* **Browser:** all

 
* **Language:** all

I can create a PR for this if needed.

cagataycivici:
We have dataKey for many data components, we can do the same here, thoughts? When it is present, ObjectUtils uses that.

meltedspark:
Specifically for my case `dataKey` is not robust enough. And I don't think I'm alone in this.
`dataKey` allows just comparison by nested field. If I want to compare by two fields, it's impossible.
What I currently do to bypass this issue (and not only in dropdown) is overriding `Dropdown.prototype.findOptionIndex` method. But frankly I would expect it to be supported on the framework level.
What do you think? I can provide an implementation.

cagataycivici:
Maybe we should just ask for equals() function in the objects? If there is equals function in the object, we can just call it.

meltedspark:
This might work. But it also has downsides.
What if the object does have `equals` function but I still want to compare it (specifically for the dropdown selection) differently?
Also, what gets the higher priority, `dataKey` or `equals`? You wouldn't know, it depends on what the user wants.
In addition, the objects being compared are mostly pure data objects, in many cases received from the server. I wouldn't want to put on each such an object equality function just to make it selectable properly in dropdown.
Anyway, even allowing object comparison by equality function will be much better than now. I just don't think it's robust enough.
Thoughts?

cagataycivici:
A custom function sounds good.

meltedspark:
Shall I create a PR?

twmobius:
I think `p-dropdown` should follow the way `SelectControlValueAccessor` is using `[compareFn]` to change the comparison algorithm

PaddingtonTheBear:
[...]

PR #3214, opened 2017-06-27
Custom equality function

meltedspark:
#3149, #3151, #3150


meltedspark:
@cagataycivici , will it merged?

cagataycivici:
We'll reconsider when more users demand it, currently there is none afaik. thank you.

lhoussin:
As in Angular : https://angular.io/api/forms/SelectControlValueAccessor
a function like compteWith in primeng is needed, hope it will be implemented very soon.

PaddingtonTheBear:
This feature would solve a lot of headaches when dealing with complex objects in dropdowns, multiselect, listbox, etc. 

I want to have my complex objects be the values of my dropdowns, but then comparing the selected item to the dropdown values becomes very specific. If any key of the objects is missing, or different, then the comparison fails (long story short, my dropdown values have more nested properties in them then the actual selected value).

Ideally, I would just do comparisons between the values and the selected using "id", but this is not simple without a custom comparison function. I can have the DataKey defined to use "id," but then this causes the primeng-utils "equals" to be messed up, as the `obj1` is my selected complex object, and then `obj2` is now simply an `id` string which fails the comparison checks done by `this.resolveFieldData` (it doesn't account for comparison an object to a string).

I can try to work around this by transforming the values in my dropdown from complex objects into the `{label, value}` format, but then when I select a value from the dropdown, the selected object is no longer in a format I expect and would have to untransform it. 

Either the `this.resolveFieldData` method needs to be fixed for this use case, or simply extend the functionality by including a `[compareWith]` custom function on all components that need it (similar to how core Angular and other UI libraries do it).

Issue #3228, opened 2017-06-28
Growl's initial messages do not disappear


nikitaulshin:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/EHeKXANxLt769hlprm69?p=preview

**Current behavior**
If array of messages have initial value then messages will not disappear.

**Expected behavior**
Initial messages disappears after [life] ms.

**Minimal reproduction of the problem with instructions**
http://plnkr.co/edit/EHeKXANxLt769hlprm69?p=preview

**What is the motivation / use case for changing the behavior?**
Growl might have initial messages

**Please tell us about your environment:**
Ubuntu 16.04, WebStrom

* **Angular version:** 2.0.X
Latest

* **PrimeNG version:** 2.0.X
Latest

* **Browser:**
Chrome
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5] TS 2.4

* **Node (for AoT issues):** `node --version` =   

Hello!

Thanks for your awesome library.
I found a small but very annoying issue with growl.
I created a pull request to fix it [(#3230)](https://github.com/primefaces/primeng/pull/3230).

Thanks for your attention!

piotrpliszko:
It also happens when empty array is initial value of messages array. I need to set it again to empty array right before adding a new message.

Example:
```typescript
export class AppComponent {
  msgs: Message[] = [];

  constructor(private logger: LoggerService) {
    this.logger.infoPopupChanged.subscribe(infoPopup => {
      this.msgs = []; // I need to add this line
      this.msgs.push({ severity: 'info', detail: infoPopup });
    });
  }
}
```

nikitaulshin:
It happens when the component itself has not been initialized yet, so @piotrpliszko 's example is also correct.

harmonmi:
FYI... still having this issue even after updating to 4.1.1

PR #3230, opened 2017-06-28
Fix disappear for initial messages

nikitaulshin:
[#3228](https://github.com/primefaces/primeng/issues/3228)


cagataycivici:
3228 seems to be fixed already.

PR #3250, opened 2017-06-29
Fix #3246 by adding a ui-treenode-content-selected class to the ui-treenode-content div when a node is selected.

### LINKED ISSUE #3246: Ability to style entire selected treenode
```
```

**Current behavior**

When a node is selected in a tree, the class ui-state-highlight is added to the span inside the node that contains the text, allowing just the text to be styled when selected. The div that contains the entire node does not change:

image

**Expected behavior**

It is desirable to allow users to optionally style the entire tree node content when selected:

image

I would suggest that a new class is added, not ui-state-highlight which would contain style already, but an empty class like `ui-treenode-content-selected` which would allow the user (and theme creators) to add their custom style to it.

I will submit a pull request shortly with this minor change.

The flexibility of styling is a great characteristic of PrimeNG and is great for most components but the tree component is generally less flexible and our designer would like to "flex his muscles" and apply a design change for a selected node in the tree that is applied to the entire node as displayed above.

OS: Windows 7 Pro SP1 64-bit  
IDE: Visual Studio Code (latest)  
node: 6.9.5  
npm: 4.3.0  

* **Angular version:** 4.2.4

* **PrimeNG version:** 4.0.3

* **Browser:** all

 
* **Language:** TypeScript 2.3.4

ceottaki:
This allows someone to style the entire node div, not just the internal span, when a node is selected.

###Defect Fixes
#3246 

###Feature Requests
This is a very small feature request that adds a class to an element when it is selected to allow for styling. No actual styles have been added.


cagataycivici:
Merged manually

Issue #1830, opened 2017-01-18
Row Grouping is not working as expected


smedavarapu:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/YvhroNjH0A9WkZUMy4kc?p=preview

**Current behavior**
Throws find of undefined error when rows are grouped by rowspan.

Taking a look at the source code of datatable.ts, seemed to be initializing the columns in ngAfterContentInit. But, 'this.columns.find' is executed in ngDoCheck, the step before the content is initialized. 

**Expected behavior**
Rows must be grouped on groupField

**Minimal reproduction of the problem with instructions**

http://plnkr.co/edit/YvhroNjH0A9WkZUMy4kc?p=preview

when groupField, sortField and rowGroupMode are passed to p-dataTable, the entire table breaks. It throws 'Cannot read property 'find' of undefined' in the console.

 
**What is the motivation / use case for changing the behavior?**
Row Grouping on the Table

**Please tell us about your environment:**
 Win 10, VS2015, NPM, Asp.Net Core RTM

* **Angular version:** 2.0.0

* **PrimeNG version:** 1.1.1/1.1.4

* **Browser:** Chrome, Firefox, Edge
 
* **Language:** TypeScript 2.0

* **Node (for AoT issues):** `node --version` =   v4.2.5

cagataycivici:
Can't replicate with 4.1, please provide a plunkr based on;

http://plnkr.co/edit/Qi2Nw2imLpDUF8NJinYw?p=preview

PR #3267, opened 2017-06-30
row grouping issues fixed

brpradeepprabhu:
fixed issues Row Grouping is not working as expected #1830
added sortByGroupField function and code formatting in webstrom creates the file changes


brpradeepprabhu:
alignment is mismatching. reverting my pull req

PR #3270, opened 2017-07-01
#1830 Row grouping issues is fixed

### LINKED ISSUE #1830: Row Grouping is not working as expected

brpradeepprabhu:
###Defect Fixes
#1830

### DIFF
--- a/src/app/components/datatable/datatable.ts
+++ b/src/app/components/datatable/datatable.ts
@@ -552,8 +552,6 @@ export class DataTable implements AfterViewChecked,AfterViewInit,AfterContentIni
     @Input() sortField: string;
 
     @Input() sortOrder: number = 1;
-    
-    @Input() groupField: string;
 
     @Input() multiSortMeta: SortMeta[];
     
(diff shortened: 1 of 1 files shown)


cagataycivici:
There is code duplication at sortByGroupField since it does the same as singleSort, we should use singleSort method or so.

cagataycivici:
Can't replicate with 4.1, please provide a plunkr based on;

http://plnkr.co/edit/Qi2Nw2imLpDUF8NJinYw?p=preview

PR #3311, opened 2017-07-05
add tableStyleClass and tableStyle to TreeTable

steve3d:
DataTable has these `tableStyleClass` and `tableStyle` input so users can customize the class and styles of the table, the TreeTable also use table but don't have these inputs, this makes it hard to make the treetable match the style of our own design.

So this PR add these two input to TreeTable.


slinlee:
@steve3d @cagataycivici  nice addition. Are there plans to merge the similar change for the TurboTables? (PR #5679  for issue #5189  )

junmao813:
i want to un subscribed this kind email,please

Slin Lee <notifications@github.com> 于2018年6月14日周四 上午7:18写道：

PR #3315, opened 2017-07-05
doc(growl): fix broken code sample

ycintre:
Method name in html template  (`clear()`) didn't match ts name (`hide()`)


sudheerj:
The commit is partial.It should be same for both component.html and component.ts

ycintre:
The ts file looks good to me and the interactive demo works.
The fix is only for the embeded code snippet which doesnt works if
copy-pasted as is.

Can you explain more what you mean ?

Le mer. 5 juil. 2017 18:22, Sudheer Jonna <notifications@github.com> a
écrit :

sudheerj:
In all the places(.html,.ts) we used it as clear. So let's use it as clear in all the places.

Issue #1733, opened 2017-01-04
Tree - child nodes may not accessible


mscudlik:
**I'm submitting a ...**  (check one with "x")
```
[ x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
#1731 is also valid for the Tree component

yigitfindikli:
I’m unable to replicate with the latest PrimeNG. If the issue still persists with the new version please create a new issue with our stackblitz template.

PR #3316, opened 2017-07-05
Tree - child nodes may not accessible #1733

CogepDominicFaucher:
Fix to a bug that made every tree node with a caret/arrow for children even when there was no children or the parameter leaf was disable if you did not import a theme. Also fix the issue: Tree - child nodes may not accessible #1733 when you have a theme imported if you remove the following lines in primeng.css (wich i can't find on github) :
.ui-tree .ui-treenode.ui-treenode-leaf > .ui-treenode-content > .ui-tree-toggler {
visibility: hidden;
}

Sorry the fix is ugly but i did not have enough time to make it better.

###Defect Fixes

###Feature Requests


CogepDominicFaucher:
I noticed after the pull requested that i forget to import a theme, this is why i had the bug. But i had the issue: Tree - child nodes may not accessible #1733 . Wich this can be use to fix that issue if you remove the following lines in primeng.css (wich i can't find on github) : 
.ui-tree .ui-treenode.ui-treenode-leaf > .ui-treenode-content > .ui-tree-toggler {
    visibility: hidden;
}

cagataycivici:
The toggler is there for alignment purposes, removing it misaligns it, I think we need a new fix.

Issue #3296, opened 2017-07-04
ContextMenu and Menubar doesn't render style or styleClass for MenuItems


ambower:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://plnkr.co/edit/endh8jqxFyZMhblrQ5zF?p=preview

**Current behavior**
I would like to set inline styles and style classes to the context menu items. When I try and set styles using the MenuItem model, they do not get rendered to the page. 

**Expected behavior**
When style or styleClass are set on the MenuItem, those styles should be applied to the respective context menu item.

**Minimal reproduction of the problem with instructions**
See the Plunker example above. I would expect the menu item in the context menu to have the 'tallyho' class applied to it.

**What is the motivation / use case for changing the behavior?**

Custom styling for menu items is a great feature.

**Please tell us about your environment:**

* **Angular version:** 4.1.3

* **PrimeNG version:** 4.1.0-rc.3

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 
* **Language:**  TypeScript 2.0.3

* **Node (for AoT issues):** `node --version` =

mlaffargue:
Same problem with the menubar, I modified the plunker to use the class on another element.
[http://plnkr.co/edit/w0fTYTzxflgyiW2XkhdQ?p=preview](http://plnkr.co/edit/w0fTYTzxflgyiW2XkhdQ?p=preview)

BustyLoli-Chan:
Why was this closed. I am still currently very much having this issue.

tleveque23:
The breadcrumb has the same issue. Cannot style a specific MenuItem

navipegator:
Seems to be fixed in 5.2.4, at least in <p-menu>

lr12003:
MenuItem style and styleClass have not been fixed in Breadcrumb yet

PR #3320, opened 2017-07-06
Feature: Update contextmenu.ts to support styleClass and style attributes #3296

ambower:
#3296 These are already passed through from the MenuItem class, so it's just a small update to the template

###Defect Fixes

###Feature Requests


cagataycivici:
Fixed at 4.1.0 already.

Issue #3345, opened 2017-07-11
Frozen columns not synced after visibility changes


HoganHu0520:
I have four columns a:frozen , b, c, d.

**Init Table**
a:Frozen b c d

**After hide column `c`**
a:Frozen a:Frozen b d

**Now Code**
`initColumns(): void {
        this.columns = this.cols.toArray();

        if(this.scrollable) {
            this.scrollableColumns = [];
            this.cols.forEach((col) => {
                if(col.frozen) {
                    this.frozenColumns = this.frozenColumns || [];
                    this.frozenColumns.push(col);
                }
                else {
                    this.scrollableColumns.push(col);
                }
            });
        }

        this.columnsChanged = true;
    }`

**Fixed Code**
`initColumns(): void {
        this.columns = this.cols.toArray();

        if(this.scrollable) {
            this.scrollableColumns = [];
            this.frozenColumns = [];
            this.cols.forEach((col) => {
                if(col.frozen) {
                    this.frozenColumns.push(col);
                }
                else {
                    this.scrollableColumns.push(col);
                }
            });
        }

        this.columnsChanged = true;
    }`

I can't check in code.
I hope someone can check it in.

* **Angular version:** 4.0.X

* **PrimeNG version:** 4.0.3

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 
* **Language:** [all | TypeScript 2.4 | ES6/7 | ES5]

* **Node (for AoT issues):** `node --version` =

PR #3346, opened 2017-07-11
Fixed bug #3345.

HoganHu0520:
When resize columns number, table will added extra frozen columns.

Code modify -> line 904 -> 909.


cagataycivici:
Sorry I can't event see the diff as it is too big. Please fix this and we'll review.

PR #3371, opened 2017-07-12
Fix for Error in Calendar when validating input value #3338

### LINKED ISSUE #3338: Error in Calendar when validating input value
```
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/3GPKSVCNUdx7XOkap2dt

**Current behavior**
If you type some invalid date value directly, validation does it job, says date is not valid and text you introduced is kept in the input box of the calendar. However, if you select a valid date using calendar dropdown, and for example, remove two digits of the date (creating an invalid date), validation works, but text in the input box disappears. This behavior is not consistent with the one described before and also confuses the end user, because invalid date message is shown, but no text is on the input field.

**Expected behavior**
Text in the input box does not disappear after introducing a valid date and removing 2 digits for example. So if an invalid value is introduced, it is kept so user understands what is wrong.

Follow the instructions in "Current behavior" in provided Plunkr

It is an inconsistent behavior and confuses end user.

Windows 7

* **Angular version:** 2.0.X and 4.0.X

* **PrimeNG version:** 2.0.X and 4.0.X

* **Browser:** Chrome 59

sabotup:
###Defect Fixes
#3338


cagataycivici:
Manual input is also cleared in my case so behavior seems consistent to me. If more users demand this, we'll reopen. Also this should be an option like keepInvalidDate or so.

sabotup:
It is only cleared after you introduce a valid date first. If you add the first time an invalid date (aaaaa, e.g.), it will not clear. Anyways IMHO, it is not consistent since the field is cleared, but the calendar stays in invalid state when you use built in validator. So user will see a message saying that the date is invalid, but there is not date in the field.

PR #3372, opened 2017-07-12
DataTable improvements

uakennysoft:
###Feature Requests
 - Added support for column ids in table column headers
- While reordering columns, make dragover handler run outside of angular and not trigger detect changes cycle on whole application

### DIFF
--- a/src/app/components/common/shared.ts
+++ b/src/app/components/common/shared.ts
@@ -65,6 +65,7 @@ export class TemplateWrapper implements OnInit, OnDestroy {
 })
 export class Column implements AfterContentInit{
     @Input() field: string;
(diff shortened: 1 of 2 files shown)


uakennysoft:
DataTable adds dragover listener to DOM and huge amount of dragover events trigger detect changes cycles. So, if you have low optimized application it will be suffer while trying to reorder columns. Especially in dev mode.
There is also bottleneck: functions that cause paint reflow are called in dragover handler which triggers huge amount of paint reflows
this.domHandler.getHiddenElementOuterWidth(this.reorderIndicatorUp) this.domHandler.getHiddenElementOuterHeight(this.reorderIndicatorUp)

PR #3402, opened 2017-07-16
fix typo in treetable doc

mix5003:
i not sure but i think it should be "-" not ":"


cagataycivici:
Merged manually thanks.

PR #3430, opened 2017-07-18
To allow custom template within the multi-select dropdown label.

akashshelke22:
Enhance the design look by allowing icon image or enhance html content
rather string value within the multi-select dropdown label.

###Defect Fixes

###Feature Requests


cagataycivici:
We should use proper Angular templating APIs instead of innerHTML tricks.

akashshelke22:
Hi, 

I have added innerHTML as p-multiselect drop down doesn't provide "ng-template" option to customize the label value. I am looking to add icon image before the text which appears in the label.

Can you elaborate on how it can be approach using Angular templating  APIs?

Uter1007:
any plans for adding image support for selected label ?

Issue #2632, opened 2017-04-28
Dynamic mask support to InputMask


jogelin:
**I'm submitting a **
```
[X ] bug report
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/uZdjjvGH3IhqYElGJvnm?p=preview

**Current behavior**
1. Set initial mask to 10 digits: "9999999999" -> the input correctly displayed: "__________"
2. Change the mask, by example to 5 digits: "99999" -> the input doesn't change, it still displayed: "_____"

**Expected behavior**
I expect to be able to change the mask of an input mask dynamically

**Minimal reproduction of the problem with instructions**
1. Set initial mask to 10 digits: "9999999999" -> the input correctly displayed: "__________"
2. Change the mask, by example to 5 digits: "99999" -> the input doesn't change, it still displayed: "_____"

* **Angular version:** 4.x.x
* **PrimeNG version:** 4.0.0-RC4

PR #3440, opened 2017-07-19
#2632 Support dynamic mask in InputMask

mix5003:
fix (agian) #2632 support dynamic mask in InputMask. 

i not sure how to use getter / setter. it my first time. but i think it work.


cagataycivici:
Merged manually

Issue #3441, opened 2017-07-19
When dataType is "string" and showTime is true the ngModel value is just a date


macroparts:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
String provided to form control is just a date when dataType is "string" and showTime is true

**Expected behavior**
It should also include a time because i specified "showTime" to be true

**What is the motivation / use case for changing the behavior?**
I don't have to implement date manipulation outside of component

PR #3442, opened 2017-07-19
Fixed updateModel method

macroparts:
Fixed time not included in ngModel value when dataType is "string" and showTime is true

###Defect Fixes
https://github.com/primefaces/primeng/issues/3441


cagataycivici:
Used formatDateTime method instead

PR #3445, opened 2017-07-19
Add the feature to disable select all option in listbox.

jmesa-sistel:
###Feature Requests
Useful if the list is small and you do not want a filter and you want a multiple checkbox listbox.
Example:
image


cagataycivici:
PR has new line character issue so many lines are changed so adding it manually.

PR #3459, opened 2017-07-20
[Fix Calendar] Make disable year and month navigator when calendar is disabled

kojisaiki:
### Defect Fixes

When a calendar declared on following code, year and month navigators was manipulatable.

```
<p-calendar [(ngModel)]="date9" [inline]="true" [disabled]="true" [monthNavigator]="true" [yearNavigator]="true" yearRange="2000:2030"></p-calendar> 
```

prev/next month button(`<`, `>`) was disabled, but two navigators was not disabled and switched calendar view.
image

---

Fixed to make disable two navigators.

image


cagataycivici:
Better to disable selects as well, will do for 4.1.2, thank you.

Issue #3476, opened 2017-07-21
Tieredmenu popup performance issue


m-tomczyk:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Using multiple popup tiered menus on page makes whole application unresponsive. Problem is created by registering document click listener in ngAfterViewInit method. Because of Angular change detection cycle being trigered by this kind of events ApplicationRef.tick method is being called multiple times. In my test case with 70 popup tiered menus every click takes about 3sec to execute.

**Expected behavior**
Application is allowed to contain multiple popup tiered menus and not take massive performance hit.

* **Angular version:** 4.2.0

* **PrimeNG version:** 4.1.2

* **Browser:** All
 
* **Language:** TypeScript

PR #3477, opened 2017-07-21
Tieredmenu performance fix

m-tomczyk:
Fixes performance issues described in https://github.com/primefaces/primeng/issues/3476

Fix shifts listener binding to show method and ensures that listener is discarded on hide method - Component listens to document click event only when its required.

For ~70 popup tiered menus on page it reduces click execution time from circa 3 seconds to almost instantaneous.


cagataycivici:
Will merge after a refactor, thank you. In many other cases, we've switched global listeners to lazy mode like this.

PR #3489, opened 2017-07-23
Update tooltip.ts

guy-eizon:
Changing updateText() method (line 167).
Changed value of this.tooltipText.innerHTML from null to '' (empty string).
I've performed that change because from some reason, IE renders the innerHtml null to 'null' (as a string), and whatever the updated tooltip text is, it gets 'null' before it. (e.g: 'FieldError' becomes  'nullFieldError').


guy-eizon:
issue #3490 is related.

cagataycivici:
Thank you.

guy-eizon:
My pleasure (:

PR #3495, opened 2017-07-24
Update tabview.ts

Seraphims:
###Defect Fixes
In tabView, if _activeIndex greater than tabs.length，this.tabs[this._activeIndex] is
 undefined
bug


cagataycivici:
Will move the check to main if, thank you.

Issue #3491, opened 2017-07-23
have to refresh browser to use file upload component 2nd time


ericxchang:
Even with the very 1st example (advance file upload) in primeng page (https://www.primefaces.org/primeng/#/fileupload), after successfully upload one file, 2nd time select file, nothing will be listed. I saw this issue on chrome and IE 11, firefox seems ok.
Just found out the 4.0.0-rc3 version does not have this issue.

nni123:
https://www.primefaces.org/primeng/#/fileupload

Above link works fine for me in chrome 59 version, firefox (most consistent performer) & Edge. In advance mode filename names remain after upload e.g. file0.jpg, file1.jpg etc stays in description box

cagataycivici:
It all works for me, description box is for demo purposes, it lists what is uploaded, if you remove ng-template with pTemplate="content", they will go away.

Issue #3385, opened 2017-07-13
FileUpload does not allow selecting same file


veplechuc:
**I'm submitting a ...**  
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
**Current behavior**

I have defined the component this way
`<p-fileUpload #fileUpload accept=".csv,.txt" maxFileSize="1000000" customUpload="true" (uploadHandler)="uploadFile($event)">
`

on my package Json I have defined

` "primeng": "^4.1.0-rc.3" `

and have this method on the js file

` uploadFile(event) {
        for (let file of event.files) {
            this.uploadedFiles.push(file);
        }
        this.uploadFileService.doSomething(this.uploadedFiles[0]).subscribe(x => {
            this.fileInfo = x;
            ..do some stuff..
        });
    }`

It works as expected the first time (showing the 3 buttons of the component choose, upload and cancel) and letting me to choose the file from the popup window, the thing is, one it finish the process, if I try to upload the same file again It does not allows me to do it. It only allows me to choose the file, but the upload button is never enabled (unless I choose a different file, in that case the upload button is enabled) and of course the uploadfile method is never called. And after that if I choose the same file as the begining it works as expected.

I am using 

` this.fileUpload.clear();` 

just for clear the part where the uploaded file it is showed to the user

Thanks

**Expected behavior**
I think I should allows me to upload the same file.

**Minimal reproduction of the problem with instructions**
The example showed on http://primefaces.org/primeng/#/fileupload has the same behavior 

**Please tell us about your environment:**
Windows 10, Visual Studio Code

* **Angular version:** 4.0

* **PrimeNG version:** 4.1.0-rc3

* **Browser:** [Chrome XX]
 
* **Language:** [ ES6/7 | ES5]

kheckelmann:
I had problems with this issue today too. The problem seems to be, that the input of type file is never cleared, so after uploading the file the input still contains it and because of that onFileSelect is not called when selecting it again. Maybe it would be enough to just clear that input in the clear method?

It seems to work in the 'basic' mode because there the input contains a ngIf so that it is only there if no files are selected. Because of that the input is always "fresh" when the dialog is opened and the onFileSelect gets called correctly. There is no such thing in the advanced mode.

[...]

PR #3519, opened 2017-07-27
Fix several bugs in fileupload component.

jmesa-sistel:
###Defect Fixes

Fix not to be able to select twice the same file even if you have press clean or remove item (clear fileinput value)
Fix to open twice the file selector dialog in multiple mode (not apply ui-fileupload-choose-selected class style in multiple mode)

Probably fix [#3491](https://github.com/primefaces/primeng/issues/3491) and [#3385](https://github.com/primefaces/primeng/issues/3385)


cagataycivici:
Merged 3385 part manually.

PR #3531, opened 2017-07-29
Emit moveAllToSource/Target only after input arrays are changed

rshchpkn:
Found this feature in Roadmap. So if it will be done in next release remove pull request.


cagataycivici:
splice already returns removed items as an array so I've done a small refactor.

```
const sourceItems = this.source.splice(0, this.source.length);
```

PR #3539, opened 2017-07-31
Tooltips - auto hide tooltip after XXX ms

aervin:
Introduces @Input() hideAfter: number to tooltip component class. If hideAfter is set to a number greater than 0, call the deactivate() method after XXX ms.


cagataycivici:
We'll consider this if more users demand it.

PR #3555, opened 2017-08-02
Adds onEndToggle event on Panel

nandowalter-lm:
Adds new event onEndToggle on Panel component, in order to trigger an event just after the panel changed its status and after the animation ended.


cagataycivici:
Just realized there is onAfterToggle so moved that instead to animation complete.

PR #3567, opened 2017-08-03
adding ability to change rowGroupMode dynamically

Ibydy:
###Defect Fixes
error when change rowGroupMode dynamically

###Feature Requests
adding ability to change rowGroupMode dynamically


cagataycivici:
There are 346 line changes and it is not clear what part has changed, so unable to review.

PR #3577, opened 2017-08-04
add search query to item template

thinhbuzz:
We are can more customizable, ex: highlight query


thinhbuzz:
I feel coding convention is not good :pensive:

cagataycivici:
There are 275 line changes for this, please fix it and we should not change the default template wrapper as it will be bloated, better to use a new wrapper directive or extend it.

Issue #3600, opened 2017-08-07
Calendar bugs if both showtime='true' and selectionmode='multiple' are chosed.


TahirMehdi:
[x] bug report 
I fixed it in https://github.com/primefaces/primeng/pull/3601, please review my pull request

If you try to use  
```
<p-calendar [(ngModel)]="dates" [showTime]="true" selectionMode="multiple" readonlyInput="true"></p-calendar> 

```
time does appear, but if you will try to use handles - primeng crashes. I found out, that the problem is in updateTime() function in calendar.ts. Seems it wasn't updated to handle isSingleSelection(). So i went ahead and updated it.

cagataycivici:
Multiple date selection with time is not supported as it is a confusing UI.

PR #3601, opened 2017-08-07
Fix for calendar with showtime='true' and selectionmode='multiple'

TahirMehdi:
Calendar bugs if both showtime='true' and selectionmode='multiple'  are chosed. 
Explained here: https://github.com/primefaces/primeng/issues/3600
This fixes it. Now showTime correctly works with selectionmode='multiple'


cagataycivici:
Multiple date selection with time is not supported as it is a confusing UI.

PR #3647, opened 2017-08-11
fixed #3046: .ui-state-filled class was not set initially for pInputT…

### LINKED ISSUE #3046: .ui-state-filled not set initially when using OnPush and [value] binding
```
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/aOzYbfRSrxc0G0NuiISL?p=preview

**Current behavior**
when using OnPush, filled inputs with pInputText don't initially have the ui-state-filled css class when setting the value with the [value] binding

**Expected behavior**
the input should have the ui-state-filled css class

1. open the plunkr above
2. inspect the input
3. it does have a value, but it doesn't have the ui-state-filled css class (background is red, if the class is missing)

because it is a bug and it breaks ultima-ng

Windows 8.1

* **Angular version:** 4.1.3

* **PrimeNG version:** 4.0.3

* **Browser:** all 
 
* **Language:** probably all

saithis:
…ext, pInputTextArea and pPassword when using ChangeDetectionStrategy.OnPush with value or ngModel binding

### DIFF
--- a/src/app/components/inputtext/inputtext.ts
+++ b/src/app/components/inputtext/inputtext.ts
@@ -1,5 +1,6 @@
-import {NgModule,Directive,ElementRef,HostListener,Input,DoCheck} from '@angular/core';
+import {NgModule,Directive,ElementRef,HostBinding,HostListener,Input,DoCheck,Optional} from '@angular/core';
 import {CommonModule} from '@angular/common';
+import {NgModel} from '@angular/forms';
 
 @Directive({
     selector: '[pInputText]',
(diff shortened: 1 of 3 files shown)


cagataycivici:
Have you tested it in a reactive form?

saithis:
I can't remember as it was over 3 months ago... but the bug doesn't exist when using reactive forms, only with value/ngModel binding. Also I think the  @Optional() was, because the NgModel injection only works when unsing ngModel binding. So for reactive forms it should be undefined, as it is with the [value] binding and so it shouldn't break it.

saithis:
@cagataycivici What is your problem with this PR? Why does it take over 6 months now to get this reviewed/merged? The uuuuultra slow response time doesnt really encourage us to buy another premium template for the next project :/

cagataycivici:
I just hope for a better solution.

saithis:
I know that the solution isn't pretty, but I spent a whole day to find a better solution and didn't find any... so is it better to what for a solution that may never come, than to fix this bug already and releave us and maybe others from all the workarounds for this bug that clutter our codebase?

cagataycivici:
My answer is super fast now, sorry sometimes lot of things get in our routine preventing a PR to proceed. Let me work on this again and get back to you. Again excuses for the unusual delay, it is one of those things that stuck and not move on.

cagataycivici:
We need a solution to cover both reactive and template driven forms.

Issue #3654, opened 2017-08-13
Chips - Tab doesn't move focus to next element when input field is empty


nmobregon:
**I'm submitting a ...**  
```
[ x ] bug report
[ ] feature request
[ ] support request 
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/K1zU6t7q6eePj4pcdUNL?p=preview

**Current behavior**
When using the new tab feature, the focus doesn't move to next field if chips input is empty

**Expected behavior**
When user stops using the field, it should automatically move to next element when tab is pressed

**Minimal reproduction of the problem with instructions**
Write some chip elements separating them with tab, and when you're done, press tab with the empty input field. It doesn't work as expected because it just stays there.

**What is the motivation / use case for changing the behavior?**
Natural behavior of an multiple input element should be to move focus to next element when nothing has been written (for example, when writing recipients in a Gmail mail)

* **PrimeNG version:** 4.1.3

* **Browser:** [Chrome 57 ]
 
* **Language:** [TypeScript 2.2.2]

PR #3673, opened 2017-08-16
add 'chips input element is empty.'

HuseyinTurkmenoglu:


cagataycivici:
See https://github.com/primefaces/primeng/issues/3654

Issue #3715, opened 2017-08-21
Fileupload dragover performance issue


jamarzka:
**The fileupload dragover event causes a performance issue as more components are added to the page**
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/Ci9yfkndEuAgu1IUKlx8?p=preview

**Current behavior**
When you drop a file it takes a couple seconds to show the dropped file

**Expected behavior**
This should be almost instant

PR #3716, opened 2017-08-21
Add fileupload dragover event outside Angular

jamarzka:
Issue: #3715 

Here is an example of the fix and what we are using as a workaround: http://plnkr.co/edit/TywW7x1fcaalkKd7nW1x?p=preview


cagataycivici:
Thanks, I've fixed it the same way this morning.

PR #3723, opened 2017-08-22
Color picker pickHue resets saturation and brightness

daniloarcidiacono:
Keep the same saturation and brightness when changing hue in colorpicker.
Use getValueToUpdate() when emitting onChange in pickHue.

###Defect Fixes
When the user changes the hue, the value passed to onChange has not the format specified by the "format" attribute. Also, changing the hue resets the current saturation (s) and brightness (b) values.


cagataycivici:
Merged manually, thanks

Issue #3724, opened 2017-08-22
Loading status for AutoComplete


cagataycivici:
Display an icon when a search is in progress.

catahoc:
Hi, @cagataycivici,
you've broken suggestions work with AsyncPipe by this change!
When I write [suggestions]="items | async", first AsyncPipe message (null) puts dropdown into "loading=false" state and second message (actual found items) is ignored. This is specifics of how ASyncPipes are working (they send null message before loading async values)

catahoc:
Created an issue for this: https://github.com/primefaces/primeng/issues/5030

PR #3735, opened 2017-08-23
3724 loading icon for autocomplete

Sumragen:
Added message "Loading" when search process is started #3724


cagataycivici:
Thank you, being implemented.

PR #3738, opened 2017-08-23
File Upload advanced mode choose button does nothing on click

radarsu:
###Defect Fixes
From version ~4.1.3 method onChooseClick was removed, so clicking "choose" button in advanced mode does literally nothing.

I restored onChooseClick method from 4.0.3 version.


cagataycivici:
A css based solution is being used now instead of clicking the input manually, the components works for me as it is now. You should first create an issue with plunkr.

PR #3740, opened 2017-08-24
Wrongful import in gmapdemo documentation

Mysame:
The Gmaps documentation shows an incorrect import statement. The example code uses the correct import.


cagataycivici:
'primeng/primeng' is also valid although not suggested.

Mysame:
It would seem weird to leave it there if it's "not suggested".
And while it may be valid, it doesn't work if the implementation uses Angular Universal. The reasoning I do not know, but the "not suggested" import breaks server side pre-rendering.

PR #3750, opened 2017-08-25
Fixed #1845

### LINKED ISSUE #1845: DataTable: expanded row styling
**I'm submitting a ...**
```
```

**Current behavior**
Expanded row has no class on it, no way to style it differently (e.g. remove padding}

**Expected behavior**
Make it possible to differentiate the expanded row from the others.

We need to change the padding in the expanded row easily.
Workaround is to use `p-datatable tbody > tr:not(.ui-widget-content)` pseudo selector, but it might be not restrictive enough.

* **Angular version:** latest

* **PrimeNG version:** 1.1.4

* **Browser:** [all]
 
* **Language:** [all]

jsmucr:
This PR adds two new style classes to be able to change styles of `DataTable` rows that are expanded:
* `ui-expanded-row` is added to the `<tr>` the user can click in order to show or hide relevant information.
* `ui-expanded-row-content` is added to the `<tr>` below.

The other classes (such as `.ui-datatable-even`) are now shared between both rows because it's impossible to reach previous sibling element via CSS (imagine you might want the odd and even expanded rows colored differently). Unfortunately, this will break the workaround stated in #1845.


cagataycivici:
(mouseenter)="dt.setHoveredRow(rowData)" (mouseleave)="dt.clearHoveredRow(rowData)"

is dangerous, will trigger changed detection and drastically slow down page performance.

PR #3762, opened 2017-08-28
support `onPush` in parent component with `ngModel` binding.

AntonLitovka:
According to angular [asynchronous implementation  of template-driven forms ](https://github.com/angular/angular/blob/c65f18a2fa4e605a0900a6ca18caece5e2bd27fc/aio/content/guide/reactive-forms.md#async-vs-sync) `onPush` strategy can't be used. Change detection should be manually activated. 

[Issue discussed here](https://github.com/angular/angular/issues/10816)

[problem example](http://plnkr.co/edit/gJva0X?p=preview)


cagataycivici:
Merged manually due to conflict

PR #3764, opened 2017-08-28
Enhancement: DataTable hovered row styling

jsmucr:
This PR adds `ui-state-hover` style class to a row that is currently hovered.


cagataycivici:
This is very dangerous as it will trigger angular change detection and kills the performance as a result so unable to accept due this reason.

Issue #2574, opened 2017-04-21
Add onPanelShow callback to MultiSelect


codechamploo:
I'm submitting a ... (check one with "x")

[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35

What is the motivation / use case for changing the behavior?
Can you add an onShow and onHide event to Mutiselect (and possibly other controls with a dropdown menu).  This will help with, but not limited to, the following scenario(s):

1.  Executing logic when the user closes (or opens) the drop down panel for the multiselect
(e.g. user selects n items from a list, closes the menu, event fires, and new data is fetched from the server, etc...)
2.  Perform operation on the select options when panel is open (e.g. sort based on selected status, etc...)

Thanks !

Makrel:
I too would like this - I had to add it myself, but would think the GenPop would like it.
Would happily submit my code...

PR #3767, opened 2017-08-28
#2574: Add onShow and onHide event emitters to MultiSelect control

codechamploo:
Added onShow and onHide events to multiselect component

Refer to issue:
https://github.com/primefaces/primeng/issues/2574


cagataycivici:
There is debugger; in the code.

Issue #3770, opened 2017-08-29
Filtering by default fails on DataTable


nhnb:
**I'm submitting a ...**
```
[X] bug report
[ ] feature request
[ ] support request
```
**Current behavior**
If `dt.filter('value', 'columnField, 'contains')` is invoked before the data is loaded the following error is thrown. On debugging, both this.value and  this.columns are undefined.

~~~~
PageAktuellComponent.html:42 ERROR TypeError: Cannot read property 'length' of undefined
    at DataTable.webpackJsonp.../../../../primeng/components/datatable/datatable.js.DataTable._filter (datatable.js:1254)
    at DataTable.webpackJsonp.../../../../primeng/components/datatable/datatable.js.DataTable.filter (datatable.js:1235)
    at PageTableComponent.set [as filter] (page-table.component.ts:39)
    at updateProp (core.es5.js:11100)
    at checkAndUpdateDirectiveInline (core.es5.js:10796)
    at checkAndUpdateNodeInline (core.es5.js:12330)
    at checkAndUpdateNode (core.es5.js:12269)
    at debugCheckAndUpdateNode (core.es5.js:13130)
    at debugCheckDirectivesFn (core.es5.js:13071)
    at Object.View_PageAktuellComponent_1._co [as updateDirectives] (PageAktuellComponent.html:47)
~~~~

**Expected behavior**
It should be possible to set a filter on page load before the data is available. The filter should be applied after the data is available.

**Minimal reproduction of the problem with instructions**

~~~~
<p-column field="produktBereich" header="P." [filter]="true">
    <ng-template pTemplate="filter" let-col>
        <input pInputText type="text" [(ngModel)]="textval" 
            (input)="dt.filter(textval,col.field,col.filterMatchMode)">
    </ng-template> 
</p-column>
~~~~

The input element fires an input event which invokes dt.filter. At this time the data is not available and therefore the DataTable is not initialized completely.

**What is the motivation / use case for changing the behavior?**

See above. There are other cases within DataTable that check for the undefined condition, but this check is missing in _filter.

* **Angular version:** 4.3.1
* **PrimeNG version:** 4.1.1

* **Browser:** all
* **Language:** all

benja-admin:
This is relevant, having the same issues trying to use Datatable in PrimeNG and getting the exact same error. No solution in sight and everything I am finding is geared toward Angular 2.

nhnb:
Pull request https://github.com/primefaces/primeng/pull/3771 fixes this issue.

PR #3771, opened 2017-08-29
prevent undefined error in DataTable on filter before data is loaded (#3770)

nhnb:
This pull request fixes issue 3770


cagataycivici:
Merged manually

Issue #3309, opened 2017-07-05
onSelect event for PanelMenu Submenus


hiteshc2:
I'm submitting a ...(check one with "x")
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35

Hi,
Could you please add event such as onSelect for a menuitem and submenu item.

majkers:
I achieved this by adding something like this on menu Show event:


```
private panelMenuComponent!: ComponentRef<PanelMenu>;

  private menuItemClickListener: (event: any) => void = (event: any) => {
    this.menuItemClick.emit(event);
  };
```


and later

```
this.panelMenuComponent = this.menuContainer.createComponent(PanelMenu);
    this.panelMenuComponent.instance.model = this.menuItems;
    this.panelMenuComponent.instance.multiple = false;
    this.panelMenuComponent.changeDetectorRef.detectChanges();

    this.panelMenuComponent.location.nativeElement
      .querySelectorAll('a[ng-reflect-router-link]')
      .forEach(
        (el: HTMLElement) => {
          el.addEventListener('click', this.menuItemClickListener);
        }
      );
```

I have to add it manually cause when I have panelMenu inside sideBar there is an issue with expanded submenus without being clicked....

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

PR #3791, opened 2017-08-30
3309 on menu item select handler

Sumragen:
#3309 PanelMenu: added two events
* onSubMenuSelect
* onMenuItemSelect
which trigger when you click on a menu item (submenu item too)

### DIFF
--- a/src/app/components/panelmenu/panelmenu.ts
+++ b/src/app/components/panelmenu/panelmenu.ts
@@ -20,7 +20,10 @@ export class BasePanelMenuItem {
             return;
         }
         
-        this.pMService.onItemSelect.next(item);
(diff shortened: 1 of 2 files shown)


src/app/components/panelmenu/panelmenu.ts
```diff
@@ -71,6 +84,10 @@ export class PanelMenuSub extends BasePanelMenuItem {
     @Input() item: MenuItem;
     
     @Input() expanded: boolean;
+    
+    constructor(private panelMenuService: PanelMenuService) {
```
scorpp: it's easier to read when constructors are grouped together

cagataycivici:
I think a bit too much to create a service for this.

PR #3797, opened 2017-08-31
Add optional equals function to TabMenu

Venthe:
Provide the ability to use custom equals in TabMenu for activeItem


cagataycivici:
We'll reconsider if more users demand it, thank you.

Issue #3504, opened 2017-07-24
Export selected rows to CSV


d-bechtel:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[X] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
Add an option to export only the selected rows to csv.

PR #3810, opened 2017-09-01
3504 added method to export selected rows to CSV

Sumragen:
#3504 DataTable: Export selected row to CSV

### DIFF
--- a/src/app/components/datatable/datatable.ts
+++ b/src/app/components/datatable/datatable.ts
@@ -2354,9 +2354,7 @@ export class DataTable implements AfterViewChecked,AfterViewInit,AfterContentIni
     
     public exportSelectedRowsToCSV() {
         let data = this.getSelection();
-        if(!!data&&data.length){
-            this.generateCSV(data);
-        }
+        this.generateCSV(data);
     }
     
     public exportCSV() {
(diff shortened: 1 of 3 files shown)


src/app/components/datatable/datatable.ts
```diff
@@ -2348,8 +2348,23 @@ export class DataTable implements AfterViewChecked,AfterViewInit,AfterContentIni
             this.updateDataToRender(this.value);    
     }
     
+    getSelection() {
+      return this._selection;
+    }
+    
+    public exportSelectedRowsToCSV() {
```
scorpp: The way this method works differs from original `exportCSV` when there are no data. `exportCSV` would still make browser download and empty CSV file containing only header. While your version silently does nothing when nothing is selected.
It would be nice to align the two IMO.

src/app/components/datatable/datatable.ts
```diff
@@ -2348,8 +2348,23 @@ export class DataTable implements AfterViewChecked,AfterViewInit,AfterContentIni
...
+        let data = this.getSelection();
+        if(!!data&&data.length){
```
scorpp: spaces ` && `

cagataycivici:
Fixed at 3504

PR #3815, opened 2017-09-01
Bug with filtertimeout cancellation on DataTable and Bug with footer column template now showing

pantonis:
###Defect Fixes
While using filter with datatable the setTimeout is not cancelled before a new one is created. As a result if you type n characters fast enough (faster than filterDelay property value) the this._filter() method is called n times instead of one.


cagataycivici:
Thank you.

Issue #2675, opened 2017-05-02
TreeNode styleClass update sets ui-tree-toggler visibility to hidden


skuby2:
**I'm submitting a ...**
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Here is a minimal Plunkr demonstrating the issue: http://plnkr.co/edit/3EzIYrLuPEYA5k54tpYO

**Current behavior**
When a parent TreeNode that has a collapse/expand ui-tree-toggler has its styleClass changed the ui-tree-toggler's visiblity is set to hidden because the parent li element is incorrectly marked with the class ui-treenode-leaf.

**Expected behavior**
When styleClass is updated the parent li element should not be marked with the class ui-treenode-leaf, which in turn will not set the visibility of ui-tree-toggler to hidden.

**Minimal reproduction of the problem with instructions**
See the above Plunkr for full reproduction of the problem.

**What is the motivation / use case for changing the behavior?**
My leaf nodes have a health status associated with them. If the status is "bad" then the tree nodes are styled red to indicate such. The feature I want to add is rolling up that status to parent folder nodes so that it is easier to tell that there is an error nested somewhere lower in the tree.

**Please tell us about your environment:**
This issue exists in all environments.

* **Angular version:**
4.0.1

* **PrimeNG version:**
4.0.0-rc.2

* **Browser:**
Fedora 21 Linux Chrome 48.0.2564.116 (64-bit)
Windows 7 Chrome 58.0.3029.96
 
* **Language:**
TypeScript 2.3.1

* **Node (for AoT issues):** `node --version` =  7.7.2

mdtempalert:
We worked around this with the following change to the template:

```
<li class=\"ui-treenode {{node.styleClass}} {{isLeaf()? 'ui-treenode-leaf' : ''}}\" *ngIf=\"!tree.horizontal\">```

I'm not an Angular expert but I suspect interpolation and [ngClass] bindings are happening at different times and there is a subtle race condition.

gschafra:
Same issue here!

**Please tell us about your environment:**
This issue exists in all environments.

-  **Angular version:**
4.3.6

- **PrimeNG version:**
4.1.3

- **Browser:**
[...]

PR #3823, opened 2017-09-03
Fix bug - TreeNode styleClass update sets ui-tree-toggler hidden

MadUser:
https://github.com/primefaces/primeng/issues/2675


cagataycivici:
Will use an array instead.

Issue #3826, opened 2017-09-04
Calendar onClose event is fired even if overlay is not visible


bossqone:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/opEfAscX5syzD1wY0KZo?p=preview

**Current behavior**
Calendar onClose event is fired even if overlay is not visible. Basically it's fired on every document click outside of calendar overlay.

**Expected behavior**
Event should be fired only when calendar/overlay is actually being closed.

* **PrimeNG version:** 4.1.3

* **Browser:** all

PR #3827, opened 2017-09-04
Calendar onClose event should be fired only when overlay is visible

bossqone:
https://github.com/primefaces/primeng/issues/3826


bossqone:
fixed by #3832

Issue #3825, opened 2017-09-03
TreeTable: Add methods to collapse/expand nodes


steigerm:
There are currently events onNodeExpand/onNodeCollapse but no methods to do it programmatically

cagataycivici:
I don't think it is necessary, at your model if you set TreeNode.expanded = true|false then UI will be updated accordingly.

steigerm:
Thanks for the information!
Good to know that this works bidirectional.

@cagataycivici Maybe another pointer to enhance the documentation.

PR #3830, opened 2017-09-04
#3825 toggle node programmatically

Sumragen:
#3825 TreeTable: added method to expand/collapse each row by single node
* method `toggleNode` of `TreeTable`

### DIFF
--- a/src/app/components/treetable/treetable.ts
+++ b/src/app/components/treetable/treetable.ts
@@ -439,6 +439,10 @@ export class TreeTable implements AfterContentInit {
         }
         return false;
     }
+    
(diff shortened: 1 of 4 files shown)


src/app/components/treetable/treetable.ts
```diff
@@ -439,11 +439,15 @@ export class TreeTable implements AfterContentInit {
         }
         return false;
     }
+    
+    toggleNode(node: TreeNode, isExpanded?: boolean) {
```
scorpp: it would be nice to have explicit `expand` and `collapse` methods in addition to toggling

cagataycivici:
See my comment at 3825

PR #3831, opened 2017-09-04
refactor(): switch to new HttpClient

danielsogl:
Just a small change switching to the new Angular HttpClient introduced with Angular 4.3.


cagataycivici:
Doesn't this break apps of Angular 4.2 users?

danielsogl:
It should only affect the PrimeNG Website.

cagataycivici:
Oh, I see now, ok, will add after 4.2.0 is out this week. Thank you.

cagataycivici:
Merged manually, thank you.

PR #3832, opened 2017-09-04
#3826 edited predicate to fired onClose event:added predicate those c…

Sumragen:
#3826 Calendar: 
Fixed the condition for firing the onClose event. In addition to verifying the click, visibility is now checked too


src/app/components/calendar/calendar.ts
```diff
@@ -1416,7 +1416,7 @@ export class Calendar implements AfterViewInit,AfterViewChecked,OnInit,OnDestroy
     bindDocumentClickListener() {
         if(!this.documentClickListener) {
             this.documentClickListener = this.renderer.listen('document', 'click', (event) => {
-                if(!this.datepickerClick) {
+                if(!this.datepickerClick&&this.overlayVisible) {
```
scorpp: add spaces around `&&`

PR #3842, opened 2017-09-05
Add onHover callback for Growl

kreuzerk:
In this PR I added a EventEmitter to get notified when I hover over a message. I think this is useful for all use cases where I want to do something with the messages on hover.


cagataycivici:
There is an issue with new lines in pr so will add manually, thank you.

kreuzerk:
Thx! I just pulled the 4.2.1 and tested this feature. Everything works as expected.

PR #3850, opened 2017-09-06
Feature: added set Paginator page programmaticlly (@Input())

### LINKED ISSUE #3849: Set Paginator page programmaticlly (@Input())
```
```

**Current behavior**
Cant set Paginator on Load (Programmaticly)

**Expected behavior**
Set Page over Input

When I go back to the table, the page I was on before should be saved and it should return back to the page.

JoCoaker:
- Paginator: currentPage is set able for the first time loading. e.g. on Navigation Back, go to the same page I was before
- Paginator Showcase: Added to Input Table

###Defect Fixes
#3849

### DIFF
--- a/src/app/components/paginator/paginator.ts
+++ b/src/app/components/paginator/paginator.ts
@@ -41,19 +41,21 @@ export class Paginator {
     @Input() style: any;
 
     @Input() styleClass: string;
-    
+
     @Input() rowsPerPageOptions: number[];
-    
+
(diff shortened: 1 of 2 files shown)


jarinwhite:
+1. Would love to see this feature merged, as it is currently preventing us from using the Pagination component in a stateful application.

cagataycivici:
This is already handled by first property, no need to complicate with new attributes.

Issue #3859, opened 2017-09-06
when multiselect or dropdown is used as datatable filter template, click on the dropdown or multiselect triggers sort


preethi29:
**I'm submitting a ...**  (check one with "x")
```
[ x ] bug report => Search github for a similar issue or PR before submitting
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://plnkr.co/edit/3G1G8JGACKpHTf37gGYT?p=preview

**Current behavior**
when p-multiselect or p-dropdown is used as datatable filter, clicking anywhere in the multiselect triggers sort

**Expected behavior**
Click on multiselect or dropdown when used as filter template, should not trigger sort of the datatable column. Sorting can be limited just to column title instead of whole column space.

**Minimal reproduction of the problem with instructions**
http://plnkr.co/edit/3G1G8JGACKpHTf37gGYT?p=preview
1. Click on 'Colors' filter
2. Select any color from the multiselect
3. Colors column is sorted.
4. Repeat the above procedure for 'Brand' column

**What is the motivation / use case for changing the behavior?**
Sort is done on (click) event of <th></th>. Instead it can be done '.ui-column-title' element. That way, filtering wont be affected. 

**Please tell us about your environment:**

* **Angular version:** 4.0.0

* **PrimeNG version:** 4.1.2

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
All

preethi29:
Pull request- https://github.com/primefaces/primeng/pull/3861

preethi29:
Instead of limiting the click event to .ui-column-title, we can stop bubbling up the event when we click on multiselect item.

jonmarozick:
We've added some custom padding in the p-multiselect and now that exposes a bit more area where clicking on the multiselect casues a sort.

preethi29:
Ideally the multiselect need not propagate its event outside its scope.

pauly815:
I have observed this as well.  I ended up adding event.stopPropagation() to the onMouseClick, onItemClick, and toggleAll functions.

preethi29:
I have added event.stopPropagation() as the interim fix but it creates other problems like stickiness of dropdown.

preethi29:
[...]

PR #3861, opened 2017-09-06
#3859 - Sort datatable column on click on title instead of whole colu…

preethi29:
…mn space

More info on : https://github.com/primefaces/primeng/issues/3859


cagataycivici:
This breaks backward compatibility so can't merge.

preethi29:
@cagataycivici Can we have a discussion how this could be made backward compatible? Can we make only title clickable based on a attribute? For example: sort-on-column-title="true".

Also, to fix the issue, I have raised another pull request which fixes even propogation in multiselect - https://github.com/primefaces/primeng/pull/3866 . Can you take a look at it?

PR #3887, opened 2017-09-08
Fixes #3855 Move all should only move visible items

### LINKED ISSUE #3855: 'Move all' buttons should not move filtered item.
```
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/yzP7UAPLnZs1BjN7iz4C

**Current behavior**
When filtering the a list of the PickList component and then clicking the 'move all' button of that list, *all* items are moved to the other list instead of only the visible items.

**Expected behavior**
Moving 'all' items of a list should only move the currently visible items.

Filter the left list of the plunkr by 'bar', then click the 'move all to right' button.
All 3 items are moved to the right list, instead of only the single 'bar' item.

Currently, the combination of filtering/moving all makes no sense

* **PrimeNG version:** 4.1.3

Fixing this should be trivial, as far as I see.

davisb10:
Fixes the moveAllRight() and moveAllLeft() methods to only push the items that are visible.

###Defect Fixes

###Feature Requests

### DIFF
--- a/src/app/components/picklist/picklist.ts
+++ b/src/app/components/picklist/picklist.ts
@@ -396,6 +396,7 @@ export class PickList implements AfterViewChecked,AfterContentInit {
           
             for(let i = 0; i < this.source.length; i++) {
                 if (this.isItemVisible(this.source[i], -1)) {
+                  this.source.splice(i, 1);            
                   this.target.push(this.source[i]);
                   sourceItems.push(this.source[i]);
                 }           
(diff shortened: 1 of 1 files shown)


davisb10:
My bad. I wiped out the old way of getting the sourceItems/targetItems and forgot to still splice out the files. I have updated my pull request.

cagataycivici:
Thanks, implemented slightly differenty.

Issue #3897, opened 2017-09-10
Slide doesn't work if the slider was initaly disabled and then got enabled


volkovIvan:
**I'm submitting a ...**  (check one with "x")
```
[x ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/jbwpX2BlewrTQCUKAMuJ?p=preview

**Current behavior**
User cannot change slider position by slide, it can be done only by clicking on the slider bar.

**Expected behavior**
Slide works and slider value changed via slide

**Minimal reproduction of the problem with instructions**
1. Create a slider and disable it initially
2. Enable it, via button for example.

**Please tell us about your environment:**

* **Angular version:** 4.2.0

* **PrimeNG version:** 4.2.0-rc.1

* **Browser:** Chrome 60.0.3112.113
 
* **Language:** TypeScript 2.5.2

PR #3898, opened 2017-09-10
Fixed issue with slider when enabling it after it was initially disabled

volkovIvan:
###Defect Fixes
Fix for #3897


cagataycivici:
Too many changes that prevent me from what has changed so will fix it on our end, thank you.

PR #3911, opened 2017-09-12
Fix slider change detection performance issue

lukasschinninger:
In this PR i fixed an performance issue in the Slider component.
The changedetection was triggered on mousemove, which causes high cpu usage. For Example i you use three Slider components the changedetection fired three times. So i excluded it.


cagataycivici:
Good one, I've also seen the room for improvement with zones when reviewing slider for another PR. Thank you for the contribution.

Issue #3358, opened 2017-07-11
Append sep={csvSeparator} to allow for seemless integration when the csvSeparator is not ',' for a datatable


kenisteward:
```
[ X] bug report => Search github for a similar issue or PR before submitting
[ X] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Currently when you export a csv and use the csv separator, it does not append sep={yourSeparator} in the file.  This causing issues with loading in excel because excel requires that you have that at the top so it knows how to separate the data.

**Expected behavior**
either simple sep={yourSeparator} is appended here 
https://github.com/primefaces/primeng/blob/41d29a59f308b7ea66b8bd136c9871ee7462e49a/src/app/components/datatable/datatable.ts#L2277-L2281

or allow a callback function that is beforeExport that happens right before the export happens that allows you to manipulate the raw data in the blob. 

**Minimal reproduction of the problem with instructions**
any datatable with exportCSV() functionality with the csvSeparator set to anything other than comma

**What is the motivation / use case for changing the behavior?**
The idea would be to allow commas in the content by changing the csvSeparator and still be compatible with csv parsers like excel.

@cagataycivici as stated above a simple change to the exportCSV() function would allow for almost seemless integration with CSV->excel.  just a 1 line change from

```

public exportCSV() {
   let data = this.filteredValue||this.value;
   let csv = '\ufeff';
...
}
```

to

```
public exportCSV() {
   let data = this.filteredValue||this.value;
   let csv = '\ufeff';
   csv += this.csvSeparator + '\r\n\'
...
```

I'd gladly make a PR if needed.

kenisteward:
@cagataycivici any update on this or exports for excel.  I know it's on the roadmap but is there a particular version target?

cagataycivici:
The change break it on my mac where I can't open the file with excel anymore.

kenisteward:
I posted the wrong code.  It should be csv += 'sep' + this.csvSeperator +
'\r\n'

My apologies

cagataycivici:
Thanks, will review again tomorrow.

kenisteward:
As i saw you post today I noticed another missing character....

I should not have wrote it on my phone haha.

 It should be csv += 'sep=' + this.csvSeperator + '\r\n'

[...]

PR #3921, opened 2017-09-12
#3358 Added separator definition for excel

kenisteward:
###Defect Fixes

#3358 True parsing of separator field for excel.

Very minor change.  I was able to test the changes in my local which worked.  

Should instead of using hard '\r\n' find a platform independent way in the browser to insert line endings.

@cagataycivici

### DIFF
--- a/src/app/components/datatable/datatable.ts
+++ b/src/app/components/datatable/datatable.ts
@@ -2368,6 +2368,7 @@ export class DataTable implements AfterViewChecked,AfterViewInit,AfterContentIni
     public exportCSV() {
         let data = this.filteredValue||this.value;
         let csv = '\ufeff';
+        csv += 'sep=' + this.csvSeparator + '\r\n';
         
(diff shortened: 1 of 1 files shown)


kenisteward:
@cagataycivici do note that at 
https://github.com/primefaces/primeng/blob/29ba431dd7fafb3ae79858156852407744d773cb/src/app/components/datatable/datatable.ts#L2383-L2395

It should definitely prefer operating system EOL as well instead of a hard '\n'

kenisteward:
@cagataycivici  do you have time to merge this?

cagataycivici:
screen shot 2017-10-03 at 11 18 56

Does not work for me on MAC

kenisteward:
I'm confused by this.  It has sep=(no character) when it should with my changes at the very least have sep=,

you ran with my changes and not with the old suggestion right @cagataycivici  ?

Issue #3933, opened 2017-09-13
OrderList throws a "Cannot read property getBoundingClientRect of undefined"


claudiuconstantin:
**I'm submitting a ...**
```
[X] bug report
[ ] feature request
[ ] support request
```

* **Minimal reproduction of the problem with instructions**
Reproducible in the official demo: https://www.primefaces.org/primeng/#/orderlist

1. Select an item
2. Unselect the item
3. Press any reorder button -> `ERROR TypeError: Cannot read property 'getBoundingClientRect' of undefined`

* **Angular version:** 4.3.6
* **PrimeNG version:** 4.2.0

Note: possibly the same problem as [here](https://github.com/primefaces/primeng/issues/1292) or [here](https://github.com/primefaces/primeng/issues/1002).

cagataycivici:
Fixed via PR, thank you.

PR #3939, opened 2017-09-14
3933 OrderList reorder of an empty array

Sumragen:
#3933 
Added a condition for checking if at least one item is selected from the list to move it.


claudiuconstantin:
Any milestone for this?

cagataycivici:
Fixed for 4.2.2, thank you.

Issue #3942, opened 2017-09-14
Visible flag support for MenuBar


parisgianluca:
**I'm submitting a ...**
```
[ x ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
There is "visible" flag on MenuModel, but the flag is not supported by MenuBar 

**Expected behavior**
MenuBar should support "visible" flag of MenuModel

**What is the motivation / use case for changing the behavior?**
Documentation says
"visible | boolean | true | Whether the dom element of menuitem is created or not."
Since this model is used in MenuBar component, the flag should be supported. In 4.2.0 it works only on Menu component.

PR #3954, opened 2017-09-15
3942 menubar: flag visible

Sumragen:
#3942 added property `visible` for each item in MenuBar

### DIFF
--- a/src/app/components/menubar/menubar.css
+++ b/src/app/components/menubar/menubar.css
@@ -1,0 +1,3 @@
+.ui-hidden-element {
+    display: none;
+}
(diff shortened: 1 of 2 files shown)


cagataycivici:
Thank you, we have ui-helper-hidden by the way.

PaulAGH:
Either I don't understand how this works, or this isn't fixed yet.  I downloaded the 4.2.2 tag of Primeng and made set the File menu item visibility in src\app\showcase\components\menubar\menubardemo.ts to false as shown below, and the menu item is still rendered.   I was really looking forward to this being resolved.  Am I missing something?

image

wagnerjts:
I also have the same problem.

Issue #3868, opened 2017-09-07
Calendar with selectionMode="multiple" show error when the array (ngModel) has value(s)


sgodoy:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**

http://plnkr.co/edit/vG1QBJQuTVyEbhQRyM7s?p=preview

**Current behavior**
When the dates array (ngModel ) has default values or dates are pushed in another method, dev console show this error message:

> ERROR Error: Uncaught (in promise): TypeError: val.getMonth is not a function
> Calendar.prototype.updateUI@https://unpkg.com/primeng@4.2.0-rc.1/components/calendar/calendar.js:757:26
> Calendar.prototype.writeValue@https://unpkg.com/primeng@4.2.0-rc.1/components/calendar/calendar.js:796:9
> setUpControl/<@https://unpkg.com/@angular/forms@4.2.6/bundles/forms.umd.js:1810:9
> FormControl.prototype.setValue/<@https://unpkg.com/@angular/forms@4.2.6/bundles/forms.umd.js:2996:65
> FormControl.prototype.setValue@https://unpkg.com/@angular/forms@4.2.6/bundles/forms.umd.js:2996:13
> NgModel.prototype._updateValue/<@https://unpkg.com/@angular/forms@4.2.6/bundles/forms.umd.js:4398:46
> ZoneDelegate.prototype.invoke@https://unpkg.com/zone.js@0.8.4?main=browser:365:17
> onInvoke@https://unpkg.com/@angular/core@4.2.6/bundles/core.umd.js:3922:24
> ZoneDelegate.prototype.invoke@https://unpkg.com/zone.js@0.8.4?main=browser:364:17
> Zone.prototype.run@https://unpkg.com/zone.js@0.8.4?main=browser:125:24
> scheduleResolveOrReject/<@https://unpkg.com/zone.js@0.8.4?main=browser:760:52
> ZoneDelegate.prototype.invokeTask@https://unpkg.com/zone.js@0.8.4?main=browser:398:17
> onInvokeTask@https://unpkg.com/@angular/core@4.2.6/bundles/core.umd.js:3913:24
[...]

PR #3962, opened 2017-09-15
#3868 Fix Calendar updateUI bug when selectionMode is not single

quaderi:
Fixes the issue where this.value is not a Date object.

The issue is documented in #3868 by @sgodoy and occurs when selectionMode is "multiple"  or "range"

### DIFF
--- a/src/app/components/calendar/calendar.ts
+++ b/src/app/components/calendar/calendar.ts
@@ -1121,7 +1121,10 @@ export class Calendar implements AfterViewInit,AfterViewChecked,OnInit,OnDestroy
     }
     
     updateUI() {
-        let val = this.value||this.defaultDate||new Date();
+        let val = Array.isArray(this.value) ? this.value[0] : this.value;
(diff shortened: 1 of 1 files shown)


cagataycivici:
Fixed by another PR, thank you.

quaderi:
Hey, @cagataycivici. I think the applied fix might still cause bugs in cases where the default value is an empty array. I believe that empty arrays are probably a valid default value for multiselecting dates.

ProgzillaRadio:
This is still an issue.  We have a form with two date fields, each to contain a range, but to provide a default range wouldn't be appropriate in this case.  Subsequently, we get the '**TypeError: Cannot read property 'getMonth' of undefined**' error on every page load.

ProgzillaRadio:
...and having tried Quaderi's code change, it has resolved this issue for us, so it might be worth reconsidering his PR.

Quaderi - thanks for your help.

PR #3967, opened 2017-09-15
Implement Calendar accessibility fully support readers and all keyboard interactions

davidkirolos:
Add full accessibility support to calendar component.
Demo can be found here : [https://davidkirolos.github.io/primeng/#/calendar](https://davidkirolos.github.io/primeng/#/calendar)

# Keyboard interaction

- **Left**: Move focus to the previous day. Will move to the last day of the previous month, if the current day is the first day of a month.
- **Right**: Move focus to the next day. Will move to the first day of the following month, if the current day is the last day of a month.
- **Up**: Move focus to the same day of the previous week. Will wrap to the appropriate day in the previous month.
- **Down**: Move focus to the same day of the following week. Will wrap to the appropriate day in the following month.
- **PgUp**: Move focus to the same date of the previous month. If that date does not exist, focus is placed on the last day of the month.
- **PgDn**: Move focus to the same date of the following month. If that date does not exist, focus is placed on the last day of the month.
- **Alt+PgUp**: Move focus to the same date of the previous year. If that date does not exist (e.g leap year), focus is placed on the last day of the month.
- **Alt+PgDn**: Move focus to the same date of the following year. If that date does not exist (e.g leap year), focus is placed on the last day of the month.
- **Home**: Move to the first day of the month.
- **End**: Move to the last day of the month
- **Tab / Shift+Tab**: If the datepicker is not in inline mode, navigate between calander grid and header/grid/button bar selection buttons, otherwise move to the field following/preceding the date textbox associated with the calendar
- **Enter**: Fill the date textbox with the selected date then close the calendar widget.
- **Space**: When focused on prev/next month it will load the prev/next month. 

# Special  interaction when focus on different elements
- **Key Down** / When focus is on the **input field** and calendar overlay is hidden:  This will open the calendar overlay.
- **KeyUp | KeyDown** / when focus is on **month/year** navigator: this will load the corresponding month/year.

### DIFF
--- a/src/app/components/calendar/calendar.css
+++ b/src/app/components/calendar/calendar.css
@@ -15,7 +15,7 @@
 .ui-calendar .ui-calendar-button:enabled:hover,
 .ui-calendar .ui-calendar-button:focus {
     border-left: 0 none;
-} 
+}
 
 /* Fluid */
 .ui-fluid .ui-calendar {
@@ -188,13 +188,13 @@
 }
 
 .ui-calendar.ui-calendar-w-btn input {
-    -moz-border-radius-topright: 0px; 
-    -webkit-border-top-right-radius: 0px; 
-    -khtml-border-top-right-radius: 0px; 
+    -moz-border-radius-topright: 0px;
+    -webkit-border-top-right-radius: 0px;
+    -khtml-border-top-right-radius: 0px;
     border-top-right-radius: 0px;
-    -moz-border-radius-bottomright: 0px; 
-    -webkit-border-bottom-right-radius: 0px; 
-    -khtml-border-bottom-right-radius: 0px; 
+    -moz-border-radius-bottomright: 0px;
+    -webkit-border-bottom-right-radius: 0px;
+    -khtml-border-bottom-right-radius: 0px;
     border-bottom-right-radius: 0px;
 }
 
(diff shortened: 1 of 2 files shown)


DavyDeDurpel:
I tested your version and I see at least 1 blocking issue.

The enter key only seems to work in Chrome.  It doesn't work in IE and Firefox.

The auto-focus that you implemented prevents anyone from manually entering the date.  Each time you click the input field the focus is set to the calendar widget.

I think it would be better to only do an auto-focus in case the input field is read only or the calendar is opened with the button.

davidkirolos:
I did fixed the issues, 
demo can be found here : [https://davidkirolos.github.io/primeng/#/calendar](https://davidkirolos.github.io/primeng/#/calendar)
- Fix Enter and space buttons are not working on IE and FireFox.
- Add Close button to calender overlay It's required for accessibility, to enable it just set (showCloseButton=true).
- Auto focus on calendar overlay now happen if:
   1- User click on Show date-picker button.
   2- Calendar input filed is readOnly.
   3- Focus is on the input field and user hit KeyDown on the keyboard.
- Add accessibility reader support for "Today","Clear","Show datepicker" and "Close" buttons.
- Now "Previous Month", "Next Month", "Close" and "Show datepicker" buttons support internationalization.

cagataycivici:
Calendar.css should not use any colors, it is up to the themes.

PR #3978, opened 2017-09-18
Emmit move events on drag and drop in picklist component

jmesa-sistel:
###Defect Fixes
When you move an element the move events are not triggered (onMoveToSource, onMoveToTarget).

PS
Not sure if this is the best way to do it. 
Also I have notice that multiselect is not working with drag and drop, not sure if this is as designed or it is a bug.


cagataycivici:
Tracked at;

3978

Issue #3964, opened 2017-09-15
Dropdown empty label shows "empty"


Ketec:
```
[ ] bug report => Search github for a similar issue or PR before submitting
[ x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Adding a item with empty string causes dropdown to show "empty".
`<span *ngIf="!itemTemplate">{{option.label||'empty'}}</span>`
Adding string with just a space causes item to lose height.

**Expected behaviour**
Shows empty entry without text.

**What is the motivation / use case for changing the behaviour?**
To add a option for deselecting/choosing none (grid filters for example).

Sumragen:
Hello @Ketec 

I think in your case the best solution will be to set property `label` as `null` or `undefined` or actual empty string. I mean string like this `""` without white space like `" "`.
But keep in mind, when you use `undefined` dropdown shows empty string BUT if that value is selected  input will consist label `undefined`. Like this
![image](https://user-images.githubusercontent.com/15640910/30492824-8599c2aa-9a4a-11e7-8dcc-4981dd284bb4.png)
Moreover, in case when value of an item is `" "` label in dropdown will be like 
![image](https://user-images.githubusercontent.com/15640910/30492779-5703eca4-9a4a-11e7-8da2-d991e8f51140.png)
instead of
![image](https://user-images.githubusercontent.com/15640910/30492795-6521dbc0-9a4a-11e7-9bca-61073a9563fb.png)


Also you should remember when value is empty `<li>` element consists class `ui-dropdown-item-empty` that gives `text-indent: -9999px;`, because of `[ngClass]="{'ui-dropdown-item-empty':!option.label||option.label.length === 0}"`

I hope this helps.

Ketec:
@Sumragen using empty string or null causes primeng to show text "empty".
This is from their source: 
`<span *ngIf="!itemTemplate">{{option.label||'empty'}}</span>`
Empty string/null both equate to false.

Sumragen:
@Ketec I'm agree with you, but look at the level above.
![image](https://user-images.githubusercontent.com/15640910/30530739-f88e5e7a-9c51-11e7-8a3c-57cd407af861.png)
And what does `ui-dropdown-item-empty` class do
[...]

PR #3982, opened 2017-09-18
3964 added flag clearedEmpty and input value empty

Sumragen:
#3964 
Added flag `clearedEmpty` to make dropdown input actual empty
Added input property `empty` to make empty label more flexible for personalization


cagataycivici:
Seems to have conflicts, an issue related to this has been resolved by the way.

PR #3985, opened 2017-09-19
Feature Request: add onSourceSelect and onTargetSelect events to picklist component

jmesa-sistel:
Add onSourceSelect and onTargetSelect events to picklist component, so you can use the picklist also as a filter.


ThiagoSegato:
Thanks...

PR #3973, opened 2017-09-18
Support keyboard events in multiselect

brachi-wernick:
will solve this issue: https://github.com/primefaces/primeng/issues/3975

PR #3987, opened 2017-09-19
suport keyboard events in listbox

brachi-wernick:
Support keyboard event in listbox. this is related also to multiselct: https://github.com/primefaces/primeng/pull/3973

In this PR I'm adding key down and key up support.

when space, option is checked.

when tabbing in, highlight the first option (skip filter and toggle all input). then arrows keys should navigate thru the options.

when tabbing again, it moves out of the list to next item in the page.

### DIFF
--- a/src/app/components/listbox/listbox.css
+++ b/src/app/components/listbox/listbox.css
@@ -8,7 +8,7 @@
 }
 
 .ui-listbox .ui-listbox-list {
-    list-style-type: none; 
+    list-style-type: none;
     margin: 0;
     padding: 0;
 }
@@ -19,6 +19,7 @@
     cursor: pointer;
     font-weight: normal;
     margin-bottom: 1px;
+    position: relative;
 }
 
 .ui-listbox .ui-listbox-item > span {
--- a/src/app/components/listbox/listbox.ts
+++ b/src/app/components/listbox/listbox.ts
@@ -163,7 +163,7 @@ export class Listbox implements AfterContentInit,ControlValueAccessor {
 
 
   onOptionKeyDown(event,option){
-    let highlightItemIndex = this.findOptionIndex(this.highlightedOption);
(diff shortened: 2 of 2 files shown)


src/app/components/listbox/listbox.css
```diff
@@ -19,6 +19,7 @@
     cursor: pointer;
     font-weight: normal;
     margin-bottom: 1px;
+    position: relative;
```
brachi-wernick: this is to enable the scroll to move down and up when press arrows keys.
that is because `ui-helper-hidden-accessible` get `position: absolute`
same thing I see in `ui-multiselect-item`

src/app/components/listbox/listbox.ts
```diff
@@ -19,25 +22,31 @@ export const LISTBOX_VALUE_ACCESSOR: any = {
...
                         <span class="ui-chkbox-icon ui-clickable" [ngClass]="{'fa fa-check':allChecked}"></span>
                     </div>
                 </div>
                 <div class="ui-listbox-filter-container" *ngIf="filter">
-                    <input type="text" role="textbox" (input)="onFilter($event)" class="ui-inputtext ui-widget ui-state-default ui-corner-all" [disabled]="disabled">
+                    <input tabindex="-1" type="text" role="textbox" (input)="onFilter($event)" class="ui-inputtext ui-widget ui-state-default ui-corner-all" [disabled]="disabled">
                     <span class="fa fa-search"></span>
                 </div>
             </div>
+          <div class="ui-helper-hidden-accessible">
```
brachi-wernick: this will be the only element that will achieve focus, actually what it does is to delegate the focus to the first option

cagataycivici:
We'll do it on our side, thank you.

Dekel-Cohen:
@cagataycivici  We saw you removed it from ```4.2.3`` milestone. 
Can you please specify on which version do you think it will get in?

Issue #3979, opened 2017-09-18
Issue with not empty target list at initial (PickList)


gtzinos:
Hi,

I found an opened issue from may on stackoverflow, can someone help us ?

https://stackoverflow.com/questions/43978134/how-to-exclude-selected-items-in-primeng-picklist

Sumragen:
@gtzinos Hello,

Could you provide more detailed information regarding this issue?
As I understood, the main idea is changing list of selected items programmatically. To make some items as selected and not. Am I correct?

gtzinos:
Hi @Sumragen, i am sorry for misunderstanding, the problem is when you define the target list as a non empty array. Then it doesn't remove from the source list the selected items, it just show them duplicate at the left and at the right side. Is it clear now ?

Sumragen:
@gtzinos 
Yes, now it clearly for me.
Thanks.

cagataycivici:
Much better is you managing this on your side, so when you provide the source, also remove the items from source.

gtzinos:
@cagataycivici Really ?

crissty:
Hi guys,
I also think that this behavior will have to be controller by the component PickList.

Any update about this matter?

Thank you!

PR #3988, opened 2017-09-19
3979 PickList: added method to check duplicate items in source and ta…

Sumragen:
#3979 PickList:
added method `checkDuplicateItems()` to check repeat target items in list of source items


cagataycivici:
Commented at 3979.

PR #3997, opened 2017-09-20
Fix #3996

### LINKED ISSUE #3996: Calendar range/multiple default values cause error
### There is no guarantee in receiving a response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeNG PRO Support* where support is provided within 4 business hours

```
```

**Plunkr Case (Bug Reports)**

Example:
http://plnkr.co/edit/djr4zpNiJyVXJqOapkHp?p=preview

**Current behavior**
Fail when you set a default value with range / multiple dates because calendar expected a Date instead of a array

**Expected behavior**
Calendar must check if is a array or date in this line:
https://github.com/primefaces/primeng/blob/master/src/app/components/calendar/calendar.ts#L1123

Set a range/multiple type date and set array Dates as default value.

Resolve the problem

Linux,  Angular 4.4. and Prime 4.2.1

* **Browser:**  ALL

MiguelAngelLV:
Fixed problem with array default values with range and multiple selection mode

### DIFF
--- a/src/app/components/calendar/calendar.ts
+++ b/src/app/components/calendar/calendar.ts
@@ -1123,7 +1123,10 @@ export class Calendar implements AfterViewInit,AfterViewChecked,OnInit,OnDestroy
     updateUI() {
         let val = this.value||this.defaultDate||new Date();
         this.createMonth(val.getMonth(), val.getFullYear());
-        
+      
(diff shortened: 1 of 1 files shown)


cagataycivici:
Merged with a different PR, thank you.

Issue #4004, opened 2017-09-21
confirm-dialog: Error when hitting Escape after dialog is closed


crathmann:
**I'm submitting a ...**
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**

http://plnkr.co/edit/ot8wLmd231lFaJNdMVgC?p=preview

**Current behavior**

When I hit the Escape button after the dialog is closed (closeOnEscape set to true), the following error occurs (Chrome console log):

`ERROR TypeError: Cannot read property 'rejectEvent' of null
    at ConfirmDialog.close (confirmdialog.ts:227)
    at eval (confirmdialog.ts:167)
    at HTMLDocument.eval (dom_renderer.ts:75)
    at ZoneDelegate.invokeTask (zone.js:398)
    at Object.onInvokeTask (ng_zone.ts:288)
    at ZoneDelegate.invokeTask (zone.js:397)
    at Zone.runTask (zone.js:165)
    at HTMLDocument.ZoneTask.invoke (zone.js:460)`

When I set closeOnEscape to false, no error occurs.

**Expected behavior**

No error occurs.

**Minimal reproduction of the problem with instructions**

Open the dialog by clicking "Confirm", then close it clicking either "No" or "Yes". Hit Escape after the dialog is closed.

* **Angular version:** 4.2.6

* **PrimeNG version:** 4.2.0

* **Browser:** [Chrome 60.0.3112.113 | Firefox 55.0.3]

PR #4006, opened 2017-09-21
4004 confirmDialog: rejectEvent of null error

Sumragen:
#4004 ConfirmDialog
Fixed error: 
`ERROR TypeError: Cannot read property 'rejectEvent' of null`


cagataycivici:
Merged manually.

Issue #4002, opened 2017-09-20
Filtered listbox toggle all fails with new items


brandonPauly:
### There is no guarantee in receiving a response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeNG PRO Support* where support is provided within 4 business hours

**I'm submitting a ...**  (check one with "x")
```
[x ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://plnkr.co/edit/2qeE3EaVRu5AB3ZMSbd5?p=preview

**Current behavior**
1. Type the letter u (or arbitrary text) into the input filter.
2. Delete the text in the input search filter.
3. click the 'Add Unit' button to add another unit to the list.
4. use the 'select all' checkbox.  Not all the items will be checked.

**Expected behavior**
Expecting all visible list items to be checked.

**What is the motivation / use case for changing the behavior?**
I cannot imagine this is expected behavior.

PR #4007, opened 2017-09-21
4002 listbox: select all after filter

Sumragen:
#4002 Listbox


cagataycivici:
Fixed in a different way, thank you. We rather not maintain visibleOptions anymore.

PR #4010, opened 2017-09-21
Fix #4009 - Add name attribute to accessible select elements for p-dropdown

### LINKED ISSUE #4009: name property for Dropdown
```
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/4Ct8Kvb3kkMPyMUlEefe?p=preview

**Current behavior**
All `select` elements generated with `p-dropdown`'s have the same value for `name` attribute ('options').

I believe this is the line that causes this:

https://github.com/primefaces/primeng/blob/master/src/app/components/dropdown/dropdown.ts#L24

**Expected behavior**
The `select` elements should use the name that was passed into the `p-dropdown`

http://plnkr.co/edit/4Ct8Kvb3kkMPyMUlEefe?p=preview

Click "Log form selects" and you'll see that the select elements all have the same name.

I'm trying to use Prime NG components on a form that is submitted using the default form submit behavior (e.g. action="" and method="" on the form). 

CentOS, Chrome

* **Angular version:** 4.3.2

* **PrimeNG version:** 4.2.0

* **Browser:** [Chrome ]

 
* **Language:** [TypeScript]

VeryBueno:
###Defect Fixes
Adds `name` attribute to the `select` elements generated by `p-dropdown`.

### DIFF
--- a/src/app/components/dropdown/dropdown.ts
+++ b/src/app/components/dropdown/dropdown.ts
@@ -21,7 +21,7 @@ export const DROPDOWN_VALUE_ACCESSOR: any = {
             'ui-state-disabled':disabled,'ui-dropdown-open':panelVisible,'ui-state-focus':focus}"
             (click)="onMouseclick($event)" [ngStyle]="style" [class]="styleClass">
             <div class="ui-helper-hidden-accessible" *ngIf="autoWidth">
-                <select [required]="required" name="options" [attr.aria-label]="selectedOption ? selectedOption.label : ' '" tabindex="-1" aria-hidden="true">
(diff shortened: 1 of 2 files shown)


cagataycivici:
The one that screenreaders read is;

```
<div class="ui-helper-hidden-accessible">
                <input #in [attr.id]="inputId" type="text" [attr.aria-label]="selectedOption ? selectedOption.label : ' '" readonly (focus)="onInputFocus($event)" role="listbox"
                    (blur)="onInputBlur($event)" (keydown)="onKeydown($event)" [disabled]="disabled" [attr.tabindex]="tabindex" [attr.autofocus]="autofocus">
            </div>
```

Issue #4016, opened 2017-09-22
Autocomplete: enable multiple selection at once


lionelmbafumoya:
**I'm submitting a ...**
```
[x] feature request
```

**Current behavior**
Given an autocomplete with the multiple attribute set to true, the values have to be selected one at the time with the panel closing each time.

**Expected behavior**
It would useful to be able to select multiple values at once similar to the MultiSelect behavior.

**Minimal reproduction of the problem with instructions**
1. Go to https://www.primefaces.org/primeng/#/autocomplete
2. Type the letters 'Bel' and try to add 'Belgium' and 'Belarus' at the same time in the Multiple autocomplete component.

**What is the motivation / use case for changing the behavior?**
The goal of an autocomplete that selects multiple values is to not only have real-time suggestions but also to be able to select as many values as needed (Similar to the MultiSelect component).
It is counter-intuitive to have to type similar queries when they share common letters.

**Please tell us about your environment:**

* **Angular version:** 4.0.2

* **PrimeNG version:** 4.2.1

* **Browser:** Chrome 60.0.3112.113

cagataycivici:
We'll reconsider if more users demand it, thank you.

bevrard:
We also need this feature.
Could you reconsider the demand?
Thanks in advance!

s13571:
It would be really great in my current project. Please reconsider, it's awesome feature

joster422:
+1

mamoun90:
+1

PCASME:
+1

bossqone:
+1

vatsalpande:
+1. Highly needed.

changulpaye:
+1 Required in current project please consider it

souravsaraf:
+1 Required. Please consider.

bhanuunrivalled:
required 
consider it

mcrtricolor:
Please consider adding this feature.

SunnoSakib:
+1

ibrdavid:
+1

MSauer89:
+1

PR #4017, opened 2017-09-22
#4016 Autocomplete multiple selection

Sumragen:
#4016 Autocomplete
changed behavior on multiple item selection


cagataycivici:
We'll reconsider if more users demand it, thank you.

PR #4033, opened 2017-09-26
fix(calendar): use ng-container instead of span

H--o-l:
I'm sumitting a small fix.
Following modification in primeng introduce date template in calendar component:
89e7d3aee0d4ca1dea870af01458d36b62d88513

In our app we don't use date template but upgrading primeng we had a regression on the display:
calendar

Not completly sure why, it's seems than the new `span` level introduced by 89e7d3aee0d4ca1dea870af01458d36b62d88513 for the 'no custom template case' didn't do any good.

I believe that's not our app, as we don't have many style on calendar, but primeng css rules for `.ui-datepicker-calendar td a` and `.ui-datepicker-calendar td span` that are applied two times now (once for the `a` and once for the new `span`).

Anyway, angular `ng-container` are here for that. It's good to used them, and recommended by angular doc, instead of adding span or div each time.

Moreover for us it does the trick and will prevent regression for others to.


cagataycivici:
Good catch, thank you.

PR #4054, opened 2017-09-28
Fix #4053 - change detection performance issue

### LINKED ISSUE #4053: Performance improvements for Tooltip
```
```

**Plunkr Case (Bug Reports)**

https://plnkr.co/sJJik8Qgi9HGU9t5Z5CM

**Current behavior**

Tooltip and Dialog calling change detection many times (each mousemove event, especially)

**Expected behavior**

Run 'heavy events' (mousemove, especially) outside change detection, unless that you need to update any "binded var" each time this events run, which is not usually the case with components. This responsibility is with who will use the library. 

See Plunkr

This is a critical performance issue. Each time change detection are called, you force a get in each attribute of each component. Imagine an application with a list of grids, for example. (for more informations, see this great article: https://blog.thoughtram.io/angular/2016/02/22/angular-2-change-detection-explained.html)

* **Angular version:** 4.0.0

* **PrimeNG version:** 4.2.0

* **Browser:** [all]

 
* **Language:** [ TypeScript 2.3.3 ]

I'am fix this issue in my repository and i'm send this issue before send a pull request. I'm create a ChangeDetectionUtil class to facilitate fix others similar issues (I see this behavior in others components)

Ps: Sorry my little poor english.

carlosearaujo:
###Defect Fixes
https://github.com/primefaces/primeng/issues/4053


cagataycivici:
Tracked at 4080 and 4053

PR #4066, opened 2017-10-01
reorderableColumns not work in scrollable dataTable

yasin-mansour:
###Defect Fixes
using scrollable attribute break the reorderableColumns behavior in dataTable


cagataycivici:
Thank you, I'll move the column refresh to a method and apply it.

yasin-mansour:
Thanks a lot @cagataycivici, but is there any way I can get this fix in the npm package?

PR #4069, opened 2017-10-01
Fix small typo in the input text showcase

kreuzerk:
In the input text documentation under the source tab there is written **{text}}** instead of **{{text}}**


cagataycivici:
Thank you.

PR #4099, opened 2017-10-04
Reduce unnecessary onVirtualScroll callbacks

machfalcon:
If the page does not change from the current page, do not emit a page change event or change the scrollTable.style.top.

###Defect Fixes

###Feature Requests


cagataycivici:
Please create an issue first with details, it is not clear what this PR does, also we should avoid introducing variables like currentPage to avoid maintaining them.

Issue #4119, opened 2017-10-06
datatable radio/checkbox selections not updated at events


Ketec:
### There is no guarantee in receiving a response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeNG PRO Support* where support is provided within 4 business hours

**I'm submitting a ...**  (check one with "x")
```
[ x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

*
**Current behavior**
selectRowWithRadio
toggleRowWithCheckbox
toggleRowsWithCheckbox

Are missing `this.selectionChange.emit(this.selection);` therefor the selection is not up to date when listening to the onRowSelect event.

wooleverj:
I ran into this same thing and changed the selection emitter to be before the onRowSelect event where it currently is after the event gets fired.  I believe this pull request will help this issue.

cagataycivici:
PR merged.

PR #4126, opened 2017-10-06
Update datatable to emit selection before rowselect

wooleverj:
I believe this will address the issue #4119.  The problem is that the datatable was emitting the selection after it fired the onRowSelect event.  This makes it seem like the selection array never got updated.

###Defect Fixes

datatable radio/checkbox selections not updating selection array #4119

###Feature Requests


TakanashiOuken:
Having the same issue here~
Hope to merge soon

vleung:
Having the same issue...
This affects the current behavior of the datatable in our application, please review it ASAP.

PR #4129, opened 2017-10-08
Add the option "addOnBlur" to the chips input component so that when …

ryanrupp:
…enabled a new chip is automatically created when the input component loses focus.


ryanrupp:
It looks like recent changes to the API/events make it probably possible to do this but it's nice to have this as a standard option. We have users where they'll click a submit button right after entering text rather than doing a tab/enter and we wan those cases to turn the outstanding input into a chip.

jbiversen:
Any idea when this will be merged and released? It would very much appreciate this feature.

ryanrupp:
Fixes https://github.com/primefaces/primeng/issues/4218

PR #4138, opened 2017-10-09
Refactor datatable: Reuse ObjectUtils.resolveFieldData

carlosearaujo:
Refactor to use objectUtils provider.


cagataycivici:
Nice catch, thanks.

Issue #4149, opened 2017-10-11
Add method to reset sizes of manually resized data table columns


Cito:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[X] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**PrimeNG version:** 4.2.3

**Feature request**

For a data table with `resizableColumns="true"` there is currently no way to restore the widths of the columns to their initial state. It would be helpfuls to have such a method.

**Motivation / use case**

* I'd like to have button in the table header that allows restoring everything to a "default view" of the table data.
* The resizableColumns feature does not play well with the colToggler feature, because the column widths set in fixColumnWidths()  do not harmonize with the widths of added ore removed columns. A solution would be to reset the column widths whenever the columns are toggled.

PR #4150, opened 2017-10-11
Add method to reset resizable DataTable columns

Cito:
This is a proposed sultion for #4149 : The DataTable currently is lacking a method to reset manually resized column widths. This patch adds a method resetResizableColumns() to the DataTable that does just that.


cagataycivici:
DataTable is deprecated in favor of Table and not maintained anymore.

PR #4181, opened 2017-10-15
removed TSLint errors

menyuscraft:
no functional changes,
only brackets, whitespaces, let -> const changes


cagataycivici:
We'll do it on our side thank you.

PR #4191, opened 2017-10-17
feat(dropdown): option to ignore keyboard events

H--o-l:
Following update to PrimeNG 4.2, I have issue with keydown events on dropdown.

I propose to add an option to disable them entirely.


cagataycivici:
Sorry, I don't see a reason how this can be useful.

PR #4203, opened 2017-10-18
Protect this.value in ngDoCheck function

jmesa-sistel:
###Defect Fixes
Protect this.value in ngDoCheck function.

If I modify  option property and then try to unselect an item I get an error in ngDoCheck
Hard to create a plunker but I can try if it is needed


cagataycivici:
#4233

Issue #3521, opened 2017-07-27
Custom filterFunction for datatable filters similar to sortFunction


Ketec:
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
**Current behavior**
Datatable filter cannot be used with objects/arrays.
If i pass in a object (or array of objects), for example `{name: 'test', id: 1}` I would like to be able to have the column filtered by name. Because in a custom column template i use the name field as  a display value but on rowclick i need to get the id to further doSomething() with the id. Therefor the object as data.

In some cases i may have an array of items or array of objects:
`[1,2,3,4,5] or [{role: 'test', id: 1},{role: 'test2', id: 2}]`

Or when using a range filter component (like from year to year).

**Expected behavior**
Can specify custom filter method similar to sortFunction.

cagataycivici:
p-dataTable is deprecated and will be removed in favor of the new p-table (aka TurboTable) of 5.1.0 so closing the issue. Please try the new p-table once 5.1.0 is released.

kareljan:
Can we reopen this? still not available in the new TurboTable. 
I would also like this.

PR #4226, opened 2017-10-21
allow custom filtering using the same mechanism as custom sort

bwobbones:
###Defect Fixes
Fixes defect #3521 

Replicates the custom sortFunction mechanism with a filterFunction.

Note that is also now allows a blank string to be a filter, this is how you would clear your fliter function.

Usage: 

`<p-column *ngFor="let columnDef of columnDefs" [filter]='columnDef.filter' filterMatchMode='custom' (filterFunction)='onFilter($event)'>`

which will then call an 

`onFilter(filterMeta: FilterMetadata)`

method on the component.  

FilterMetadata looks like this:

```
export interface FilterMetadata {
    field?: string;
    value?: any;
    matchMode?: string;
}
```

### DIFF
--- a/src/app/components/common/filtermetadata.ts
+++ b/src/app/components/common/filtermetadata.ts
@@ -1,4 +1,5 @@
 export interface FilterMetadata {
+    field?: string;
     value?: any;
     matchMode?: string;
 }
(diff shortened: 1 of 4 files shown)


cagataycivici:
Why change isFilterBlank?

bwobbones:
I found that if I didn't allow empty strings, the filter wouldn't clear when the textbox became empty.  FilterMetadata retained the first character in the filter.

Changing it didn't seem to have any negative impact.

cagataycivici:
DataTable is deprecated in favor new p-table, feel free to port this PR to TurboTable.

Issue #4252, opened 2017-10-24
ui-btn displayed as default text when there is no label set on buttons.


BustyLoli-Chan:
https://github.com/primefaces/primeng/blob/d06afd823cbac923b5cf3042b33efa64ba4ba53e/src/app/components/button/button.ts#L34

It would be nice to have the option to have an empty label, instead of all buttons just defaulting to have the text `ui-btn` when `btn.label` is falsey. I can think of a lot of situations in which an end user would prefer to just have a button that's _just empty_. I can think of almost **0** situations where the end user would prefer to have their button say `ui-btn`.

Edit: it is known by issue https://github.com/primefaces/primeng/issues/1657 that you can hide the text by adding 
```
.ui-button-text-only .ui-button-text{
    display: none !important;
}
```
but I'd be lying if I said this doesn't seem dumb and like it shouldn't even be necessary or an issue to begin with.

Edit edit: I have opened a pull request for this issue here.
https://github.com/primefaces/primeng/pull/4254

cagataycivici:
I can't replicate the issue to begin with, create a plunkr please.

BustyLoli-Chan:
http://plnkr.co/edit/dSx8dBkXUQJXWtZHIOvc?p=preview
Ok done.

PR #4254, opened 2017-10-24
Default to an empty label instead of text that says ui-btn when btn.label is falsey.

BustyLoli-Chan:
### Defect Fixes
This fixes the fact that the button text defaults to ui-btn instead of just being empty when a label isn't set.
https://github.com/primefaces/primeng/issues/4252


cagataycivici:
Commented at issue

BustyLoli-Chan:
I also commented on the issue and updated it with repro steps.

robbiemu:
what was the resolution to this issue?

PR #4260, opened 2017-10-25
Range filter match mode for DataTable

sappo:
Solution: Add between filterConstraint which allows the user to pass
a filter object with start and end attributes.

The filter object must be like: {'start': '15', 'end': '50'}

The second commit updates the showcase!


sappo:
@cagataycivici is there any reason this PR is pushed back once again? I just tried it with latest master and it works fine :)

cagataycivici:
DataTable is deprecated in favor of p-table so not maintained anymore, feel free to provide a PR to the new TurboTable instead.

PR #4265, opened 2017-10-25
Horizontal virtual scrolling

donriver:
added horizontal virtual scrollinfg
added attribute virtualScrollCellLoadingText to show text instead of "obsolete" data when virtual scroll enabled


cagataycivici:
There are so many unrelated files, please fix the PR to make sure only related files are included.

Issue #4293, opened 2017-10-26
Internal span of file upload component remains disabled 


VDubber:
```
[X ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/wkcf4RCj03ecMZJfIZ7V?p=preview

**Current behavior**
When the input attribute disabled is set to **false**, a internal span **remains disabled**.
```
<span class="ui-fileupload-choose ui-button ui-widget ui-state-default ui-corner-all ui-button-text-icon-left" icon="fa-plus" pbutton="" ng-reflect-klass="ui-fileupload-choose" ng-reflect-ng-class="[object Object]" ng-reflect-label="Choose" ng-reflect-icon="fa-plus"
          disabled="false"> 
       ...
</span>
```

**Expected behavior**
Internal span does not have disabled attribute applied when component input disabled attribute is false.

**Minimal reproduction of the problem with instructions**
Create a fileUpload component with [disabled]="false"
DOM indicates that the span is disabled.
By w3c standards, the presence of the disabled attribute is what actually disables the element.

**What is the motivation / use case for changing the behavior?**
Styles meant to be applied to disabled elements are misapplied.

PR #4294, opened 2017-10-26
Fix disabled attribute on internal span (has class 'ui-fileupload-choose')

VDubber:
### Defect Fixes
Styles meant to be applied to disabled elements are misapplied.

See issue: https://github.com/primefaces/primeng/issues/4293


cagataycivici:
This issue was fixed, thank you.

Issue #4301, opened 2017-10-27
Vertical Trees can be forests, but horizontal Trees can't


Cito:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[X] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

I'll also submitting pull request #4302 to fix this.

**Plunkr Case (Bug Reports)**

http://plnkr.co/edit/GRsvgNuh8iql4HzIK3I0?p=preview

**Current behavior**

Currently, if you show a Tree with more than one root (i.e. a forest) in vertical layout, it works as expected. However, if you switch to horizontal layout, only the first tree is shown.

**Expected behavior**

Vertical and horizontal layout should work the same, showing all trees of a forest.

**Minimal reproduction of the problem with instructions**

See [Plunkr](http://plnkr.co/edit/GRsvgNuh8iql4HzIK3I0?p=preview).

You can also use the PrimeNg [Tree demo](https://www.primefaces.org/primeng/#/tree) to see this. In the demo, the horizontal tree uses filesTree11, which has been set up with an additional single root element. If you set the tree attribute of the horizontal tree to filesTree10 instead of filesTree11, you only see the first of the three trees.

**What is the motivation / use case for changing the behavior?**

PrimeNG should behave more consistently. I want my users to be able to switch between the two layout modes. They would be surprised when in horizontal mode suddenly some trees are missing.

* **Angular version:** 4.4.3

* **PrimeNG version:** 4.3.0

* **Browser:** all

* **Language:** TypeScript 2.3.4

cagataycivici:
No plans to support this, horizontal tree starts with a single root. If there is more demand, we'll reconsider, thank you.

Cito:
Please reconsider, as the fix is very simple and obvious, does not need additional code, and makes horizontal layout work the same as vertical. Otherwise,  the fact that horizonzal trees have restrictions compared with vertical ones should be documented, because this is not what the user of the component expects.  In my view this is a bug since the `layout` attribute should only affect *how* things are displayed, not *which* things are displayed.

One use case for me is a button that allows users switching from horizontal to vertical layout mode. It is not acceptable that when switching the layout mode, suddenly whole branches are missing.

paddysr:
+1

[...]

PR #4302, opened 2017-10-27
Display horizontal Trees with multiple roots

Cito:
This is a fix for issue #4301.


cagataycivici:
No plans to support this, horizontal tree starts with a single root. If there is more demand, we'll reconsider, thank you.

Cito:
This is sad because it is an extremely simple and straightforward fix in lines 335/336, that makes horizontal trees work the same as vertical trees (consistency is always a good thing). The diff only looks large because of whitespace changes (my editor stripped all the trailing whitespace).

PR #4309, opened 2017-10-30
Support for html content in select button icon.

nikisak:
This change will enable adding custom image or icon with the select button label which will give better visual experience.

###Defect Fixes

###Feature Requests


cagataycivici:
Better to do it with templating instead of insecure innerHTML.

jibin-tj:
how do we do this with templating in selectButton? Please help.

PR #4340, opened 2017-11-01
Update the password strength to use zxcvbn

### LINKED ISSUE #2949: Use zxcvbn as Password Strength Estimation
```
```

**Current behavior**
When using the PrimeNG Input Password, typing a 1234567 password is marked as a medium password, and when you type 123456789123456789, this password is marked as a strong password.
But clearly it is a weak password.

**Expected behavior**
Mark passwords that are repeats, sequences or keyboard patterns such as weak passwords.

Using [Password Input
](https://www.primefaces.org/primeng/#/password)

When using [zxcvbn ](https://github.com/dropbox/zxcvbn/)(from the Dropbox development team). Most of these problems would be solved.
The lib can recognize patterns, common passwords, names and surnames. WordPress and some other CMS's already use zxcvbn.
And this could bring more security to the user, because instead of having 3 levels (weak, medium and strong), we would have 5 levels:

- 0 # too guessable: risky password. (guesses < 10^3)
- 1 # very guessable: protection from throttled online attacks. (guesses < 10^6)
- 2 # somewhat guessable: protection from unthrottled online attacks. (guesses < 10^8)
- 3 # safely unguessable: moderate protection from offline slow-hash scenario. (guesses < 10^10)
- 4 # very unguessable: strong protection from offline slow-hash scenario. (guesses >= 10^10)

**Angular version:** 2.0.X
4.1.X

**PrimeNG version:** 2.0.X
4.0.3

mfcar:
Updates the Password strength indicator, to use the [zxcvbn](https://github.com/dropbox/zxcvbn) library, which provides a better indication if the password entered is weak or strong.

I used this library because it can identify patterns, sequences, common words, among others: [zxcvbn: realistic password strength estimation](https://blogs.dropbox.com/tech/2012/04/zxcvbn-realistic-password-strength-estimation/)

Close #2949

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,13 +1,13 @@
 {
   "name": "primeng",
-  "version": "4.2.2-SNAPSHOT",
+  "version": "4.3.1-SNAPSHOT",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
     "@angular-devkit/build-optimizer": {
-      "version": "0.0.21",
-      "resolved": "https://registry.npmjs.org/@angular-devkit/build-optimizer/-/build-optimizer-0.0.21.tgz",
-      "integrity": "sha512-YK+wss6MFpLbuf1hxcgiwFJITTRcvjDq0MQkrMHbX2zonmgBT/olcvm6JoMsvrm9iQ/HETUgUY55koih/pwlJg==",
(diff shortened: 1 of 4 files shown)


cagataycivici:
Sorry it creates a dependency on this library, can you make this optional?

Issue #1657, opened 2016-12-22
Button with empty label and text shows "ui-button"


Emdee89:
**I'm submitting a bug report.

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/Vtp8OTakHAekaZOc6kUl?p=preview

**Current behavior**
I have a button with no label and no further content (only want to show its icon):
`<button pButton id="menuButtonWeights" type="button" label="" icon="fa-balance-scale" class="ui-button-secondary"></button>`
The button shows the text "ui-button" right to the icon.

**Expected behavior**
Only the icon should be visible, thus the text "ui-button" should not occur.

**Minimal reproduction of the problem with instructions**

**What is the motivation / use case for changing the behavior?**
I only need a button with an icon and w/o text.

**Please tell us about your environment:**
PrimeNG 1.1.1, Angular 2.4.0

* **Angular version:** 2.0.X
2.4.0

* **PrimeNG version:** 2.0.X
PrimeNG 1.1.1

All browsers

Mrtcndkn:
Hi,

It's the default label value of p-button.If you describe a label="" like this it will show default value.If you dont want anything to write just delete label property.

rafaelcoutinho:
Think it's related to this issue. I want to use Material icons (other than font-awesome ones). If I create a tag like this:
` <button pButton type="button" ><i class="material-icons">accessible</i> MyBtn</button>`

It is displaying the ui-button text
![image](https://cloud.githubusercontent.com/assets/908621/22120092/4ca0c13e-de65-11e6-9a26-4f378ecd594f.png)

Mrtcndkn:
You should do that via css you can take look at ultima-ng

http://www.primefaces.org/ultima-ng/#/sample

JuanELo-O:
Hi!,

I have a similar issue when I try to interpolate a label value, in older releases I don't have this problem.

This is my code:

`````
<button type="button"
                pButton
                icon="fa-close"
                label="{{'no' | translate}}"
                class="secondary"
                (click)="confDialog.reject();"></button>
``````

And this is my button:

![buttonno](https://cloud.githubusercontent.com/assets/6380354/22446926/f4e888ec-e714-11e6-931b-e2124609a884.png)

Only if I put the label directly it works.

nguyenjk:
I have no label but it is still showing ui-button as a text. 

angular: 2.4.4
primeng: 2.0.1

Kappyh:
[...]

PR #4342, opened 2017-11-01
Default to no button text when label is not set.

BustyLoli-Chan:
This is a pull request to fix the issue demonstrated by this Plunkr.
http://plnkr.co/edit/dSx8dBkXUQJXWtZHIOvc?p=preview

I've modified this line:
https://github.com/primefaces/primeng/blob/d06afd823cbac923b5cf3042b33efa64ba4ba53e/src/app/components/button/button.ts#L34
To allow for situations in which an end user would prefer to just have a button without a label so that they may edit the actual HTML inside of the button tag.

It is currently known by issue https://github.com/primefaces/primeng/issues/1657 that you can currently hide this text with the following styling 
```
.ui-button-text-only .ui-button-text{
    display: none !important;
}
```
but this seems like it should be unnecessary.

### DIFF
--- a/src/app/components/button/button.ts
+++ b/src/app/components/button/button.ts
@@ -31,7 +31,7 @@ export class Button implements AfterViewInit, OnDestroy {
         
         let labelElement = document.createElement("span");
         labelElement.className = 'ui-button-text ui-clickable';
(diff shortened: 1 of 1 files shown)


BustyLoli-Chan:
Has there been any update on this?

PR #4382, opened 2017-11-07
fix travis build regarding node cache, version and webpack dependency

nhnb:
This pull requests fixes the travis test run. It consist of three changes that are all required. 

* change Travis environment to use the LTS version of node because Travis defaults to Node 9, which is incompatible to npm.
* disabled caching of the node_modules directory because npm fails with weired errors with active cache on the second run
* added webpack to devDependencies

I am not really happy about adding webpack to devDependencies because it may need to be adjusted with the next update of Angular. The explicit dependency on webpack is required since commit 9bdca3f780268 (update to Angular 5). Older version of primeng run successfully with the other two changes only. Furthermore, I do not need the webpack-dependency locally even with the latest primeng master.


BustyLoli-Chan:
This seems to be the only pull request that has passed in days... 
Does anyone at PrimeNg actually care about open source contributions or maintaining this product?

nhnb:
Hello @BustyLoli-Chan I am a bit confused about your comment. As far as I can tell, the PrimeNG developers accept pull request on a regular basis. There are 699 closed PRs (the majority of them was either merged directly or accepted in a slightly modified form) and only 30 open.

So far, all of my other PRs have been accepted. Regarding this pull request: As I stated earlier, I am not completely happy about it myself. From my point of view, it is completely understandable that the PrimeNG developers prefer to deactivate Travis for now instead of introducing a bogus dependency.

cagataycivici:
PRs are reviewed on a timely manner, please note that we are working on PrimeNG and bunch of other things, please understand that tasks take time.

cagataycivici:
Is there a way to avoid adding webpack to devDependencies, I wonder what other libs do.

nhnb:
I had another look and noticed a package-lock.json, which was committed before the update to Angular 5 and not modified since. Dependencies specified in package.json override entries in package-lock.json: therefore Angular 5 was pulled in. But webpackage was not.

Deleting this file (and the changes to .travis.yml to use node lts) fix the Travis build.

cagataycivici:
Will do after 5.2. final.

cagataycivici:
Removed package-lock

Issue #4389, opened 2017-11-08
Autocomplete does not honor forceSelection="false"


mbaer3000:
The autocomplete component does not honor forceSelection="false". Only by leaving out the property can this be achieved.

sudheerj:
The demo example is updated with data-binding property.

PR #4390, opened 2017-11-08
Make forceSelection="false" work

mbaer3000:
Setting forceSelection="false" attribute for <p-autoComplete> does now work, bringing it in line with the documentation for the forceSelection property at https://www.primefaces.org/primeng/#/autocomplete

Fixes https://github.com/primefaces/primeng/issues/4389

### DIFF
--- a/src/app/components/autocomplete/autocomplete.ts
+++ b/src/app/components/autocomplete/autocomplete.ts
@@ -511,7 +511,7 @@ export class AutoComplete implements AfterViewInit,AfterViewChecked,DoCheck,Cont
         this.onModelTouched();
         this.onBlur.emit(event);
         
-        if(this.forceSelection && this.suggestions) {
+        if(this.forceSelection && this.forceSelection !== 'false' && this.suggestions) {
             let valid = false;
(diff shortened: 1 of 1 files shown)


nhnb:
The cause of the issue seems to be a confusing about HTML-attributes and Angular properties. In HTML, an attribute evaluates to true, if it exists without value-definition or if it has any value (including the string "false"). 

Therefore all of the following lines set forceSelect to true:

~~~~
<p-autoComplete forceSelection>
<p-autoComplete forceSelection="true">
<p-autoComplete forceSelection="something">
<p-autoComplete forceSelection="false">
~~~~

The only way to set an attribute to false in HTML is to omit it. In Angular properties, however, can be set to false explicitly:

~~~~
<p-autoComplete [forceSelection]="false">
~~~~

The linked example uses `forceSelection="true"`. Although this is correct, I suggest to change it to either `[forceSelection]="false"` or just `forceSelection` to minimize confusion.

mbaer3000:
Thanks @nhnb for the clarification. Yes, it'd be good to update the documentation as you suggest.

PR #4410, opened 2017-11-09
Edited emptyMessage as InnerHTML for datatable

ManojSethuraman:
Have edited the EmptyMessage config as innerHTML so that if any styling required for empty message can be done.


Ketec:
Doesn't this already exist when using pTemplate="emptymessage" for styling.

cagataycivici:
DataTable is deprecated in favor of p-table, please use p-table instead. It offers an emptyMessage template to do this properly.

PR #4415, opened 2017-11-10
can dynamic scrollHeight in Datatables

### LINKED ISSUE #4114: Datatable's scrollHeight does not change on a binding anymore (broken since 4.2.2)
```
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/4YrOKBik8IIX6uguThsE?p=preview
(Wait for 5 and then 10 seconds after running)

**Current behavior**
When a datatable's scrollHeight is set to a fixed value, the height and scrollbar is set as expected. But when the input is set via a binding variable (thus, [scrollHeight]="myvariable"), the scrollHeight is only set once at the beginning and never updates -- but it did update in PrimeNg <= 4.2.0.

**Expected behavior**
I would expect the scrollHeight to update again.

* **Angular version:** 4.2.6

* **PrimeNG version:** 4.2.2

* **Browsers:** All

mix5003:
i think it should fix #4114

### DIFF
--- a/src/app/components/datatable/datatable.ts
+++ b/src/app/components/datatable/datatable.ts
@@ -305,6 +305,8 @@ export class ScrollableView implements AfterViewInit,AfterViewChecked,OnDestroy 
     @Input() frozen: boolean;
     
     @Input() width: string;
+
+    _scrollHeight: any;
     
     @Input() virtualScroll: boolean;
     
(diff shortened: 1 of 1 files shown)


bbuehrer:
I have the same problem with primeng 5.0.2.

cagataycivici:
DataTable is deprecated in favor of new p-table so not maintained anymore.

Issue #2157, opened 2017-02-26
[SimpleFeatureRequest]Add a Cell style function like row style function "rowStyleClass" in DataTable component.


Lyoko-Jeremie:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[ X ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**

Current have a row style function "rowStyleClass", but not have a cell style function maybe like "cellStyleClass" and so on.

**Expected behavior**

add a cellStyleClass that : 
"gets the row data and row index as parameters and returns a style class for the <b>cell</b>".
that function can return a style class to change cell background-color and so on.

**Minimal reproduction of the problem with instructions**

only add a cellStyleClass like rowStyleClass at the &lt;td&gt; element in datatable.ts 's line 162: TableBody compoment's template html. and add support function in the DataTable compoment like getRowStyleClass function.

**What is the motivation / use case for changing the behavior?**

example, change cell background-color to make a color-full timetable.
use <code>div</code> in template cannot fill the full cell.

**Please tell us about your environment:**

latest WebStorm

**Language:** [all | TypeScript X.X | ES6/7 | ES5]
all

Lyoko-Jeremie:
i use follow hack directive to solution it, so, i think i dont need it now !!

```ts
import {Directive, ElementRef, OnInit, Renderer, Input, OnDestroy, Output, EventEmitter} from '@angular/core';
import * as _ from 'lodash';
import {ReplaySubject} from "rxjs";

@Directive({
    selector: '[ccParentsSetter]'
})
export class ParentsSetterDirective implements OnInit,OnDestroy {

    @Input('classObject') set classObject(value: any) {
        this.classObjectSubject.next(value);
    }

    @Input('styleObject') set styleObject(value: any) {
        this.styleObjectSubject.next(value);
    }

    @Input('attributeObject') set attributeObject(value: any) {
        this.attributeObjectSubject.next(value);
    }

    // (click) function
    @Input() onClickParam: any;
    @Output('ccParentsSetterOnClick') onClick: EventEmitter<any> = new EventEmitter();
    private onClickListenUnhookFunc: Function|undefined;

    // cold Subject
    // use to dynamic flush all the thing
[...]

PR #4463, opened 2017-11-16
Adds cellStyleClass to Column class to allow contextual cell styling

protonyx:
See #2157


cagataycivici:
This will slow down the table due to constants calls I'm afraid, we need a performant solution.

PR #4466, opened 2017-11-17
Adjust site.css to show last items of sidebar.

CarlinhosSousaJunior:
The last items of sidebar were not be visible because 100% value on height property.

###Defect Fixes
When height property was 100% without subtract top property value, the last items of sidebar was hidden.

### DIFF
--- a/src/assets/showcase/css/site.css
+++ b/src/assets/showcase/css/site.css
@@ -186,7 +186,7 @@ body {
     position: fixed;
     left: 0;
     top: 70px;
-    height: 100%;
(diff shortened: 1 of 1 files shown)


sudheerj:
Previously I observed this issue with last menu group(misc). Need to check once whether it cause any layout misalignments by removing the top property.

CarlinhosSousaJunior:
@sudheerj The problem is not only with last menu group(misc), but with all sidebar, because the top property is hidden last elements of menu, and when you subtract the top value of height (calc(100% - 70px)), all works as expected.

PR #4469, opened 2017-11-17
Add onChange callback to Galleria

PH89:
- Makes it possible to receive callback events when the displayed image has changed.
- Removes whitespaces.


cagataycivici:
Will add manually.

PH89:
Is there a reason why you add this manually? I mean this way its not noticeable that I did these changes!
A better approach would be to inform the contributor about the weak points in PR so he/she can fix it.

However glad that the feature is now available....

Issue #4483, opened 2017-11-20
Error in prime 5.0.0 .. autocomplete -> Cannot read property 'nativeElement' of undefined


huineng:
**I'm submitting a ...**  (check one with "x")
```
[x ] bug report
```
**Current behavior**
```
Cannot read property 'nativeElement' of undefined
```

**Minimal reproduction of the problem with instructions**

this line is producing an error
https://github.com/primefaces/primeng/blob/master/src/app/components/autocomplete/autocomplete.ts#L610

```
return (target === this.dropdownButton.nativeElement || target.parentNode === this.dropdownButton.nativeElement);
```

there's no such thing as a dropdownButton did you mean ?
```
return (target === this.el.nativeElement || target.parentNode === this.el.nativeElement);
```

thanks

sudheerj:
The dropdownButton object is the view child of auto complete container. So it should be defined for dropdown options. I can't replicate the issue. So Can you please provide the plunker to replicate the issue.

huineng:
well it's right here 

![image](https://user-images.githubusercontent.com/3482445/33019949-afa43b26-cdfc-11e7-8c46-280175ed6c37.png)

mdmazhar:
same here.

![image](https://user-images.githubusercontent.com/10234464/33041658-15ef3858-ce0d-11e7-97a7-b2e40b35a174.png)

lukasweber:
I made a little fix for this issue. 

Hope it gets approved as soon as possible.

aamir1995:
Please merge this PR for the upcoming release.

jizza:
+1 (#4514)

cagataycivici:
Fix checked in, we'll do 5.0.1 later today or tomorrow morning.

aamir1995:
Thanks @cagataycivici .

PR #4492, opened 2017-11-21
fix issue #4483

lukasweber:
This pull requests fixes the issue #4483 which caused some problems with the autocomplete component.

My fix just checks if there is a dropdown button existing, because it can't get the 'nativeElement' of the dropdown-button if it doesn't exists.


cagataycivici:
Thanks, I'll also add dropdown prop check.

PR #4503, opened 2017-11-21
Default to no button text when label is not set.

BustyLoli-Chan:
This is a pull request to fix the issue demonstrated by this Plunkr.
http://plnkr.co/edit/dSx8dBkXUQJXWtZHIOvc?p=preview

I've modified this line:

primeng/src/app/components/button/button.ts

Line 34 in d06afd8
 labelElement.appendChild(document.createTextNode(this.label||'ui-btn')); 

To allow for situations in which an end user would prefer to just have a button without a label so that they may edit the actual HTML inside of the button tag.

It is currently known by issue #1657 that you can currently hide this text with the following styling

.ui-button-text-only .ui-button-text{
    display: none !important;
}

but this seems like it should be unnecessary.


sudheerj:
It looks like that the PR includes other changes as well. Can you please recheck it.

BustyLoli-Chan:
The other commit merged in the latest primeNg stuff from Master because I hope it would fix the Travis build. It didn't. If you look at the files changed tab though there is only one file that is different from Master.

cagataycivici:
We still some sort of content otherwise button shrinks, I'll add a css solution instead.

BustyLoli-Chan:
Alright, sounds good. I just hope it fixes it 😱

Issue #1683, opened 2016-12-26
Filtering for Tree


agusdutra:
It would be really nice to have a filtering box that can be added to Trees and filters the TreeNode[] on the component without losing the selection and the original TreeNode[].

The filter could work with the label of the Tree Nodes.
 
In a project I'm using checkable tree nodes and sometimes it gets a bit hard to find some nodes.

I'm talking about some filter like this: 
http://ee.github.io/angular-ui-tree-filter/

Thanks.

ajmair01:
Agreed :point_up:. We are using the tree component for selecting across a large hierarchy. Would be very nice to have filtering support without having to write custom code to support it.

cagataycivici:
@mertsincan did it for PrimeFaces, we can do similar I believe.

olegkon:
This Search/Filter is very important for us.
Can we vote on that issue?  +1

When can we expect it done?

TIA,
Oleg.

PrajwalGonsalves07:
+1

beaugogh:
+1

pepe-escrich:
+1

mkseng:
+1

saikiranmandhala:
+1

seeshadows:
+1

banejo:
+1

kkcode24:
+1

mchauhan:
+1

claudiuconstantin:
Any milestone for this?

ethnoplex:
+1

skormel:
+1

CarlinhosSousaJunior:
+1

j1j:
+1

mediaaminian:
+1

cal2sms:
+1

alaa9jo:
+1

akashdeepg12:
It would be very good to  have global functionality in PrimeNg tree.Atleast user can search any file/folder globally by using global filter box..

Fiszi:
+1

claudiuconstantin:
Any updates on this one?

cagataycivici:
Not in the scope of new TreeTable of 6.0.

chennoam7:
Hi,
We need it too, it will be a great feature.
Thanks, 
Chen

lucky1ravi:
Any update on this feature.

i23591326:
[...]

PR #4504, opened 2017-11-21
#1683 Filtering tree. Update docs.

dybek:
#1683 Filtering added to Tree based on filtering in Datatable

###Defect Fixes

###Feature Requests

### DIFF
--- a/src/app/components/common/filterstateenum.ts
+++ b/src/app/components/common/filterstateenum.ts
@@ -1,0 +1,5 @@
+export enum FilterState {
+  FOUND,
+  ON_FOUND_PATH,
+  NOT_FOUND
+}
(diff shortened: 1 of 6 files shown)


cagataycivici:
The changes are a bit much so I'll review next week thanks.

dybek:
If you need any explanation, please ask. My main idea was to travers all nodes, check label and change they filterState to one of three possible value: 
- FOUND - label of node contains searched value,
- ON_FOUND_PATH - one of descendant have FOUND state
- NOT_FOUND - otherwise.
Based on filterState corresponding css class is set, so we can marked or hide some nodes.

cagataycivici:
We had our own tree filter api derived from PrimeFaces Tree, we'll use that instead. Thank you.

PR #4508, opened 2017-11-22
Changed loading icon default class

LucasAndrad:
### Loading Icon
Hi I was trying to add a custom loading icon to my project using the loading property but I got some many bug's, than I notice that the loading property was concatenating the 'fa fa-spin' class with my custom icon class. 

So I remove this concatenation `"'fa fa-spni' + loading"` and left only the loading variable, because every body should be able to add a custom loading spinner icon, and also, if some one want to use the 'fa fa-spin' class he can add by himself.


sudheerj:
I tried some of the font-awesome icons for custom icons and it works good. You shoudn't remove the prefix classes which creates the animation behavior.

cagataycivici:
Better not remove to avoid breaking apps already, I think you should override css rules of fa-spin instead.

PR #4540, opened 2017-11-27
#3901 - Show total item count on pagination

hcansiz:
Show total item count on pagination when set to "showTotalItemCount" as true . Default value is false.

This is reletad to #3901


cagataycivici:
Contains hard code values, should be customisable to other languages as well.

arjendevries:
I updated to primeng 5.2.5 and set showTotalItemCount to true, but I'm not seeing a total count.  Is there documentation on this?

PR #4543, opened 2017-11-27
Pass previousindex to onchange event of tabview

SwintDC:
Does not fix any issues.

This just adds another member (previous tab index) to the event object passed to the onchange event emitter in the tabview. The parent component can then use this previous index to know which was the previous tab.


cagataycivici:
Misses docs and apparently the object key so unable to accept.

PR #4560, opened 2017-11-28
Fix wording

ite-klass:


ite-klass:
I edited this on GitHub, and the GitHub editor added a newline to the end of the file (which is a common practice and POSIX standard for text files but unrelated to what I intended to do with this PR). If that's an issue for you, please tell me if you will fix this up yourself, or whether I should do so.

PR #4565, opened 2017-11-29
Add explain regrading events after DataTable Edit

eyal7773:
There are a lot of issues, becuase the explenation is not clear.
I added a line that explain it.


cagataycivici:
p-dataTable is deprecated and will be removed in favor of p-table.

Issue #4306, opened 2017-10-30
DataTable Race Condition Crash in Chrome 62 (#4233 did not fix)


starlightknight:
### There is no guarantee in receiving a response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeNG PRO Support* where support is provided within 4 business hours

**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
**Current behavior**

A race condition causes the following issue when column toggling in Chrome 62. This issue was previously reported in #4233, but the referenced commit 3b84cf7 that it was closed with did not actually fix the problem. 

@cagataycivici mentioned being unable to reproduce, so I have attached a minimal test project to reproduce the issue.

Test System:

Linux Mint 18.2
Chrome Version 62.0.3202.62 (Official Build) (64-bit)

```
ERROR TypeError: Cannot read property 'remove' of undefined
    at _DuplicateMap.webpackJsonp.../../../core/@angular/core.es5.js._DuplicateMap.remove (core.es5.js:7490)
    at DefaultIterableDiffer.webpackJsonp.../../../core/@angular/core.es5.js.DefaultIterableDiffer._unlink (core.es5.js:7187)
    at DefaultIterableDiffer.webpackJsonp.../../../core/@angular/core.es5.js.DefaultIterableDiffer._remove (core.es5.js:7178)
    at DefaultIterableDiffer.webpackJsonp.../../../core/@angular/core.es5.js.DefaultIterableDiffer._mismatch (core.es5.js:6964)
    at DefaultIterableDiffer.webpackJsonp.../../../core/@angular/core.es5.js.DefaultIterableDiffer.check (core.es5.js:6859)
    at DefaultIterableDiffer.webpackJsonp.../../../core/@angular/core.es5.js.DefaultIterableDiffer.diff (core.es5.js:6830)
    at NgForOf.webpackJsonp.../../../common/@angular/common.es5.js.NgForOf.ngDoCheck (common.es5.js:1691)
    at checkAndUpdateDirectiveInline (core.es5.js:10837)
    at checkAndUpdateNodeInline (core.es5.js:12330)
    at checkAndUpdateNode (core.es5.js:12269)
[...]

PR #4602, opened 2017-12-03
Make ObjectUtils.equals() work with frozen objects (fix #4600)

### LINKED ISSUE #4600: Problem when using DataTable with data loaded via Apollo GraphQL client ("object is not extensible" error)
```
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/jCwS5OadJK7KHP0wptat

**Current behavior**
When you load data via the Apollo GraphQL client and use it for a selectable DataTable that has no dataKey set, you get an error "Cannot add property _$visited, object is not extensible" when selecting rows.

**Expected behavior**
There should be no error, like with data loaded from other sources or when setting a dataKey.

see Plunkr case above

I want to display data loaded with the Apollo Client for GraphQL directly in a PrimeNG DataTable.

* **Angular version:** 5.0.5

* **PrimeNG version:** 5.0.2

* **Browser:** Chrome, Edge, Firefox
 
* **Language:** TypeScript 2,4

**Analysis of the Problem**:

The problem is caused by Apollo Client returning ["deep-frozen"](https://github.com/apollographql/apollo-client/blob/master/packages/apollo-utilities/src/util/maybeDeepFreeze.ts) query results in test and development. Particularly, no new properties can be added to these objects.

However, when there is no dataKey set on a DataTable, PrimeNG tries to compare rows using its `objectUtils.equals()` method. This method tries to set a `_$visited` property on the examined objects, which does not work when the objects are frozen.

Cito:
In order to avoid infinite recursion, the method added a "visited" flag as property to the compared objects. This caused the method to raise an error when "frozen" (read only) objects were compared that cannot be extended. This fix uses a stack to memorize visited objects instead of setting a property on the objects, and therefore allows frozen objects to be compared as well.

In real life, this can happen when feeding a PrimeNG data table with data queried using the Apollo GraphQL client (see issue #4600).

This fix also adds some unit tests to confirm proper behavior of ObjectUtils.equals().

### DIFF
--- a/src/app/components/utils/objectutils.spec.ts
+++ b/src/app/components/utils/objectutils.spec.ts
@@ -2,17 +2,17 @@ import { ObjectUtils } from './objectutils';
 import { inject, TestBed } from '@angular/core/testing';
 
 describe('ObjectUtils Suite', () => {
-    
+
     let objectUtils: ObjectUtils = null;
-    
+
     beforeEach(() => TestBed.configureTestingModule({
         providers: [ ObjectUtils ]
     }));
-    
+
     beforeEach(inject([ObjectUtils], s => {
         objectUtils = s;
     }));
-    
+
     let data: any = [
         {brand: "VW", year: 2012, color: {name:"Orange"}, vin: "dsad231ff"},
         {brand: "Audi", year: 2011, color: {name:"Black"}, vin: "gwregre345"},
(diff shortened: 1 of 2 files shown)


starlightknight:
This should also fix the root cause of #4306 according to the diagnosis from the Chrome developers in the upstream issue

PR #4613, opened 2017-12-06
Listen mouseup not only on header

argnist:
This fixes the same problem as #2550

You should listen mouseup on full browser, not only on header of the dialog to stop dragging. Sometimes mouseup invokes on different places if you move fast.

### DIFF
--- a/src/app/components/dialog/dialog.ts
+++ b/src/app/components/dialog/dialog.ts
@@ -149,7 +149,9 @@ export class Dialog implements AfterViewInit,AfterViewChecked,OnDestroy {
     
     id: string = `ui-dialog-${idx++}`;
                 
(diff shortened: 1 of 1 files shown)


argnist:
Forgot remove listener...
So you need to create 
`this.listener = renderer.listen("body", "mouseup", event => this.endDrag(event));`
and remove it in ngOnDestroy:
`this.listener();`

argnist:
My investigation led me to the conclusion that the problem appears because of selectable text in dialog header in IE. You just need to add 
`.ui-dialog-titlebar {-ms-user-select: none; user-select:none}` 
to dialog styles and it will fix problem at all without listening mouseup. IE doesn't fire mouseup on selected text.

cagataycivici:
This will trigger change detection always whenever a user does mouseup, so very bad for performance.

PR #4638, opened 2017-12-08
Dropdown: Add `filterMode` property

nlcano:
Add the ability to set the filter mode for the searchbox.   Default filtering mode is "contains" and alternative is "startsWith".

Example: <p-dropdown filterMode="startsWith" ...>


nlcano:
Replaced with a new one.

PR #4654, opened 2017-12-12
Fix/menu item style class

### LINKED ISSUE #1097: style and styleClass attributes for TieredMenu MenuItem
i would like to pull-right a menu item but the class MenuItem does not have an attribute to extend the style.

amiram:
Adds styleClass and style to TieredMenu. Partially fixes #1097


cagataycivici:
Merged manually, thanks.

Issue #681, opened 2016-07-29
DataTable: sorting null not properly working


mleonhartsberger:
when sorting a column which contains some rows with "value is null", the sort algorythm just stops.

it could be solved by adding null checks at
https://github.com/primefaces/primeng/blob/master/components/datatable/datatable.ts#L575

![image](https://cloud.githubusercontent.com/assets/20722294/17245211/c0b2dc88-5585-11e6-97ae-8f9116098df0.png)

edangerfield:
I experience the same.

hzach:
+1

ctrl-brk:
Same issue still exists int multisort mode

afras21:
Yeah the actual problem is null property is string and others are numbers .... change them to string it ll work fine

PR #4667, opened 2017-12-14
Update datatable.ts

antkers:
Fix for multiple fields sort when a field's value is null.
It was already fixed for single field sorting, see issue #681


cagataycivici:
p-dataTable is deprecated and will be removed in favor of p-table.

Issue #3906, opened 2017-09-12
Refactor DataScroller for Performance and Fix Lazy Loading


burner-account:
Request: There is no guarantee in receiving a response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeNG PRO Support* where support is provided within 4 business hours

Response: I am currently test driving PrimeNG. I am not yet convinced, that we should buy support for this product. I can't even tell if your request is legit or fishy. Is it a hint, a warning or a threat?

**I'm submitting a ...**  (check one with "x")
```
[ x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
If the datascroller fires a lazy load event, it assumes an instantaneous, successful lazy load happens, thus immediately incrementing  "this.first = this.first + this.rows;". 
Since there is no way to unbind the lazy load event while lazy loading, a normal scroll finger movement will fire huge amounts of lazy load events... while the lazy load event you actually care about is not yet finished.

**Expected behavior**
Maybe allow deactiving the lazy load event while lazy loading. Seriously. Is there any use case in which in the current behaviour is required or useful? What do i miss?

alexander37137:
👍

Bohdan-Tereta:
Fixed this issue, see my pull request #4676

cagataycivici:
We'll implement virtual scroll for lazy mode.

mroskamp:
I'm still running into this issue in v5.2.4. Here's a plnkr where I introduce a 250 ms timeout before loading the next page of results, which illustrates the issue: http://plnkr.co/edit/5EyojPuaoLg9ox5olk4z?p=preview

PR #4676, opened 2017-12-16
Fix for "Datascroller lazy load implementation broken #3906"

Bohdan-Tereta:
Added two properties to datascroller component that allow to correctly wait for external data load callbacks ( see issue #3906): 

1. isDataLoaded - when set to true, allows receiving next portion of data, when set to false outside component, means component is waiting for data to load, should be set back to true after data finished loading (for example, inside subscribe event of data loading method)

2. isNoMoredata - set outside component to true means all data was loaded, and we reached the end of data, so no need to fire load()


cagataycivici:
Will fix it differently, thank you.

Issue #4682, opened 2017-12-18
Tooltip is aligned wrong when tooltipStyleClass changes the tooltip size


Ldoppea:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report
[ ] feature request
[ ] support request
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/SgOQW2iiIFzEeTrgBRnr?p=preview

**Current behavior**
In the plnkr, the input's tooltip is not aligned with the input. This is due to custom CSS in tooltipStyleClass applying a different font size (also works with custom paddings or everything that affects tooltip size)

**Expected behavior**
The tooltip should be aligned with the input field.

**Minimal reproduction of the problem with instructions**
Apply pTooltip to an input. 
Set the tooltipStyleClass with a custom class. 
Add the corresponding CSS class and set a big font size.

**What is the motivation / use case for changing the behavior?**
Stylizing the tooltip

**Please tell us about your environment:**

* **PrimeNG version:** 
5.0.2

* **Browser:**
all
 
* **Language:**
all

KerollosMelad:
Still not fixed !!!

tamlogic:
Hi @cagataycivici , I see it not fixed now.
Please help me check it again.
I don't change the style of tooltip (background color, width and align). Thanks.

PR #4683, opened 2017-12-18
Fix/Tooltip Apply container.className before align()

Ldoppea:
In pTooltip: tooltipStyleClass is now applied before calculating tooltip alignment.

Fix for https://github.com/primefaces/primeng/issues/4682


cagataycivici:
Implemented differently thanks.

PR #4728, opened 2017-12-27
Css fix for some Android.

kojisaiki:
## description
Optimize brower's painting process.
For instance, my case is Android 4.4.2 (Device published by SHARP) .

## detail
Use `text-ident: -supersize px`(and complex some animation, like fade-in, fade-out...) some browser paint all position, solemnly. It makes more slow performance.


kojisaiki:
Cause of commit and difference is complexed, I recreate the PR later.

PR #4742, opened 2017-12-29
Add dropdown appendTo to paginator and data components

BustyLoli-Chan:
### Defect Fixes
I've noticed that when adding tables to modals, the paginator seems to calculate the direction it should drop in (down or up) based purely on screen height. This can sometimes result in a paginator dropdown which still drops down and makes your modal scroll.  
I have attached a repo plunkr of the issue here: http://plnkr.co/edit/gvesU2PhChbN321tMeER?p=preview

I have attempted to fix this by adding appendTo support to the paginator dropdown. I believe if you append the paginator dropdown directly to the body you can at least have it float over the bottom of the modal instead of being stuck inside of it to scroll. This was my personal take on a solution to this issue. If there are other/better ways to fix this, I am open to all suggestions.

### DIFF
--- a/src/app/components/datatable/datatable.ts
+++ b/src/app/components/datatable/datatable.ts
@@ -478,7 +478,7 @@ export class ScrollableView implements AfterViewInit,AfterViewChecked,OnDestroy 
             
             <p-paginator [rows]="rows" [first]="first" [totalRecords]="totalRecords" [pageLinkSize]="pageLinks" styleClass="ui-paginator-bottom" [alwaysShow]="alwaysShowPaginator"
                 (onPageChange)="onPageChange($event)" [rowsPerPageOptions]="rowsPerPageOptions" *ngIf="paginator && (paginatorPosition === 'bottom' || paginatorPosition =='both')"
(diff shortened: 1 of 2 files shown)


akorchev:
The same occurs if the component is inside an HTML element with `overflow: auto` - the dropdown is clipped. Please either set appendTo="body" or expose it in the paginator and DataTable components.

akorchev:
FWIW here is an ugly workaround which mitigates the problem to some extent:

```
.ui-paginator .ui-dropdown-panel {
  position: fixed;
  left: auto !important;
  top: auto !important;
  min-width: auto !important;
  display: block !important;
}
```

Of course this makes the dropdown fixed - if the user scrolls the page the dropdown will remain fixed.

BustyLoli-Chan:
We've worked around this by setting `top: unset;` and setting `bottom: 32px` (the height of our input) but yes it would be super handy to have the appendTo exposed to hopefully mitigate this issue.

cagataycivici:
The PR missed documentation so added manually with some changes.

PR #4780, opened 2018-01-09
Dropdown: Add `filterMode` property

nlcano:
Add the ability to set the filter mode for the searchbox. Default filtering mode is "contains" and alternative is "startsWith".

Example: <p-dropdown filterMode="startsWith" ...>


nlcano:
@cagataycivici: Can you please review and merge this PR?   

_FYI, The original one (#4638) was sent on Dec 8, 2017 but it wasn't merged. It was replaced with this new one which includes  the latest code from the master branch._

cagataycivici:
These filters should better be used, I'd suggest moving them to objectUtils so other form components like Listbox, multi select can use them.

PR #4790, opened 2018-01-11
allow configuration of csv linebreak character

olivermue:
###Feature Requests

This is a quite small feature. It simply adds the possibility to configure the used linebreak character when exporting to csv. The default is still the existing Linux style (LF only), but could be changed to windows style (CR/LF), Mac OS style (CR only) or something exotic (pipe sign '|') if desired.

### DIFF
--- a/src/app/components/datatable/datatable.ts
+++ b/src/app/components/datatable/datatable.ts
@@ -573,6 +573,8 @@ export class DataTable implements AfterViewChecked,AfterViewInit,AfterContentIni
     @Input() contextMenu: any;
     
     @Input() csvSeparator: string = ',';
+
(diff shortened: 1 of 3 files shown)


cagataycivici:
p-dataTable is deprecated in favour new p-table so we're not doing any changes on p-dataTable. Would be glad if you port this to p-table.

olivermue:
The commit already contains the changes within the p-table also. So you could simply cherry pick the changes you like. But I'm surprised that p-dataTable is deprecated. I couldn't find any hint at the [documentation](https://www.primefaces.org/primeng/#/datatable) nor there are any examples based on p-Table. They're all based on p-dataTable.

PR #4877, opened 2018-01-19
Add table sort compareFunction feature

treythomas123:
This allows a compareFunction property to be provided for a table column (working the same way as with Array.prototype.sort), to allow customization of the sort ordering. Implemented for both single-column and multi-column sorting.

Also added this to both the old datatable and the new turbo table.


cagataycivici:
p-dataTable is deprecated in favor p-table so we're unable to accept this request. p-table will get a custom sort function in 5.2.RC2 though this week.

Issue #4915, opened 2018-01-22
TurboTable refresh data resets current page


pantonis:
### There is no guarantee in receiving a response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeNG PRO Support* where support is provided within 4 business hours

**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://plnkr.co/edit/3fCevmLlcgpYntK2Fo98?p=preview

**Current behavior**
When you have sorting enabled for a column and you are not in first page but in any other page (2,3,4 etc...) and data are refreshed, paginator immediately jumps in first page. 
The same thing happens when you have something typed in global filter current page is not preserved

**Expected behavior**
Current page must be preserved

**Minimal reproduction of the problem with instructions**
Please run the plnkr above

**What is the motivation / use case for changing the behavior?**
I have seen a lot of datatables and they all preserve current page when data are refreshed

cagataycivici:
Resetting page is by design, these should be options at least, not all user may like it.

pantonis:
@cagataycivici Can you add the optional parameter for resetting the paging and sorting on data update?

pantonis:
@cagataycivici any comment?

pantonis:
I think I know why he wont fix it. This fix exploits more bugs to the turbotable. When you are in fifth page and filter the data are not shown when this fix is applied.

PR #4916, opened 2018-01-22
Fixed bug #4915

pantonis:


cagataycivici:
Resetting page is by design, these should be options at least, not all user may like it.

pantonis:
@cagataycivici I have vast experience with business intelligence and tabular data for years.  I have never seen this behavior before.
In any case If you wish to keep this behavior It would be good to have it optional as you said. I will be waiting for this change. Thanks

jarinwhite:
Seconding @pantonis, the leading datagrid (ag-Grid) does not reset the page when sorting. See their [Pagination example](https://www.ag-grid.com/javascript-grid-pagination/). The current behavior of TurboTable is simply counter-intuitive to almost every enterprise-grade datagrid I've used. 

Would be good to at least make this an option.

pantonis:
Yes. For me every time I npm install have to go and manually change it. Sometimes it does not AOT build because of my changes.

Hope @cagataycivici changes this.

PR #4918, opened 2018-01-23
Make sticky property work immediately in Growl when set to true

ddvkid:
I just came across a requirement that the Growl panel needs to be sticky if the message is not successful.
With current version, if you set sticky property on the fly, it won't take effect immediately because the timer is already running.
I think this could be a common one and this pr will fix it.


cagataycivici:
@Input() get sticky(): boolean {
        return this.sticky;
    }

This calls for a stack overflow error by the way.

ddvkid:
Opps, missed a _😓

PR #4953, opened 2018-01-24
Fixed angular formGroup example

kroeder:
.get['cities'] should be .get('cities') according to https://angular.io/guide/reactive-forms#inspect-formcontrol-properties
I've tested this and can aprove this.


cagataycivici:
We've settled on myFormGroup.controls['cities']

PR #4965, opened 2018-01-25
Fix virtualScrollCallback not triggering for ui virtualization

### LINKED ISSUE #4964: dataTable virtualisation without lazy does not update scrollTable.top.y
```
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/m7x0OzvikVjrifzT8lGS?p=info

**Current behavior**
If you scroll down a dataTable with virtualisation but not with lazy, the content gets virtualised correctly but it's also out of view, becaue the scrollTable.top.y never gets changed, due to virtualScrollCallback never being called.

**Expected behavior**
Virtual content scrolls down with the user

Scroll down just enough so you can see the last number in the plnkr change, indicating that the content has indeed been changed. However, you continue scrolling into total white.

Mysame:
virtualScrollCallback is used to set scrollTable.style.top.
virtualScrollCallback currently only gets called on data change, which only happens on lazy load.
It should also trigger when rendering data has been updated.

This fix just showcases the problem and is probably not production-ready, however it does fix the issue.

Other things that don't work properly, but that are not included in this PR:
- Filtering does not reduce the totalRecords, allowing you to scroll as if there is no filtering
- It seems to randomly glitch out and keep showing the same records over and over

###Defect Fixes
#4964 dataTable virtualisation without lazy does not update scrollTable.top.y

### DIFF
--- a/src/app/components/datatable/datatable.ts
+++ b/src/app/components/datatable/datatable.ts
@@ -1155,7 +1155,13 @@ export class DataTable implements AfterViewChecked,AfterViewInit,AfterContentIni
                 if(this.lazy)
                     this.onLazyLoad.emit(this.createLazyLoadMetadata());
                 else
+                {
                     this.updateDataToRender(this.filteredValue||this.value);
+                  
+                    if(this.virtualScrollCallback){
+                      virtualScrollCallback();
+                    }
+                }
             }, this.virtualScrollDelay);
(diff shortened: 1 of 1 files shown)


cagataycivici:
DataTable is not maintained anymore in favor of p-table.

PR #4971, opened 2018-01-25
Feature add input property first to datalist component

RogerioTeixeira:
###Feature Add input property [first] to datalist component.


cagataycivici:
DataList is deprecated in favor of DataView

PR #4996, opened 2018-01-26
Feature #4995 - Double click expand/collapse option in a Tree

### LINKED ISSUE #4995: Double click expand/collapse option in a Tree
**I'm submitting a ...** 
```
```
**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/oMpfZQrOOw0myyCnytDw?p=preview

**Current behavior**
For the moment it is only possible to have some kind of double click tree expand/collapse when attaching it to the node template. This obligates users to double click on the text; they cannot double click expand by clicking the folder icon or other locations that do trigger a select event

**Expected behavior**
Double clicking a tree node expands/collapses it. Combined with checkbox selection mode, this feature will need a slightly different approach (e.g. only toggle checkboxes when the user clicks on the checkbox itself)

Having a built-in optional user-friendly behavior for expanding/collapsing the nodes of a tree. Pretty much every file explorer has this functionality.

* **PrimeNG version:** 5.0.2
* **Browser:** all

rmschots:
https://github.com/primefaces/primeng/issues/4995

Adds an option to Tree to expand/collapse on double clicking a node.
When the expandOnDoubleClick option is enabled together with checkbox selection mode, nodes will have to be checked by clicking on the checkboxes instead of anywhere on the line.
Also disabled text highlighting in the tree by default (can still be overwritten by implementing custom node content by usage of a template.

### DIFF
--- a/src/app/components/tree/tree.ts
+++ b/src/app/components/tree/tree.ts
@@ -27,7 +27,7 @@ import {BlockableUI} from '../common/blockableui';
                     ><span [class]="getIcon()" *ngIf="node.icon||node.expandedIcon||node.collapsedIcon"></span
                     ><span class="ui-treenode-label ui-corner-all" 
                         [ngClass]="{'ui-state-highlight':isSelected()}">
-                            <span *ngIf="!tree.getTemplateForNode(node)">{{node.label}}</span>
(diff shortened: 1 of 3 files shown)


cagataycivici:
I don't understand various design decisions in this PR so closing like blocking text selection and changes to onNodeClick vs, we'll implement it ourselves in a future release.

PR #5000, opened 2018-01-27
fix: make the pReorderableColumn compatible with row expansion

### LINKED ISSUE #4999: ColReorder support for Row Expansion in TurboTable
### There is no guarantee in receiving an immediate response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeNG PRO Support* where support is provided within 4 business hours

```
```

**Plunkr Case (Bug Reports)**
Please check my own primeng fork at: https://github.com/loicsalou/primeng.git branch "bugfix/tablerowexpansion-colreorder-issue" and launch showcase then check turbo table's expansion row demo to reproduce the issue.

**Current behavior**
when drag'n'dropping a column the resulting table is not correct: column is not at the correct place. Moving the last column, moving color before brand, produces an error (ERROR TypeError: Cannot read property 'header' of undefined).

**Expected behavior**
no error expected + dragged column should be placed at the drop position

drag a column to a new position and check final order of columns, it's wrong

Problem analysis: easy to understand actually. The method "Table.onColumnDrop(event, dropColumn)" reorders the declared columns (for instance Vin, Year, Brand, Color) but the domHandler provides dragIndex and dropIndex from the dom and as I have setup a column for the toggle of expansion row I have 5 columns, not 4. So the columns involved in the reorder are not correct and if I move the last column (Color) I get the index 5 which does not exist in the declared columns).

I'm trying to find a solution which I can propose.

reorder is an important feature for my users

MacOs or windows 7, Chrome (latest, Angular 5.2, prime 5.2.0-rc.3, node 8.4, npm 5.6.0)

* **Angular version:** 5.2.0
the issue is definitely coming from Primeng

* **PrimeNG version:** 5.2.0-rc.3

* **Browser:** Chrome latest, Safari
 
Typescript

loicsalou:
closes #4999

###Defect Fixes
using pReorderableColumn in a table which uses Row Expansion feature does not work because of the row toggler (the extra first column in the table) is seen by the API as a regular column sot he dragIndex and dropIndex calculated by domHandler.index(...) function are wrong.

I enclose a change in the demo for table, row expansion feature, in order to demonstrate the solution. without the fix it does not work, with it, it works. See #4999 for more information


loicsalou:
I merged all changes into my branch, probably I should have rebased...
Anyway, changes are small, they involve only DomHandler and table.

cagataycivici:
Thanks, added after refactor.

PR #5005, opened 2018-01-29
Add export for new table component

lukasweber:
I just added an export for the new table component because it was missing yet.


cagataycivici:
primeng/primeng is deprecated so we're not adding the new TurboTable to it to promote 'primeng/table';

Issue #5042, opened 2018-02-01
Atribute visible from MenuItem is not work in SplitButton


netstart:
```
[ x ] bug report

```

**What is the motivation / use case for changing the behavior?**
Atribute visible from MenuItem is not work in SplitButton

**Please tell us about your environment:**
Windows 10
IDE: Webstorm

* **Angular version:** 5.2.1
* **PrimeNG version:** 5.2.0

* **Browser:**
Versão 63.0.3239.132 (Versão oficial) 64 bits
 
* **Language:** 
    "typescript": "~2.6.1"
    "target": "es5",

* **Node (for AoT issues):** 
v9.2.1

skormel:
I updated primeng from 5.0.2 to 5.2.0 and visible property from menuitem of all the contextmenu stop working (always it's shown). I revert to 5.0.2 and start working again.

PR #5044, opened 2018-02-01
fix to 5052

netstart:
###Defect Fixes to #5042


cagataycivici:
Fixed with a different syntax, thanks.

Issue #2601, opened 2017-04-25
New Component: Ellipsis


ctrl-brk:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
I'm looking for a general way to show some kind of tooltip (on mouse hover or icon click or something...) if the data overflows the cell.
I'm using text-overflow:ellipsis for my cells in order to keep rows height consistent and want to be able to show the full content somehow.

Thanks.

bmstambaugh:
I would like to same. While resizable columns are an option sometimes its just easier for the user to hover over the cell without affecting the whole table.

bmstambaugh:
My workaround for this is to use the Tooltip component and a template cell.

`<p-column field="Col1" header="Col1" [sortable]="true" [style]="{'width':'200px'}">
        <ng-template let-col let-data="rowData" pTemplate="body">
            <span [pTooltip]="data[col.field]" tooltipPosition="top">{{ data[col.field] }}</span>
        </ng-template>
    </p-column>` 

and override the class
`.ui-datatable .ui-datatable-thead>tr>th, .ui-datatable .ui-datatable-tfoot>tr>td, .ui-datatable .ui-datatable-data>tr>td  {
    text-overflow: ellipsis; 
    overflow : hidden; 
    white-space: nowrap;
}
`

ctrl-brk:
Yes, but it's annoying to put everywhere

bmstambaugh:
I completely agree. Having it built in would save a lot of HTML markup

kenisteward:
I have built a component that does it exact thing.  it limits itself to 1 line and if that overflows it automatically uses a bootstrap 4 or angular material tooltip to tell you what it is.  I'll have to plop it on github at some point

ctrl-brk:
Would be nice to see. Can you make it work with primeng tooltip?
But it's still a component you have to add at every datatable column. Still too bulky.

kenisteward:
Yeah I can have it work with primeng . Also it is an extremely lightweight
component that does everything contained.

dhniels:
@kenisteward can we see it?

kenisteward:
Oh shoot! I went on vacation soon after this and never did update . Now
[...]

PR #5064, opened 2018-02-04
Ellipsis Feature p-table - Fix for #2601

malhar12:
###Feature Requests
This PR is being created with respect to the Issue #2601. Demo for the Feature can be found here:

https://malhar12.github.io/#/table/ellipsis

The Ellipsis feature is used as an attribute directive [pTextEllipsis] on the columns. When the directive attribute applied, it will find overflows in that column and apply ellipsis. The directive works well, with Basic p-table, as well as Scrollable [Fit & Expand] and Non-Scrollable [Fit & Expand] mode, as it can be seen in the demo.

### DIFF
--- a/src/app/components/table/table.css
+++ b/src/app/components/table/table.css
@@ -65,7 +65,7 @@
 
 .ui-table-scrollable-body > table > .ui-table-tbody > tr:first-child > td {
     border-top: 0 none;
-} 
+}
 
 .ui-table-virtual-table {
     position: absolute;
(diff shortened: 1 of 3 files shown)


cagataycivici:
Sorry too many changes for us to consume and accept, it is one of those features that should be implemented by our team as stated on the pull request description.

mohitlakhera:
@cagataycivici anything added for this in new version of primeNg

r1x3l:
We recently stumpled upon the same requirement. Anyone considering to reopen the pull request? The solution from @malhar12 seems to be straightforward and useful.

PR #5089, opened 2018-02-06
New calendar milliseconds

GermanoGuerrini:
###Feature Requests
Adds milliseconds to the calendar widget.
It works just like showSeconds (which is required), that is:
- add `[showMilliseconds]="true"`
- configure an optional `stepMillisecond` to handle incrementing/decrementing


cagataycivici:
Sorry but as mentioned on PR policy, new feature requests should be handled by our team. We accept PRs for defect fixes mostly.

PR #5108, opened 2018-02-08
Dropdown do not close on enter if no match

H--o-l:
Proposal:
When using the filter option, if all options are filtered, and so
no match found, I believe that keyboard enter key shouldn't close the
dropdown.
It give the wrong feeling that a new value had been selected where it's
not the case.

This is a solution to not close the dropdown in such a case.

### DIFF
--- a/src/app/components/dropdown/dropdown.ts
+++ b/src/app/components/dropdown/dropdown.ts
@@ -581,9 +581,11 @@ export class Dropdown implements OnInit,AfterViewInit,AfterContentInit,AfterView
             
             //enter
             case 13:
-                this.hide();
-
-                event.preventDefault();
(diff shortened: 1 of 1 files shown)


H--o-l:
Hi !
What do you think about this PR ?

adrienverge:
I agree.

In my opinion pressing "enter" at that point shouldn't close the dropdown:
image1

To go even further, I think "enter" should close the dropdown **only if** an option is selected, for example:
image2
-> on the left, "enter" should select option and close the dropdown
-> on the right, "enter" shouldn't do anything

H--o-l:
Updated following https://github.com/primefaces/primeng/pull/5108#issuecomment-367956238

adrienverge:
@H--o-l Very nice!!

@cagataycivici do you think it could be included in the next release?

cagataycivici:
Added differently, thanks.

Issue #5111, opened 2018-02-09
metaKeySelection option for TurboTable


borisilic:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**What is the motivation / use case for changing the behavior?**
We have a setup where the fields of a row are displayed below the table itself and can be edited as required. We don't want to provide the user with the option to unselect any particular row. If they click on an already selected row, it should just remain selected as before. 

* **PrimeNG version:** 2.0.X
5.2

PR #5112, opened 2018-02-09
Suggested code for #5111

borisilic:
###Feature Requests
Adds the option to disable row deselection in TurboTable while in _selectionMode="single"_.


cagataycivici:
The upcoming metaKeySelection option can cover this as it does in p-dataTable.

PR #5128, opened 2018-02-12
Update p-dataTable reference in docs to p-table

hughanderson4:
Let's update this outdated `p-dataTable` reference to `p-table`, to keep the docs consistent


cagataycivici:
Rest of the code is wrong actually.

Issue #5080, opened 2018-02-06
p-dropdown [group]=true not set default value


leonetosoft:
[x ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35

I am performing a simple record of occurrence,
When I use the update option I load the form values from the database, like this: (ignore the rest of the missing code)
`this.ocorrenciaService.Load(id).subscribe(rs => {

      const procedimentos = [];
      rs.data.procedimentos.forEach(pro => {
        procedimentos.push(pro.id);
      });

      this.form.patchValue({
        id: rs.data.id,
        nome_paciente: rs.data.nome_paciente,
        quarto: rs.data.quarto,
        data_ocorrencia: moment(rs.data.data_ocorrencia).toDate(),
        medico_id: rs.data.medico_id,
        duracao: rs.data.duracao,
        entrada: moment(rs.data.entrada, 'HH:mm:ss').toDate(),
        saida: moment(rs.data.saida, 'HH:mm:ss').toDate(),
        obs: rs.data.obs,
        procedimentos: procedimentos,
        local: rs.data.status === OCORRENCIA_STATUS.ESPERA ?
          0 : rs.data.status === OCORRENCIA_STATUS.QUARTO ?
            -1 : rs.data.status === OCORRENCIA_STATUS.CANCELADA ? -2 : rs.data.sala_id
      });

      this.carregando = false;
    }, err => {
      this.carregando = false;
    });`

Note that I have a variable called "loading" to indicate that the form has been fully loaded, and making the elements visible on the screen.
my formControll is initialized this way: (ignore the rest of the missing code)
`this.form = fb.group({
      'id': [null],
      'nome_paciente': [null, Validators.required],
      'quarto': [null],
      'data_ocorrencia': [null, Validators.required],
      'medico_id': [null],
      'procedimentos': [null],
      'duracao': [null],
      'entrada': [null],
      'saida': [null],
      'obs': [null],
      'local': [null, Validators.required]
    });`

Here is the layout of my dropdown: (ignore the rest of the missing code)

`<div class="ui-g-12 ui-md-4 ui-lg-4">
        {{form.controls['local'].value}}
        <span>Local (<span style="color: red">*</span>):</span>
[...]

PR #5171, opened 2018-02-19
Fix 5080

jakob-bebop:
Trying to fix behaviour of Dropdown when used with [group]=true

- As described in #5080 , the grouped dropdown was not able to update it gui after the value had been set programatically.
- when the options were being set asynchronously (i.e. [options]="groupedOptions$ | async" ), a previously selected option - even though still present in the new list of options would - disappear, and the placeholder would appear instead.

re https://github.com/primefaces/primeng/issues/5080

###Defect Fixes

###Feature Requests


cagataycivici:
Fixed differently, thank.

PR #5205, opened 2018-02-23
Make configurable Messages's icon.

kojisaiki:
# description
Adding `icons` input interface to `Messages` component, for overwrite serverity icons;

# why
In my case, serverity `error` icon looks like the close button.
So I want to custom message icon like following:
image

### DIFF
--- a/src/app/components/messages/messages.ts
+++ b/src/app/components/messages/messages.ts
@@ -36,6 +36,18 @@ export class Messages implements OnDestroy {
     
     @Input() styleClass: string;
 
+    @Input() icons: {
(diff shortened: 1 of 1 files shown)


cagataycivici:
Better to have separate attributes like infoIcon, messageIcon, by the way you can just override then with css :before.

kojisaiki:
My case was solved use following selector, thank you.
```
.ui-messages-error .ui-messages-icon.fa-close:before
```

EkkoLin:
Can you please provide an example of how to override the icon?

PR #5216, opened 2018-02-26
 Fixed splitbutton display in IE 11, Firefox 57 and Edge

jgosar:
Setting overflow: hidden in the previous change caused splitbuttons to display incorrectly in all browsers except Google Chrome.

The bug can be seen on https://www.primefaces.org/primeng/#/splitbutton


cagataycivici:
Dangerous change to change all buttons as it has no splitbutton related part, the fix is already done earlier this week.

Issue #5253, opened 2018-03-02
slideMenu ignores styleClass on MenuItem model properties


blackholegalaxy:
[Copy from forum]

**I'm submitting a ...**  (check one with "x")
```
[x] bug report 
[ ] feature request
[ ] support request
```

SlideMenu seems to ignore some MenuItem model properties. For example, defining an item `styleClass` is totally ignored.

Repro here:
https://stackblitz.com/edit/prime-slidemenu-bug-r8gjno?file=app%2Fapp.component.ts

As you can see if I add `styleClass: 'any-custom-class'` on the language menu item element, it's completely ignored and is not applied anywhere on the menu.

![Image](https://i.imgur.com/6y3MbSd.png)

I found the same applies for badge or badge style class. The others seem to work properly.

blackholegalaxy:
@cagataycivici  thanks for fix. Sorry about missplacement of these on my PR.

Please note the `badge` and `badgeStyleClass` properties of MenuItem is also ignored here and maybe should be added in all menus using MenuItem model.

cagataycivici:
badges are used in menus premium templates at the moment as far as I know, we sure can add support to p-menu* as well in the future.

diamond-darrell:
@cagataycivici please do 🙂
> we sure can add support to p-menu* as well in the future.

PR #5254, opened 2018-03-02
add MenuItem styleClass and style properties to slidemenu

blackholegalaxy:
### Defect Fixes
Closes: #5253

Add styleClass and style properties from MenuItem on slideMenu component.

@cagataycivici @Merve7


cagataycivici:
Should apply to li

PR #5276, opened 2018-03-07
Document table selectionChange output event

miniplus:
#5277


miniplus:
Suddenly realized this was for the 2-way binding

PR #5305, opened 2018-03-09
TurboTable with virtualScroll not display records when totalRecords is set after ngOnInit

lukgit17:
###Defect Fixes
In all my projects, where are datatables with virtualscroll,  totalRecords it's initialized dynamically when records and total of records are received from server request. Without this feature I cannot migrate to TurboTable. In present turbotable run only if totalRecords are initialized static or in ngOnInit.

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -1805,6 +1805,7 @@ export class ScrollableView implements AfterViewInit,OnDestroy {
         }
 
         if(this.dt.virtualScroll) {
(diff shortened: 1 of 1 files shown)


cagataycivici:
Not the best location to do it, as it will run at every scroll, causing perf issues.

lukgit17:
And how will resolve this problem. I'm buyer of Ultima Theme, and maybe another themes in future, if this problem is corrected. I think in virtual scroll it's normal to get totalRecords, together with data records, in real time when it's many clients run in same time.

PR #5317, opened 2018-03-11
RTL Support for Scrolling Table

gilhanan:
Table header nested under `direction: rtl` ancestor behave currently unexpected:
2018-03-14 10_54_16-primeng

After fix:
2018-03-14 10_53_36-primeng


cagataycivici:
Breaks horizontal scrolling and has issues with frozen columns so reverted.

PR #5319, opened 2018-03-12
TurboTable with virtualScroll not display records when totalRecords is set after ngOnInit - patch 2 with Observable

lukgit17:
Below descripted issue are corrected now using observables. I hope this changes are ok. Performance are very good with these changes and are not be affected before i maked.
###Defect Fixes
In all my projects, where are datatables with virtualscroll, totalRecords it's initialized dynamically when records and total of records are received from server request. Without this feature I cannot migrate to TurboTable. In present turbotable run only if totalRecords are initialized static or in ngOnInit.

###Defect Fixes

###Feature Requests


cagataycivici:
Merged manually

Issue #5285, opened 2018-03-07
Add showClear property to Dropdown


brian428:
The new clear icon for Dropdown is a nice feature, but I don't think it should just automatically be applied. We have plenty of dropdowns where a value **must** be chosen, so the new clear icon is confusing and also breaks our logic (before the clear icon, there was no way for the user to select "nothing").

I would say make the clear functionality opt-in via an input property. Or, less desirable would be an input property to turn it off (but that would mean we'd have to manually go through and disable it throughout our code).

ASK83:
I came across the similar issue (#5286) and making it optional would solve that one too.

hughanderson4:
Agree with @brian428 that this should be opt-in.

In addition, please consider adding a css class to the parent ui-dropdown, whenever the child contains the clear icon (suggest `ui-dropdown-clearable`).  That way my app's css can adjust, or even hide, the presentation of the clear icon based on the parent class, instead of having to target all dropdowns.

PR #5328, opened 2018-03-13
Update paginator.ts

RLevenets:
Do not show clear button on "rowsPerPageItems" drop-down since it doesn't make any sense

###Defect Fixes
Fixed an issue when clear button is displayed on "rowsPerPageItems" drop-down in Paginator component after upgrading PrimeNg to 5.2.1


bossqone:
it should be fixed in the newest version: https://github.com/primefaces/primeng/issues/5285 (showClear should be false by default)

cagataycivici:
Fixed already yes.

Issue #5322, opened 2018-03-12
Bug in p-chart.ts template


pierfreeman:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
Looking at chart.ts source code (https://github.com/primefaces/primeng/blob/master/src/app/components/chart/chart.ts) I noticed that the html template of the component has an error in the input "attr" passed to the canvas object:

```
<div style="position:relative" [style.width]="responsive ? null : width" [style.height]="responsive ? null : height">
            <canvas [attr.width]="responsive ? null : width" [attr.width]="responsive ? null : height" (click)="onCanvasClick($event)"></canvas>
        </div>
```

The second [attr.] should have height, not width.

PR #5341, opened 2018-03-14
#5322   bug in p chart.ts template

jupin-r:
###Defect Fixes
This PR fixes bug #5322 - "Bug in p-chart.ts template". The property height of p-chart accidentaly set the property width of canvas.


cagataycivici:
There is spinner code so will fix manually

PR #5343, opened 2018-03-14
Update tablerowexpansiondemo.html

potato47:
<a> tag click event is invalid


cagataycivici:
It is not clear to me what this PR does at all.

PR #5344, opened 2018-03-14
Update tablerowexpansiondemo.ts

potato47:
add expand click event


cagataycivici:
It is not clear to me what the PR does.

Issue #5356, opened 2018-03-15
Documentation error on Dialog


swatijain1812:
Hi,
![image](https://user-images.githubusercontent.com/37396973/37449833-c3f78ab0-2851-11e8-8513-791946c3c034.png)

![primeng_dialogbox_minheightproperty](https://user-images.githubusercontent.com/37396973/37449693-405e79ca-2851-11e8-8bcc-3adf638b2336.PNG)

I found an issue on "**PrimneNG Dialog box minHeight property**" you '**write**' width in the description of this property instead of '**height**'. Please correct the description of this.
I also attached the screen shots for this issue.

Thanks,
Swati Jain
Swatij1989@gmail.com

PR #5358, opened 2018-03-15
#5356 - PrimeNG DialogBox Property

jupin-r:
###Defect Fixes
The documentation of dialog 
There was  word width instead height in minHeight property documentation


cagataycivici:
Thank you

PR #5414, opened 2018-03-23
Update keyfilter

jeffwu85182:
Add pBlockKeyOnly if key block only.


cagataycivici:
This causes a backward compatibility issues so unable to accept.

PR #5418, opened 2018-03-23
Adding an updateOption function for the FullCalendar

SteamWind:
Adding an updateOption function to update the options after the view init.

Useful when your config is dependent of the return of a service.

###Feature Requests


SteamWind:
Maybe instead of this, we should create a way to destroy and reinitialize the calendar as looooot of options aren't hot updatable.

cagataycivici:
How about ngOnChanges? Let's discuss this over an issue.

SteamWind:
I made an issue here: https://github.com/primefaces/primeng/issues/5502

PR #5422, opened 2018-03-26
Add tooltip on menu items

sourabh8003:
#### Current behavior
Currently the MenuModel includes a title property which sets the title element attribute on the various pMenu* templates, providing basic tooltip functionality via the browser's native support for displaying the title attribute of an HTML element.

#### Desired behavior
Display the pTooltip component on hover of a pMenu* menu item, via properties passed from a new tooltip property on the MenuItem API with properties corresponding to the pToolTip API:

pTooltip
tooltipPosition
tooltipEvent
positionStyle
tooltipDisabled
appendTo
hideDelay
showDelay
tooltipStyleClass
escape
tooltipZIndex

What is the motivation / use case for changing the behavior?
Consistent user (and developer) experience in displaying tooltips, regardless of the component in which the tooltip is embedded.

### DIFF
--- a/src/app/components/common/menuitem.ts
+++ b/src/app/components/common/menuitem.ts
@@ -21,4 +21,16 @@ export interface MenuItem {
     title?: string;
     id?: string;
     automationId?: any;
+    toolTipMessage?: string;
+    toolTipPosition?: string;
+    toolTipEvent?: string;
+    toolTipPositionStyle?: string;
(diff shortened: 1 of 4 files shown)


cagataycivici:
This bloats the menuitem api with tooltip, instead there should be tooltipOptions, also this PR only adds it to MenuBar. If more users demand it, we'll implement this on our side. Thank you.

PR #5467, opened 2018-04-02
Fixed undefined variable referenced from megamenu template

### LINKED ISSUE #5466: Undefined variable "child" in the template breaks the Megamenu
```
```
**Current behavior**
The megamenu breaks if the model contains a value for **routerLink** at the top level menu item.
This happens because the ***ngIf="category.routerLink"** condition evaluates to true and the template which contains an undefined variable (child) is rendered.

**Expected behavior**
The megamenu should not break when this condition evaluates to true.

Pass the following model to the megamenu to see this issue:

```
this.items = [
            {
                label: 'TV', 
                icon: 'fa-check',
                routerLink: ["/home"]
             }
    ]
```

h-arora:
**Defect Fixes**

Fixes #5466


cagataycivici:
This is fixed already at; 5169

PR #5469, opened 2018-04-02
fullTemplateTypeCheck issue in TabView

steverob2k:
###Defect Fixes
When running aot and fullTemplateTypeCheck is set to true in Angular 5 it is throwing the exception

> ngc -p tsconfig-aot.json
node_modules\primeng\components\tabview\tabview.d.ts.TabPanel.html(6,31): : Property 'columns' does not exist on type 'TabPanel'.


cagataycivici:
Merged manually

Issue #5449, opened 2018-03-29
contextMenu MenuItem visible=false not working anymore


xarfai:
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35

**Current behavior**
MenuItem.visible has no effect in contextMenu

**Expected behavior**
MenuItem.visible = false makes the menu item not appear

**Infos**
#4819 seems to have removed it (maybe for a good reason?)
see: https://github.com/primefaces/primeng/blame/master/src/app/components/contextmenu/contextmenu.ts#L15
and 
https://github.com/primefaces/primeng/commit/66e1498114ba2118f8aeec02ce740733d9aeacd6#diff-71ecd97be5773697a23711e122a54b33R15

* **PrimeNG version:** 5.2.4

* **Browser:** all
* **Language:** all

skormel:
This should be related with #5042

cagataycivici:
Unable to replicate with upcoming 5.2.5

PR #5488, opened 2018-04-04
Fix MenuItem visible on Contextmenu

skormel:
ContextMenu did not use MenuItem.visible property. 
This patch use ng-container to wrap all the code and prevent to inject it if visible property is not set.

#5449


cagataycivici:
Visible is used to hide it with css.

PR #5517, opened 2018-04-10
Removing href="#" attribute from accordion header <a> tag

abidhmuhsin:
Fixed [bug #5518 ](https://github.com/primefaces/primeng/issues/5518)

Since accordion header content is wrapped in <a href...> tag, hovering on the header displays the hyperlink on the page's left-bottom corner.
This is even visible in the demo pages as the hypelink https://www.primefaces.org/primeng/#.
This looks weird in an angular app which itself doesnt have hyperlinks.
Could disable this by styling 'pointer-events:none;', but that also disables any custom click events binding or anything on the header.
Also while trying to bind custom click events it triggered app reload due to href='#' (similar to https://github.com/angular/angular/issues/7294)
**All the functionalities are working without href tag and  it seems more apt not to have an href tag in an angular 2+ app.**
Kindly review the same.

Thanks


cagataycivici:
Without the href, keyboard accessibility is lost

PR #5541, opened 2018-04-11
Correct typing in example use of `modal` property

### LINKED ISSUE #5540: DialogComponent p-dialog modal property use incorrect in example code
```
```

**Current behavior**
The Source tab of the [DialogComponent documentation](https://www.primefaces.org/primeng/#/dialog) shows use of the `modal` property with a string value, but the Properties table of the same `modal` property type as `boolean`, which is the actual [source code typing](https://github.com/primefaces/primeng/blob/master/src/app/components/dialog/dialog.ts#L72).

**Expected behavior**
Documentation should be consistent & correct.

Not confuse developers.

KeithGillette:
###Defect Fixes
#5540

### DIFF
--- a/src/app/showcase/components/dialog/dialogdemo.html
+++ b/src/app/showcase/components/dialog/dialogdemo.html
@@ -6,7 +6,7 @@
 </div>
 
 <div class="content-section implementation">
-    <p-dialog header="Godfather I" [(visible)]="display" modal="modal" [responsive]="true" [width]="350" [minWidth]="200" [minY]="70">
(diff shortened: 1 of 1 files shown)


fengliang0630:
dialog modal has bugs, two Nested dialog set modal is true,  the inner dialog can not click,  the reason is the shade `s style z-index set bug.  please try again and repaire it.  3q

cagataycivici:
@fengliang0630 Never next dialogs, keep them separate instead.

PR #5549, opened 2018-04-12
Allow disabling dialog button focus

AustinMatherne:
The dialog focus functionality forces focus on the first button in the dialog. That functionality often isn't desired.

In our use case, we use a variety of text inputs in our dialogs which need focus instead of the cancel and confirmation buttons at the bottom of the dialog. This PR allows a consumer to disable the focus functionality and to handle it in the parent component.


cagataycivici:
Rename to focusOnShow

PR #5570, opened 2018-04-17
New filter constraint `containsIn`

MI53RE:
#### Feature request

Added a new constraint to Table allowing user to multi-select over a list of elements containing several possibilities.

context ex: a list of cars with several colors options and users who want to select only cars having at least one of the selected colors.
(please look at commit message 01d4f9b which contains a deeper presentation and examples)

#### How this impact the current code ?

This pull request simply add a new attribute to `Table.filterConstraints` : `containsIn` that can be passed 
as an option to `dt.filter` and return a boolean such as the other filters currently in use.

#### How this work ?

The filter will take the value as a string and compare it to each filter as a regexp, if one element in the filters array is present in the value then it return true else if none has been found  it return false.

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -1205,24 +1205,23 @@ export class Table implements OnInit, AfterContentInit {
             return value > filter;
         },
 
-        containsIn: function (value, filter) {
+        containsIn(value, filter): boolean {
             if (filter === undefined || filter === null || filter.length === 0) {
                 return true;
             }
             if (value === undefined || value === null) {
                 return false;
             }
-            for (let _i = 0; _i < filter.length; _i++) {
-                if (filter[_i] === undefined || filter[_i] === null || (typeof filter[_i] === 'string' && filter[_i].trim() === '')) {
+            for (let i = 0; i < filter.length; i++) {
+                if (filter[i] === undefined || filter[i] === null || (typeof filter[i] === 'string' && filter[i].trim() === '')) {
                     return true;
                 }
                 if (value === undefined || value === null) {
                     return false;
                 }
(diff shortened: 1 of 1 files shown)


cagataycivici:
Can you provide a plunkr or a demo to show an example use case, it is not very clear yet.

MI53RE:
Sorry for the late answer, as I am not enough familiar with Plunkr, allow me to provide you a demo through some screenshots:

First thing, the "contains" for "containsIn" doesn't mean that the behaviour is the same as in the existing "contains" . The filter here is looking for an entry in a 2 dimentional array or a list presented as a string.

The foolowing scrrenshot should present it better I think :

So here is my table with some people, each of them having several groups :
firefox_2018-04-27_12-07-36

And here is my filter list made from all those groups : 
firefox_2018-04-27_12-07-49

If I'm selecting gdf2, filter will retrun 2 rows as thos both have gdf2 in their groups list

firefox_2018-04-27_12-08-17

firefox_2018-04-27_12-09-27

If I select gdf3 and gdf11, filters will retrun all row having at least one of each value : 

firefox_2018-04-27_12-08-57

firefox_2018-04-27_12-09-10

But if I select gdf1 only 1 row will be returned as gdf1 !== gdf11 where original "contains" filter would have accepted both

firefox_2018-04-27_12-09-58

firefox_2018-04-27_12-09-42
 

Does that short demo is sufficient to illustrate this new filter ?

I did a small fix update as some special character like `|` would have broken the filters if present in the string, now it will proprely be escaped without risk of being broken.

cagataycivici:
I was expecting value to be a collection as well like;

```
in(value[], filter: any[])
```

Also the PR has no support for Date objects which is required as other filters have it.

PR #5624, opened 2018-04-24
Prevent change if radio button is disabled

thinhbuzz:
## Environment

Angular version: 4.x, 5.x
PrimeNG version: 4.x, 5.x
Browser: all
Language: all

### Current behavior

The user still able change (click to label, using arrow keys) when radio button is disabled, but model or form not update.
http://plnkr.co/edit/tMeADFBlymPDyLVnDeou?p=preview

### Desired behavior
User can not change when radio button is disabled

## Action

- add `disabled` attribute into input element for issue using arrow key change value
- prevent event if radio button is disabled


cagataycivici:
Will fix differently

PR #5651, opened 2018-04-27
Added attribute "disableClickEvent" to p-accordion

rsieja:
Type is Boolean. TODO: check if "disableClickEvent" is true, then change CSS property "cursor" to "default", instead of using "pointer".

I did it because it was needed to expand and hide the accordion using another components like checkboxes or radio buttons instead of the default behavior ("click anywhere over the tabs to open/close").


cagataycivici:
If more users demand it, we'll reconsider, thank you. Not taking action now.

PR #5671, opened 2018-05-03
add feat for support Asian calendar

doyeonOh:
###Feature Requests
titleYearFirst in LocaleSettings
- It is common for Asian calendar to display year first than month first. Primeng has been pushing PR because there is no part to support for that element.

yearNameSuffix in Calendar
- And since it is common to add '년 for Korea and '年' for China and Japan after the year, we also added that field to LocaleSettings.


cagataycivici:
Sorry, create a new PR please, it is not possible to understand what has changed easily.

PR #5690, opened 2018-05-05
Migration to Angular 6

sacquet49:
###Feature Requests
Migration to Angular 6 and fix RXJS import


cagataycivici:
Sorry too many changes for us to review, we'll do it on our side.

PR #5703, opened 2018-05-08
Add support for adding a header to the multi-select in the same fashi…

augwa:
…on as the footer


cagataycivici:
We'll do it on our side for 6.0.2.

Issue #5706, opened 2018-05-09
SplitButton's overlay not being removed


tiagodws:
**I'm submitting a ...**  (check one with "x")
```
[ X ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please demonstrate your case at stackblitz by using the issue template below. Issues without a test case have much less possibility to be reviewd in detail and assisted.

https://stackblitz.com/edit/splitbutton-error

**Current behavior**
When you append the overlay of the SplitButton to the "body" element, if the button gets removed from the screen the overlay stays open, unusable.

**Expected behavior**
The overlay should be destroyed with the button.

**Minimal reproduction of the problem with instructions**

https://stackblitz.com/edit/splitbutton-error

1 - Open the overlay of any splitbutton
2 - With the overlay still open, change the grid's current page
3 - Observe that the overlay stays open, impossible to close, and clicking on any of it's items only fires a navigation event

* **Angular version:** 5.2.9

* **PrimeNG version:** 5.2.5

* **Browser:** All
 
* **Language:** All

sidneibiz:
+1

leonardolessa:
👍

sivnz:
+1

rafaeldmpacheco:
+1

felipexw:
+1

felipefaust:
👍

sut123:
If anyone absolutely needs a temporary fix for this, here's what we did:

  @ViewChild(SplitButton) splitButtonComponent: SplitButton;
  ngOnDestroy() {
    if (this.splitButtonComponent) {
      let overlay = this.splitButtonComponent.overlayViewChild.nativeElement;
      if (overlay && overlay.parentNode) { overlay.parentNode.removeChild(overlay); }
    }
  }

PR #5707, opened 2018-05-09
Ensures that the SplitButton's overlay will be removed when the component gets destroyed #5706

tiagodws:
###Defect Fixes

###Feature Requests


cagataycivici:
Will fix it differently and let angular destroy it.

Issue #5095, opened 2018-02-07
"disabled" property not working on tableRadioButton and tableCheckbox


UrzaUrza:
**I'm submitting a ...**
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
When the "disabled" property is set to true on tableRadioButton or tableCheckbox, the row is still selectable.

**Expected behavior**
When the "disabled" property is set to true on tableRadioButton or tableCheckbox, the row is not selectable.

* **PrimeNG version:** 5.2.0

sashaSavych:
The same for me :(
"disabled" property sets just style

jun58:
It same for me ~~

noamichael:
I submitted a PR:

#5515

cagataycivici:
Set to 5.2.5.

cagataycivici:
PR merged, thank you.

jmaruiz:
Can you do the same with tableHeaderCheckbox? That does not obey disabled styling or functionality.

PR #5761, opened 2018-05-18
disabled property to TableHeaderCheckbox

noamichael:
### Defect Fixes
As discussed in #5095, the table radios and checkboxes could not be disabled. This commit allows the `<p-tableHeaderCheckbox/>` to be disabled, which I missed in #5515


cagataycivici:
Merged manually

Issue #5638, opened 2018-04-26
Change checkbox mode behavior in ListBox


Mynken:
I'm submitting a...
```
[X ] bug report
[X ] feature request 
```
**Current behavior**
I am using pListBox Advanced with multiply flag, it "requires" to click exactly on the checkbox.
When I have some values selected and click on next value(on the text of this value or between checkbox and text) the ListBox value delete all previous values and have value only of the last element which I clicked.

**Expected behavior**
When click on "all" active zone for item => excpeted behavior is to add this item to ListBox, not to clear it and add last element.

**Minimal reproduction of the problem with instructions**

Mynken:
1) How it now works:
https://ibb.co/fMUf1x
2) Expected behavior
https://ibb.co/cWxB8c

POLAK93:
I have the same issue, it's annoying.

PR #5770, opened 2018-05-21
Listbox with checkboxes click behaviour

h1949158:
If checkboxes are enabled the Listbox behaves as if the ctrl-key is pressed so the list wont be cleared if the checkbox is missed.
Issue #5638 on primefaces/primeng describes that problem.
A demonstration of the problem and it's fix here:
https://stackblitz.com/github/primefaces/primeng-issue-template?file=src%2Fapp%2Fapp.component.html

###Defect Fixes

###Feature Requests


cagataycivici:
Will do it differently in 6.0.2.

PR #5771, opened 2018-05-21
Dropdown: allow for implicit form submission

pawelbialaszczyk:
This PR enables implicit form submission using the enter key in the context of the dropdown component.

Implicit form submission is only possible when the native input element has focus and the dropdown is closed.


cagataycivici:
It is not clear to me what the PR does and we should not force focus if outside is clicked to hide. It would be confusing.

PR #5808, opened 2018-05-28
#5806 - restoring dialog when using custom route reuse strategy

### LINKED ISSUE #5806: dialog is not restored when using custom angular reuse strategy
**I'm submitting a ...**  
```
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/github-tnguf2

**Current behavior**

I am trying to use a dialog component inside an app with routing, using a custom reuse strategy. It is a complex use case, with some other code firing changes at some specific moments during route navigation, so the code example up here can seem a little tricky.

When I leave a component with an open dialog, and some changes happen after the component was removed from the DOM, the dialog is hidden (display: none) when angular router reuses component instance.

**Expected behavior**

The dialog should be visible again on the page.

To see the issue:

1. click on "Open dialog"
2. click on "Navigate to 2nd component", the current component is stored to be reused by angular router (see customReuseStrategy.ts)
3. on the 2 component, click on "Reuse previous route", to go back to the first component : angular will use the previously stored component instance, with the same state.

You will notice the dialog is not visible, yet its input state is visible: as a consequence, if you click on "Open dialog" again, nothing happens as the boolean is already true and the popup should already be visible.

The issue comes from this line, in the `center` method of the dialog, called from `ngAfterViewChecked` through `positionOverlay`: https://github.com/primefaces/primeng/blob/master/src/app/components/dialog/dialog.ts#L286

While navigating, if the component has been removed from the DOM when `ngAfterViewChecked` is fired, its width and height equal 0, so the line forces display mode to none which does not correspond to visible input anymore. One simple fix could consist in storing the previous value of display style property, to reuse it here instead of forcing none. The patched method body can be found as an example in my demo, in `c1.component.ts`. You can just uncomment the code to notice it fixes the issue.

I can send a pull request if you think this fix is appropriate.

* **Angular version:** 5.2.0

* **PrimeNG version:** 5.2.6

* **Browser:** Chrome
 
* **Language:** TypeScript

m1ch3lcl:
fixes #5806 

In dialog component :
- store display style property in case dialog container's width and height equal 0
- restore previous display mode after centering modal

### DIFF
--- a/src/app/components/dialog/dialog.ts
+++ b/src/app/components/dialog/dialog.ts
@@ -279,11 +279,12 @@ export class Dialog implements AfterViewInit,AfterViewChecked,OnDestroy {
         let elementWidth = this.domHandler.getOuterWidth(this.containerViewChild.nativeElement);
         let elementHeight = this.domHandler.getOuterHeight(this.containerViewChild.nativeElement);
         if(elementWidth == 0 && elementHeight == 0) {
+            const displayMode = this.containerViewChild.nativeElement.style.display;
             this.containerViewChild.nativeElement.style.visibility = 'hidden';
             this.containerViewChild.nativeElement.style.display = 'block';
             elementWidth = this.domHandler.getOuterWidth(this.containerViewChild.nativeElement);
             elementHeight = this.domHandler.getOuterHeight(this.containerViewChild.nativeElement);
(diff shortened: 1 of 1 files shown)


cagataycivici:
It is not clear what this change does and it may cause regressions.

PR #5839, opened 2018-06-03
resolveFieldData should be able to resolve functions as well

NicholasMaestrello:
Hi everyone, I have changed the method resolveFieldData and created another method called isFunction so that when I have a field that is defined as a function, the resolveFieldData method can extract its data


heliobeirao:
+ 1

cagataycivici:
So many lint changes so will merge manually

PR #5852, opened 2018-06-05
accessibility(paginator): add aria-label descriptions for buttons

gilhanan:
add aria-label descriptions for buttons in order to improve its accessibility (ARIA14)


cagataycivici:
We should not add static English texts to the components. They must be properties of the paginator.

PR #5869, opened 2018-06-07
Support for rowexpand functionality also for frozen rows

SPKorhonen:
This PR adds support for rowexpand also for frozen rows via frozenrowexpansion template which mirrors rowexpansion template. If no frozenrowexpansion is defined the behaviour of the table component remains unchanged.

Also included is simple demo of the new functionality.


SPKorhonen:
Sorry, got overzealous when cleaning old branches

cagataycivici:
We'll do it on our side based on your PR for 7.0.1, thank you;

Please monitor: https://github.com/primefaces/primeng/issues/6596

sten82:
why is that not part of v8? this is exactly what I'm looking for

Issue #5581, opened 2018-04-18
Slider with step 0.1 don't working


bbuehrer:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://plnkr.co/edit/BDbKBUgPbLMhOqC1YPxZ?p=preview

**Current behavior**
Slider: step=0.1 don't working

* **Angular version:** 5.2.10

* **PrimeNG version:** 5.2.4

bbuehrer:
I think the "handleStepChange"-function is invalid:

    handleStepChange(newValue: number, oldValue: number) {
        let diff = (newValue - oldValue);
        let val = oldValue;
        
        if(diff < 0) {
            val = oldValue + Math.ceil((newValue - oldValue) / this.step) * this.step;
        }
        else if(diff > 0) {
            val = oldValue + Math.floor((newValue - oldValue) / this.step) * this.step;
        }
        
        this.updateValue(val);
        this.updateHandleValue();
    }


Here's the valid code:

    handleStepChange(newValue: number, oldValue: number) {
        let diff = (newValue - oldValue);
        let val = oldValue;
        
        if(diff < 0) {
            val = oldValue + Math.ceil(newValue / this.step - oldValue / this.step) * this.step;
        }
        else if(diff > 0) {
            val = oldValue + Math.floor(newValue / this.step - oldValue / this.step) * this.step;
        }
        
        this.updateValue(val);
        this.updateHandleValue();
    }

PR #5879, opened 2018-06-08
 Fixed bug with step <1 in Slider

bbuehrer:
see "handleStepChange" funtion.

###Defect Fixes
issue: Slider with step 0.1 don't working #5581


bbuehrer:
I have now updated to version 6.1.3 and it does not work!

Rathdor:
I have tried `step 0.5` in PrimeNg 7.0.3 and it does not work.

Any solution, please?

Issue #5885, opened 2018-06-11
TextArea directive bug using ui-inputtext instead ui-inputtextarea


brauliodeleon:
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Right now the textarea when I use pInputTextarea adds `ui-inputtext` class instead of `ui-inputtextarea` class (as described in [docs](https://www.primefaces.org/primeng/#/inputtextarea))

**Minimal reproduction of the problem with instructions**
Add a TextArea component and pInputTextarea as directive.
Open the inspector and you can check that `.ui-inputtext` is added instead of `.ui-inputtextarea`

**What is the motivation / use case for changing the behavior?**
The motivation is matching with documentation, and keep isolated behavior between inputText and textArea

**Please tell us about your environment:**

MacOS
WebStorm
npm

* **Angular version:** 6.0.2

* **PrimeNG version:** "6.0.0-alpha.2"

* **Browser:** all
 
* **Language:** all

* **Node (for AoT issues):** `node --version` =   8.11.2

cagataycivici:
Fixed the docs, it uses ui-inputtext as well.

PR #5886, opened 2018-06-11
fix inputTextarea directive, according to docs

bradelev:
### Defect Fixes
related to #5885


cagataycivici:
Well this removes all styling from pInputTextarea so can't merge, I've just fixed the docs instead.

PR #5902, opened 2018-06-12
Re-position TieredMenu when window resizes

### LINKED ISSUE #5901: Popup TieredMenu should hide itself on window resize
```
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/github-wtne7e

**Current behavior**
If the menu is opened and the window is resized, the menu doesn't re-position itself to it's container.

**Expected behavior**
When resizing the window, the menu and its container should always be "stuck" together, like the regular Menu is.

1. Click the button to show the menu
2. Change the browser window's width. Notice how the menu drifts away from the button as the browser changes.

In an application with many menus (standard menus and tiered menus), they should all behave consistently. Additionally, even with just a single tiered menu, if the menu does stick to its container, then it's easy for the menu to get lost off screen as the window gets narrower.

* **Angular version:** 5.2.10

* **PrimeNG version:** 5.2.6

mroskamp:
Fixes #5901

### DIFF
--- a/src/app/components/tieredmenu/tieredmenu.ts
+++ b/src/app/components/tieredmenu/tieredmenu.ts
@@ -114,7 +114,8 @@ export class TieredMenuSub {
             <p-tieredMenuSub [item]="model" root="root" [baseZIndex]="baseZIndex" [autoZIndex]="autoZIndex" [hideDelay]="hideDelay"></p-tieredMenuSub>
         </div>
     `,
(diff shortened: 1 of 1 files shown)


cagataycivici:
This is a performance penalty, it will trigger change detection. Better to do it with an NgZone outside of angular and bind/unbind it on show.

PR #5907, opened 2018-06-13
Format and fix tslint warnings in schedule component

vially:
Just some minor cleanups to make the linter happy.

Most of the changes were fixed automatically using: 

```bash
tslint -c tslint.json src/app/components/schedule/schedule.ts --fix
```

PR #5908, opened 2018-06-13
Bump fullcalendar to v4.0.0-alpha

vially:
This should allow dropping [jQuery as a dependency](https://fullcalendar.io/blog/2018/04/alpha-release-jquery-removal).

PS: This PR is built on top of #5907, so once that one gets merged I'll rebase this one.


cagataycivici:
We've done it ourselves already, thank you.

PR #5938, opened 2018-06-19
Resize columns to min-width if resize was smaller than that

### LINKED ISSUE #5937: Turbo Table Column Resize Is Ignored If Smaller Than minWidth
```
```

**Current behavior**
1. Set [style.minWidth] for a resizeable column
1. Try to resize the column to be narrower than its min-width
Result: resize is ignored, the column has the same width as before the resize.

**Expected behavior**
The column should be resized to its min-width.

This behavior is not user friendly.

Dragsaw:
###Defect Fixes

Fix #5937

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -1396,7 +1396,7 @@ export class Table implements OnInit, AfterContentInit {
     
         const newColumnWidth = columnWidth + delta;
     
-        if (columnWidth + delta > minWidth) {
+        if (newColumnWidth >= minWidth) {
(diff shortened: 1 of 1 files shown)


cagataycivici:
Sorry I don't get this PR

Dragsaw:
@cagataycivici , let me explain:
This is how it works right now:

The column has a min width of 300px. If it is resized to be shorter, the width **does not change** at all.

This is how it works with the PR:

Again, the column has a min width of 300px. If it is resized to be shorter, the width is **changed to min width** (300px in this case).

PR #5941, opened 2018-06-20
I added readonly to the calendar component.

alinsimy:
###Feature Requests
I added readonly property to calendar component.


cagataycivici:
States seems to be mixed up with this, for example readonly disables the button. We'll reconsider if there is more demand.

Issue #5040, opened 2018-02-01
Multiple filters on a field in TurboTable


sebaiona:
### There is no guarantee in receiving an immediate response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeNG PRO Support* where support is provided within 4 business hours

**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[X] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**

When you filter a field and then you want to add another filter to this same field, it replaces the previous one.

**Expected behavior**

The possibility to set two filters on a same field. 
For example I'd want to set `gt` a value and `lt` another value on a date field.

diegoflores:
any update on this?

alpernakin:
up?

alpernakin:
I sorted the problem on this pull request: https://github.com/primefaces/primeng/pull/5976

matt14m:
Need this as well. For filtering on a time range.

ElliotSchmelliot:
@matt14m Check out [this SO post](https://stackoverflow.com/questions/50098545/primeng-datatable-date-range-filter) on achieving date range filtering using this control. Worked great for me, although I do agree that being able to add multiple filters more directly would be a plus.

fwin02208:
any update for this?

yigitfindikli:
You can use our new [p-columnFilter](https://primefaces.org/primeng/showcase/#/table/filter) for this.

PR #5976, opened 2018-06-27
Multiple filters for a column

alpernakin:
As turbo-table allows single local filter for a column, the code has been edited for applying multiple filters at the same time

###Feature Requests
https://github.com/primefaces/primeng/issues/5040

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -153,7 +153,9 @@ export class Table implements OnInit, AfterContentInit {
 
     @Input() exportFilename: string = 'download';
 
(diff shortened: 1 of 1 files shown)


cagataycivici:
Feature of this size should be implemented by the PrimeNG team.

umdstu:
Is there an ETA/release for this to happen then @cagataycivici ?

alpernakin:
> Feature of this size should be implemented by the PrimeNG team.

I don't understand why the team still ignores that important enhancement, though it is just an easy commit

PR #5980, opened 2018-06-28
add ui-datepicker-timeonly class

televic-education:
This css class can be used to style the datepicker differently (different icon e.g.) when the timeonly property is set.


cagataycivici:
Renamed to ui-calendar-timeonly.

PR #5981, opened 2018-06-28
add missing events on confirmdialog

televic-education:
We needed to know when the confirmdialog is closed by using the cross in the upper right corner without accepting nor rejecting.
Our use case is as follows: when a user tries to navigate to another page, but has made some changes, we wanted to ask the user whether or not those changes should be saved. When the confirmation is simply closed, we need to prevent the navigation and stay on the current page. Without these events, it's impossible to know when that happens.


cagataycivici:
This causes an API inconsistency with Dialog which has onHide only. Why not set "closable" as false in your case to force confirmation.

televic-education:
Adding onHide and onShow to the confirmationDialog, like they exist on the Dialog, is also perfectly fine for us. It doesn't have to be beforeHide or afterHide, we saw those events existed on the overlayPanel, so we copied them from there.

PR #5982, opened 2018-06-28
add ui-state-completed class to steps

televic-education:
This css class can be used to skin all the previous steps of a wizard differently.
In our case we wanted to show checkmarks:
image


cagataycivici:
Renames to ui-steps-complete and also added ui-steps-incomplete / ui-steps-current as well.

PR #5985, opened 2018-06-29
Solve the problem：Table is messing up array order with more than 10 i…

cnnicknick:
…tems

We do not want to sort the table if the param (this.sortField) is null or undefined. this is my suggestion, maybe you can optimize the code.

###Defect Fixes

###Feature Requests

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -564,7 +564,8 @@ export class Table implements OnInit, AfterContentInit {
                     });
                 }
                 else {
-                    this.value.sort((data1, data2) => {
+                    if( this.sortField ){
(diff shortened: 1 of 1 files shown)


cnnicknick:
This is my first time to pull a request. I have a lot to learn from you.

cnnicknick:
15.4.4.11 Array.prototype.sort (comparefn)：The elements of this array are sorted. The sort is not necessarily stable (that is, elements that compare equal do not necessarily remain in their original order). copy from 【ECMAScript Language Specification Edition 3 24-Mar-00】

cagataycivici:
But line has this check already?

`if(this.sortField && this.sortOrder) {`

Issue #4891, opened 2018-01-19
Support for MenuItem icons in breadcrumb


KeithGillette:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[X] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
The `icon` property of `MenuItem` is ignored by `BreadcrumbComponent`.

**Expected behavior**
`BreadcrumbComponent` displays specified `icon` to left of `label` like most other PrimeNG menu components.

**What is the motivation / use case for changing the behavior?**
* Provide additional information to user about the items displayed in the breadcrumb navigation.
* Create more consistent API/developer experience across PrimeNG menu components.
* Make BreadcrumbComponent even more awesome.

PR #5989, opened 2018-06-29
Support icon field of MenuItem

hughanderson4:
If `MenuItem` structure specifies an icon, let's draw that on the Breadcrumbs.  Applying a class of `ui-menuitem-icon` would allow user to customize display, if desired.
Additional fix is provided for customizing `home.icon` in the case where `home.routerLink` is defined.
It can solve #4891


hughanderson4:
this has a typo.  will submit it again

Issue #3269, opened 2017-06-30
Calendar widget should have a way to display only month and year


nareshurs:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

http://plnkr.co/edit/Wj39h1?p=preview

**Current behavior**

**Expected behavior**

**Minimal reproduction of the problem with instructions**

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **Angular version:** 2.0.X

* **PrimeNG version:** 2.0.X

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

* **Node (for AoT issues):** `node --version` =   
https://www.primefaces.org/primeng/#/calendar

There should be a way to display only months and year in calendar. Can you consider this in your future releases?

coppolaop:
There is any news about it?

ChristianSchaiter:
I would love to see this feature in a future release too!

sridhark158:
+1

leonardolessa:
👍

herychemo:
+1

victorguillon:
+1

Goldflow:
+1

bandedos:
+1

csvinhal:
+1

MicheleMola:
+1

EricHislop:
+1

ssarkisy:
+1

kmiszcz:
+1

senad87:
@cagataycivici what about year only calendar?

PR #6020, opened 2018-07-08
Added month only datepicker

YounesM:
The calendar is lacking a month only picker ( #3269 ), so I tried adding one to it.


cagataycivici:
Thanks but features of this size must be implemented by our team.

PR #6031, opened 2018-07-10
Enable custom filtering on p-dropdown

gbourant:
###Feature Requests
Enable custom filtering on p-dropdown


cagataycivici:
Sorry but feature of this size should be implemented by the team.

gbourant:
The filter comparison is happening based on indexOf method.
Sometimes the user doesn't want the values that contains the filterValue to be displayed.

It would be great if you can add startsWith method for comparison.

PR #6048, opened 2018-07-11
Update Dialog documentation to include blockScroll

andyg3:
###Defect Fixes
Added documentation for blockScroll property of Dialog


cagataycivici:
It seems it has been included by then.

Issue #6059, opened 2018-07-13
Showcase table selection missing sample for mixing with frozen columns


thedanielgray:
It was not evident to me how to mix frozen columns with a row selection so I read through and figured it out and thought I might share how to do it with others via a showcase sample (provided in PR).

cagataycivici:
If more users demand it, we'll add it. Thank you.

PR #6060, opened 2018-07-13
Add frozen columns + row selection showcase sample

thedanielgray:
It was not evident to me how to mix frozen columns with a row selection so I read through and figured it out and thought I might share how to do it with others via a showcase sample (related issue - #6059).


cagataycivici:
If more users demand it, we'll add it. Thank you.

thedanielgray:
Why is that? What do you have to lose from adding an example and making your documentation/help more detailed? It would make sense if there is some maintenance or other additional work involved on your side, but I honestly don't see any reason not to just add it in.

Issue #4380, opened 2017-11-07
PickList: disabled property only on control buttons


pierfreeman:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
Hi, first of all thanks for adding disabled property on the PickList, is really useful.

Another suggestion to enhance this component would be giving more control about this property, like having a more 'fine-grained' disabled on the control buttons.

One use case would be to disable the 'move-to-target' or 'move-to-source' buttons if a minimum or maximum value is reached on one of the two lists.

stephanie-dm:
Another important thing to notice about this question is that it's not only the buttons that should be disabled in this example, but also moving items by double clicking or dragging them should not be possible anymore.

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

PR #6074, opened 2018-07-16
Add fine-grained controls over what move buttons to disable/hide

thedanielgray:
###Defect Fixes
This PR adds more fine-grained controls over disabling individual buttons (move one or all in either direction) allowing the programmer to individually disable any of the buttons, as requested in issue #4380. I also added a `moveDisabled` attribute that will only disable movement from one list to another but still allow other operations (like reordering) on the list.

I have also added an input to hide any of the item moving buttons individually.

If you approve the PR, I'm also happy to add documentation to the showcase section on the new attributes!


cagataycivici:
We'll do it on our side, thank you.

thedanielgray:
And what's the reason for that?

Issue #6076, opened 2018-07-16
Separate templates for Source and Target of PickList


danielthegray:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
The PickList only allows the programmer to define a single item template that will be applied for both the source and target lists.

**Expected behavior**
The PickList component should allow independent templating of the source and target list items if it is desired.

**What is the motivation / use case for changing the behavior?**
To improve the flexilibity of the PickList.

* **PrimeNG version:** 5.X
6.0.1

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
All
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
all

cagataycivici:
We'll reconsider this if more users demand it, thank you. Currently no request has been made except this one.

JaviManteiga:
@cagataycivici +1

danielthegray:
@JaviManteiga I posted code that does it here:
https://github.com/primefaces/primeng/pull/6077

JaviManteiga:
Thank you @danielthegray I'll take a look.

johnmarian:
So this is what I've done as a workaround:

- I extended the interface for the items in the list to include a property to indicate what list it's on
- I created methods to handle the events from the `onMoveToTarget`, `onMoveAllToTarget`, `onMoveToSource`, and `onMoveAllToSource` events
- I parse the event to get the items that have been moved (note, the shape of this event is different when using the buttons to move an item vs. when an item is dragged, so you'll have to handle that too if you're using drag/drop)
- I iterate over the source or target list (as applicable) and update the custom property for every item on the list of moved items
- I use an `ngSwitch` in the template to render different templates based on the value of the custom property

This is not efficient at all, but it's the only workaround I've found until the very simple PR from @danielthegray gets merged (why wasn't this already merged??).

francowhyline:
Hi! We are also interested in this feature because we need to add an input on the target list!

cpt-agemman:
[...]

PR #6077, opened 2018-07-16
Add separate templating parameters for source and target lists

thedanielgray:
###Defect Fixes
This adds a feature described in issue #6076, where there are targetItem and sourceItem template types that allow the programmer to separately template items in the source and target list.

### DIFF
--- a/src/app/components/picklist/picklist.ts
+++ b/src/app/components/picklist/picklist.ts
@@ -31,7 +31,7 @@ import {ObjectUtils} from '../utils/objectutils';
                             (click)="onItemClick($event,item,selectedItemsSource,onSourceSelect)" (dblclick)="onSourceItemDblClick()" (touchend)="onItemTouchEnd($event)"
                             [style.display]="isItemVisible(item, -1) ? 'block' : 'none'"
                             [draggable]="dragdrop" (dragstart)="onDragStart($event, i, -1)" (dragend)="onDragEnd($event)">
-                            <ng-container *ngTemplateOutlet="itemTemplate; context: {$implicit: item, index: i}"></ng-container>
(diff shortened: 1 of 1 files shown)


cagataycivici:
We'll do it on our side if more users demand it, thank you.

thedanielgray:
Is there some sort of stylistic/architectural difference that prevents you from merging in the pull request? I put effort into separating all the changes I did into small deltas specific for every feature and to match your coding style as closely as possible.

zLAs3R:
Hi,
i'm also highly interested by this feature !!

Naveed1842:
Hi I am also looking for this feature.We have use case we are showing all the user in sourcelist but when client take some user from list we need to show the dropdown with those in target list.If there are two separate template available it will easy.Will you please add this feature.

aammfe:
Hello!! I m looking for this feature as well..
we have to show all product and user will select those products and in target like he will add quantity to it.
so yes this feature is must.

danorr:
Would very much welcome this feature as I'm needing to show the items that have already been selected by a user in the targetList.
Yes please!

johnmarian:
Would also very much like this feature!

francowhyline:
Hi! We need this feature also!

cloggy45:
And update on this?

kralphs:
Would like this as well

danielthegray:
It's kinda hilarious/sad that even though I did the work for this feature, over 4 years ago, and packaged it in as small a form as possible, as respectful as possible with the existing code styles, it was closed with a simple "not-invented-here" argument of "we'll do it on our side if there is demand", and even after so much demand from people both here and in the issue, no response has been given.

BryceSimtars:
> We'll do it on our side if more users demand it, thank you.

Many people have requested this and @danielthegray provided a very clean and simple implementation, why is it being ignored?

Issue #5853, opened 2018-06-05
Remove UTC mode from Calendar


cagataycivici:
This property makes the calendar 2x complex, almost everything has to be coded again for UTC. Instead provide a demo to show how to work with UTC and potentially any time zone.

Menecats:
The #5823 simplifies the code, you don't have to code twice because it access automatically to the UTC and non-UTC methods by using  the utc flag.

Eg:

```javascript

doSomething(): void {
    // Before
    const pm = this.utc ? this.date.getUTCHours() > 11 : this.date.getHours() > 11;
    // After
    const pm = this.getHours(this.date) > 11;

    // Before
    this.utc ? this.date.setUTCHours(13) : this.date.setHours(13);
    // After
    this.setHours(this.date, 13);
}

getHours(date: Date): number {
    if (this.utc) {
        return date.getUTCHours();
    }
    return date.getHours();
}

setHours(date: Date, hours: number): Date {
    if (this.utc) {
        date.setUTCHours(hours);
    }
    date.setHours(hours);

    return date
}
```

Mysame:
This just completely broke my whole project, great.
You can't just pull functionality like this. An informative line in the console is definitely not enough.

I see no reason why @Menecats PR wasn't accepted. It fixes the concern of "double-logic", and allows us flexibility with timezones.

kstruyven:
Why the hell would you remove this feature?? Is there at least a work around to have it back in UTC? Our project is UTC time only.

sgentile:
ugh, same issue here, what happen to the utc flag ???

mparker-ft:
guys any workarounds for this? it's broken our whole project as well

mparker-ft:
@cagataycivici 

> Instead provide a demo to show how to work with UTC and potentially any time zone

Where can we find this demo?

anaq1:
Broken project here as well... we use the utc flag for correct conversion of date without time part from p-calendar to LocalDate on java server. Now it's not working anymore and we have to face the "day minus one" issue. In our case the timezone has no meaning.

mahald:
Removing such a core feature is silly. How about to provide a directive in the primeNg lib similar to anaq1 one's, but with time support ?

MattiJarvinen-BA:
Just came through to check if I could update primeng 5 -> 7 ... apparently I can't this really sucks.

MattiJarvinen-BA:
[...]

PR #6081, opened 2018-07-16
New p-datepicker element

Menecats:
Hi,

I've just created this brand new component, mainly to fix the gap left when removing the 'utc' flag (#5853) but also to fix some issues in the time picker when minDate and maxDate are set.

This component use Luxon under the hood and allows to pick time in the specified timezone.

It also allows single, multi and ranges selections with years, months and days; for example the code below allows to pick different months.

```HTML
<p-datepicker [(ngModel)]="value"
              selectionMode="multiple"
              precision="month"
              dateFormat="MM/yyyy">
</p-datepicker>
```

**Note:** I've also added an EventEmitter to the p-inputmask element in order to obtain the click event from the datepicker element


digaus:
Wouldnt it be better to start using the DatePipe from Angular? Had to modify the current calendar to be able to use millis as input and use the DatePipe for better and easier formatting.

Menecats:
@digaus I've tried to keep as much as possible the same behaviour as the original calendar component.
The datePipe is useful for formatting dates but not to parse them (AFAIK), I've used DateTime from Luxon because it's good for timezone management and for the ease of formatting capabilities (it uses browser native Intl API).
Regarding the locale, in some projects I've worked on it was useful to have the possibility to change every single i18n entry of the calendar, but that depends on each use case.

Menecats:
With these new changes it's possible to specify some default configuration directly when the module is imported.

### Module import w/ configuration

Module:
```typescript
@NgModule({
  imports: [
    ...

    DatePickerModule.configure({
      zone: 'utc',
      dateFormat: 'yyyy/MM/dd'
    }),

    ...
  ],
  ...
})
export class MyAwesomeModule() {
}
```

Template:
```html
<p-datepicker [(ngModel)]="myValue"></p-datepicker> <!-- zone=utc, dateFormat=yyyy/MM/dd -->
<p-datepicker [(ngModel)]="myValue" zone="local"></p-datepicker> <!-- zone=local, dateFormat=yyyy/MM/dd -->
<p-datepicker [(ngModel)]="myValue" dateFormat="dd/MM/yyyy"></p-datepicker> <!-- zone=utc, dateFormat=dd/MM/yyyy -->
```

### Plain module import

Module:
```typescript
@NgModule({
  imports: [
    ...

    DatePickerModule,

    ...
  ],
  ...
})
export class MyAwesomeModule() {
}
```

Template:
```html
<p-datepicker [(ngModel)]="myValue" zone="utc" dateFormat="yyyy/MM/dd"></p-datepicker> <!-- zone=utc, dateFormat=yyyy/MM/dd -->
<p-datepicker [(ngModel)]="myValue" dateFormat="yyyy/MM/dd"></p-datepicker> <!-- zone=local, dateFormat=yyyy/MM/dd -->
<p-datepicker [(ngModel)]="myValue" zone="utc"></p-datepicker> <!-- zone=utc, dateFormat=dd/MM/yyyy -->
```

cagataycivici:
Sorry but due to policy we don't accept new components or feature implementations that apply huge changes. Plus we already have p-calendar and there is no need for a separate p-datepicker.

arnotes:
> Sorry but due to policy we don't accept new components or feature implementations that apply huge changes. Plus we already have **p-craplendar** and there is no need for a separate p-datepicker.

Issue #6092, opened 2018-07-18
editable treetable documentation error


asagajda:
I'm submitting an error found in documentation: https://www.primefaces.org/primeng/#/treetable (Editing section)

* **PrimeNG version:** 6.0.1

Trying this is not working (tried according to documentation):
`<td [ttEditableColumn]="rowData" [ttEditableColumnField]="'year'">`

This example should be replaced with:
```
<p>If you require the edited row data and the field at onEditComplete event, bind the data to pEditableColumn and pEditableColumnField properties of ttEditableColumn directive</p>
<td ttEditableColumn [pEditableColumn]="rowData" [pEditableColumnField]="'year'">
```
After changing code like this I got expected behaviour

asagajda:
Created PR [](https://github.com/primefaces/primeng/pull/6093)

cagataycivici:
Merged

PR #6093, opened 2018-07-18
Tiny fix in documentation on how to use editable treetable

asagajda:
Issue:
[https://github.com/primefaces/primeng/issues/6092](https://github.com/primefaces/primeng/issues/6092)


cagataycivici:
Should be the other way, code should be fixed instead.

Issue #6103, opened 2018-07-19
Improving Accessibility for Visually Impaired Users Via Aria Attributes


mattcfilbert:
**I'm submitting a ...**
```
[ ] bug report => Search github for a similar issue or PR before submitting
[X] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Many input components can be made more accessible to the visually impaired by adding "aria-label" and "aria-required" attributes. The "required" attribute does a lot of what the "aria-required" attribute does, but the latter is still useful for user agents that do not yet support HTML5.

**Proposed improvement**
I am submitting a PR that adds "aria-label" and "aria-required" attributes to all applicable input components. 

**What is the motivation / use case for changing the behavior?**
This PR will improve PrimeNG's overall accessibility for visually impaired users who rely on screen readers.

mattcfilbert:
My pleasure! Thank you for improving PrimeNG’s accessibility, I am a big
fan of your excellent library.

maniiiit:
is aria-label enhancement added to the calendar component ??

maniiiit:
is aria-label property included in calendar component ??

cagataycivici:
The issue is too broad, please create per component tickets.

PR #6104, opened 2018-07-19
Aria accessibility improvement

mattcfilbert:
### Defect Fixes
This PR addresses the accessibility improvements outlined in [issue #6103 ](https://github.com/primefaces/primeng/issues/6103).

### Feature Requests
This PR adds "aria-label" and "aria-required" attributes to applicable PrimeNG input components. The changes are small, but impactful for visually impaired users with screen readers. This will improve PrimeNG's overall accessibility.

The biggest change made to any of the input components was the addition of the following code to the typescript:

The addition of [attr.aria-label]="ariaLabel" [attr.aria-required]="ariaRequired" to the input element in the template and the addition of @Input() ariaLabel: string; and @Input() ariaRequired: boolean; to the list of inputs.

NOTE - The "required" attribute does a lot of what the "aria-required" attribute does, but the latter is still useful for user agents that do not yet support HTML5.


cagataycivici:
Please create separate tickets for each component as each component is treated separately, this will let us see the changes better as well, right now it is way too much for a single commit.

Issue #6142, opened 2018-07-23
Fix for suggestions not showing up due to this.loading = false on second time load


nimeshvaghasiya:
I'm using p-autocomplete for postcode lookup. it is work like below.
search with M4C 1M5 post code it will give me 1 suggestion with placeholder 13 addresses.
When i'm selecting that I'm changing the suggetions with that 13 addresses from api result.
at that time below line is stopping me to show 13 addresses in suggestion dropdown. Specially this.loading is false so it is not entering inside of condition in handleSuggestionsChange function.

Code Line :  if (this.panelEL && this.panelEL.nativeElement && **this.loading**) {

this.loading should be true as i did in below code, otherwise it is not allowing me to show changed suggestions on second time because i'm selecting it from dropdown and based on selected value i set the suggestions again.

`Object.defineProperty(AutoComplete.prototype, "suggestions", {
        get: function () {
            return this._suggestions;
        },
        set: function (val) {
            this._suggestions = val;
            if (this.immutable) {
                if (this._suggestions && this._suggestions.length) {
                    this.loading = true;
                }
                this.handleSuggestionsChange();
            }
        },
        enumerable: true,
        configurable: true
    });`

cagataycivici:
Please provide a stackblitz template first to demonstrater the issue and create a new issue ticket with your case.

PR #6143, opened 2018-07-23
Fix for suggestions on second time load.

nimeshvaghasiya:
I'm using p-autocomplete for postcode lookup. it is work like below.
search with HU1 1UU post code it will give me 1 suggestion with placeholder 13 addresses.
When i'm selecting that I'm changing the suggetions with that 13 addresses from api result.
at that time below line is stopping me to show 13 addresses in suggestion dropdown. Specially this.loading is false so it is not entering inside of condition in handleSuggestionsChange function.

Code Line :  if (this.panelEL && this.panelEL.nativeElement && this.loading) {

###Defect Fixes
[https://github.com/primefaces/primeng/issues/6142](https://github.com/primefaces/primeng/issues/6142)

###Feature Requests

### DIFF
--- a/src/app/components/autocomplete/autocomplete.ts
+++ b/src/app/components/autocomplete/autocomplete.ts
@@ -202,6 +202,7 @@ export class AutoComplete implements AfterViewChecked,AfterContentInit,DoCheck,C
         this._suggestions = val;
         
         if(this.immutable) {
+            this.loading = true;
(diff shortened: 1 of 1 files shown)


cagataycivici:
Please provide a stackblitz template first to demonstrate the issue and create a new issue ticket with your case.

Issue #6098, opened 2018-07-18
Table: Virtual scroll broken


codyb11989:
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Virtual Scroll changes data but the rows stay at the top of the grid

**Expected behavior**
Virtual scroll loads appropriate data in the appropriate location

**Minimal reproduction of the problem with instructions**
Go to the demo site and the issue is visible there
https://www.primefaces.org/primeng/#/table/scroll

**Please tell us about your environment:**

* **PrimeNG version:** 6.0.1

damatri:
woahh I was trying to get virtual scroll to work, and it took me so long to figure out the problem was not with my code. 

Lazyloading is working correctly but as said the rows are not shown within the view, making virtual scroll unusable

codyb11989:
https://github.com/primefaces/primeng/blob/master/src/app/components/table/table.ts#L1768

This should be giving the grid a class of `ui-table-virtual-table` which would make the table position absolute but this class is not getting added.

damatri:
ah you are right! It was as simple as that. When you add:
```
.ui-table-scrollable-body-table{
  position:absolute;
}
```
it seems to work.  It is a nice quickfix I guess.

wolves:
I have created a PR to fix this - https://github.com/primefaces/primeng/pull/6163

PR #6163, opened 2018-07-26
Fixed virtual scroll with lazy loading visibly jumping on data change

wolves:
### Defect Fixes
Currently the TurboTable when configured to use Lazy Loading and Virtual Scrolling has a visual scroll position jump when new row data is loaded in. This is caused because the scrollTableViewChild "top" styling calculation done [here](https://github.com/primefaces/primeng/blob/master/src/app/components/table/table.ts#L1955) cannot be applied since that element does not have the `position: absolute;` css applied to it currently. This PR fixes that.

**Issues Addressed**
https://github.com/primefaces/primeng/issues/6098

### DIFF
--- a/src/app/components/table/table.css
+++ b/src/app/components/table/table.css
@@ -70,7 +70,6 @@
     border-top: 0 none;
 }
 
-.ui-table-scrollable-body-table,
 .ui-table-virtual-table {
     position: absolute;
 }
--- a/src/app/components/table/table.css
+++ b/src/app/components/table/table.css
@@ -70,6 +70,7 @@
     border-top: 0 none;
 }
 
+.ui-table-scrollable-body-table,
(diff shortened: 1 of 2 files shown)


codyb11989:
Is this a good fix? The ` .ui-table-virtual-table` class already exists and should be getting appended to the table class list [here](https://github.com/primefaces/primeng/blob/master/src/app/components/table/table.ts#L1768). It's just not getting added properly.

By using `.ui-table-scrollable-body-table` won't that break other scrolling features?

wolves:
That is a great point. I will take a deeper look today and see if I am able to resolve the issue by addressing the virtual-table class not being added properly rather than via the added css rule.

wolves:
Nvm looks like you addressed the root issue. Thank you for this! 🤘

PR #6179, opened 2018-07-29
Fixed #4397

### LINKED ISSUE #4397: Autocomplete - paste via mouse doesn't work
To reproduce: 
Go to the autocomplete demo page (https://www.primefaces.org/primeng/#/autocomplete) 
Type an "a" in any of the autocomplete boxes. This works and the dropdown will show. Now copy the "a" and delete it (or cut it). 
Right click and choose "Paste" - the "a" appears but the dropdown doesn't open. 

If you paste via the keyboard the dropdown opens correctly.

This sounds very similar to a primefaces autocomplete bug: https://github.com/primefaces/primefaces/issues/831 which has apparently been fixed. Also somewhat similar to a primeng bug in the spinner: https://github.com/primefaces/primeng/issues/3970 . If it's appearing in two different components then it's very possible that it occurs in additional places.

sswebcoder:
Fixed for paste and cut events. Tested in Google Chrome Version 67.0.3396.99.


sswebcoder:
Maybe I should create event emitter for cut/paste events like blur?

cagataycivici:
The PR had no effect in my tests sorry.

sswebcoder:
Please tell me, how I can reproduce your test for check solution?

Issue #6189, opened 2018-07-31
MenuBar controlling hide delay


mcherb:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
When mouving the mouse from the first level of the menu to second level, i want to do it in diagonal maner. Indeed, when the mouse mouves horizontally, the second level is still open and i can mouve the mouse over it, but the second level is hidden when i mouve from the first level to the second or third option in the second level.

Ex: In the showcast https://www.primefaces.org/primeng/#/menubar 
This works: With AutoDisplay > Help > Search > Text
This wont work: With AutoDisplay > Help > Search > File 

**Expected behavior**
Choose any option in the second level whitout hidding the menu.

**What is the motivation / use case for changing the behavior?**
Ease navigation on menubar

**Please tell us about your environment:**
Linux, intellij idea

* **Angular version:** 5.2.X

* **PrimeNG version:** 5.2.X

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
 
* **Language:**  TypeScript 2.6.X

* **Node (for AoT issues):** `node --version` =   v6.11.3

mcherb:
I don't know if the request become a bug report !

cagataycivici:
The new menubar will not close on mouseout so this issue will be obselete when the new one is ready.

PR #6239, opened 2018-08-05
[#6189] MenuBar controlling hide delay

mcherb:
The problem is not the hide delay on it self but when we enter an active element it become's inactive (because of the removed condition).

The activated element when hovering is the last one who get's the mouse.


cagataycivici:
Issue is not clear, please provide a stackblitz template or detailed descriptions first next time.

mcherb:
the problem is related to https://github.com/primefaces/primeng/issues/6189, it's reported as a feature but in fact it's a bug.

PR #6248, opened 2018-08-06
Update galleria.ts

Durga456:
Customized Image width and height by adding width & height properties to "Images"

###Defect Fixes

###Feature Requests


cagataycivici:
Don't think we should give the size of panel to the images.

PR #6264, opened 2018-08-07
Enable sorting of filtered items in DataView

mlevinem6:
This enables the sorting of filtered items on the DataView component.

###Defect Fixes
6263


cagataycivici:
We better fix it the way table does.

Issue #6246, opened 2018-08-06
Overlay dialog not  maximize without footer template


hidrogenos:
[X ] bug report 

**Current behavior**

when i try to maximizable a dialog without footer template "<p-footer>", it doesn't work generate this error:
![image](https://user-images.githubusercontent.com/4935507/43729884-4903d31a-996f-11e8-8a7d-d317151e907e.png)

And rdirect to "localhost:4200/#" 

**Expected behavior**
the overlay dialog must be maximizable with or without footer template.

* **Angular version:**  6.0.4

* **PrimeNG version:**  6.1.0

* **Browser:**  all

anmoljainatcs:
Its still issue with PrimeNG 6.1.0. Please let me know when this fix be available.

PR #6274, opened 2018-08-09
#6246 fixed

yigitfindikli:
"Overlay dialog not maximize without footer template." fixed.


cagataycivici:
const cannot be defined again so unable to accept.

Issue #6206, opened 2018-08-01
P-Calendar doesn't update Navigator fields when date is changed by text input or ngModel


scostagabriel:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
https://www.primefaces.org/primeng/#/calendar

**Current behavior**
When the date is changed by ngModel or typing the date the month and year of Navigator doesn't reflect the changes. 

**Expected behavior**
The month and year of the navigator should be changing as the user types another month or year in the date

**Minimal reproduction of the problem with instructions**
Click on the input field of the P-Calendar that contains a Navigator (Navigators example)
Select any day in the popup
Change the month by the input field (not clicking in the popup)
Change the year by the input field (not clicking in the popup)
The Month and Year displayed on the popup of the calendar doesn't update to the value in the input field.

**What is the motivation / use case for changing the behavior?**
Not an expected behavior

**Please tell us about your environment:**
All OSes

* **Angular version:** 6.0.5

* **PrimeNG version:** 6.0.1

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
Tested on Chrome
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

* **Node (for AoT issues):** `node --version` =

JacobSiegle:
Looks fixed in latest version.

adevkota:
I still see the issue with version 6.1.0, on the demo site itself (https://www.primefaces.org/primeng/#/calendar)

The model associated with the date is updated, but the year and month pickers do not change on manual entry or when the model is changed by a parent component.

I think the reason is that the updateUI method in calendar.ts file (https://github.com/primefaces/primeng/blame/master/src/app/components/calendar/calendar.ts) does not update currentMonth and currentYear. 

I can create a PR for the fix if you can point me to what steps I need to take before submitting the PR.

adevkota:
created a PR, however the master branch itself has failing tests which causes CI job to fail.
https://github.com/primefaces/primeng/pull/6291

[...]

PR #6291, opened 2018-08-10
Update month and year dropdown on "navigator" view (issue #6206)

adevkota:
This addresses #6206 by updating the currentMonth and currentYear variables inside updateUI method.


cagataycivici:
See my comment at issue please.

PR #6351, opened 2018-08-18
Perform the export taking into consideration subfields of the objects in Turbotable

RicardoZomb:
The method is modified so that at the time of exporting the objects shown in the table, subfields of the object are taken, since currently the table does show subfields of matched objects, but when exporting those subfields are not taken.

###Defect Fixes
Object properties are not exported when accessing a subfield of the object

###Feature Requests
The method is modified to validate if what is found to export is an object, if it is true, access the property of the subfield, exporting the information correctly.

> Structure of the columns for the table
structure of information object

> With the native method, if it shows the subfields in the table correctly, but when exporting it is displayed as in the attached image.
view

> Subfields are displayed incorrectly, when exported with native method.
native method

> With the proposed method, this error is solved.
correctly form

Thanks for your time! Greetings.

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -1365,7 +1365,14 @@ export class Table implements OnInit, AfterContentInit, BlockableUI {
             for (let i = 0; i < this.columns.length; i++) {
                 let column = this.columns[i];
                 if (column.exportable !== false && column.field) {
-                    let cellData = this.objectUtils.resolveFieldData(record, column.field);
+                    let data = record;
+                    let field = column.field;
+                    if (typeof record[field] == 'object' ) {
+                        data = record[field];
+                        field = column.subfield;
+                    }
+
+                    let cellData = this.objectUtils.resolveFieldData(data, field);
                     
                     if (cellData != null) {
                         if (this.exportFunction) {


cagataycivici:
Provide a stackblitz template that demonstrates the problem first.

RicardoZomb:
Currently there is an issue with stackblitz because there is an error when importing primeng, which does not allow to visualize some components but it allowed me to reproduce the aforementioned error.
[Issue StackBlitz](https://github.com/stackblitz/core/issues/300)

Next I leave the link to visualize the commented error at the moment of exporting, for the previous issue the export button is not displayed correctly, but if it is clicked if it works.

https://stackblitz.com/edit/angular-g8xmcl

Again, I thank you for your attention and I hope for an answer, Thank you.

cagataycivici:
But this adds a new subfield property, why not use field: 'building.name' and use dot operator to define your field e.g. 'capacityCode.name'

RicardoZomb:
https://stackblitz.com/edit/angular-g8xmcl


The change is made as you comment and in fact if you export in the correct way, but the information in the table is not displayed, apparently it is not supporting properties of an object to show the information.

Stackblitz is appended for exemplification

RicardoZomb:
https://stackblitz.com/edit/angular-g8xmcl

The change is made as you comment and in fact if you export in the correct way, but the information in the table is not displayed, apparently it is not supporting properties of an object to show the information.

Stackblitz is appended for exemplification @cagataycivici 

Thank's for your attention.

PR #6363, opened 2018-08-20
Add optional possibility to use reorderable column handle

vadamk:
Add optional possibility to specify specify which element is used to initiate the dragging for columns ( same as for rows, but optional )

###Defect Fixes

###Feature Requests


cagataycivici:
If there is more demand, we'll reconsider, no plans at the moment. Thank you.

PR #6369, opened 2018-08-21
Update dropdown.ts

### LINKED ISSUE #6368: Disabled option(s) in p-dropdown are selectable
```
```
**Current behavior**
In the p-dropdown component a disabled option is still selectable by clicking on it (the disabled is only a visual effect).

**Expected behavior**
When an option if the p-dropdown is marked as disabled it shouldn't be selectable, i.e. a click on the item should have no effect at all.

Create a module with a p-dropdown using SelectItem array with one option disabled. The disabled option is selected by clicking on it.
https://stackblitz.com/edit/github-jc5fhw

When an option is disabled the user shouldn't be able to select it, otherwise I still have to check that the option is disabled and give a visual feedback that the selected option isn't correct.

* **Angular version:** 6.1.3
* **PrimeNG version:** 6.1.2
* **Browser:** all
* **Language:** all

tfd:
fix #6368

Simply ignore clicks on disabled options.


cagataycivici:
Unable to accept a PR with 14 unrelated file changes.

PR #6389, opened 2018-08-24
Fix #6334

### LINKED ISSUE #6334: Password component throws an error when handling keyup events and feedback is false
```
[ X ] bug report => Search github for a similar issue or PR before submitting
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/angular-datnyq?embed=1&file=src/app/app.component.html

**Current behavior**
When you set `feedback` to `false` on a `pPassword` component and try to handle `keyup` events, the application throws the following error:

imagen

It seems that the component tries to access to some properties of the password strength helper, but it can't because it doesn't exist.

**Expected behavior**
It should manage internally whether the feedback component exists or not, to avoid accessing it in case that `feedback` is `false`.

To reproduce the problem, just drop a `pPassword` component with `[feedback] = false` and bind a `keyup` event handler. For example:

`<input name="password" type="password" pPassword [feedback]="false" (keyup)="checkForm($event)" [(ngModel)]="pass"/>`

Kubuntu 18.04 64 bits, NPM 6.4.0

* **Angular version:** 5.X

Angular: 6.1.3

* **PrimeNG version:** 5.X

PrimeNG 6.1.2

I tested it on Chrome and Firefox and it's the same on both.
 
TypeScript

Node v10.6.0

skormel:
###Defect Fixes
#6334 

###Feature Requests


skormel:
Why travis failed?

PR #6404, opened 2018-08-28
Fixed a bug when navforward (or navbackward) was clicked on "Month" view

kafo85:
Fixed a bug when navforward (or navbackward) was clicked on "Month" view


kafo85:
When the icon back or next is clicked on the calendar with "Month" view mode. it have no effect but it work with an another mode

in fact, le refresh of the view is not done.

cagataycivici:
Sorry, I can't replicate any problem with current code. Please create an issue ticket with a stackblitz template to demonstrate the case. Thank you.

Issue #5971, opened 2018-06-27
Disable animations or allow customization of animations


ksdev-pl:
**I'm submitting a ...** 
[x] feature request

**Current behavior**
Animations are hard-coded (`400ms ease-in` etc).

**Expected behavior**
There are no animations or they are customizable. I'm referring specifically to the `p-calendar` component.

nfdevil:
This should indeed be an input property on the component: 'disableAnimations'

rafa-as:
+1 both: customization and disable

mialy:
_Maybe_ a work-around for disabling animations with PrimeNG 6.1.2:

- In .ts file add ```@HostBinding('@.disabled') disabled = true;``` line to child components; adding it to parent component will disable animations on all child components too
- In .html (template) file add ```[@.disabled]="true"``` to child component, like ```<p-calendar [@.disabled]="true" ...></p-calendar>```

cagataycivici:
https://github.com/primefaces/primeng/issues/6652

PR #6432, opened 2018-08-31
Add option to disable animations on single components

iprignano:
### Defect Fixes
https://github.com/primefaces/primeng/issues/5971

### Description

As per the linked issue, I'd like to implement an option for disabling animations selectively.

**This is just an example of what I came up with to solve the problem.** If maintainers think it's viable, I'll extend the change to all other components that implement animations and update the PR.

Example usage: 

```html
<p-dialog [animate]="false">
  This will not animate.
</p-dialog>

<p-dialog>
  This will animate.
</p-dialog>
```


maorcc:
> transitionDuration="0"

@cagataycivici ,  Can you please give more details where and how to set the transitionDuration?

tomsoja978:
> We're handling it different to make it an option where users can change the speed-easing of the animation, setting it to 0 disables it e.g.
> 
> transitionDuration="0"

I would also like to use this option but don't know where to set it either. Please help

PR #6435, opened 2018-08-31
calendar multi date or range picked time changes not applying to the …

### LINKED ISSUE #6416: TimePicker (hh:mm:ss) in p-calendar is non editable.
**I'm submitting a ...**  
```
[ ] bug report 
[x] feature request 
[ ] support request 

Plunkr Case :
[https://github-eewso5-xiqsoh.stackblitz.io](https://github-eewso5-xiqsoh.stackblitz.io)

**Current behavior**
The time displayed in timepicker component is not editable .

**Expected behavior**
The user can only increase /decrease the hour,minute and seconds value in the timepicker component
using only the up/down arrows keys.This is really troublesome for the user. It would be great if
these  fields are editable.

Customer request.

**Angular version:** 4.0

**PrimeNG version:** PrimeNG 6.1.0

**Browser:** [ Chrome, Firefox, IE  , Safari  ]
 
**Language:** [TypeScript 3.0.1]

volkanakbayir:
Defect Fixes
#6416

Description
When selectionMode is range and showTime enabled, changes to the time with mouse interaction will not be applied to the backing model which will lead inconsistent behaviour

To Reproduce:
 <p-calendar
   selectionMode="range" 
   [appendTo]="'body' 
   [showTime]="true"     
   [(ngModel)]="value"></p-calendar>

### DIFF
--- a/src/app/components/calendar/calendar.ts
+++ b/src/app/components/calendar/calendar.ts
@@ -1435,7 +1435,7 @@ export class Calendar implements OnInit,OnDestroy,ControlValueAccessor {
     updateUI() {
         let val = this.value||this.defaultDate||new Date();
         if (Array.isArray(val)){
-            val = val[0];
+            val = val[val.length -1];
         }
 
         this.currentMonth = val.getMonth();


cagataycivici:
Range + TimePicker is generally not a supported combination

PR #6449, opened 2018-09-04
Allow passing of per-chart plugins

SPKorhonen:
Chart.js allows (since version 2.5) per chart plugins but Primeng does no allow passign of this parameter. This PR adds input to chart component to allow passing of this parameter to underlying Chart.js object.


cagataycivici:
Nice, thank you.

Issue #6448, opened 2018-09-04
MessageService does not refresh the component when called by Angular


Wanchai:
**I'm submitting a ...**  
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**

https://github-primengmsg.stackblitz.io

**Current behavior**
I have an error handler that calls the MessageService to display messages. 
It works on some errors, if I call _throw new Error('Test');_ for instance ...
But if it's a navigation error or http call, the message component is not refreshed and the message not displayed.

**Expected behavior**
The message is displayed each time MessageService.add(...) is called, no matter what.

**Minimal reproduction of the problem with instructions**

https://github-primengmsg.stackblitz.io

On my case, you click the "HTTP call" button THEN "Message" to see both.

**What is the motivation / use case for changing the behavior?**
The purpose is to have an error handler who can send messages from every kind of errors.

I tried to use ChangeDetectorRef.detectChanges() but it creates new problems (delays when you close the message)

I guess it has to do with some kind of UI thread and HTTP calls are not part of it.

**Please tell us about your environment:**
- win10
- Visual Studio Code
- MessageService
- MessageModule
- MessagesModule
- ToastModule

* **Angular version:** 6.0.3

* **PrimeNG version:** 6.1.3

* **Browser:** 
Chrome,
Firefox
 
* **Language:** 
Typescript 2.7.2

cagataycivici:
I'm unable to replicate with the latest PrimeNG. If the issue still persists with the new version please create a new issue with our stackblitz template.

PR #6450, opened 2018-09-04
MessageService does not refresh the component when called by Angular

Wanchai:
###Defect Fixes
When called programmatically, the MessageService Observable list is not represented in the message or toast component.

Issue #6448


cagataycivici:
Not sure if using ChangeDetector directly here is the best way as it may cause unnecessary cd cycles as well, if there is more request with the same problem, we'll reconsider. We'll be monitoring 6448.

PR #6451, opened 2018-09-04
fix(inputswitch): trigger change in writeValue #6434

### LINKED ISSUE #6434: InputSwitch not firing change detection on external value change with Push Strategy
```
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-kxjjv5?file=src%2Fapp%2Fapp.component.ts

**Current behavior**
The `writeValue` method does not fire change detection.
When using ChangeDetectionStrategy.OnPush value gets set later and change doesn't propagate.

**Expected behavior**
The component should work same with and without the OnPush change detection strategy.

When you disable ChangeDetectionStrategy.OnPush it works, but with change detection on UI doesn't refresh, because `writeValue` doesn't trigger the change event.

**What is the motivation/use case for changing the behavior?**
InputSwitch used within components with OnPush strategy do not get value properly initialized. This is because the method `writeValue` is called after the component has been drawn and initialized, and change is never propagated.

Linux Ubuntu, Yarn, Webstorm

* **Angular version:** 5.X
6.1.1

* **PrimeNG version:** 5.X
6.1.3

* **Browser:** all
 
* **Language:** all

meeroslav:
Fixes #6434

### DIFF
--- a/src/app/components/inputswitch/inputswitch.ts
+++ b/src/app/components/inputswitch/inputswitch.ts
@@ -81,8 +81,9 @@ export class InputSwitch implements ControlValueAccessor {
         this.onModelTouched();
     }
 
-    writeValue(checked: any) : void {
+    writeValue(checked: any): void {
         this.checked = checked;
+        this.onModelChange(this.checked);
     }
 
     registerOnChange(fn: Function): void {


cagataycivici:
The change will write the value back which is not ideal, instead we need to use ChangeDetectorRef.

PR #6455, opened 2018-09-05
Add stopPropagation to pRowToggler

FrEaKmAn:
This pull requests adds stopPropagation to pRowToggler. We have expandable row and use this directive, while we have custom events for table row. We could manually add stop propagation

```
(click)="$event.stopPropagation()"
```
but I find this solution more DRY. Also this is backward compatible.

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -2403,6 +2403,8 @@ export class RowToggler {
 
     @Input() pRowTogglerDisabled: boolean;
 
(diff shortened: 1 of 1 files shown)


cagataycivici:
But suppose there is an open dropdown somewhere on the page, clicking row expand will prevent the dropdown panel to be closed. Instead custom events in the table should check if the event target is an expandable row toggler or not instead. If you can provide a stackblitz template, we can discuss further, please create a new issue ticket with the sample case.

PR #6459, opened 2018-09-06
ContextMenu - Trigger event

Nawsen:
Make it possible for the user to select for which event the context menu should be displayed.


cagataycivici:
Great, thank you.

andreeapNRO:
@cagataycivici can you let me know in which version this will be fixed?

Issue #6442, opened 2018-09-03
Dialog/OverlayPanel onHide triggered when destroying an hidden dialog


Silentscripter:
**I'm submitting a ...**  (check one with "x")
```
[X ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**

https://angular-3fwuct.stackblitz.io

**Current behavior**
The onHide event in p-dialog is always triggered when the p-dialog component is destroyed, whether the dialog is opened or not.

**Expected behavior**
The onHide event should trigger only when an open dialog is closed or destroyed. This was working normally in primeng v.5

**Minimal reproduction of the problem with instructions**
1. Add a p-dialog with an onHide callback associated.
2. Destroy the dialog container, whether by ngIf or by routing to another page
3. You should notice that the callback is executed even if the dialog has never been opened.
https://angular-3fwuct.stackblitz.io

**Please tell us about your environment:**
Windows 10 x64

* **Angular version:** 6.1.4

* **PrimeNG version:** 6.1.3

* **Browser:**all

cagataycivici:
PR merged, thank you.

PR #6471, opened 2018-09-07
Update overlayPanel and dialog. Emit onHide only if panel/dialog is visible.

Hunroll:
###Defect Fixes
We have a panel on layout, subscription to onHide. Once we navigate to another form, onDestroy is triggered and then onHide is emmited, but it was not supposed to as dialog was not even shown.

###Feature Requests
N/A

### DIFF
--- a/src/app/components/dialog/dialog.ts
+++ b/src/app/components/dialog/dialog.ts
@@ -598,7 +598,9 @@ export class Dialog implements OnDestroy {
     }
 
     onOverlayHide() {
-        this.onHide.emit({});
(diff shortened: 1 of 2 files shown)


Hunroll:
Was reported in issue #6442

cagataycivici:
Thank you.

huineng:
the variable isvisible is already set to false before onOverlayHide is being called, so this never get's executed

Hunroll:
Thank you! Created new PR to fix both problems.

PR #6481, opened 2018-09-08
Added the ability to configure icon classes for a nodes when the component is called.

vzx7:
When I was creating a Tree on the server with a large number of nodes a problem appeared. To avoid the problem, I had to add the ability to configure icon classes for a node directly into the template when the component is calledIn my opinion, it will separate the logic of the client from the logic of the server, and also it will help to get rid of duplicated properties of classes for a node in the tree data, during nodes creation.
At the same time I could preserve the ability to generate data on the server, including classes for a node. It means that you can keep transmitting data as you used to do before...
But now you can send data like this:
```
{
    "data":
    [
        {
            "label": "Documents",
            "data": "Documents Folder",
            "children": [{
                    "label": "Work",
                    "data": "Work Folder",
                    "children": [{"label": "Expenses.doc", "data": "Expenses Document"}, {"label": "Resume.doc", "data": "Resume Document"}]
                },
                {
                    "label": "Home",
                    "data": "Home Folder",
                    "children": [{"label": "Invoices.txt", "data": "Invoices for this month"}]
                }]
        }
    ]
}
// call
<p-tree 
    [value]="filesTree12"
    [classEndNodeIcon]="'pi pi-file'"  
    [classExpandedIcon]="'pi pi-chevron-circle-down'" 
    [classCollapsedIcon]="'pi pi-chevron-circle-right'" 
    ></p-tree>
```

### DIFF
--- a/src/app/components/tree/tree.ts
+++ b/src/app/components/tree/tree.ts
@@ -24,7 +24,7 @@ import {BlockableUI} from '../common/blockableui';
                     ><div class="ui-chkbox" *ngIf="tree.selectionMode == 'checkbox' && node.selectable !== false"><div class="ui-chkbox-box ui-widget ui-corner-all ui-state-default">
                         <span class="ui-chkbox-icon ui-clickable pi"
                             [ngClass]="{'pi-check':isSelected(),'pi-minus':node.partialSelected}"></span></div></div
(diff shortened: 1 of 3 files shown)


cagataycivici:
I'm sorry, new feature requests should be implemented by our team, we only accept defect fixes in PRs. Please create an enhancement request instead and we'll review.

PR #6485, opened 2018-09-10
Add textarea to the exception list on draggin

vadamk:
Add possibility to use TEXTAREA instead INPUT when we use drag&drop for column transfer. Now, when we try to highlighted text in the TEXTAREA, the column moves.

There are cases where we can not use INPUT.

###Defect Fixes

###Feature Requests


cagataycivici:
Formatting is broken so will merge manually.

PR #6516, opened 2018-09-18
Perform the export & data visualization taking into consideration subfields of the objects in Turbotable 

RicardoZomb:
https://stackblitz.com/edit/angular-g8xmcl

The change is made as you comment and in fact if you export in the correct way, but the information in the table is not displayed, apparently it is not supporting properties of an object to show the information.

Stackblitz is appended for exemplification

###Defect Fixes
The properties of the objects are not shown in the table

###Feature Requests
The method is modified to validate if what is found to export is an object, if it is true, access the property of the subfield, exporting the information correctly.
Similarly, it can be handled with the property of the subfield mentioned above showing the data correctly.

This issue refers to the closed issue #6351 , please take into consideration the new information provided for the handling of the data.

Thank you @cagataycivici


cagataycivici:
You have full control over presentation;

```
<td *ngFor="let col of columns">
             {{rowData[col.field]}}
            </td>
```

Instead of using above, you can resolve multiple fields on your side, this is not Table related.

PR #6529, opened 2018-09-20
fixed #6520

### LINKED ISSUE #6520: onHide event not triggered in version 6.1.4 for dialog and overlayPanel
```
```
Current behavior**
The onHide event is NOT triggered anymore even if the panel/dialog is visible

**Expected behavior**
The onHide event should be  triggered as before.

This simple code can be used tor reproduce the problem:

```
<p-overlayPanel [dismissable]="true" [showCloseIcon]="true" #op [style]="{'width':'320px'}" appendTo="body" (onHide)="hideTest()">
    <h1>CIAO</h1>
</p-overlayPanel>

<button type="text" pButton label="Basic" (click)="op.toggle($event)"></button>
```

and, of course, a hideTest() method should be defined in the Component.

Now click on the button and then close the OveralyPanel... the hideTest method is not called because the onHide event is not emitted.

yigitfindikli:
if statement changed in overlaypanel.ts and dialog.ts.


cagataycivici:
This PR causes regressions so closing.

PR #6534, opened 2018-09-20
Fix #6520 - overlayPanel's and dialog's onHide

### LINKED ISSUE #6520: onHide event not triggered in version 6.1.4 for dialog and overlayPanel

Hunroll:
###Defect Fixes
Fix #6520 - overlayPanel's and dialog's onHide. Revert previous changes (PR #6471) and suspend calling onOverlayHide event in destructor if overlay is not visible.

###Feature Requests
N/A


coffeetea-synchestra:
Please merge it, anybody ;(

cagataycivici:
We'll handle it differently, I don't think it makes sense to call onHide at ondestroy e.g. route change.

igor04:
@cagataycivici any changes about `onHide`?

because it looks like `onHide` event is not triggered and there is no references to other issues/pr's

rafa-as:
I think that makes sense to call onHide on destroy because sometimes you can have an ngIf in a component that have a dialog inside and if have some logic when de dialog disappears in this case this logic will not fired.

Please reconsidere this pull request @cagataycivici

PR #6567, opened 2018-09-25
Better Schedule Change Detection

### LINKED ISSUE #6386: Changing p-schedule option causes no refresh
**I'm submitting a ...
```
```

**Current behavior**
I'd like to set week's first date dynamically in a dropdown, because some user find default value unfit.
Since Schedule has no firstDate property, I set it in options:
 [options]="{'firstDay': this.config.calendarFirstDay}"
After I change dropdown value, the schedule is not refreshed, first day remains the same.

**Expected behavior**
Changing the first day of week the schedule displays calendar according to the selected value.
I think schedule should have firstDay property to be easy configurable, or the ngChange() method should listen to the option values too.

noamichael:
### Defect Fixes
#6386

### Description
Previously, the `<p-schedule/>` component:

1. Only checked the settings found in the `options` attribute on initialization
2. Only checked any other settings changes *if the calendar was initialized*

So, if someone dynamically changed their `options` object, the calendar would not reflect it. Further, if the calendar was not in view, changes would not be detected. The following now works:

```typescript
@Component({
    template: `<p-schedule [events]="events" [options]="options"></p-schedule>`
})
export class MyComponent {
    options: any = {editable: false}

    ngOnInit(){
        setTimeout(() => this.options = {editable: true}, 5000)
    }
}

```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,12047 @@
+{
+    "name": "primeng",
+    "version": "6.1.5-SNAPSHOT",
+    "lockfileVersion": 1,
+    "requires": true,
+    "dependencies": {
+        "@angular-devkit/architect": {
+            "version": "0.7.1",
+            "resolved": "https://registry.npmjs.org/@angular-devkit/architect/-/architect-0.7.1.tgz",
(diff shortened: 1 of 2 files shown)


cagataycivici:
Please remove the lock json.

noamichael:
@cagataycivici Done!

SteamWind:
Nice, where will it be merged? Is it released?

Issue #6419, opened 2018-08-29
Strong Types for Turbo Table


MikeMatusz:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

Current behavior
A lot of the API for TurboTable uses "any" as the type for arguments, return values, and even column definitions.  Not all properties/methods are documented, and those that are could potentially become out of date.  Options could also change when updating to a newer version, and currently there would be no compile-time checking around this.

Expected behavior
Ideally, everything would have interfaces or other strong types, and "any" would be rarely aside from the row data itself, which should allow for anything.
Here are some specific cases I've encountered that have forced me to dig into source code, along with some low-hanging fruit encountered while looking through the source:
* Table.columns input property
** An interface that specified properties like header, field, exportable would be very helpful
* Table.filter method
* Table.contextMenu
** Presumable this has to be an instance of p-context-menu, but if you don't want to limit it to that, it definitely needs be something that implements a show() function.  Either should be typed as ContextMenu or some IContextMenu { show(event: Event): void; }
* LazyLoadMetadata (Table.onLazyLoad)
** Typing this event emitter would be provide intellisense when you go to handle the event. The createLazyLoadMetadata function uses a very specific contract.
* Table.exportCSV(options?: any)
** I think currently the only option is selectionOnly, but it would be nice to make that explicit.  And if you add additional options in the future, they'll become immediately apparent.
* Table.columnResizeMode is typed as string, but changing it to be 'fit' | 'expand' (or an enum) would better validate input and provide intellisense.
* Other event emitters that would benefit:
** onPage, sortFunction, onSort, selectionChanged (Array), onRowSelect, onRowUnselect, onContextMenuSelect, onHeaderCheckboxToggle, onFilter, onRowCollapse, onRowExpand, onColResize, onColReorder, onRowReorder, onEditInit, onEditComplete

What is the motivation / use case for changing the behavior?
Better IntelliSense while developing. Compile time error reporting for many possible mistakes. Assurance when you upgrade to a new version whether contracts have changed.  Potentially less defects created when changing the code.

PrimeNG version: 6.1.2 (still present in 14.2.1)

Language: TypeScript

ciesielskico:
[...]

PR #6576, opened 2018-09-26
Adding stronger types to Table component

MikeMatusz:
###Defect Fixes
Supplies types to satisfy #6419 

Please also let me know if you'd like me to split the types/interfaces out into separate files (and where they should go, if so), or if there are any other guidelines you'd like me to conform to.


cagataycivici:
Thanks, I seem some apis that we already have and not sure about the location, we'll do it on our side based on your PR. Thank you.

Issue #4985, opened 2018-01-26
Keyboard accessibility for MultiSelect Header Checkbox


skhattry:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Tab doesn't stop at the select all checkbox in p-multiselect.

**Expected behavior**
Select all checkbox in p-multiselect should be reachable via keyboard

**What is the motivation / use case for changing the behavior?**
This will improve the control's accessibility

**Please tell us about your environment:**
* **Angular version:** 2.0.X
4.1.3

* **PrimeNG version:** 2.0.X
4.2.1

* **Browser:** 
Tested on Microsoft Edge and Mozilla Firefox

leonliang33:
I am also blocked on this issue, is there plans for this fix soon?

cagataycivici:
https://github.com/primefaces/primeng/issues/6025

PR #6612, opened 2018-10-02
Add keyboard navigation for multiselect component.

minkowalentin:
1. Focus on multiselect by tab-ing to the component .
2. Alt + Down expands list.
3. Tab until a list item is selected
4. User can now navigate using UP and DOWN keys, select/deselect using SPACE and collapse dropdown using ESC. User can expand component again by pressing Alt+Down.

Also added 'nativeElement.scrollIntoView' so that checkboxes are always displayed when user navigates with arrow keys

###Reported issue on subject:  #4985

###Defect Fixes
Previously pressing Tab collapsed dropdwon and navigating trough list items was not possible

###Feature Requests
-- I don't think its a significant change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,12226 @@
+{
+    "name": "primeng",
+    "version": "6.1.5-SNAPSHOT",
+    "lockfileVersion": 1,
+    "requires": true,
+    "dependencies": {
+        "@angular-devkit/architect": {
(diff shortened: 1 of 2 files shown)


cagataycivici:
There are too many changes for us to merge. We'll implement it ourselves for the next release. Thank you.

PR #6627, opened 2018-10-03
Make frozen width work dynamically

### LINKED ISSUE #5617: Non-frozen columns don't move or resize when frozen columns' width changes
```
```

**Plunkr Case (Bug Reports)**

http://plnkr.co/edit/7Mns2fJdQ111mqHMVMS4?p=preview

**Current behavior**
When you have a table with some frozen columns, the div containing the second table (unfrozen) has the "left" property calculated once.  The left property doesn't get updated if the frozen columns' width changes.  (**EDIT:** The div containing the second table also needs to have its width updated to calc(100% - [frozenWidth])) 

Thus, in this Plunk, changing the window size results in situations like these: 

frozen_demo_2
frozen_demo_3

**Expected behavior**
The second table should always appear directly next to the first table, even if the first table's width changes.  (This would most likely be accomplished by having "left" updated.)
**EDIT:** The div containing the second table also needs to have its width updated to calc(100% - [frozenWidth])

* **Browser:** Chrome, Firefox

thedanielgray:
The currently rejected PR had a very large and complex fix. I have created a much simpler one that also works, by removing it from the ngAfterViewInit() and moving it straight to the attributes, which makes changes in the width trigger a re-render.

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -86,7 +86,7 @@ export class TableService {
 
             <div class="ui-table-scrollable-wrapper" *ngIf="scrollable">
                <div class="ui-table-scrollable-view ui-table-frozen-view" *ngIf="frozenColumns||frozenBodyTemplate" [pScrollableView]="frozenColumns" [frozen]="true" [ngStyle]="{width: frozenWidth}" [scrollHeight]="scrollHeight"></div>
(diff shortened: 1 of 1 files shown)


cagataycivici:
I was afraid that it may broke regular scrollable tables with no frozen columns but style does not get applied since I guess frozenWidth is undefined so works fine.

thedanielgray:
Thanks a lot!! I just realized that I hadn't linked it to the issue it fixes. Sorry!!

This fixes issue #5617 .

PR #6639, opened 2018-10-05
OverlayPanel - Add onShow and onHide event Emiter

serweck:
###Defect Fixes

###Feature Requests


cagataycivici:
They are handled during animations at the moment which is the correct place right after change detection so dom is available. Due to this, not taking any action regarding this PR, thank you.

PR #6645, opened 2018-10-07
Added autocomplete scroll event

carlosearaujo:
###Feature Requests
Added Autocomplete onScroll(running outside change detection, to improve performance) and onScrollEnd(useful for infinite scroll) events.

*I don't add unbind because suggestions have a ngIf directive, so this don't are necessary.


cagataycivici:
Not sure if these two events are in demand at the moment so not taking action, if there is more user demand we'll reconsider, thank you.

PR #6667, opened 2018-10-11
adding RTL support on Column Resize

a28028:
###Defect Fixes
if current table is RTL fix Resize table column


cagataycivici:
We'll cover this one when we add built-in RTL support to all Table features, thank you.

Issue #6675, opened 2018-10-12
feat-request: Table Column Index for column toggle 


fuyucn:
Add Table Column index feature for column toggle.
Currently, the column toggle will reorder the index of each column in table, but i want them to always shows in order.
```
[ ] bug report => Search github for a similar issue or PR before submitting
[X ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

cagataycivici:
MultiSelect, the component used to implement column toggle demo passes index to its item template so you already have access to it in your item template.

fuyucn:
I think i mislead-ed, from my topic. Already fix the topic.
I just try to add index for table columns.
Ex: As in demo, if you uncheck `vin` field, and then check it. The `vin` column will show at end. I want it shows as in order.

PR #6676, opened 2018-10-12
[feat] Add column index for table

fuyucn:
feat added for issue #6675.
The `index` helps you order columns in table, even if you toggle it.

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -48,7 +48,12 @@ export class TableService {
         this.totalRecordsSource.next(value);
     }
 
+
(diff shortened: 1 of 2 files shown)


cagataycivici:
But this would enforce a new property like index, which will break existing apps as it tries to sort based on index. Better to handle it on the app side since there is no built-in functionality like column toggler, it is just a demo.

NicholasJonesCode:
@fuyucn Hey, did you ever figure out how to implement this from the project side? As in, implement this in an angular project rather than internally from the primeNG code?

Issue #6503, opened 2018-09-17
[Calendar] The onClose Event doesn't emit when TouchUI is on.


lhysrc:

cagataycivici:
7.0.0 should fix this one, the animation callback calls onClose regardless of how it is closed.

PR #6717, opened 2018-10-18
p-calendar: Emit onClose event everytime the calendar is closed

vedard:
Sometime the onClose event is not emited when you close the calendar:
- Using the touch ui
- Using the `<tab>` key
- Using the clear button

### Defect Fixes
- #6503


cagataycivici:
7.0.0 should fix this one, the animation callback calls onClose regardless of how it is closed.

PR #6719, opened 2018-10-19
 fixed #6712 - Carousel allow header templating

### LINKED ISSUE #6712: Carousel allow header templating
[ X ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
```

It would be convenient to have option for templating carousel header in case we need to add some buttons inside to perform specific actions like add sort etc. 

It would be great like this:
<p-carousel>
    <ng-template pTemplate="header">
    <!-- header content -->
    </ng-template>
</p-carousel>

or like this:
<p-carousel>
    <p-header>
    <!-- header content -->
    </p-header>
</p-carousel>

Also, you should provide ready to use navigation (something like DataView has <p-dataViewLayoutOptions> for selectiog between grid and list layout)

Win10, VS Code, npm

* **Angular version:** 5.X
6.1.8

* **PrimeNG version:** 5.X
6.1.4
 
Typescrypt 2.7.2

npm version 6.4.1
node version 8.12.0

yigitfindikli:
added header pTemplate.

### DIFF
--- a/src/app/components/carousel/carousel.ts
+++ b/src/app/components/carousel/carousel.ts
@@ -40,8 +40,6 @@ export class Carousel implements AfterViewChecked,AfterViewInit,OnDestroy{
     
     @Input() numVisible: number = 3;
 
-    @Input() firstVisible: number = 0;
-
     @Input() headerText: string;
 
     @Input() circular: boolean = false;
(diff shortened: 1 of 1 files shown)


cagataycivici:
Not merging as we'll do a new Carousel.

PR #6721, opened 2018-10-19
Editable field are not "blur" when clicking outside the field

yigitfindikli:


cagataycivici:
Will cause performance issues so not merging.

PR #6814, opened 2018-11-02
Update width to min-width to avoid wrapping

### LINKED ISSUE #6153: Pagination Page Numbers Wrapping After 100 Pages
```
```

**Current behavior**
Paginator page numbers are wrapping outside of the display box (.ui-paginator-page) when going into the triple digits.

**Expected behavior**
Page numbers should be fully encompassed by the paginator page borders.

Create any datatable with more than 100 pages of data. Once you get past the 100th page, you should see the page numbers starting to wrap outside of their parent containers.

Note that this **does not** happen for the page that's currently selected, although that simply hides the rest of the page number which isn't much better.

* **Angular version:** 6.0.3
* **PrimeNG version:** 6.0.0
* **Browser:** all

Steven-Harris:
### Defect Fixes
#6153

### DIFF
--- a/src/app/components/paginator/paginator.css
+++ b/src/app/components/paginator/paginator.css
@@ -28,7 +28,9 @@
 .ui-paginator .ui-paginator-prev,
 .ui-paginator .ui-paginator-current {
     display: inline-block;
-    min-width: 2em;
+    min-width: 1.5em;
+    padding-left: 5px;
+    padding-right: 5px;
     height: 1.5em;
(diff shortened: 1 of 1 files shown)


Steven-Harris:
the master build must be failing, I don't think my css could have broke anything. I am not seeing a linting problem either.

cagataycivici:
PR may break some themes due to padding, I'll just replace to min-width and let themes define the padding.

Issue #5676, opened 2018-05-03
TurboTable filter works only for the second time


MrBra1nwash:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Here is [example](https://stackblitz.com/edit/angular-hhqptu?file=app/hello.component.ts), but unfortunately i can't reproduce this issue in example. It is reproducible only in my project...

**Current behavior**
If i use the same code as in the stackblitz example table is filtering only for the second time. For example i entered "1" in input field - nothing happened. Then, i am press any other key - and table is filtering by previous value - "1" and so on.

**Expected behavior**
Table should be filtered after every inputted value

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**
Windows 10, IDE - VSCode, Webpack

* **Angular version:** 5.2.7

* **PrimeNG version:** 5.2.5

* **Browser:** [Chrome XX | Firefox XX | IE XX |]
Didn't try all browsers
 
* **Language:** TypeScript 2.7.2 | ES6/7]

JSeed:
If you can't reproduce the issue on stackblitz can you please provide us with some of your code?  
It will be hard for anyone to help you without any being able to see something

MarMic1:
Hi @MrBra1nwash.

I have the same problem.

The only difference between your example and my code is that I'm using:

`  changeDetection: ChangeDetectionStrategy.OnPush, `

If I add this line to your example, the example displays the issue.

Hope somebody can help us? 
@JSeed  maybe this is enough to help us
 👍

MarMic1:
should be fixed in #6831

cagataycivici:
Provide a stackblitz template please that demonstrate the issue.

MarMic1:
https://stackblitz.com/edit/angular-opcdqk

the table is in the hello component. 

The problem is that the filter is lagging or does not work every time. I hope you can help us now, after you have closed the commit #6831

PR #6831, opened 2018-11-07
updated turbotable to use ChangeDetectionStrategy onpush

MarMic1:
#5676 
solved like multiselect commit:  b46f36c58dc6a7924422de8350c584b8f6f00688


cagataycivici:
Provide a stackblitz template please that demonstrate the issue.

MarMic1:
https://stackblitz.com/edit/angular-opcdqk

the table is in the hello component.

The problem is that the filter is lagging or does not work every time. I hope you can help us now, after you have closed the commit #6831

@cagataycivici

PR #6839, opened 2018-11-07
Fixed #3261 - prevent contextmenu sub menu going out of view

### LINKED ISSUE #3261: Improve ContextMenu collision detection
```
```

**Current behavior**
Submenu is position properly only when it is opened on the right side of the viewport. It is displayed correctly on the left side then. (this was fixed in #1596)

**Expected behavior**
Could you please also fix the position of the submenu when it is open at the bottom of the page? It overflows the bottom of the viewport.
contextmenu-issue

Submenu should reposition itself on all sides of the viewport, not just when open near the right edge

* **Angular version:** 4.2.2

* **PrimeNG version:** 4.1.0

danieljones740:
Prevents contextmenu sub menu from going off screen if it's longer than the base menu.

See issue: [https://github.com/primefaces/primeng/issues/3261](https://github.com/primefaces/primeng/issues/3261).


danieljones740:
Based on the test output and other pull requests it I think the failed test is the result of a core change.

PR #6860, opened 2018-11-13
fixed #6859 - Pagination values mismatch after filtering

### LINKED ISSUE #6859: Pagination values mismatch after filtering
Reported by a PRO user;

>  We are using global filter and the pagination. We are displaying the number of records on the right side using the below code.
> 
> <ng-template pTemplate="paginatorright" let-state>
> state.first+1 - {{state.first + state.rows >
> rooms.length ? state.totalRecords : state.first + state.rows}} 'general-txt-of' | translate
> state.totalRecords
> </ng-template>
> 
> Output is displayed something like "1 - 25 of 941"
> 
> Whenever we type text in global filter the table filters displays the records based on the search output but the pagination numbers are not changing.
> 
> Let us know how to fix this issue.

yigitfindikli:


cagataycivici:
Better to move it to totalRecords setter;

```
    set totalRecords(val:number) {
        this._totalRecords = val;
        this.updatePageLinks();
        this.updatePaginatorState();
    }
```

Issue #6862, opened 2018-11-13
Tooltip - tooltipEvent support more than hover/focus events


justinotero:
## Tooltip - tooltipEvent support more than hover/focus

**I'm submitting a ...**
```
[   ] Bug report
[ x ] Feature request
[   ] Support request
```

Currently tooltipEvent supports only `hover(default) | focus`. Would be really helpful if it supported at minimum `click` also. I can imagine minimum support for the following events:

- `hover` (default) 
- `focus`
- `click`
- `touchstart`

**Motivation/Use Case**
More usable on mobile
Click ensures the tooltip stays open until the user clicks on the element again, or anywhere else in the document.

rafa-as:
@cagataycivici We also need this functionality

SarahJane87:
@cagataycivici +1

justinotero:
@cagataycivici Provided we get enough community support can I resubmit my PR, then go through the regular PR process?

yaoquanyu:
Cool, useful functionality to have. Can't wait to see it in the Tooltip 👍 !

cardeo:
I would love to see this!

gregAnd:
@cagataycivici
Is this demand for new features available today ?

ViniciusCaminhaDeveloper:
=1

karabaesh:
+1

nlappe:
would love to have to other events aswell. click for example

moussa-b:
also needed

vetoshk1n:
+1

serega270:
+1

codePitcher:
+1

cdskill:
+1

devfake:
+1

ChiranjeeviKandukuri:
tooltip(version 11.1.0) not working on images in responsive, is the issue is exists? can u please help
thanks

greg59650:
+1

AlexanderVega:
+1

D-Biela:
+1

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

PR #6871, opened 2018-11-14
Issue #6862: Tooltip - tooltipEvent support touchstart and click events also

justinotero:
Issue #6862: Tooltip - tooltipEvent support more than hover/focus events

- Tooltip - tooltipEvent now supports `hover, focus, touchstart, click` events.

**Note:** When a `touchstart | click` event is registered for that element, a `touchstart | click` event is also registered on the `document.body`, this ensures that when a user clicks outside of the element we *deactivate* the tooltip.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,12083 @@
+{
+  "name": "primeng",
+  "version": "7.0.0-rc.1-SNAPSHOT",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@angular-devkit/architect": {
+      "version": "0.10.5",
+      "resolved": "https://registry.npmjs.org/@angular-devkit/architect/-/architect-0.10.5.tgz",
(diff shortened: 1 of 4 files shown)


justinotero:
*"The touch interfaces enable applications to create enhanced user experiences on touch enabled devices. However, the reality is the vast majority of today's web content is designed only to work with mouse input. Consequently, even if a browser supports touch, the browser must still emulate mouse events so content that assumes mouse-only input will work as is without direct modification."*
- https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent

 
Can potentially remove support for `touchstart`. As the **[event order](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent)** will still trigger a `click` event.

justinotero:
Build is failing due to an issue in

Issue #6803, opened 2018-10-31
DOM Handler causes incorrect offset in components that use "relativePosition"


JacobSiegle:
**Current behavior**
Under a specific set of conditions the calendar will incorrectly offset itself due to the DOM handler setting the left style property incorrectly in the function relativePosition(). This does not happen on the normal PrimeNG website because the calendar will make the page larger instead of overflowing.

I can try to make a stackblitz when I get time and see if I can figure out a PR if you guys haven't already done it.

**Minimal reproduction of the problem with instructions**
1) Go to the PrimeNG website and load under chrome dev tools as Galaxy S5.

2) Change the metaview port to - 
`<meta name="viewport" content="user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1, width=device-width, viewport-fit=cover">`

3) Shrink the size of the calendar input, the smaller it is the worse the offset is.

4) An example of the calendar input shrunk down significantly
![image](https://user-images.githubusercontent.com/5766272/47803031-9175fc80-dcff-11e8-839a-d155acca2f40.png)

JacobSiegle:
Do you have an example of what the positioning issue is? I tried it again on every component that uses relativePosition and it seems to work as i'm expecting. Maybe you just want the overlay calc to be more complex instead of just cutting the left in half?

cagataycivici:
SplitButton position had issues after the merge;

https://www.primefaces.org/primeng/#/toolbar

PR #6882, opened 2018-11-17
#6803

JacobSiegle:
Fixes DOM handler moving elements off screen when size is too small and screen has set size (see issue number of example).

example-of-fix


JacobSiegle:
Do you have an example of what the positioning issue is? I tried it again on every component that uses relativePosition and it seems to work as i'm expecting. Maybe you just want the overlay calc to be more complex instead of just cutting the left in half?

PR #6945, opened 2018-12-03
Adds 'Closable' Feature to TabMenu

### LINKED ISSUE #3018: TabMenu - close tab ability
```
[ x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
```

**Plunkr Case (Bug Reports)**

**Current behavior**
In TabMenu component there is no closeable ability for each tab
**Expected behavior**
I'm using prime 4 for angular
Could you add same functionality as tabview has to tabmenu? For example ability to close tabs
Following for #688

gg**Please tell us about your environment:**

windows
* **Angular version:** 4.0.X

* **PrimeNG version:** 4.0.X

* **Browser:** [all | Chrome XX | Firefox XX | IE 11| Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 Chrome XX |  IE 11
TypeScript 

joshrathke:
Adds a 'closable' feature to the TabMenu API by exposing an input to display the 'close' icon, along with exposing an 'onTabCloseClick' output to allow developers to react to closures. Should fix #3018 if implemented.

Rather than mutating the list of MenuItems, I opted to simply expose an output that would allow developers to determine what kind of "closing" needs they may have. I'd be open to suggestions if this isn't too big of a feature for a non-employee to implement.

### DIFF
--- a/src/app/components/common/menuitem.ts
+++ b/src/app/components/common/menuitem.ts
@@ -21,4 +21,5 @@ export interface MenuItem {
     title?: string;
     id?: string;
     automationId?: any;
+    closable?: boolean;
 }
--- a/src/app/components/tabmenu/tabmenu.css
+++ b/src/app/components/tabmenu/tabmenu.css
@@ -35,6 +35,13 @@
     vertical-align: middle;
 }
 
+.ui-tabmenu .ui-tabmenu-close {
+    color: #848484;
+    cursor: pointer;
+    margin: 0.5em 0.5em 0 0;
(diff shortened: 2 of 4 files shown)


cagataycivici:
Reverted after a second though, if we provide custom content support, that will be easier.

joshrathke:
So provide an interface for a custom tab template? Is there a component that already has this functionality that I could look at to get an idea of how you guys like to handle something like that?

PR #6947, opened 2018-12-04
fix #6931

### LINKED ISSUE #6931: VirtualScroller wrong behaviour with lazy loading
```
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/primeng-virtualscroller-issues?file=src%2Fapp%2Fapp.component.html

**Current behavior**
The whole area that can be scrolled is a multiple of the `rows` attribute, and not defined by the `totalRecords` attribute, once the end of the dataset is reached.

Last but not least, if the whole dataset lenght is less than - but close enough to - a multiple of the `rows` attribute, lazy loading does not trigger for the last few items.

**Expected behavior**
VirtualScroller dataset container height should only be dependent on the `totalRecords` attribute.
All items should be loaded into VirtualScroller.

Just scroll the stackblitz and look at the console.

This is both a template and logic broken behaviour.

Win10, Webstorm.

* **Angular version:** 7.x

* **PrimeNG version:** 7.0.0

* **Browser:** Tested on Chrome.

* **Language:** Typescript.

AndreaBarbasso:
Now every item of the lazy-loaded VirtualScroller is accessible.

You can test it by adding a single item to the `cars` array in `virtualscrollerdemo.ts:87`:

```
for (let i = 0; i < 10001; i++) {
    this.cars.push(this.generateCar());
}

//in a real application, make a remote request to retrieve the number of records only, not the actual records
this.totalLazyCarsLength = 10001;
```

### DIFF
--- a/src/app/components/virtualscroller/virtualscroller.ts
+++ b/src/app/components/virtualscroller/virtualscroller.ts
@@ -92,10 +92,14 @@ export class VirtualScroller implements AfterContentInit,BlockableUI {
         if (this.lazy) {
             if (val) {
                 let arr = this.cache ? [...this.lazyValue] : Array.from({length: this._totalRecords});
-                for (let i = this.first, j = 0; i < (this.first + this.rows); i++, j++) {
+                let lastIndex = Math.min(this._totalRecords, this.first + this.rows);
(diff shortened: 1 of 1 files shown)


cagataycivici:
Thank you for the PR, we'll do it on our side based on this due to amount of changes.

Issue #2876, opened 2017-05-24
Typed TreeNode and SelectItem


william-lohan:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[X] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**

```ts
let node: TreeNode;
node.data; // <- any

let item: SelectItem;
item.value; // <- any
```

**Expected behavior**
```ts
interface FileSystem { name: string; size: string; type: string; }
interface Column{ field: string; size: header; }

let node: TreeNode<FileSystem>;
node.data.name; // <- string

let item: SelectItem<Column>;
item.value.field; // <- string
```

I can make a pull request for this but there are multiple ways to accomplish this and I don't know which will more likely be accepted.

1. Update api models with generics, causes breaking changes and pull request with significant changes.

```ts
export interface SelectItem<T> {
    label: string;
    value: T;
}

export interface TreeNode<T> {
    label?: string;
    data?: T;
    icon?: any;
    expandedIcon?: any;
    collapsedIcon?: any;
    children?: TreeNode<T>[];
    leaf?: boolean;
    expanded?: boolean;
    type?: string;
    parent?: TreeNode<T>;
    partialSelected?: boolean;
    styleClass?: string;
    draggable?: boolean;
    droppable?: boolean;
    selectable?: boolean;
}
```

2. Update api models with generics with default parameter, no breaking changes, requires typescript 2.3.x.

```ts
export interface SelectItem<T = any> {
    label: string;
    value: T;
}

export interface TreeNode<T = any> {
    label?: string;
    data?: T;
    icon?: any;
    expandedIcon?: any;
    collapsedIcon?: any;
    children?: TreeNode<T>[];
    leaf?: boolean;
    expanded?: boolean;
    type?: string;
    parent?: TreeNode<T>;
    partialSelected?: boolean;
    styleClass?: string;
    draggable?: boolean;
    droppable?: boolean;
    selectable?: boolean;
}
```

3. extend api interface, less desirable.

```ts
export interface TypedSelectItem<T> extends SelectItem{
    value: T;
}

export interface TypedTreeNode<T> extends TreeNode {
    data?: T;
    children?: TypedTreeNode<T>[];
    parent?: TypedTreeNode<T>;
}
[...]

PR #7018, opened 2018-12-18
Add generic SelectItem and TreeNode (fix #2089 and #2876)

### LINKED ISSUE #2089: Generic SelectItem
```
```

**Current behavior**
SelectItem interface has 'value' field of type _any_. Assigning values does not force any type check in this case.
In most cases values are either string or some form of enum (enum type or string literal type acting as enum) for every value of SelectItem's 'value' field in array. Having _any_ there makes keeping type safety in dropdowns difficult.

**Expected behavior**
SelectItem is generic interface. Example:

```
interface SelectItem<T> {
    label: string;
    value: T;
}

let dropdownItems: SelectItem<ExampleEnum>[] = [
    {
        label: 'example', 
        value: ExamleEnum.EXAMPLE_VALUE //<-- This is now ensured by typescript
    }
] 
```
Still if somebody requires to mix values in 'value' field you can use _any_ as type

More robust applications. 

* **Angular version:** 2.X.X

* **PrimeNG version:** 2.X.X

* **Browser:** All browsers
 
* **Language:** Typescript

danielthegray:
Issue #2089 was closed with the promise of adding the generic signature when Typescript 2.3 has wider adoption. Given that we're on Angular 7, with Typescript 3.1.x, this should not be a problem any more, and the improvement should be implemented.
I have also incorporated the suggested TreeNode signature improvement proposed on issue #2876.

Also, I'd like to ask: why don't the files have a newline at the end? It dirties the diff with displays like this:
```
diff --git a/src/app/components/common/selectitem.ts b/src/app/components/common/selectitem.ts
index 71123ccd..ad615c36 100644
--- a/src/app/components/common/selectitem.ts
+++ b/src/app/components/common/selectitem.ts
@@ -5,4 +5,4 @@ export interface SelectItem<T = any> {
     icon?: string;
     title?: string;
     disabled?: boolean;
-}
\ No newline at end of file
+}
```
which are kind of pointless. I would incorporate that change in my branch but I want to make sure that there isn't some other technical reason I don't know about for having no newlines at the end of the files.

### DIFF
--- a/src/app/components/common/treenode.ts
+++ b/src/app/components/common/treenode.ts
@@ -1,14 +1,14 @@
-export interface TreeNode {
+export interface TreeNode<T = any> {
     label?: string;
-    data?: any;
+    data?: T;
     icon?: any;
     expandedIcon?: any;
     collapsedIcon?: any;
-    children?: TreeNode[];
+    children?: TreeNode<T>[];
     leaf?: boolean;
     expanded?: boolean;
     type?: string;
-    parent?: TreeNode;
+    parent?: TreeNode<T>;
     partialSelected?: boolean;
     styleClass?: string;
     draggable?: boolean;
--- a/src/app/components/common/selectitem.ts
+++ b/src/app/components/common/selectitem.ts
@@ -1,6 +1,6 @@
-export interface SelectItem {
+export interface SelectItem<T = any> {
     label?: string;
-    value: any;
+    value: T;
     styleClass?: string;
     icon?: string;
     title?: string;


cagataycivici:
We shall do it in PrimeNG 8.0.

src/app/components/common/treenode.ts
```diff
@@ -1,14 +1,14 @@
...
     icon?: any;
     expandedIcon?: any;
     collapsedIcon?: any;
-    children?: TreeNode[];
+    children?: TreeNode<T>[];
     leaf?: boolean;
     expanded?: boolean;
     type?: string;
-    parent?: TreeNode;
+    parent?: TreeNode<T>;
```
cghislai: Im concerned with this one and the children one. Why would the parent node be of the same type?
Can't we have a `TreeNode<MyFile>` as a child of a `TreeNode<MyFolder>`?
Having the parent/child node types in the generic signature is probably cumbersome, so maybe it should be left as `any`

cghislai: Or maybe add the parent/child type in the generic signature, defaulting to the node item:
`TreeNode<T = any, P = T, C = T>`.

thedanielgray: With this change, your use case is left undisturbed (just make it `TreeNode<any>` -- or just `TreeNode`). It will just allow for stricter typing control for those who would like it/need it. 

Your other multi-typed generic proposal could be interesting! I'd like to see it. Make a PR! :)

cghislai: Ok. I was afraid that implementing this interface with different types would not compile, but it seems to compile just fine. Something like
`class MyNodeA implements TreeNode { value: string; children: MyNodeB[]; /**...**/}`
`class MyNodeB implements TreeNode { value: number; children: MyNodeB[]; /**...**/}`

william-lohan: To the original example, you could also:

```ts
abstract class MyFileSystemNode { ... }
class MyFolder extends MyFileSystemNode { ... }
class MyFile extends MyFileSystemNode { ... }
```

Then a `TreeNode<MyFileSystemNode>` could have value of `MyFolder` and have child with value of `MyFile`.

mark-at-tusksoft:
This was added to the milestones then removed. Will this be added back to the roadmap anytime soon?

OldBuddy:
@cagataycivici can you comment why it is still not implemented?

cagataycivici:
Review for 10.0.1.

cagataycivici:
Fixed in #2876

Issue #7024, opened 2018-12-19
DynamicDialog should not use translate3d and translateX/Y to center itself


mselerin:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
The DynamicDialog center its content using `translateX(-50%) translateY(-50%)`. It also use a `translate3d` to animate the dialog.

When using Chrome, blurry text could occurs after a `translate3d` animation.

**Expected behavior**
The `translate3d` animation should be removed to fix the blurry text. 
The centering could be done using flex and a wrapper div.

I have made the fix inside my fork. Will do a PR ASAP.

* **Angular version:** 7.1

* **PrimeNG version:** 7.0

* **Browser:** all
 
* **Language:** all

mselerin:
Don't work on IE, I'm on it...

cagataycivici:
I'm unable to replicate the issues mentioned so rather not change anything on our side, there are no similar reports as well.

davidpotvincode:
we are experiencing the same issue after updating from 6.1.5 to 7.1.0.

huorwhisp:
The same issue happens for me, including ConfirmDialog, on Chrome v73.0 and Edge v42.17. I think you should use flex css

mertsincan:
Fixed for 7.1.3. Please see; https://github.com/primefaces/primeng/issues/7598

djay1977:
Not fixed in version 7.1.3, the following code is used in DynamicDialog and now in ConfirmDialog : `transform: 'translateX(-50%) translateY(-50%)'`
=> In Chrome, text, line, font are blurred (see [https://ibb.co/nz0xPFq](printscreen))

It's OK with classic Dialog because there is no css transformation.
For information, I'm using theme Ultima 7.0.1

mickeyjohn:
I am also using Utima 7.0.1 and experienced the blurry text in Chrome version 73.0.3683.103 (Official Build) (64-bit). However, after do the setting reset, It seems to be fixed in Chrome but not with Edge v44.*. In addition, It does not appear on Edge Dev version 76.*, may be it because of the Chromium engine under the hood..
Thanks.

PR #7025, opened 2018-12-19
fix for #7024 - DynamicDialog should not use translate3d and translateX/Y to center itself

mselerin:
Fix for #7024.


mselerin:
Don't work on IE, I'm on it...

cagataycivici:
I'm unable to replicate the issues mentioned so rather not change anything on our side, there are no similar reports as well.

PR #7026, opened 2018-12-19
Add property style and styleClass #7017

### LINKED ISSUE #7017: ConfirmDialog set width
```
```

**Current behavior**
I've just upgraded to primeNG 7.0.3, and notice that some inputs of Dialog and confirmDialog are deprecated (width, height, minWidth, minHeight) by style input. This works correctly with Dialog but not with confirmdialog. I can't set its width without width input (but this it's deprecated).

I notice that confirmDialog Component doesn't have style or ngStyle input, but dialog yes.

**Test case**
https://github-w4h5cc.stackblitz.io

**Expected behavior**
Set width with new style input.

gasparbarancelli:
Bug fix #7017
Add property style and styleClass in component Confirm Dialog


cagataycivici:
My colleague @yigitfindikli has already done it recently as well, thank you.

PR #7036, opened 2018-12-22
fixed #5614 using multiple lines in p-message(s) breaks layout

### LINKED ISSUE #5614: using multiple lines in p-message(s) breaks layout
```
```

**Current behavior**
if you use line breaks or just a text which breaks automatically the whole text gets pushed below the icon.

image

**Expected behavior**
I think the correct/better behavior would be to either let the text float around the icon or keep it intended.

the layout is off when using multiple lines.
In our case we needed a warning to explain to user what will happen if he proceeds.
This can't be done in one line.

* **Angular version:** 5.X
5.2.10

* **PrimeNG version:** 5.X
5.2.3

Chrome  66.0.3359.117 
Firefox 59.0.2
not tested with other browsers

yigitfindikli:


cagataycivici:
Merged manually

PR #7044, opened 2018-12-27
Update steps to include tooltip and autowidth

DarkAurore:
Hi,
For my project I wanted this component to take the full width of the container.
I suggest adding this autoWidth input property to automatically set the same width in % depending on the number of items.
To take the full width of the container, it also needs to be defined on the component, example with 100% : 
[autoWidth]="true" [style.width]="100 + '%'"
If set to false, it will set the li width as auto. If the width is set via css, then the value in css will be used, so your example with 25% will still work. Not sure if there are any side effects. Only tested with Chrome.
I also suggest adding the title element of "a" to use the menu item title property to display a tooltip on the step.
Thanks,
Aurore

###Defect Fixes

###Feature Requests


cagataycivici:
Better way would be I guess to use flex so browser can scale them properly without using props e.g. flex: 1 1 0

Issue #7074, opened 2019-01-04
TurboTable is not properly unbinding the body scroll listener


thedanielgray:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
The TurboTable ScrollBody component's `scroll` event listener is not being unbound correctly, it is rather being bound twice.
In `table.ts`:
```
2275     unbindEvents() {
2276         if (this.scrollHeaderViewChild && this.scrollHeaderViewChild.nativeElement) {
2277             this.scrollHeaderBoxViewChild.nativeElement.removeEventListener('scroll', this.headerScrollListener);
2278         }
2279
2280         if (this.scrollFooterViewChild && this.scrollFooterViewChild.nativeElement) {
2281             this.scrollFooterViewChild.nativeElement.removeEventListener('scroll', this.footerScrollListener);
2282         }
2283         //                                     vvv HERE
2284         this.scrollBodyViewChild.nativeElement.addEventListener('scroll', this.bodyScrollListener);
2285     }
```

**Expected behavior**
The event listener should be unbound instead of re-bound.

**Minimal reproduction of the problem with instructions**

**What is the motivation / use case for changing the behavior?**
More correct behavior.

**Please tell us about your environment:**
any

* **Angular version:** 5.X
7.x

* **PrimeNG version:** 5.X
7.x

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
all

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
all

cagataycivici:
Fixed via PR

PR #7075, opened 2019-01-04
Fix incorrect unbinding of scroll event

thedanielgray:
###Defect Fixes
This PR fixes issue #7074.


thedanielgray:
The Travis build failed with something not related to my code it seems... I only changed one line (completely unrelated to where the error is).

Issue #5793, opened 2018-05-24
p-dropdown onItemSelect event


dolphinsd:
The code consumes onItemClick, it would be nice to emit it as well so parent might do something.

cagataycivici:
You can use onChange for this.

PR #7090, opened 2019-01-07
p-dropdown onItemSelect event #5793

gasparbarancelli:
Added onItemSelect event in dropdown component.


cagataycivici:
This is not called when item is selected with keyboard.

cagataycivici:
There is also onChange callback for this.

Issue #4945, opened 2018-01-23
Use HTTPClient in FileUpload


netstart:
```
[ x  ] feature request
```
Please change upload request from p-FileUpload to use HttpClient (Angular 4.3+), 
In this case I can  use interceptor to show my LoadingComponent in any upload requests.

Is better to add Bearer authentication header too.

atmjesus:
I'm also waiting for this. It would be nice for adding headers and for using Angular HttpInterceptor.

Thanks

kkdattaGIT:
@cagataycivici  This change broke OnUpload method. Now, event does not have the Response object coming back from server.

mfcar:
> @cagataycivici This change broke OnUpload method. Now, event does not have the Response object coming back from server.

Same here.

juri33:
That changes a lot and breaks a lot. Not a good idea to to this in a minor update ...

jeslomo:
Same here, the method beforeUpload crash when read "event.xhr".

 onBeforeSend(event) {   
      event.xhr.setRequestHeader('bla bla bla');
  }

jotae1e:
This is not minor update, folks.  If you use .upload() method to send a file and variables, this will not work.  

It's a very big update.

namnhcntt:
I upgraded to 7.0.5, but the upload event does not call my HTTP_INTERCEPTORS in app.module.ts. Anyone can fix this? 
(I have another library that call my HTTP_INTERCEPTORS in the same application)

szabolcssimonyi:
Today I realized xhr surprizingly disappeared from onBeforeSend $event in 7.0.0 and 7.0.3 too!! Is this possible to change released versions like that? It breaks my app since i append authentication to the header at this event.

szabolcssimonyi:
Must agree with @jotae1e, made me a bad day too...

PR #7091, opened 2019-01-07
Use HTTPClient in FileUpload #4945

gasparbarancelli:
Changed the XHR request to now use the HookpClient of the Angular, so that it is possible to use the interceptors. The onBeforeUpload, onBeforeSend, onUpload and onError events were also changed, in the first 3 events the XHR object was removed, and in the 4 event the request error was added.


gasparbarancelli:
@cagataycivici 
It is not passing the tests because now the file-upload request is performed by httpClient, we would have to change the tests by making a mock of httpClient.

Issue #4285, opened 2017-10-26
Multiselect - Select only when checkbox is clicked


GuzzoLM:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[X] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
In a multiselect, whenever the user clicks an option, the option is selected.

**Expected behavior**
It could be useful if there was a wayto only enable option select if the user clicks the checkbox.

sudheerj:
Can you please elaborate about expected behavior. Do you want to select through checkbox only(not through label)?

GuzzoLM:
@sudheerj yes, I want an option to disable selecting through label.

kvabhishek:
I am also looking for similar functionality in the Datatable. Did you find a solution? Please help.

cagataycivici:
We'll reconsider if more users demand this behavior.

PR #7092, opened 2019-01-07
Multiselect - Select only when checkbox is clicked #4285

gasparbarancelli:
Added clickInCheck property with its default value to false so as not to break old versions. When checked, the item is selected only if the user checks the box.


cagataycivici:
We'll reconsider if more users demand this behavior.

Issue #2318, opened 2017-03-16
Responsive Lightbox


bogacg:
**I'm submitting a ...** 
```
[X] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
```

**Current behavior**
Full size lightbox images are displayed in given size as is, even on mobile devices. If image size is larger than available window size, image visibile area is cropped & centered (but not really centered, there seems to be some padding)

**Expected behavior**
It would be nice to have `responsive` attribute, if defined fits image by width, to available window area.

**Minimal reproduction of the problem with instructions**
[Docs Lightbox](https://www.primefaces.org/primeng/#/lightbox)

**What is the motivation / use case for changing the behavior?**
Better mobile experience

bossqone:
any updates?

PR #7097, opened 2019-01-08
Responsive Lightbox #2318

gasparbarancelli:
Adjustment of dynamic image centering.


cagataycivici:
It breaks after a lightbox is closed and tried to reopen.

PR #7153, opened 2019-01-21
fixed #7123 closeOnEscape property not working on p-dialog

### LINKED ISSUE #7123: closeOnEscape property not working on p-dialog
*I'm submitting a ...*  (check one with "x")
[x] bug report => Search GitHub for a similar issue or PR before submitting
[ ] feature request => Please check if the request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit a support request here, instead see http://forum.primefaces.org/viewforum.php?f=35

This is the official PrimeNG showcase for Dialog-
https://www.primefaces.org/primeng/#/dialog

*Current behavior*
the dialog doesn't close on hitting escape.
I am using [visible]=true on all my dialogs and at the same time I am also using [closeOnEscape]=ture which doesn't allow the dialog to close on esc key as the visible property is set to true. We also tried to close the dialog from typescript but that didn't help. Can you please provide the POC for this.  

Below is one of the example of the properties I used to set on my dialog
<p-dialog  [visible]="true" [closable]="false" (onShow)="onAfterShow($event)" [modal]="true" [responsive]="true" [draggable]="true" [resizable]="false">

*Expected behavior*
The dialog should be allowed to be closed on hitting the Escape key

*Minimal reproduction of the problem with instructions*
Open the dailog and simply click on escape to close the dialog, it doesn't close.

Angular version:** 5.0.5

PrimeNG version:** 5.2.7

This can be reproduced in all the browsers

Language: [all | TypeScript 2.7.2 ]

yigitfindikli:


NekitCorp:
Similar problem with [dynamicdialog](https://github.com/primefaces/primeng/blob/39bb387c3139c918c759a9cc1a4afbb8b5890f7b/src/app/components/dynamicdialog/dynamicdialog.ts#L146)?

PR #7154, opened 2019-01-21
fix disabled option in multiple selectButton

PavloBosoi:
Added additional functionality for disabled option of selectButton component. Now it can work normally in maltiple mode. Earlier it works normally only in single mode.


cagataycivici:
We should not enforce the custom object to have a disabled property, disabled options are only supported when a SelectItem interface is provided. Better solution is to use optionDisabled property similar to optionLabel to get the disabled property name from the custom object.

PR #7158, opened 2019-01-22
Multiselect - clear input on filter cross click

murtuzamacdev:
We have a specific requirement in the multiselect component that when we click the cross at the left hand side, rather than closing the popup, it should just clear the filter input. 

I have added a flag 'clearInputOnCross' which when will be true, will only clear the filter text input and will not close the popup.

###Defect Fixes

###Feature Requests


cagataycivici:
If more users demand it this behavior, we'll reconsider thank you.

Issue #6144, opened 2018-07-23
Tree: checkbox selection, items where 'selectable:false' don't have a check box 


nanafocus:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
Tree: checkbox selection, items where 'selectable:false' still have a check box #5359


**Current behavior**
Items that have 'selectable' as false in the tree don't have checkbox.

**Expected behavior**
Checkboxes should be displayed on items where 'selectable' is false.

**Minimal reproduction of the problem with instructions**
Create a tree with a selection mode as checkbox. Set the items to 'selectable' false, note you will not see the checkbox.

**What is the motivation / use case for changing the behavior?**
I want to see the checkbox when selectable is false because I use this property to disable the tree. I want to see the complete tree with the checkbox also when it is disable.

* **Angular version:** 6.0.8

* **PrimeNG version:** 6.0.0

estenp:
This doesn't actually appear to be a bug, but rather an intentional change made by the team in the 6.0 release (`selectable` option added to `*ngIf` on the `.ui-chkbox` : https://github.com/primefaces/primeng/compare/1e66a7c2acb0a77ec102ab9d44caa8f5fbff27fd...master).

I guess the question is why was this change made and would the team be willing to accept a pull request adding the old disabled checkbox display back in, or at least possible adding an option to tree node for turning it's display on or off.

KhaledGabr:
This doesn't really fix the issue.
Seletable Property should be used to **show** or **hide** the selection checkbox. 

Please, add a new property called, **disabled** that would disable a checkbox. However, the selectable button needs to be reverted back like it was before.

PR #7183, opened 2019-01-28
removed ngIf condition for selectable === false on ui-chkbox in order…

estenp:
This code seems to have been removed with the 6.0.0 release. Someone already created an issue, but this seems intentional. 

Previously, if a node has the selectable option set to false, an opaque, "disabled" checkbox is displayed. Currently, there is simply an absence of a checkbox.

What is the reasoning behind the change? If the team is not interested in replacing the old functionality, would it be open to adding an option to turn on or off the disabled checkboxes under the "!selectable" condition?

###Defect Fixes
https://github.com/primefaces/primeng/issues/6144


cagataycivici:
We disable it instead now.

PR #7205, opened 2019-02-05
IMP - global filter - now filters with objects inside objects in all the tables

Sofy18:
###Defect Fixes
When you use the global filter in a table with objects inside an object, the filter doesn't work with the subfields
###Feature Requests
The method is modified to validate if the element of each column is an object that include more objects inside, if the field has subfields, the filter search inside the property of the object aggregated in the general object. Method resolveFieldData with the fix, return the correct property of the subfield.


Sofy18:
When you have columns in the tables with objects inside an object, the filter now works.

cagataycivici:
Case is not clear, please provide a stackblitz demonstrating the case first.

PR #7211, opened 2019-02-06
Fixed #6822 - Fixing double click in Splitbutton and Tiered menu

### LINKED ISSUE #6822: SplitButton & TieredMenu - double click on item reloads the page
**Current behavior**
Double click any of the splitbutton's items, it will reload the page.

**Expected behavior**
On both disabled and enabled items the double click should not trigger the refresh of the entire page.

https://www.primefaces.org/primeng/#/splitbutton
Double click any of the splitbutton's items, it will reload the page.

SplitButton - double click on item reloads the page, for both enabled and disabled items.

Win64

* **Angular version:** 6.1.0

* **PrimeNG version:** 6.1.5

* **Browser:** [all]
 
* **Language:** [all]

shogogan:
Fixing #6822 on SplitButton and TieredMenu's

I removed the `href='#'` and treated the href as attr.href, so the <a> tag doesn't have a href tag if it doesn't exists, fixing the problem with double clicks. Then i adjusted the CSS to add the cursor:pointer on them


cagataycivici:
Merged manually, thanks.

Issue #7213, opened 2019-02-07
Add comma as separator for chips component


Sheraff:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
*Chips* component only allows for a few keys to trigger the addition of a *chip*.

**Expected behavior**
Ideally we'd setup a new `@Input()` to allow user to set their own list of keys to trigger an addition of chips (or even add their own parsing on the text instead of key triggers). But the most requested one seems to be the *comma* key.  

**What is the motivation / use case for changing the behavior?**
Many apps using a similar component allow for a CSV type of input where separating your values by commas will be recognized as different chips.

Sheraff:
For example, the comma was requested here: https://github.com/primefaces/primeng/issues/5333 (though in combination with a `(paste)` event).

Allowing the developer to define their own parsing was requested here: https://github.com/primefaces/primeng/issues/3419

cagataycivici:
If more users demand it via this ticket, we'll consider, thank you.

djdev:
I think having the option to paste comma separated values into the chip input and the input handling the transformation of the input based on separators will be an excellent piece for UX. Where can I upvote for it?

tigrenok00:
We also wish there was support for pasting a list of values with custom delimiter (slash or comma)

sweatC:
@cagataycivici I think there are a lot people who need this feature :+1:

cagataycivici:
Set to 9.0.2

cagataycivici:
#5333

PR #7214, opened 2019-02-07
adding comma as a separator for chips component

Sheraff:
Related to https://github.com/primefaces/primeng/issues/7213

This is a feature request PR with very minimal changes. More could be done as described in the issue linked above.

### DIFF
--- a/src/app/components/chips/chips.ts
+++ b/src/app/components/chips/chips.ts
@@ -56,6 +56,8 @@ export class Chips implements AfterContentInit,ControlValueAccessor {
     
     @Input() addOnTab: boolean;
 
+    @Input() addOnComma: boolean;
+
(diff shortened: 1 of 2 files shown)


cagataycivici:
If more users demand it via the original ticket, we'll consider, thank you.

sweatC:
It would be great to have this functionality. I have old UI which uses [tags input component](https://mbenford.github.io/ngTagsInput/). It has the ability to add values via comma. When I started redesign using PrimeNG I have found a similar component called Chips. But it allows adding the only when a tab key pressed. My point is to make it compatible with old UI functionality in which values should also be added by comma and tab keys. @cagataycivici

cagataycivici:
Please see;

https://primefaces.org/primeng/showcase/#/chips

Issue #7179, opened 2019-01-28
p-fileUpload: after upgrade to primeng 7 there is no possibility to get response


RushPVA:
there is no possibility to get response
primeng 6:
onUpload(event: any) {
        const uploadedDocuments = JSON.parse(event.xhr.responseText);

primeng 7:
onUpload(event: any) {
        const uploadedDocuments = JSON.parse(event.xhr.responseText); 
        **event.xhr is undefined.**

But i don't see any other possibility to obtain response in primeng7.(in our case response contains file id that was created on the backend side)

TriplicateGirl:
I believe this was caused by a change 7.0.5.

https://github.com/primefaces/primeng/pull/7091

I am also curious how to get the response, I was doing something very similar to you.

I was also using event.formData.append in the onBeforeSend, which no longer is working either.

andreLagacione:
I had the same issue today in version 7.0.5, the response was an empty formData and hadn't the xhr, I installed the version 7.0.5 and the response it's OK, I needed get the response for add the token in the header in onBeforeSend callback:
public addHeader(event) {
	event.xhr.setRequestHeader('Authorization', `Bearer ${this.uploadService.getToken()}`);
}

I suggest to you use the version 7.0.4, set "primeng": "7.0.4" in package.json

vrt2015:
Is there going to be a fix for this? We would like to get access to the response back from our API

i255:
Every second minor release breaks something in my app. That is sad. Could you please add the HttpResponse variable to the onUpload event handler like so:
this.onUpload.emit({files: this.files, event: event});
instead of this.onUpload.emit({files: this.files}); in line 329

gasparbarancelli:
> I had the same issue today in version 7.0.5, the response was an empty formData and hadn't the xhr, I installed the version 7.0.5 and the response it's OK, I needed get the response for add the token in the header in onBeforeSend callback:
> public addHeader(event) {
> event.xhr.setRequestHeader('Authorization', `Bearer ${this.uploadService.getToken()}`);
> }
> 
> I suggest to you use the version 7.0.4, set "primeng": "7.0.4" in package.json

You can now add the token by the httpclient interceptor.

gasparbarancelli:
> there is no possibility to get response
> primeng 6:
> onUpload(event: any) {
> const uploadedDocuments = JSON.parse(event.xhr.responseText);
> 
> primeng 7:
> onUpload(event: any) {
[...]

PR #7231, opened 2019-02-12
Added Response Event to onUpload call #7179

eltonmmoreira:
Added Response Event to onUpload call #7179


cagataycivici:
Renamed to originalEvent due to convention.

Issue #7219, opened 2019-02-08
 Prevent duplicates and Prevent open duplicates for Toast


distributev:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

This could add value Prevent duplicates and Prevent open duplicates

like here https://foxandxss.github.io/angular-toastr/

cagataycivici:
We'll consider if there is more demand, thank you.

ankuradhey:
This is interesting feature and user can have an option to enable duplicates or disable.

JacobSiegle:
Nice QOL feature I think, it might be nice if when a duplicate was found it reset the life on the existing message and gave some sort of signal you hit it again like highlight.

ManzerHa-Kore:
How to prevent duplicates toast message in primeng?

JacobSiegle:
You can pull the PR  https://github.com/primefaces/primeng/pull/7243 and include it in your project manually, otherwise there's currently no built in way.

cagataycivici:
Planned for 9.0.2

PR #7243, opened 2019-02-14
Added the prevent open duplicates and prevent duplicates in toast #7219

gasparbarancelli:
#7219 
Added the preventOpenDuplicates and preventDuplicates properties. If the first one is true when adding a new message it is not displayed if there is already a toast with the same detailt, when the second property is true all the messages already added are stored in a list and adding a new message is checked if a message with the same detailt already exists in that list, and the message is only displayed if no messages are found.

### DIFF
--- a/src/app/components/toast/toast.ts
+++ b/src/app/components/toast/toast.ts
@@ -157,6 +157,10 @@ export class Toast implements OnInit,AfterContentInit,OnDestroy {
 
     @Input() hideTransitionOptions: string = '250ms ease-in';
 
+    @Input() preventOpenDuplicates: boolean = false;
+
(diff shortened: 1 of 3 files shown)


lucini:
Nice feature, thanks Gaspar!

cagataycivici:
We'll consider if there is more demand, thank you.

FernandaGM:
Nice feature

diogosatdl88:
Really useful feature

VictorRChagas:
Awesome, it will help a lot to fix some bugs we have in our projects

cagataycivici:
We'll implement it on our side based on your PR, thank you.

PR #7252, opened 2019-02-15
fix(multiselect): filter with virtualscroll enabled.

### LINKED ISSUE #7251: MultiSelect with filtering and virtual scroll enabled doesn't work properly.
```
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/github-wpyzrc

**Current behavior**
In the `<p-multiSelect>`-element component you have created enabling the virtual scroll combined with filtering creates an issue, see the stackblitz link above.

When inserting a search in the filter field, the response only shows a result from the items currently inserted in the DOM.

**Expected behavior**
When enabling `[virtualScroll]="true"` and  `[filter]="true"` on a `<p-multiSelect>`-element, one would on filtering amongst all the values, expect a list of all matches in the multi-select-dropdown.

Insert the following snippet of code and have at least some hundred values in the `options`-array.
```
<p-multiSelect
    [options]="options"
    [virtualScroll]="true"
    [panelStyle]="{ minWidth: '15em' }"
    [filter]="true"
    itemSize="30"
></p-multiSelect>
```

As a user would most likely appreciate all possible matching values when filtering on a set of values, this issue would imo need a fix.

Windows, Visual Studio Code, yarn

* **Angular version:** 7.0.1

* **PrimeNG version:** 7.0.5

* **Browser:** Chrome 72.0.3626.96

 
* **Language:** [ TypeScript 3.1.4 | ES6 ]

PeterSkopal:
Change utilized array in p-multiSelectItem displaying actual filter results for users.

###Defect Fixes
#7251 

###Feature Requests

### DIFF
--- a/src/app/showcase/components/multiselect/multiselectdemo.html
+++ b/src/app/showcase/components/multiselect/multiselectdemo.html
@@ -27,7 +27,7 @@
     <p>Selected Cars: {{selectedCars2}}</p>
 
     <h3>Virtual Scroll (10000 Items)</h3>
-    <p-multiSelect [options]="items" [(ngModel)]="item" [panelStyle]="{minWidth:'12em'}" [virtualScroll]="true" [itemSize]="34" [filter]="false"></p-multiSelect>
+    <p-multiSelect [options]="items" [(ngModel)]="item" [panelStyle]="{minWidth:'12em'}" [virtualScroll]="true" [filter]="true" [itemSize]="34" [filter]="false"></p-multiSelect>
 </div>
 
 <div class="content-section documentation">
--- a/src/app/components/multiselect/multiselect.ts
+++ b/src/app/components/multiselect/multiselect.ts
@@ -115,7 +115,7 @@ export class MultiSelectItem {
                         </ng-container>
                         <ng-template #virtualScrollList>
                             <cdk-virtual-scroll-viewport #viewport [ngStyle]="{'height': scrollHeight}" [itemSize]="itemSize" *ngIf="virtualScroll">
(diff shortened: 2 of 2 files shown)


src/app/showcase/components/multiselect/multiselectdemo.html
```diff
@@ -27,7 +27,7 @@ <h3>Templating</h3>
     <p>Selected Cars: {{selectedCars2}}</p>
 
     <h3>Virtual Scroll (10000 Items)</h3>
-    <p-multiSelect [options]="items" [(ngModel)]="item" [panelStyle]="{minWidth:'12em'}" [virtualScroll]="true" [itemSize]="34" [filter]="false"></p-multiSelect>
+    <p-multiSelect [options]="items" [(ngModel)]="item" [panelStyle]="{minWidth:'12em'}" [virtualScroll]="true" [filter]="true" [itemSize]="34" [filter]="false"></p-multiSelect>
```
Mysame:
```
[filter]="true" [itemSize]="34" [filter]="false"
```
That seems odd.

PeterSkopal: @Mysame  Agree, honest mistake, should we create a new issue for it?

First `filter` input currently overrides the second one.

Mysame: You probably should, as of now it isn't clear and looks wrong (even if it functions correctly)

PeterSkopal: #7339

Issue #7253, opened 2019-02-16
Autodisplayfirst setting works incorrectly in dropdown and doesn't change the selected value.


olegbabkevich:
**I'm submitting a bug**  

https://stackblitz.com/edit/issue-with-autodisplayfirst-setting-in-dropdown

**Current behavior**

By default Autodisplayfirst = true. That setting is changing the selected option inside of dropdown, but doesn't change the selected value.

**Expected behavior**

**Minimal reproduction of the problem with instructions**
https://stackblitz.com/edit/issue-with-autodisplayfirst-setting-in-dropdown
-->

**What is the motivation / use case for changing the behavior?**
Correct possibility to use control on a form

**Please tell us about your environment:**
* **Angular version:** 5.X
  7.0.0
* **PrimeNG version:** 5.X
  7.0.5
  Chrome: 72.0.3626
* **Node (for AoT issues):** `node --version` =  8.9.1

PR #7256, opened 2019-02-17
Fix issue 7253 - Autodisplayfirst and Value

mattlgroff:
###Defect Fixes
Fixes issue 7253 for the Autodisplayfirst setting. 

Issue: When options are set, the `this.value` is passed in, but is empty since it was never selected manually.
https://github.com/primefaces/primeng/issues/7253


cagataycivici:
This doesn't check if a value already exists though, generally not sure about it, need more feedback.

Issue #7261, opened 2019-02-18
p-spinner: readonly attribute value ignored


cghislai:
The `p-spinner` `[readonly]` attribute presence makes the input disabled, regardless of its value.
This is due to the use of `[attr.readonly]` rather than `[readOnly]` to bind the attribute value.
As seen in the following plunker, `[attr.readonly]="false"` actually makes the input readonly.

plunkr: https://vbwilqlo.github.stackblitz.io (code should be at https://stackblitz.com/github/vbwilqlo but does not seem to work)

<details>
<summary>plunkr code</summary>
<pre><code><![CDATA[
&lt;h2&gt;PrimeNG Issue Template&lt;/h2&gt;
&lt;p&gt;Please create a test case and attach the link of the plunkr to your github issue report.&lt;/p&gt;

&lt;button pButton label="PrimeNG" (click)="onClick($event)"&gt;&lt;/button&gt;
{{counter}}
Readonly inputs: {{ counter % 2 === 1 }} 
&lt;div&gt;
  &lt;p-spinner [readonly]="counter % 2 === 1"
  [(ngModel)]="counter"
  &gt;&lt;/p-spinner&gt;
&lt;/div&gt;

&lt;div&gt;
  input[type=number] [readonly]=boolean
&lt;/div&gt;
&lt;div&gt;
  &lt;input type="number" [readonly]="counter % 2 === 1"
  [(ngModel)]="counter"
  &gt;
&lt;/div&gt;

&lt;div&gt;
  input [readonly]=boolean
&lt;/div&gt;
&lt;div&gt;
  &lt;input [readonly]="counter % 2 === 1"
  [(ngModel)]="counter"
  &gt;
&lt;/div&gt;

&lt;div&gt;
  input [attr.readonly]=boolean
&lt;/div&gt;
&lt;div&gt;
  &lt;input [attr.readonly]="counter % 2 === 1"
  [(ngModel)]="counter"
  &gt;
&lt;/div&gt;
]]></code></pre>
</details>

PR #7263, opened 2019-02-18
Spinner: fix readonly binding & tests

### LINKED ISSUE #7262: p-spinner readonly do not affect buttons
Duplicate of #7261, see also comment on commit https://github.com/primefaces/primeng/commit/3aa15371da206a31446132ac5dbcabb29b9c07ea.

Spinner button should be disabled using `[disabled]="disabled || readonly"`

cghislai:
Use `[readOnly]` rather than `[attr.readonly]` to bind the readonly value, as only the former one results in HtmlInputElement.readOnly to be false when the bound value is false.

Disable the buttons when readonly.

Fix #7262 and #7261


cagataycivici:
Merged manually, thanks.

PR #7279, opened 2019-02-22
add optionValue to dropdown, multiselect, listbox and selectButton

yelhouti:
###Defect Fixes

###Feature Requests
As you can see this a very minor change 6 lines if you don't include documentation, please accept it :)


cagataycivici:
Thank you but a feature of this size should be implemented by the team internally. For example, I don't see any tests in the PR.

yelhouti:
@cagataycivici I didn't add some because i didn't find any for optionLabel (maybe I didn't check enough), I would gladly add them if it means the PR would be accepted

PR #7280, opened 2019-02-22
Add visibleChanged event emitter in menu component

MikeMnD:
This way you can subscribe and do stuff on menu close or on menu open


src/app/components/menu/menu.ts
```diff
@@ -107,12 +109,14 @@ export class Menu implements OnDestroy {
         else
             this.show(event);
 
+        this.visibleChanged.emit(this.visible);
```
cghislai: this will fire the event twice has both method above will fire it already

MikeMnD: Yep that's right I've left it only in show and hide.

cagataycivici:
Renamed to onShow and onHide

PR #7302, opened 2019-02-27
Fixed #6657 and #7084 - mouseenter closes menubar with autodisplay false

### LINKED ISSUE #6657: Menubar gets closed immediately after mouse enter on menu-item when AutoDisplay property as false
I am using PrimeNG Menubar v5.2.5 with AutoDisplay property as false.
The issue is, when you open a menu item and getting away from it, and then move the mouse over the same menu item, it gets closed.

But, this was not the case in v5.2.4. There it is working perfectly.

You can also check this issue on https://www.primefaces.org/primeng/#/menubar, in their without AutoDisplay Demo. It is also replicating there.

### LINKED ISSUE #7084: Menubar Without AutoDisplay - need 2 click to open menu after focus out
```
```

Go to https://www.primefaces.org/primeng/#/menubar,
1. Click on a navbar Without AutoDisplay - "File" - it's open menu.
2. Click on not navbar anywhere - it's close menu.
3. Click on a navbar "File" - menu open is not work now.
4. Second click on a navbar "File" - it's work again

Tryforce:
Menubar now doesnt close on mouseenter if autodisplay is set to false and will remain open until focus is taken elsewhere, as expected. Also the menu will reopen with 1 click, as opposed to 2 as brought up in #7084.

Fixes #6657 and fixes #7084

### DIFF
--- a/src/app/components/menubar/menubar.ts
+++ b/src/app/components/menubar/menubar.ts
@@ -65,7 +65,7 @@ export class MenubarSub implements OnDestroy {
                 return;
             }
             
-            this.activeItem = this.activeMenu ? (this.activeMenu.isEqualNode(item)? null: item) : item;
+            this.activeItem = this.activeMenu ? (this.activeMenu.isEqualNode(item) && this.autoDisplay ? null : item) : item;
             let nextElement = <HTMLLIElement>item.children[0].nextElementSibling;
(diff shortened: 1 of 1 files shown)


cagataycivici:
Thank you!

Issue #7331, opened 2019-03-04
Improve showcase/documentation for Responsive


Cito:
Please have a look at the [showcase for the responsive interface](https://www.primefaces.org/primeng/#/responsive).

You will notice that the HTML source does not show the most crucial thing in this demo, namely that the `ui-fluid` class must be set somewhere on a higher level element to enable responsive mode.

Also, it would be nice if there was a switch in the showcase that allows toggling between normal and responsive mode so that you can see and explore the difference in behavior.

Motivation / use case for changing the behavior: Making the responsive mode in the documentation better understandable.

PrimeNG version: 7.1

cagataycivici:
Not sure if we need a switch for responsive mode, still will update the docs to add ui-fluid. Thank you.

PR #7332, opened 2019-03-04
Improve the showcase for the responsive interface

Cito:
Makes the page switchable between normal and responsive mode.

Also, the ui-fluid class was set in the template of the demo, but missing in the displayed HTML source. The displayed source now properly shows how the ui-fluid class is set at the top.

This solves issue #7331.


cagataycivici:
Not sure if we need a switch for responsive mode, still will update the docs to add ui-fluid. Thank you.

PR #7354, opened 2019-03-08
fix(multiselect): initialize property value, enabling selectionLimit.

### LINKED ISSUE #7221: Multiselect with [selectionLimit] = "1" throws exception
Current Waiting Time: ~8 weeks.

```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-m1kf75

**Current behavior**
The error occurs when limiting the MultiSelect

**Expected behavior**
Select the object from the list

add [selectionLimit] = "1" to any multiselection and produce the error
ERROR TypeError: Cannot read property 'length' of undefined
    at MultiSelect.onOptionClick (multiselect.ts:374)

Visual Code

* **Angular version:** 5.X
Angular 7.0.1

* **PrimeNG version:** 5.X
PrimeNG 7.0.5

All

Temporal Solution:
multiselect.js line 201
`if (!this.selectionLimit || (!this.value) || (this.value.length < this.selectionLimit)) {`

PeterSkopal:
Add initialization for the property value in multi-select-component, making if-cases to pass/fail as intended in function `onOptionClick`.

###Defect Fixes
#7221

### DIFF
--- a/src/app/components/multiselect/multiselect.ts
+++ b/src/app/components/multiselect/multiselect.ts
@@ -246,7 +246,7 @@ export class MultiSelect implements OnInit,AfterViewInit,AfterContentInit,AfterV
     
     @Output() onPanelHide: EventEmitter<any> = new EventEmitter();
     
-    public value: any[];
+    public value: any[] = [];
     
     public onModelChange: Function = () => {};
     


cagataycivici:
Not sure if we should do an empty array, maybe better to initialize it when needed if value is empty.

PR #7393, opened 2019-03-17
Fix #7389 calendar bindDocumentClickListener in order to avoid flickering.

### LINKED ISSUE #7389: Calendar TouchUI doesn't open
[x] bug report

Bug can be reproduced on PrimeNG calendar demo for TouchUI in Chrome 73 - Windows and Mac
https://www.primefaces.org/primeng/#/calendar

**Current behaviour**
Click on the input to open calendar overlay - The overlay flickers then closes again.

**Expected behaviour**
The overlay should open as it does in any other browser or mode

Kled9:
###Defect Fixes
Fix #7389 Nevertheless, I'm not if It was the best solution to avoid flickering on calendar.


jensengar:
Great find. Trying to get this moved up in priority.

cagataycivici:
Fixed with a different approach thank you.

Issue #7398, opened 2019-03-18
onFileSelect EventEmitter not returning files


hidigitech:
onFileSelect event emitter is returning the variable `files` it should return `this.files`. The EventEmitter is not outputting the files. 

https://github.com/primefaces/primeng/blob/05e28dedfac6c3a11d50e60e4707ddb6b748ad24/src/app/components/fileupload/fileupload.ts#L214

rafa-as:
Hi @hidigitech your fix seems to be incorrect because the onSelect emitter must emit only the new ones and your approach emit all the files.

I think that the solution should be something like:

Before

```typescript
if(!this.multiple) {
    this.files = [];
}

let files = event.dataTransfer ? event.dataTransfer.files : event.target.files;
for(let i = 0; i < files.length; i++) {
    let file = files[i];

    if(!this.isFileSelected(file)){
      if(this.validate(file)) {
          if(this.isImage(file)) {
              file.objectURL = this.sanitizer.bypassSecurityTrustUrl((window.URL.createObjectURL(files[i])));
          }

          this.files.push(files[i]);
      }
    }
}

this.onSelect.emit({originalEvent: event, files: files});
```

After

```typescript
const newFiles = Array.from(event.dataTransfer ? event.dataTransfer.files : event.target.files);
const selectedFiles = [];

newFiles.forEach(file => {
    if(!this.isFileSelected(file) && this.validate(file)){
        if(this.isImage(file)) {
            file.objectURL = this.sanitizer.bypassSecurityTrustUrl((window.URL.createObjectURL(file)));
        }

        selectedFiles.push(file);
    }
})

if(selectedFiles.length > 0) {
    if(!this.multiple) {
        this.files = selectedFiles;
    } else {
        this.files = [...this.files, ...selectedFiles];
    }
    this.onSelect.emit({originalEvent: event, files: selectedFiles});
}
```

rafa-as:
@cagataycivici @mertsincan what do you think?

hidigitech:
> @cagataycivici @mertsincan what do you think?
> 
> If you want, I can do the PR quickly.

I'll close my PR.

hidigitech:
> > @cagataycivici @mertsincan what do you think?
> > If you want, I can do the PR quickly.
> 
> I'll close my PR.

Sorry, I thought you had done the PR. I'll make the changes and update my PR. Thanks @rafa-as

rafa-as:
Thank you @hidigitech

[...]

PR #7399, opened 2019-03-18
Fixed onSelect eventemitter to return list of files. #7398

hidigitech:
###Defect Fixes

###Feature Requests


rafa-as:
@cagataycivici some news here?

cagataycivici:
We just pass whatever is selected from the event so current behavior looks correct to me. Feel free to create a new ticket with a stackblitz that demonstrates the problem in detail.

rafa-as:
@cagataycivici the onSelect should be pass only the new ones and only if the validations are passed. Now is returning the previous values of the input, also when the input is populated by an endpoint.

You think that is correct?

Issue #6812, opened 2018-11-02
Accessibility: aria support for paginator


danielcollettepb:
**I'm submitting a ...**  (check one with "x")
```
[x ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
When tabbing through the paging controls with a screen reader, the first/previous/next/last buttons do not read text. Tabbing to the items per page dropdown does not show focus state.

**Expected behavior**
The screen reader should speak (vi aria-label) when tabbed to each control "First", "Previous", "Next", "Last". When tabbing to the items per page dropdown a focus state should be shown to indicate the current control.

**Minimal reproduction of the problem with instructions**
![image](https://user-images.githubusercontent.com/6116867/47897650-0143ce80-de49-11e8-8694-994d16b96c92.png)

**What is the motivation / use case for changing the behavior?**
Support for screen reader accessibility

* **Angular version:** 6

* **PrimeNG version:** 6.1.5

* **Browser:** MacOS: Chrome, Firefox, Safari

tharahan:
We are facing the same problem in PrimeNG 11.x. Any ETA on when will this be fixed?

rueeggl:
same here with PrimeNG 13.0.4. The fix would be very much appreciated.

komby:
We are seeing this error in our Accessibility testing, also using 13.0.4.  We also verified that the error occurs on https://www.primefaces.org/primeng/showcase/#/table/filter  when using the Axe tool it reports  
`{ 
    "ruleId": "button-name",
    "description": "Ensures buttons have discernible text",
    "help": "Buttons must have discernible text",
    "helpUrl": "https://dequeuniversity.com/rules/axe/4.3/button-name?application=AxeChrome",
    "impact": "critical",
    "isManual": false,
    "selector": ["#pr_id_3 > p-paginator > .p-paginator-bottom.p-paginator.p-component > .p-paginator-first.p-disabled.p-paginator-element"],
[...]

PR #7422, opened 2019-03-21
Accessibility: add aria labels to paginator

mattcfilbert:
###Defect Fixes
This PR addresses part of the accessibility improvements outlined in [issue #6812](https://github.com/primefaces/primeng/issues/6812), which refers to the paginator. This PR focuses on improvements that are localized to the paginator component (adding aria labels to each of the buttons).

A separate PR will address adding additional aria label support for the drop down.

###Feature Requests
This small PR adds aria labels to the paginator buttons, which allows users with screen readers to be better informed about each button's purpose (such as "Go to previous page")


padmini-chilappagari:
When can we expect this change?

Issue #5835, opened 2018-06-01
Accessibility: Dropdown needs customizable aria-labels or to be able associate aria-label with value


djpmsft:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
When using the screen reader with p-dropdown, the value selected will be read. ex: 'Ford'

**Expected behavior**
When using the screen reader with p-dropdown, a user must have the option of both the label and dropdown being read ex: 'Model: Ford'

**Minimal reproduction of the problem with instructions**
Turn on screen reader and move through a dropdown with keyboard navigation

**What is the motivation / use case for changing the behavior?**
Visually impaired users may get confused if they do not know what the value of a dropdown is associated with.

**Please tell us about your environment:**
Windows, NPM, Angular, VSCode

* **Angular version:** 5.1.3

* **PrimeNG version:** 5.2.5

* **Browser:** [all ]
 
* **Language:** [  TypeScript 2.8.0 ]

* **Node (for AoT issues):** `node --version` =

PR #7424, opened 2019-03-21
Accessibility: dropdown aria label category

mattcfilbert:
###Defect Fixes
This PR addresses accessibility improvements outlined in [issue #5835](https://github.com/primefaces/primeng/issues/5835), which refers to the drop down, and [issue #6812](https://github.com/primefaces/primeng/issues/6812), which refers to the paginator. This PR focuses on allowing the user to set the category for the aria labels of options in the drop down (as requested in #5835) and setting a default aria label category to clarify the purpose of the paginator dropdown (as requested in #6812)

A separate PR [(#7423)](https://github.com/primefaces/primeng/pull/7423) addresses adding additional accessibility support for the paginator nav buttons.

###Feature Requests
This small PR adds an aria label category to the drop down options, which allows users with screen readers to be better informed about each option's purpose (such as "Rows per page" in the paginator)

PS - If you prefer a function for adding the ': ' in the aria labels rather than nested ternaries, I am happy to make that change.


cagataycivici:
We'll do it on our end due to conflicts, thank you.

Issue #7358, opened 2019-03-08
Calendar touchUI on mobile devices - Cannot use due to keyboard popup


JacobSiegle:
When using touch UI on the calendar input should not happen to prevent the keyboard from displaying.

![Screenshot_20190308-155439](https://user-images.githubusercontent.com/5766272/54058208-cd900100-41ba-11e9-8dd1-15f715810597.png)

cyberrranger:
the same on ios, use demo https://www.primefaces.org/primeng/#/calendar to test ist

cagataycivici:
Fixed via #7555

cagataycivici:
I set inputmode="off" if touchui is on for 10.

PR #7467, opened 2019-03-29
#7358 - Keyboard popup using touchUI on mobile devices

JacobSiegle:
Note: https://github.com/primefaces/primeng/pull/7393 is related to how touchUI flickers, which obviously you will notice here also.

Touch UI should force the input to become readonly to prevent keyboard popup.

Alternatively if you can think of a scenario where touchUI could be used without setting the input to readonly I would recommend updating the touchUI docs to encourage setting readonly mode for mobile device support.

Note: I only confirmed this was fixed on an android device.

See screenshot in #7358.


cyberrranger:
push @JacobSiegle thx for your change

shogogan:
this solution blocks typing if you use `[showIcon]='true'` and `[showOnFocus]='false'`

JacobSiegle:
Maybe I don't understand your comment @shogogan, but it's supposed to block typing no matter what since it's setting readonly when touchUI is enabled. Are you saying you want the person to be able to type behind the popup like the screenshot below for some reason?

Screenshot_20190416-114027

shogogan:
if you set [showOnFocus]='false' it will only open if you press the calendar, so the user can still type on it, i use a mask on it so i want to display the calendar ONLY if the user wants it by pressing the icon on the input. that's why it shouldn't assume that if it's set as touchUI it will always use the calendar

JacobSiegle:
Thanks for the feedback I tried it quickly and see what you mean. I will instead make another PR for the docs which recommends enabling readonly mode when using touchUI like I originally mentioned.

shogogan:
fixed it on the #7358 PR :)

Issue #6921, opened 2018-11-27
OverlayPanel content projection causes bugs in numerous components


JacobSiegle:
**Example**
![tooltip](https://user-images.githubusercontent.com/5766272/49091266-0d8f3180-f225-11e8-906e-425acb77d6b4.png)

I'll make a stackblitz later.

JacobSiegle:
https://stackblitz.com/edit/github-jzz1ht

To replicate -
1) Open the overlay by clicking the button.
2)  Hover over the button in the overlay which will show the tooltip. 
3) Click outside the overlay on the right side.
4) The overlay will begin to close since you clicked outside.
5) Quickly move your mouse back on top of the button in the closing overlay.
6) The tooltip will still be showing (like the screenshot above) even when the overlay has been removed.

JacobSiegle:
Upon further investigation it seems like this is caused due content projection by the overlayPanel (https://bit.ly/2zUW1h4) and the fix is to use ngTemplateOutlet like the turbo table does. Currently it seems like overlayPanel will leak memory from failure to call life cycle looks (correctly me if this is wrong) which is also why the tooltip will stick around because onDestory isn't being called. 

I don't think I can attempt to fix this any further due to significant changes it needs (new overlayPanel component).

JacobSiegle:
I'm assuming this can happen with other components, but calendar is breaking when being used in overlay due to this bug.

Open the overlay, open the calendar, while both have the popup shown click on the page body -> calendar won't open again.
https://stackblitz.com/edit/github-wvne9h

JacobSiegle:
Some other components i've tested - 
Dropdown, multiselect, color picker will require double click for it to open.
p-dialog perma breaks like calendar.

cagataycivici:
Shouldn't this be fixed by Angular then, this means content projection is defective.

cagataycivici:
Event with ngTemplateOutletContext, tooltip stays on quick mouse move, that should be fixed on tooltip side I guess. 

But lifecycle hooks being ignored is a problem, we've decided to migrate to ngTemplateOutlet and deprecate-drop content-projection.

https://github.com/primefaces/primeng/issues/8677

cagataycivici:
We're migrating all ng-content to ngTemplateContext while maintaining backwards compatibility until PrimeNG 11.

PR #7468, opened 2019-03-29
#6921 - Content projection causes errors by not calling on destroy properly - Breaking Change

JacobSiegle:
Note: This is a breaking change so this is more of a quick hack fix copy paste from turbo table. 

#6921 is listed as a defect but after investigating content projection I think it's a bigger issue then I originally thought and edited it accordingly. 

I have not attempted to fix this everywhere since large/breaking change PRs usually are done internally, but you can let me know if you would like a full PR otherwise. 

This issue leaks over to any other components which use ng-content instead of template outlet and have components that can be destroyed in them. You can read #6921 for the full notes.

You will notice in the HTML example I added it looks weird when the destroy happens since it goes from a button to a small dot. This is just due to how the button component destroys itself. You can add a breakpoint in the on destroy of button to easily replicate.


cagataycivici:
10 used ng-template instead of ng-content for all components where necessary, both are supported for backward compatibility.

Issue #7464, opened 2019-03-28
readonly chips


dolphinsd:
It will be nice to have readonly chips for display purposes or badges component

JacobSiegle:
I was looking for something like this the other day actually. I ended up just using a bunch of rounded buttons that were disabled.

PR - https://github.com/primefaces/primeng/pull/7469

luis-paulo-tdo:
It's possible turn Chips readonly following these steps:

1. Disable <p-chips> component
`<p-chips [disabled]="true">`

2. Set display property to none for ui-chips-token-icon class on CSS
`.ui-chips-token-icon {
      display: none;
    }`

KaKi87:
> Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation.

It has not : https://primevue.org/chips/

Thanks

PR #7469, opened 2019-03-29
#7464 Chips readonly

JacobSiegle:
#7464 - Allows chips to be readonly.

Also fixed a bug I noticed where when readonly or disabled was used I could still hit the backspace to delete chips.


cagataycivici:
We'll do it on our end based on new styling thank you.

Issue #7445, opened 2019-03-26
p-spinner binds strings instead of numbers on manual input in PrimeNG 7.1.0


simdevmon:
**I'm submitting a ...** 
```
[x] bug report 
```

**Plunkr Case (Bug Reports)**

https://github-pqpa62.stackblitz.io/

**Current behavior**
When a number is entered manually in a p-spinner, the attribute in the class is a string instead of the number. If the up and down buttons of the spinner are used, the spinner works correctly. 

**Expected behavior**
I would expect that the attribute is always a number for both manual input and usage of the buttons.

**Minimal reproduction of the problem with instructions**
Open the stackbiz example and the browser console: https://github-pqpa62.stackblitz.io/

Change the number with the button -> it will output a number correctly
Enter a number manually, it will be disabled as a string.

![2019-03-26_09h24_29](https://user-images.githubusercontent.com/16277781/54983067-5f19b400-4fac-11e9-9cca-82163775fb93.png)

* **Angular version:** 7.2.10

* **PrimeNG version:** 7.1.0

* **Browser:** all

Issue #7244, opened 2019-02-14
spinner bug


sourdoth:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting

**Demo bug report:**
https://github-dmcqc2-xc72xz.stackblitz.io/

**Current behavior**
hi, when i set decimalSeparator="," and thousandSeparator="." not working, when i write "1,5" the "," change to "." (1.5) and thousandSeparator not show any "." separate

**Expected behavior**
when i set decimalSeparator="," spinner not chanche "," to "."
when i set thousandSeparator="." i expected to show the "." when i write 1000 chanche to 1.000

**Minimal reproduction of the problem with instructions**

demo -> https://github-dmcqc2-xc72xz.stackblitz.io/

<p-spinner size="30" [(ngModel)]="val" thousandSeparator="." decimalSeparator="," [min]="0" [step]="0.01"></p-spinner>


* **PrimeNG version:** 7.0.5
<!-- Check whether this is still an issue in the most recent Angular version -->

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
<!-- All browsers where this could be reproduced -->
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

* **Node (for AoT issues):** `node --version` =

Issue #6610, opened 2018-10-02
Thousand separator missing


Pikadili83:
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
https://nbwiamqn.github.stackblitz.io

**Current behavior**
Thousand separator is missing

**Expected behavior**
Thousand separator should be shown. This works on previous versions.

**Minimal reproduction of the problem with instructions**
Add p-spinner component to the template.

**Please tell us about your environment:**

* **Angular version:** 5.X
6.1.9

* **PrimeNG version:** 5.X
6.1.4

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

* **Node (for AoT issues):** `node --version` =

Pikadili83:
This problem is still present in version 6.1.5. Is there a plan to fix it?

PR #7480, opened 2019-04-01
#7445 - Spinner can emit string

JacobSiegle:
#7445 - Fixes spinner emitting a string when being typed into. Additional fixed the docs and made the spinner always a number type since separators have been removed.

These can also be closed since seperaters were removed - 
https://github.com/primefaces/primeng/issues/7244
https://github.com/primefaces/primeng/issues/6610

I think this one is fixed also since it refs strings causing the issue which can no longer be used in spinner - 
https://github.com/primefaces/primeng/issues/3058


cagataycivici:
Will use onInput instead, thank you.

Issue #6445, opened 2018-09-03
Ratings - Show half units instead of integer


rafa-suagu:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Rating component can only show integer units.

**Expected behavior**
Show half units.

**What is the motivation / use case for changing the behavior?**
This component can be used to show ratings and the ratings normally are allowed to get a half number.

rafa-as:
@cagataycivici Would you accept a PR for this feature?

JacobSiegle:
Useful ref - https://stackblitz.com/angular/nnjppodbojn?file=app%2Frating-decimal.ts

rafa-as:
Thank you @JacobSiegle

rafa-as:
@cagataycivici some news?

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

rafa-as:
@mertsincan I can't reopen the issue and we're continuing with no option to display half units (or a concrete value)

cetincakiroglu:
Issue tracker is used for defects only as part of our commitment to quality and continuous improvement in all areas. Enhancements are collected as valuable community feedback and managed internally so moving this enhancement ticket to our internal project management backlog.

PR #7484, opened 2019-04-01
#6445- Allow half star ratings

JacobSiegle:
#6445 - Allows rating component to do half star ratings. 

I use this for my own personal use for a dashboard, just wanted to share incase anyone else might need it. This has some accessibility issues since you cannot tab on to a half star rating or anything like that. It calcs the half star rating by detecting if you clicked on the first or second half of the icon. I implemented it this way since it seemed like the easiest way for no breaking changes in case someone else wanted to pull it in.

### DIFF
--- a/src/app/components/rating/rating.css
+++ b/src/app/components/rating/rating.css
@@ -1,3 +1,14 @@
 .ui-rating {
     font-size: 1.25em;
 }
+
+.empty-icon {
+    position: relative;
+    display: inline-block;
+}
+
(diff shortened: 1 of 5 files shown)


JacobSiegle:
I only put it here for reference in case someone wanted to manually pull it into their own repo. You can still ref it from the linked issue. I assume the maintainer won't use this anyway and will make a version with more breaking changes to get the half stars to work with proper accessibility.

Issue #7490, opened 2019-04-02
onPush detection for TreeTable component


boryskryvous:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
There is to much ngFor calls, when where is no changes for `pTreeTableBody`

**Expected behavior**
ngFor should run only when changes detected

**Minimal reproduction of the problem with instructions**
To reproduce and also to analys perfomence of treetable just add this function into your component:
```
trackByFn(index, item) {
    console.log(`${index}`, item);
    return item.node.data.id; 
  }
```
and add this function into your template 

```
<p-treeTable
  #treeGrid
  class="tree-table flexcolumn"
  [value]="data"
  [columns]="columns"
  (onNodeCollapse)="collapsedRow($event)"
  (onNodeExpand)="expandedRow($event)"
  (onEditCancel)="returnFocus()"
  (onEditComplete)="returnFocus()"
  [selection]="currentRowNode"
  [scrollable]="true"
  scrollHeight="100%"
  [resizableColumns]="true"
  columnResizeMode="expand"
  dataKey="id"
  [rowTrackBy]="trackByFn">
```

**What is the motivation / use case for changing the behavior?**
I was working on data presentation into primeng treetable.  Because it was a lot of data i have had a bad perfomance and to solve it i'v added onPush detection to `pTreeTableBody` component. And now i have better perfomance with cell and row selection.
Given code just to be needed to add into treetable.ts file:
```
@Component({
    selector: '[pTreeTableBody]',
    template: `
        <ng-template ngFor let-serializedNode let-rowIndex="index" [ngForOf]="tt.serializedValue" [ngForTrackBy]="tt.rowTrackBy">
            <ng-container *ngIf="serializedNode.visible">
                <ng-container *ngTemplateOutlet="template; context: {$implicit: serializedNode, node: serializedNode.node, rowData: serializedNode.node.data, columns: columns}"></ng-container>
            </ng-container>
        </ng-template>
        <ng-container *ngIf="tt.isEmpty()">
            <ng-container *ngTemplateOutlet="tt.emptyMessageTemplate; context: {$implicit: columns}"></ng-container>
        </ng-container>
    `,
[...]

PR #7493, opened 2019-04-03
onPush detection for treetable component #7490

boryskryvous:
#7490


cagataycivici:
All components will be push in 9.1.0 and we'll review this on our side, thank you.

Issue #7499, opened 2019-04-04
problem with p-dialog


wassim21:
I have many dialog
![image](https://user-images.githubusercontent.com/27426327/55561183-a0673d80-56e9-11e9-9b4a-9e4c799eb855.png)
for exemple when I maximize one of dialog, the other dialog don't dissapear 
![image](https://user-images.githubusercontent.com/27426327/55561307-d73d5380-56e9-11e9-9f60-e0a26ed679cb.png)
how can I solve this problem ?

toddsr:
We have the same issue. And it's not just when you maximize a dialog. The need exists to be able to raise a dialog to the top when simply clicked on so it's not potentially covered by other dialogs.

JacobSiegle:
If you wanted to fix it asap you would detect when "maximized" changes in the dialog and then call the function moveOnTop. When the dialog is de-maximized I assume it should go back to the previous order it appeared in (under the other elements)?

JacobSiegle:
I PRed it like mentioned above.

PR #7507, opened 2019-04-04
#7499 - Maximized dialog to appear over other dialogs

JacobSiegle:
See issue #7499. When the dialog maximize is removed the z index returns to its prev set state to keep the dialog ordering.


cagataycivici:
I think simply calling moveOnTop on maximize() can fix it as well, not sure if we need to restore the old zindex, operating systems do not restore to old layer when a window is maximized and then restored.

Issue #7520, opened 2019-04-07
Add HTML support to Toast message detail


sdoubey:
```
[ ] bug report => Search github for a similar issue or PR before submitting
[*] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
**Current behavior**
We can't put html tags in message detail property for Toast message, they are displayed a regular characters

**Expected behavior**
Be able to add html tags like `<strong>`, `<i>`, ...

**What is the motivation / use case for changing the behavior?**
Sometimes i need to highlight some parts of message

**Please tell us about your environment:**

* **Angular version:** 7.2.7

* **PrimeNG version:** 7.1.0

* **Browser:** all
 
* **Language:** TypeScrip 3.2.4

cagataycivici:
Toast provides an ng-template for HTML support so please use it instead.

PR #7522, opened 2019-04-07
Add HTML support to Toast message detail - #7520

sdoubey:
Add HTML support to Toast message detail using innerHtml attribute - #7520


cagataycivici:
Toast provides an ng-template for HTML support so please use it instead.

Issue #7540, opened 2019-04-11
p-multiselect disabled and selected option gets un-selected by the select all checkbox in header


rdustin:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
https://github-nanxuo.stackblitz.io

**Current behavior**
If a SelectItem is disabled and also selected by default, if you choose the select all checkbox in the p-multiSelect element header it will un-select the disabled SelectItem and it cannot be re-selected.

**Expected behavior**
Any disabled selectItems should not be affected by the select all checkbox.

**Minimal reproduction of the problem with instructions**
Set a SelectItem to disabled.
```
this.cars = [
            {label: 'Audi', value: 'Audi'},
            {label: 'BMW', value: 'BMW'},
            {label: 'Fiat', value: 'Fiat', disabled: true},
            {label: 'Ford', value: 'Ford'},
            {label: 'Honda', value: 'Honda'},
            {label: 'Jaguar', value: 'Jaguar'},
            {label: 'Mercedes', value: 'Mercedes', disabled: true},
            {label: 'Renault', value: 'Renault'},
            {label: 'VW', value: 'VW'},
            {label: 'Volvo', value: 'Volvo'},
        ];
```
Set that same SelectItem as selected.
```
selectedCars1: string[] = [
      'Audi',
      'BMW',
      'Fiat'
    ];
```
Check the select all checkbox in the multiselect header and see that the disabled SelectItem has been un-selected and cannot be re-selected.

**What is the motivation / use case for changing the behavior?**
In certain instances an interface may need an item that is displayed as checked but should not be changed. For example, an interface where assigning users to roles, but all users have a default role that should not be changed by anyone..

**Please tell us about your environment:**
MacOS, Visual Studio Code, npm

* **Angular version:** 7.2.8

* **PrimeNG version:** 7.1.0
This can be replicated on the master branch

* **Browser:** Chrome 73.0.3683.86
 
* **Language:** TypeScript 3.2.4 | ES5

PR #7541, opened 2019-04-11
Add fix for disabled checkboxes - see issue #7540

rdustin:
###Defect Fixes

###Feature Requests

This will fix the issue reported in #7540


cagataycivici:
Focus is applied to the th element so aria props should be there.

https://github.com/primefaces/primeng/issues/7766

PR #7542, opened 2019-04-11
RTL support for Dropdown

liozassa:
Full support rtl for dropdown
* Add property "rtl" to the component dropdown.
* Add example to demo page and description to properties table.

I started with dropdown component, but i continue to all components.


cagataycivici:
This is something we'd like to do on our own with the new PrimeNG design system, thank you for your feedback.

PR #7553, opened 2019-04-16
Fixed #7358 - Removing focus on showOverlay if touchUI is activated

### LINKED ISSUE #7358: Calendar touchUI on mobile devices - Cannot use due to keyboard popup

shogogan:
###Defect Fixes
Solved the #7358 problem with a blur on the input when showing the overlay (only if the touchUI is enabled)


cagataycivici:
Hi, what does the PR do? I thought we had documented it with @JacobSiegle 's patch.

shogogan:
The problem still exists as i mentioned on @JacobSiegle PR (setting fixed the readonly).

This PR just calls the blur on the input when showing the overlay with the touchUI enabled (with no focus on the input the keyboard will not be shown, making readonly really an option and not a requirement to make it work properly).

JacobSiegle:
I closed #7467 and made #7555 which only updates the docs instead of forcing readonlyInput when touchUI is used. I'm not sure what this PR does exactly, maybe @shogogan can explain or there's just some confusion since I replaced #7467 with #7555.

shogogan:
@JacobSiegle While your solution was blocking it, mine was making sure that it doesn't have focus. if the input doesn't have focus the smartphone doesn't show the keyboard, and the touchUI will work normally (i tested on an iphone 6 and a moto g2), fixing the issue without needing to make it readonly.


The tests that travis is failing are from radiobuttons, not any changes that i did here btw.

JacobSiegle:
Thanks I believe I understand. @shogogan PR is basically just saving the user having to set readonlyInput by always calling blur. In his scenario this also allows the user to type when using the combo of [showIcon]='true' and [showOnFocus]='false' because showOverlay is not called which means the blur he put in is never called.

So overall I believe #7555 could be undone for this as this saves the user having to set readonlyInput. I think the only reason to keep #7555 over this would be some other scenario where you needed to manage the readonlyInput state but I cannot think of one off the top of my head.

shogogan:
Just so it can be fully understood why I need the showIcon true and showOnFocus false:
https://stackblitz.com/edit/github-erqvb2-wvtlhj

on this stackblitz there is a mask on the calendar (try typing exactly 05182015), and the button is the one that opens the calendar if the person needs it

cagataycivici:
I set inputmode="off" if touchui is on.

PR #7555, opened 2019-04-16
Touch UI Mobile Docs

JacobSiegle:
###Defect Fixes
https://github.com/primefaces/primeng/issues/7358

Recommend to users that use touchUI to set readonlyInput to true to prevent keyboard popup for mobile users.

I made a previous PR to always force readonlyInput for touch UI but a use case where that was not wanted was shown here - https://github.com/primefaces/primeng/pull/7467


cagataycivici:
I went ahead and make it built-in with the following, thank you.

```html
[readonly]="readonlyInput||touchUI"
```

Issue #7583, opened 2019-04-24
p-dropdown not closing when clicking outside of the control itself


DrJoStiwa:
If you have a PrimeNG PRO Support subscription please post your issue at;

https://pro.primefaces.org 

where our team will respond within 4 business hours.

If you do not have a PrimeNG PRO Support subscription, fill-in the report below. Please note that
your issue will be added to the waiting list of community issues and will be reviewed on a first-come first-serve basis, as a result, the support team is unable to guarantee a specific schedule on when it will be reviewed. Thank you for your understanding.

Current Waiting Time: ~8 weeks.

**I'm submitting a ...**  (check one with "x")
```
[ x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
May be related to #2739.
The dropdown doesn't close when clicking outside of it. In the specific case the dropdown is
part of a table. In the same table there are some p-multiSelect controls in use, which work as expected.

**Expected behavior**
The dropdown should close when any other control or body is clicked.

**Minimal reproduction of the problem with instructions**
The issue is not easy to reproduce, but by debugging and comparison I noticed that the DocumentListener of the p-dropdown is not receiving the first click. 
When comparing to p-multiSelect, there's a difference in the method **show()**, which is not binding the document click listener.
When adding the according call, the p-dropdown works in all situations I've tested.

* **PrimeNG version:** 7.0.3

DrJoStiwa:
Created proposed fix and pull request for this issue https://github.com/primefaces/primeng/pull/7584

DrJoStiwa:
Found an additional problem with ChangeDetection.OnPush, where the hide event is not triggered and the users of the control have no possibility to register for a OnHide event.
Proposed a second fix based on handling similar to multiselect control.

cagataycivici:
I'm unable to replicate with the latest PrimeNG. If the issue still persists with the new version please create a new issue with our stackblitz template.

PR #7589, opened 2019-04-24
Allow to handle hide() with ChangeDetection.OnPush

DrJoStiwa:
Second part of fix for #7583

###Defect Fixes

###Feature Requests


cagataycivici:
But at hide, the dom is not hidden yet.

PR #7605, opened 2019-04-26
Possibility to save custom settings for table.

andriikuts:
I have developed columns toggle directive (show/hide columns).
I want to save information about visible columns in object with all table settings (in future i will save it in database), but i can not do this because all custom properties are removed after save.
I propose to create a property where developer can save some custom table settings.


cagataycivici:
See https://github.com/primefaces/primeng/issues/7666

PR #7607, opened 2019-04-26
Table state saved event

andriikuts:
I need this event to get settings from storage on full update and send JSON to database.
Now i do not have possibility to do this.


cagataycivici:
The PR has no docs so unable to accept but we'll do it on our side.

cagataycivici:
https://github.com/primefaces/primeng/issues/7666

Issue #5713, opened 2018-05-10
Dialog can be resized/dragged outside the viewport


lasimone:
This is my first github issue post. I hope this contains enough information:

**Current behavior**
Dialog can be resized/dragged beyond the right and bottom edge of the viewport. 

**Expected behavior**
The dialog should not be allowed to be dragged or resized beyond the top, left, bottom and right edges of the viewport. (The top and left work accordingly, its just the right and bottom than need to be fixed)

**Minimal reproduction of the problem with instructions**
STEPS TO REPRODUCE:
1. Open a draggable and resizeable dialog

2. Drag the dialog to the top left corner until it hits the edge of the viewport (based on https://github.com/primefaces/primeng/issues/5312 the dialog should not be able to go beyond these edges)

3. Resize the dialog as far right off the browser screen as you can (I have multiple monitors so I can stretch it quite far. Making the browser smaller than a single monitor screen and resizing works too)

4. Notice that the dialog is now bigger than the viewport and there is no way to close or resize it. This is because the left and top of the viewport restrict the dialog from moving in those directions. No matter how you drag the dialog, there is no way to fit it back within the viewport.

**What is the motivation / use case for changing the behavior?**
To make resizing and dragging dialogs more user friendly and stable.

**Please tell us about your environment:**
Operating system: Windows
IDE: Microsoft Visual Studio
Package manager: NPM 

* **Angular version:**
5.2.10

* **PrimeNG version:**
5.2.6

* **Browser:**
All
 
* **Language:**
TypeScript 2.6.2
* **Node (for AoT issues):** 
`node --version` =   6.12.3

Issue #7476, opened 2019-04-01
Dialog can't be resized/dragged outside the viewport


hakim-dc:
this issue is basically the original behavior of dialog before the modification that has been done in this issue #5713.

**I'm submitting a ...**
```
[ ] bug report
[X] feature request
[ ] support request
```

**What is the motivation / use case for changing the behavior?**
To make resizing and dragging dialogs outside the viewport optional.

**Please tell us about your environment:**
Operating system: ubuntu
IDE: Microsoft Visual Studio
Package manager: NPM

- **Angular version:** 7.2.11

- **PrimeNG version :** 7.1.0

- **Browser:** All

- **Language:** TypeScript 3.3.4

JacobSiegle:
Please don't duplicate issues - #7420

yigitfindikli:
I’m unable to replicate with the latest PrimeNG. If the issue still persists with the new version please create a new issue with our stackblitz template.
https://stackblitz.com/github/primefaces/primeng-issue-template

Issue #7420, opened 2019-03-21
Drag p-dialog outside of viewport


toddsr:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Issue #5713 fixed being able to drag a dialog outside of the viewport in version 7. But in our case, we require the ability to drag the dialog outside of the viewport which allows users to see/access content easily behind the dialog - especially for non modal dialogs.

We can use minX and minY to allow the dialog to be dragged off the viewport to the left and top by setting those to negative values, but we need the same ability to drag the dialog off the viewport to the right and bottom.

**Expected behavior**
Drag p-dialog outside of the viewport in any direction.

**Please tell us about your environment:**

* **Angular version:** 7.2.7

* **PrimeNG version:** 7.0.5

* **Browser:** [Chrome XX ]

 
* **Language:** [TypeScript 3.2.2]

hakim-dc:
1+

Yoranium238:
I concur with this request.  Those applications that require significant dialog interaction should be able to move those dialogs around to positions that extend beyond the viewport.  I think a valid approach to accommodate this would be to limit this movement to some configurable  boundary restriction where you cannot move the dialog completely out of view.

hakim-dc:
@cagataycivici  Are there any plans to fix this in the incoming release?

yigitfindikli:
I’m unable to replicate with the latest PrimeNG. If the issue still persists with the new version please create a new issue with our stackblitz template.
https://stackblitz.com/github/primefaces/primeng-issue-template

PR #7638, opened 2019-05-05
Dialog Movement - Body offset is always 0 in PrimeNG

JacobSiegle:
#5713 fixed moving a dialog out of the view port. This causes a bug with domhandler offset because the PrimeNG document body always returns 0 as the scrollTop and scrollLeft. If you have an app where the body scroll return values for scrollTop and scrollLeft it breaks dragging the dialog on the page once any scrolling has happened.

https://stackblitz.com/edit/github-hsl3wk?file=src/styles.css

Steps (note only shows drag up and down failing):
Open the dialog
Move it around
Scroll down and it will fail to move around
Scroll back up to top and it will move around

### DIFF
--- a/src/app/components/dom/domhandler.ts
+++ b/src/app/components/dom/domhandler.ts
@@ -339,8 +339,8 @@ export class DomHandler {
         let rect = el.getBoundingClientRect();
         
         return {
-            top: rect.top + document.body.scrollTop,
-            left: rect.left + document.body.scrollLeft
+            top: rect.top,
+            left: rect.left
         };
     }
 


cagataycivici:
Dangerous PR as getOffset is used in many places which require scroll info, also I can't replicate it on showcase with 7.1.2.

JacobSiegle:
As mentioned above the PrimeNG showcase will never have this problem because the showcases document body scroll will always return 0 so the removed code never makes a difference in the context of the primeng showcase.

Testing every other component that uses the offset and it seems to make no difference to them, they all interact the same way in my limited testing.

Here is an updated plunker that show more components - https://stackblitz.com/edit/github-hsl3wk
Example in repo - https://github.com/JacobSiegle/primeng/tree/Jacob/BodyScrollFixExample

At the very least would you accept a PR which would add an additional param which would allow moving the dialog outside the viewport again (undoing #5713 via a param)? Other requests exist for this same feature - https://github.com/primefaces/primeng/issues/7476, https://github.com/primefaces/primeng/issues/7420

hakim-dc:
Are there any plans to fix this in the incoming release?

Issue #7640, opened 2019-05-06
[Table] Filtering with ChangeDetectionStrategy.OnPush show the previous state


rafa-as:
**I'm submitting a ...** 
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Stackblitz project: https://stackblitz.com/edit/table-error-onpush
Demo APP: https://table-error-onpush.stackblitz.io

The error is on the **src/app/components/table/table.ts:1234** :

``` Typescript
this.filteredValue.push(this.value[i]);
```

should be:

``` Typescript
this.filteredValue = [...this.filteredValue, this.value[i]];
```

**Current behavior**
When filter with ChangeDetectionStrategy.OnPush, the table shown inconsistent data (the previous state, not the current).

**Expected behavior**
The table show the new values properly without any extra implementation.

**Minimal reproduction of the problem with instructions**
Use the table component inside another one with **ChangeDetectionStrategy.OnPush** active

**Please tell us about your environment:**
MAC OS X Mojave 10.14.4
WebStorm 2019.1.2
Yarn 1.15.2

* **Angular version:** 7.2.1

* **PrimeNG version:** 7.1.2

* **Browser:** all
 
* **Language:** all

cagataycivici:
Duplicate https://github.com/primefaces/primeng/issues/6236

Issue #6236, opened 2018-08-03
p-table: filtering doesn't work with ChangeDetection: OnPush


ghost:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Minimal reproduction of the problem with instructions**
https://stackblitz.com/edit/angular-tgcclb?file=src/app/app.component.html

1. set Vin filter to `a1`
![image](https://user-images.githubusercontent.com/25089646/43674015-621a0b1c-97d5-11e8-9338-a67c381b1d35.png)
2. Remove entered value
![image](https://user-images.githubusercontent.com/25089646/43674012-53149d94-97d5-11e8-9095-ee04375bd323.png)

Workaround: https://stackblitz.com/edit/angular-cvqxne?file=src/app/app.component.html

* **Angular version:** 6.1.1
* **PrimeNG version:** 6.0.2
* **Browser:** all

maxkorz:
@fxmfr the workaround provided in the first post still works

jamieathans:
Looks like just the presence of an empty onFilter function will make on OnPush change detection work ...

https://stackblitz.com/edit/angular-jgeky3

rafa-as:
This solution is not correct because the problem is the array.push solved in my PR https://github.com/primefaces/primeng/pull/7641/files

lucioleandro:
Rafa, i tested your solution, it works, but with some bugs

rafa-as:
@lucioleandro can you be more specific? thanks

annnoo:
Any update on this? It still does not work for me ...

lucioleandro:
> @lucioleandro can you be more specific? thanks

Your solution worked, but sometimes failed, not filtering the table or not filtering correctly.
As I had no time, I didn't investigate the cause and looked for another solution.

rafa-as:
> > @lucioleandro can you be more specific? thanks
> 
> Your solution worked, but sometimes failed, not filtering the table or not filtering correctly.
> As I had no time, I didn't investigate the cause and looked for another solution.

I don't know exactly but seems another different bug. My approach only solves the onPush.

PR #7641, opened 2019-05-06
#7640 fix table filtering using ChangeDetectionStrategy.OnPush

rafa-as:
https://github.com/primefaces/primeng/issues/7640

* Use spread operator instead of array.push, to prevent errors with ChangeDetectionStrategy.OnPush

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -1231,7 +1231,7 @@ export class Table implements OnInit, AfterViewInit, AfterContentInit, Blockable
                     }
     
                     if (matches) {
-                        this.filteredValue.push(this.value[i]);
+                        this.filteredValue = [...this.filteredValue, this.value[i]];
                     }
                 }
     


cagataycivici:
Duplicate https://github.com/primefaces/primeng/issues/6236

rafa-as:
@cagataycivici the solution https://github.com/primefaces/primeng/issues/6236 is wrong. The problem with array.push is a basic to get work ChangeDetection.OnPush....

rafa-as:
@cagataycivici This change causes memory leaks and a very bad performance on complex tables with frozen columns / rows, editable cells, and a lot of table features. The reason is because the detectChanges function, marks for check the component and all the children components recursively. 

Please, use the spread operator instead and remove the detectChanges.

You can read more here for example:  https://chrislo.ca/angular-2345-change-detection-on-data-bound-array-pushunshift-popshift-or-splice/

rafa-as:
@cagataycivici can you check this please? it's very easy to solve the memory leaks. In our case, we have a very big complex table and with this change it has become unusable.

lucioleandro:
Rafa, i tested your solution, it works, but with some bugs

rafa-as:
@lucioleandro, what bugs?

Issue #7149, opened 2019-01-19
p-fileUpload component has bug on select file


wszgrcy:
Current Waiting Time: ~1weeks.

**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**

 When I click choose btn(like button but a span ),I click on icon but can't open select file.
so I saw the source ,I find the icon has keep out the input Tag,so I add css `z-index:1`,and click ,
success open dialog
**Expected behavior**

input `file` add css `z-index:1` 
**Minimal reproduction of the problem with instructions**

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

* **Angular version:** 

official website

* **PrimeNG version:** 

4.x 7.x (I think all version has this because the latest version has this bug)
* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 Chrome latest
* **Language:** 
official website

* **Node (for AoT issues):** `node --version` =

rafa-as:
@cagataycivici some news here?

PR #7660, opened 2019-05-09
#7149 Clicking icon in advanced fileupload button does not open filechooser

matthiasblaesing:
The icon in the "Choose" button of the advanced fileupload button is
positioned absolute and is defined after the <input type="file">
element, which is also position=absolute.

So when a click is done in the region of the icon, the event is not
recorded on the file upload element, but on the icon and thus swallowed.

The solution is to raise the file element over the other elements via
a z-index style. This is considered to be save, as the button creates
a new stacking context.

The referenced issue was created independend of my analysis, but we
reached the same conclusion.

A sample project demostrating the problem can be found here:

https://github.com/matthiasblaesing/primengupload-icon-insensitive

### DIFF
--- a/src/app/components/fileupload/fileupload.css
+++ b/src/app/components/fileupload/fileupload.css
@@ -77,6 +77,7 @@
     filter: alpha(opacity=0);
     direction: ltr;
     cursor: pointer;
+    z-index: 1;
 }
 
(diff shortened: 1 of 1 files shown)


cagataycivici:
Thank you

Issue #7669, opened 2019-05-10
Accessibility: SortIcon calculates aria text, but does not bind the template to it


matthiasblaesing:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
When setting `ariaLabel`, `ariaLabelAsc` and `ariaLabelDesc` properties on sortIcon, no `aria-label` is created.

**Expected behavior**
The `aria-label` attribute should show the text, that matches the associated property.

**What is the motivation / use case for changing the behavior?**
Support visible impaired users.

* **PrimeNG version:** 7.1.2

* **Browser:** all
 
* **Language:** all

cagataycivici:
Focus is applied to the th element so aria props should be there.

https://github.com/primefaces/primeng/issues/7766

PR #7670, opened 2019-05-10
#7669: SortIcon calculates aria text, but does not bind the template to it

matthiasblaesing:
It was noted that sortable tables do not give useful info about the
sort state for blind users. We tried to use the ariaLabel, ariaLabelAsc
and ariaLabelDesc properties, but they did not improve the situation.

Analysis in the inspector showed, that the aria-label attribute is not
updated, as I would expect it to be.

This change binds the calculated SortIcon#ariaText to the templates
aria-label attribute.


cagataycivici:
Duplicate

Issue #7684, opened 2019-05-14
Issue with indexing of current page in currentPageReport


BogdanMaier:
**I'm submitting a ...** 
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
https://github-47e8gw.stackblitz.io

**Current behavior**
When displaying the pageReport the current page index is 0 based. 

**Expected behavior**
My suggestion is to use 1 based indexing for the current page value. 

**Minimal reproduction of the problem with instructions**
https://github-47e8gw.stackblitz.io

**What is the motivation / use case for changing the behavior?**
I want to use the pagination report. This happens for independently defined paginator and included paginators aswell.

**Please tell us about your environment:**

* **Angular version:** 8.0.0-beta.12

* **PrimeNG version:** 7.1.2

* **Browser:** [all]
 
* **Language:** [all]

BogdanMaier:
Fixing PR: https://github.com/primefaces/primeng/pull/7685

PR #7685, opened 2019-05-14
primeng-7684: fix issue with 0 based index

BogdanMaier:
Fixed issue with  0 based indexing for the current page in currentPageReport template value setting.


BGBRWR:
Fixes #7684

PR #7695, opened 2019-05-16
Proposed fix in dropdown.ts for Issue: #7694 (handle clicking outside of dropdown)

### LINKED ISSUE #7694: p-dropdown not closing when clicking outside of the control itself
```
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-mgnfo9

**Current behavior**
May be related to #2739.
The dropdown doesn't close when clicking outside of it. In the specific case the dropdown is
part of a table. In the same table there are some p-multiSelect controls in use, which work as expected.

**Expected behavior**
The dropdown should close when any other control or body is clicked.

Steps to Reproduce:
- Click the dropdown to open
- Click on an item (dropdown not closing)
- Click a second time (dropdown closes)
- Click to open dropdown again
- Click outside the dropdown, dropdown not closing
- Click Button "Run Changedetection" to trigger manual CD and this will close the dropdown

I noticed that the DocumentListener of the p-dropdown is not receiving the first click.
When comparing to p-multiSelect, there's a difference in the method show(), which is not binding the document click listener.
When adding the according call, the p-dropdown works in all situations I've tested.
Additionally the onHide Event is not triggered until the View is closed, which is problematic when using ChangeDetectionStrategy.OnPush and detached Changedetection. The client of the Dropdown can not run ChangeDetection (which would the dropdown-view), unless it gets notified in any way.
This is also working with the p-multiselect component and done similarly there, according to the proposed pull request.

* **PrimeNG version:** 7.1.2

DrJoStiwa:
###Defect Fixes
#7694.

### DIFF
--- a/src/app/components/dropdown/dropdown.ts
+++ b/src/app/components/dropdown/dropdown.ts
@@ -513,6 +513,7 @@ export class Dropdown implements OnInit,AfterViewInit,AfterContentInit,AfterView
     
     show() {
         this.overlayVisible = true;
+        this.bindDocumentClickListener();
     }
 
     onOverlayAnimationStart(event: AnimationEvent) {
@@ -543,7 +544,6 @@ export class Dropdown implements OnInit,AfterViewInit,AfterContentInit,AfterView
             break;
 
             case 'void':
-                this.onHide.emit(event);
                 this.onOverlayHide();
             break;
         }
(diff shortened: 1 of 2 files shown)


cagataycivici:
I can't replicate the issue to begin with.

Issue #7693, opened 2019-05-16
Only inside p-overlayPanel, p-calendar has month/year navigation problem


darius-iko:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-mkrr8t

**Current behavior**
Clicking on the month navigation button either forward or backward will cause overlay panel behind and the calendar itself to close. (This issue does not exist on PrimeNG version 7.0.4 and older).

**Expected behavior**
Calendar should navigate to next/previous month instead of closing overlay component and the calendar itself.

**Minimal reproduction of the problem with instructions**
Step to reproduce:
1. Open URL https://stackblitz.com/edit/github-mkrr8t
2. Click on the button "Open Overlay"
![image](https://user-images.githubusercontent.com/8023913/57848400-5f476a80-780b-11e9-9441-20a247ba1909.png)
3. Click on the calendar button to open the calendar overlay.
![image](https://user-images.githubusercontent.com/8023913/57848535-b64d3f80-780b-11e9-862e-1a0da565d148.png)
4. Click on the month navigation button either forward or backward will cause overlay panel behind and the calendar itself to close.
![image](https://user-images.githubusercontent.com/8023913/57848648-f90f1780-780b-11e9-830e-2669eb53d754.png)

**What is the motivation / use case for changing the behavior?**
The calendar inside an overlay panel can be used as complex filter for a column in the turbo table, e.g: to filter with date range and created person name in a column.

**Please tell us about your environment:**
Windows 8.1, VS Code

* **Angular version:** 7.1.0

* **PrimeNG version:** 7.1.2

* **Browser:** all
 
* **Language:** TypeScript

**Suggestion to fix**: 
Please use setTimeout to delay modifying the property in the calendar component when navigating forward or backward. This is to make sure that the navigation button will not be replaced directly that will cause the overlay panel behind to mistakenly assume the navigation button is not descendant of overlay panel while listening on document click event that hide the panel.
![image](https://user-images.githubusercontent.com/8023913/57849231-58215c00-780d-11e9-8174-30c42fc4ddec.png)

redouane20063:
[...]

PR #7699, opened 2019-05-17
Fix for #7693 calendar nav issue in overlayPanel

darius-iko:
###Defect Fixes
Issue #7693 https://github.com/primefaces/primeng/issues/7693


cagataycivici:
7693 seems to be fixed already, thank you.

PR #7705, opened 2019-05-17
Changes in index() and indexWithinGroup() methods of DomHandler

bhuvhi:
Defect:
While resizing/reordering columns in a table with grouped columns, column index calculation was not considering the grouped columns.

Fix:
In DomHandler.js, added a check for colSpan in index() and indexWithinGroup() methods


cagataycivici:
Grouped columns currently do not support resizing.

Issue #7741, opened 2019-05-28
p-fileupload onBeforeSend not working properly.


ipuiu:
**I'm submitting a ...**
```
[x] bug report 
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/github-8qbdgu

**Current behavior**
onBeforeSend event of p-fileupload has no xhr object to add headers.

**Expected behavior**
onBeforeSend event of p-fileupload has should have a xhr object to add headers, as the documentation states in it's event definition: "Callback to invoke before file send begins to customize the request such as adding headers.".

**What is the motivation / use case for changing the behavior?**
I need to add a custom header on the upload request.

* **Browser:** [all]

* **Language:** [all]

JacobSiegle:
onBeforeSend seems bugged further then just the headers not being exposed.  The comment on the function that emits onBeforeSend says `The request was sent out over the wire.` so you cannot customize the header or formData on it because the web request has already been sent.

Using onBeforeUpload seems to allow you to customize the formData but doesn't expose customizing headers.

To expose the headers you need to make the header a generic object since using Angulars HttpHeader object doesn't update the reference emitted when calling append (it creates a new header obj).

```
            let headers = {};
            this.onBeforeUpload.emit({
                'formData': formData,
                'headers': headers
            });

            for(let i = 0; i < this.files.length; i++) {
                formData.append(this.name, this.files[i], this.files[i].name);
            }

            this.http.post(this.url, formData, {
                reportProgress: true,
                observe: 'events',
                headers: headers
            }).subscribe( (event: HttpEvent<any>) => {
```

I can PR it if I get time, otherwise anyone else feel free to. Maybe emitting an entire RequestOptions obj instead of headers would fix the reference issue mentioned before.

cagataycivici:
See;

https://github.com/primefaces/primeng/issues/7805
https://github.com/primefaces/primeng/issues/7803

JacobSiegle:
Closed PR.

PR #7765, opened 2019-06-01
Fix fileupload exposing headers #7741

JacobSiegle:
Ignore the branch name I accidently forgot to switch. 

1) Exposes http options from onBeforeUpload which allows customizing headers.
2) Renames onBeforeSend to onSend because the docs are out of date. onBeforeSend was being called after the request was made and headers cannot be customized at the point it was called.

Some of the test were broken so I changed the code to at least be able to run the basic test I added.

Fixes - https://github.com/primefaces/primeng/issues/7741


JacobSiegle:
https://github.com/primefaces/primeng/issues/7741#event-2409774079

cagataycivici:
7741 seems to be fixed already, thank you.

Issue #7357, opened 2019-03-08
P-Calendar fires onBlur event upon clicking a date in the calendar


wewedgeworth:
**I'm submitting a ...**  (check one with "x")
```
[ X ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
We have a p-calendar with an onBlur event. When we expand the calendar, and click a date to select it, the onBlur event is getting fired. Afterwards, the p-calendar field is re-focused. Then, once the field is actually blurred by the user, onBlur fires again.

**Expected behavior**
There is already a onSelect event which is meant to fire when the user selects a date through the calendar. Also, the field remains focused after the user selects a date. For these two reasons, it does not make sense for onBlur to fire upon clicking a date in the calendar. onBlur should only fire when the field is actually blurred by the user.

**Minimal reproduction of the problem with instructions**
Create a p-calendar with onBlur event.

* **Angular version:** 5.X
Angular 5.2.1

* **PrimeNG version:** 5.X
PrimeNg 5.2.5

* **Browser:** 
Chrome

nitoloz:
Having the same problem with primeng 7.1.2 and angular 7.2.14.
This seems quite critical for me as I would like to set the "default" date if user has given incorrect date and leaves the input, but it turns out that I override user's input when he selects a date in calendar.

Please find steps to reproduce below:
* Put cursor in calendar input and erase the date
* Select any date in calendar
* (onBlur) event is fired and selected date in calendar gets overwritten with "default" date

nitoloz:
Here are some results of a short investigation:
The `<p-calendar>` template contains an input that has `(blur)="onInputBlur($event)"` callback attached to it that leads to `onBlur` event being triggered if user leaves input to click smth in calendar dropdown (datepicker). The `onInputBlur` function code is following:
```
    onInputBlur(event: Event) {
        this.focus = false;
        this.onBlur.emit(event);
        if (!this.keepInvalid) {
            this.updateInputfield();
        }
        this.onModelTouched();
    }
```

[...]

PR #7823, opened 2019-06-19
#7357 add onClickOutside output to calendar

nitoloz:
Add new output to avoid onBlur being fired on datepicker click

###Defect Fixes
This PR should improve behavior described in #7357. 
`onClickOutside()` output is added to help understand whether user has finished working with calendar and clicked anywhere outside of calendar template, `onClickOutside()` can be used instead of `onBlur()` output in case calendar with `datepicker` (not inline) is used,


yigitfindikli:
Hi,
Thanks for the contribution, unfortunately we cannot accept this pr because there is conflicts. Will be added manually.

Issue #6924, opened 2018-11-28
only upload a maximum of 2 files fileLimit="2"


azaddeveloper:
<p-fileUpload 
          [files]="externalFiles" 
          required="true" 
          [multiple]="true" 
          accept="video/mp4"
          **fileLimit="2"** 
          maxFileSize="1000000000"
            [showUploadButton]="false" [showCancelButton]="false" (onSelect)="onUpladFile($event.files)" (onRemove)="onRemoveFile($event)">
          </p-fileUpload>

file limit not working

dinesh-here:
Hi @azaddeveloper ,
There is no "fileLimit" option available in primeng. You can use **onSelect** output method event to get no of selected files and apply validation

PR #7855, opened 2019-06-26
Included the fileLimit on the FileUpload component. 

JO3-W3B-D3V:
See [this issue](https://github.com/primefaces/primeng/issues/6924), to make life a little easier, the fileLimit property has been added & relevant tests included, as @dinesh-here mentioned, you can use **onSelect** to do this, but I find this approach much easier & cleaner.


cagataycivici:
Tracked at https://github.com/primefaces/primeng/issues/7906

PR #7910, opened 2019-07-08
 Optionally retain selection when toggling header checkbox in paginated data

cirovladimir:
This is somewhat related to #1206 which was implemented on #1206 and later improved in #2424.

I think selection with header checkbox should behave the same as if we were selecting rows individually, they add up to the current selection. This behaviour is more notorious in paginated data, which I added an example in the documentation for.


Ng-India99:
is there any update on this ??

cagataycivici:
Seems to be resolved according to the original tickets.

PR #7917, opened 2019-07-09
Fixes #7841: When forceSelection is true, change model to null instead of empty string

### LINKED ISSUE #7841: forceSelection not clearing invalid single numbers/strings
```
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/github-9pxw5z?file=src/app/app.component.ts

**Current behavior**

If I type a number/single string that would not bring any valid value on an Autocomplete(forceSelection=true), it does not clear the invalid value.
If I type a number with a letter, it does clear (this should also happen with the single string/number) and after that it does also clear the single number (as shown on the example bellow), but on the first time I type the number, it does not clear.

**Expected behavior**

It should clear the invalid value just like it does after typing a number plus another character.

type a number on a p-autocomplete (forceSelection=true), for exemple "1" then press tab, it should clear the value because it is an invalid value.
https://stackblitz.com/edit/github-9pxw5z?file=src/app/app.component.ts

* **Angular version:** 7.0.0

* **PrimeNG version:** 7.1.2

* **Browser:** [Chrome XX | Firefox XX]

 
* **Node (for AoT issues):** `node --version` =   10.15.1

rpdmiranda:
Fixes #7841: When using autocomplete with forceSelection=true, sets the model to null instead of empty string when the user clears the input field. 
The old behaviour, setting to an empty string, can cause problems if the value of the field is to be submitted to a server.

### DIFF
--- a/src/app/components/autocomplete/autocomplete.ts
+++ b/src/app/components/autocomplete/autocomplete.ts
@@ -347,7 +347,10 @@ export class AutoComplete implements AfterViewChecked,AfterContentInit,DoCheck,O
         if (value.length === 0 && !this.multiple) {
            this.hide();
            this.onClear.emit(event);
-	   this.onModelChange(value);
+           if (this.forceSelection)
+               this.onModelChange(null);
+           else
+               this.onModelChange(value);
         }
 
         if (value.length >= this.minLength) {


nhnb:
Please note, that the current behaviour is a regression. In PrimeNG 5, the value was correctly set to `null`. But in PrimeNG 8, it is set to an empty string. In our cases, this breaks invocation of REST services that expect a number or null.

cagataycivici:
7841 seems to be fixed already.

Issue #7963, opened 2019-07-17
Selected item in drop-down with virtual scrolling does not always stay into view


monim67:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report
```

**Current behavior**
- Selected item in drop-down with virtual-scrolling does not stay into view while navigating
via up/down arrow keys.
- Selected item in drop-down with virtual-scrolling does not stay into view while typing in filter input-box.
- Selected item in drop-down with virtual-scrolling shows up at the top of the list when
drop-down is opened while it shows up at the bottom while not using virtual scrolling.

**Expected behavior**
- Selected item should stay into view while filtering/navigating via arrow keys.

**Stackblitz Demo**

https://stackblitz.com/edit/primeng-dropdown-issues

**Minimal reproduction of the problem with instructions**
- Go to the stackblitz link above, click on the drop-down select to open up the list items.
- Press down arrow key to scroll down the list, the selected item changes, continue to press it,
the selected item is lost at the bottom of the view (bug 1).
- Now select an item from the bottom of the list, start typing in the filter text box, even if the
selected item matches the text it does not stay into view always (bug 2).
- You can see whenever the drop-down is closed and re-opened the selected item is the first
of the list, but in case of drop-downs without virtual scrolling, the select item shows up as the
last item of the list when drop-down is opened (bug 3).

**Please tell us about your environment:**

* **Angular version:** 7 (as in stackblitz, it's same in version 8)
* **PrimeNG version:** 8.0.1
* **Browser:** all
 * **Language:** all
* **Node (for AoT issues):** `node --version` =  N/A

mertsincan:
Duplicate of #8110

mertsincan:
Hi @monim67,

Thanks a lot for the PR! I made some changes in component. Thanks again!

Best Regards,

monim67:
Close the PR if it's fixed.

PR #7964, opened 2019-07-17
drop-down fixes to keep selected item into view

monim67:
### Defect Fixes using `ListRange` from `@angular/cdk/collections`:

- Selected item out of view issue while navigating via up/down arrow keys.
- Selected item out of view issue while typing in filter input-box.

These issues are detailed in issue #7963 .


mertsincan:
Thanks a lot for the PR!

PR #7966, opened 2019-07-18
Fix #7965 BlockUI won't call block() in the blocked setter

### LINKED ISSUE #7965: BlockUI failed if nested in *ngIf/*ngFor
```
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-kegvvs?embed=1&file=src/app/app.component.html

**Current behavior**
If `<p-blockUI>` is nested in `*ngIf` or `*ngFor`, it won't be appended to the `target` or `body`(default) and set auto-Zindex. It because of the `ViewChild` static is false, the `#mask` query will be resolved after change-detection runs and couldn't be queried in `blocked` setter in the first time.

**Expected behavior**
It should be appended to the `target` and set auto-Zindex if needs.

As the stackblitz case.

* **Angular version:** 5.X
8.0.0~8.1.2

* **PrimeNG version:** 5.X
8.0.1

passerbyid:
Mark the `static` value as true to resolve query before change detection runs, then the `block()` will be called in `blocked` setter correctly.

### DIFF
--- a/src/app/components/blockui/blockui.ts
+++ b/src/app/components/blockui/blockui.ts
@@ -18,7 +18,7 @@ export class BlockUI implements AfterViewInit,OnDestroy {
     
     @Input() baseZIndex: number = 0;
     
-    @ViewChild('mask', { static: false }) mask: ElementRef;
+    @ViewChild('mask', { static: true }) mask: ElementRef;
     
     _blocked: boolean;
         


yigitfindikli:
Seems to be fixed https://github.com/primefaces/primeng/blob/master/src/app/components/blockui/blockui.ts

Jin-K:
It seems that static was reset to false by mistake:
https://github.com/primefaces/primeng/commit/9908406e1e8078ccc433d152c0dae25c627be43e#diff-9057fb19f09e62ec40d3c010792b67eef99788808dd3fc696142c46181670312

Could you reopen the bug ?

passerbyid:
Cannot reopen #7965 by me, maybe @yigitfindikli can help to reopen?

Issue #7971, opened 2019-07-19
Table filter using functions doesn't work


Muji86:
**I'm submitting a ...** 
```
[*] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please demonstrate your case at stackblitz by using the issue template below. Issues without a test case have much less possibility to be reviewd in detail and assisted.

https://stackblitz.com/edit/angular-c2mx4n

**Current behavior**
When filtering columns which uses a function to work out the cell values all rows are filtered out.

**Expected behavior**
We should be able to filter on columns which use functions for cell values as well as ones which use attributes

I could see in the code that we can filter on row attribute and it also works for nested objects where we can provide the column objects path.
For example we can set the column field value to be "attribute1.attribute2" and it would correctly display and filter rows where a row in this example is like:
```
{
attribute1: 
   {
      attribute2: 'value'
    }
}
```
However sometimes a more complex data structure might be used (with arrays in my case) where the filtering no longer supports it.
However I could see in the ObjectUtils.resolveFieldData function that we're trying to resolve the field in case it's a function. If identified as function then this issue wouldn't exist.

The problem is the field param passed to the resolveFieldData function is always a string even if it's originating from a function upstream in the call stack because of the way the logic constructs and applies the filter. For this reason I think it's intended to work with function column fields (if I may call it that)

**Minimal reproduction of the problem with instructions**
The stackblitz link has a minimal and easy to understand demo. I could provide a more complex and better use case but I don't believe it's needed here.
https://stackblitz.com/edit/angular-c2mx4n

**What is the motivation / use case for changing the behavior?**
This is extending/enabling the behavior rather than changing it.
The reason I have is using complex json objects which I want to keep that way when displaying to utilize the binding which Angular provides to keep the data current and not having to flatten the object the then reconstruct it on every change.

**Please tell us about your environment:**
No relevant in this case 

* **Angular version:** 5.X
8.0.1

* **PrimeNG version:** 8.0.0

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

 
[...]

PR #7972, opened 2019-07-19
#7971 Table filter using functions doesn't work

Muji86:
###Defect Fixes

###Feature Requests


cagataycivici:
We'll consider if there is more demand, thank you.

Issue #7187, opened 2019-01-30
Multiselect doesn't disable options if selectionLimit reached by default value


FrEaKmAn:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please demonstrate your case at stackblitz by using the issue template below. Issues without a test case have much less possibility to be reviewd in detail and assisted.

https://stackblitz.com/edit/ngprime-multiselect-qvvtfu?file=src/app/app.component.ts

**Current behavior**

I have 2 values already passed as default value with selectionLimit to 2. Once I open dropdown, other options are still enabled. Once I click on one of them, it closes the dropdown, disables options and doesn't select any value.

**Expected behavior**

I would expect that when I open dropdown, options should already be disabled.

**Minimal reproduction of the problem with instructions**

Follow stackblitz test link and open dropdown. You will see other options as selectable.

* **Angular version:** 7.X

* **PrimeNG version:** 7.X

* **Browser:** all

 
* **Language:** all

* **Node (for AoT issues):** `node --version` =

PR #7976, opened 2019-07-22
Multiselect fix: disable options if selectionLimit reached by default value

monim67:
### Defect Fixes
- #7187

#### Why use `ngDoCheck`?
I could set the value of `maxSelectionLimitReached` in `writeValue` method. But using `ChangeDetection`
ensures the check-boxes are updated even if `selectionLimit` or `options` are updated later. I could not
use `ngAfterViewChecked` for that as the value of `maxSelectionLimitReached` is used in the template.

#### Why remove some code?
As the value of `maxSelectionLimitReached` is set though `ChangeDetection`, we don't need to set it
on checkbox click event anymore.


cagataycivici:
doCheck will cause performance issues, should better to it at writeValue.

monim67:
@cagataycivici the checking can be moved to `ngOnChanges`, is there any performance issues? My motive here is to keep the `selectionLimit` working even if I set/update `selectionLimit` or the options dynamically after the multi-select is initialized.

PR #7991, opened 2019-07-25
Fix #7868 - Galleria not updating as expected; also add ability to move right to left

### LINKED ISSUE #7868: Galleria not updating as expected
**I'm submitting a bug report**

Stackblitz:  https://stackblitz.com/edit/angular-cqgi9z

**Current behavior**
Galleria controls aren't updated correctly after images array is modified.

**Expected behavior**
Galleria UI should update when images array are changed.

**Minimal reproduction of the problem with instructions (https://stackblitz.com/edit/angular-cqgi9z)**
1. Navigate through the galleria with three images (works nice)
2. Press the button (will update the images array to one image only)
3. Navigate through the galleria images again (the image is updated, but the controls aren't updated, the arrow thinks that we have four items, one for the new array and three of the older array)

**Angular version:** 7.X
**PrimeNG version:** 7.X
**Browser:** Chrome, Chromium and Firefox Developer Edition (not tested in others)
**Language:** Typescript
**Node Version:** 10

The not so good solution:), setting timeout on ngAfterViewChanged (galleria.js) (probably there is a better solution and that's why I'm opening this issue):
```
Galleria.prototype.ngAfterViewChecked = function () {
        if (this.imagesChanged) {
            this.stopSlideshow();
           // added setTimeout
            setTimeout(() => {
                this.render();
                this.imagesChanged = false;
            }, 0);
            // this.render();
            // this.imagesChanged = false;
        }
    };
```

feardarkness:
It fixes issue #7868.
Also added the ability to move from the first panel to the last one (from right to left).

### DIFF
--- a/src/app/components/galleria/galleria.spec.ts
+++ b/src/app/components/galleria/galleria.spec.ts
@@ -89,4 +89,17 @@ describe('Galleria', () => {
 
         expect(x).toEqual(1);
     });
+
+    it('should move from right to left', () => {
+        fixture.detectChanges();
+
+        const leftButton = fixture.debugElement.query(By.css('.pi-chevron-left'));
+        const clickNavLeftSpy = spyOn(galleria, "clickNavLeft").and.callThrough();
+        leftButton.nativeElement.click();
+        fixture.detectChanges();
+
+        expect(clickNavLeftSpy).toHaveBeenCalled();
+        expect(galleria.activeIndex).toEqual(galleria.images.length - 1);
+
+    });
 });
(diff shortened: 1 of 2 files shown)


runderworld:
@feardarkness if you rename this issue ticket to something descriptive like `Fix #7868 - Galleria not updating as expected; also add ability to move right to left`  you might have more luck getting it merged.

yigitfindikli:
Hi @feardarkness,
Thanks for the pr but ability to move right to left cause some problems (etc: filmstrip wrapper not updating to the right direction). Updating issue will be added manually.
Best Regards.

Issue #7996, opened 2019-07-29
yearRange is not set


marcusklein:
**I'm submitting a ...**  (check one with "x")
```
[x ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-mwq3km

**Current behavior**
The `yearRange` variable is not set.
When setting the `yearRange`, the setter method only sets the options, but doesn't update the component's `_yearRange` for the getter to access.
e.g. `console.log(this.cal.yearRange) // always undefined`

**Expected behavior**
When set, `yearRange` should be accessible from the component's API. 

* **PrimeNG version:** 8.0.2

* **Browser:** [all]
 
* **Language:** [all]

yigitfindikli:
fixed in https://github.com/primefaces/primeng/commit/6738b4b633c06693e7c7b1dba2a77172f1a7455b

mertsincan:
Refactor on PR. 

Thanks, @marcusklein! Best Regards,

PR #7997, opened 2019-07-29
set private yearRange in setter method

marcusklein:
###Defect Fixes
https://github.com/primefaces/primeng/issues/7996


yigitfindikli:
Thanks for contributing.

PR #8016, opened 2019-08-03
Fixed #7273: p-calendar does not use dateFormat from locale although …

### LINKED ISSUE #7273: p-calendar does not use dateFormat from locale although the docs imply it
the docs suggest that you can provide the dateFormat as part of the locale.
```typescript
export class MyModel {

    en: any;

    ngOnInit() {
        this.en = {
            firstDayOfWeek: 0,
            dayNames: ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"],
            dayNamesShort: ["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"],
            dayNamesMin: ["Su","Mo","Tu","We","Th","Fr","Sa"],
            monthNames: [ "January","February","March","April","May","June","July","August","September","October","November","December" ],
            monthNamesShort: [ "Jan", "Feb", "Mar", "Apr", "May", "Jun","Jul", "Aug", "Sep", "Oct", "Nov", "Dec" ],
            today: 'Today',
            clear: 'Clear',
            dateFormat: 'mm/dd/yy'
        };
    }
}
```

However p-calendar uses a defined interface of localeSettings that doesn't include dateFormat. Instead it only uses the dateFormat from the direct input.

```typescript
export interface LocaleSettings {
    firstDayOfWeek?: number;
    dayNames: string[];
    dayNamesShort: string[];
    dayNamesMin: string[];
    monthNames: string[];
    monthNamesShort: string[];
    today: string;
    clear: string;
}
```

Either remove dateFormat from the docs and explicitly state that the dateFormat needs to be provided seperately or use the dateFormat from the LocaleSettings.

I personally would prefer the second option.

ryvo:


cagataycivici:
Please see

https://github.com/primefaces/primeng/issues/7273

aeb-sia:
We need this fix. We current implementation makes sure the dateFormat from the locale is not used unless dateFormat is set to a falthy value. It should default to the locale

PR #8038, opened 2019-08-08
Add missing attributes

MikeMnD:
The id and automationId properties from the menuItem are missing from the template as attr.id and attr.automationId

###Defect Fixes

###Feature Requests


yigitfindikli:
Hi,
Added manually because there is conflicts with branch.
Thanks for contributing.
Best Regards.

PR #8041, opened 2019-08-09
Fix #7811: DynamicDialog - draggable?

### LINKED ISSUE #7811: Draggable and Resizable DynamicDialog
Is there a property that can be set to allow the dynamic dialog to be draggable in the same way the dialog is draggable ?

lsamalea:
###Defect Fixes
Fixing issue #7811

###Feature Requests


cagataycivici:

Issue #8046, opened 2019-08-10
Table cell editor not switching correctly from input to output with OnPush


sdoubey:
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Stackblitz exemple**
Here is demo of this issue.

https://stackblitz.com/edit/github-ictgok

**Current behavior**
When a component containing a table is using OnPush the edition mode is not correctly left when we click outside the table. MarkForCheck is missing.

**Expected behavior**
Hide input layout and show output layout when user click outside

**Minimal reproduction of the problem with instructions**
https://stackblitz.com/edit/github-ictgok

**Please tell us about your environment:**

* **Angular version:** 8

* **PrimeNG version:** 8.0.2

* **Browser:** all
 
* **Language:** TS

cagataycivici:
Unable to replicate with PrimeNG 9, if the issue persists please create a new ticket with a test case reproducing the issue e.g. stackblitz or a github repo and it will be reviewed by our team once again.

PR #8047, opened 2019-08-10
#8046 Table cell editor not switching correctly with OnPush

sdoubey:
### Code fix
For #8046  - Table cell editor not switching correctly from input to output with OnPush


cagataycivici:
Thanks.

PR #8050, opened 2019-08-12
Fixed  #4370: Add supporting HTML tags in menu label item.

### LINKED ISSUE #4370: Html support for Menu Components
```
```

**Plunkr Case (Bug Reports)**
http://plnkr.co/edit/t4IjvgHqKPFzpdGBVhL6?p=preview

**Current behavior**
MenuItem current accepts a _label_ property containing the text for the item, which is added to the component via interpolation. This means that an HTML formatted label is not possible. Any HTML in the property will be displayed in its original markup.

**Expected behavior**
_label_ should allow HTML content, or a new option (htmlLabel perhaps?) could be used in an innerHTML binding on the LI element

Many menus have some degree of formatting of the item - whether for simple styling, or, in my case, adding a "open in new tab" icon

Windows 7, VS2017, Node

* **Angular version:** 4.2.3

* **PrimeNG version:** 4.1.3

* **Browser:** all 
 
* **Language:** all

* **Node (for AoT issues):** `node --version` =   6.11.4

iarudakov30:
This PR is to close the issue 4370. Let's add possibility to use html tags in the menu item label, to add icons, counter etc. 

The issue https://github.com/primefaces/primeng/issues/4370


jmpeltier:
I vote to merge the Pull Request that was submitted for this open issue.

cagataycivici:
We should use escape property or provide templating instead. That would be safer.

PR #8051, opened 2019-08-13
Closes 3212 shift key selection to picklist, orderlist

### LINKED ISSUE #3212: OrderList and PcikList Multi-Select feature should support range (Shift+select) selection
```
```

Request List-box (OrderList and PcikList )'s Multi-Select feature support range (Shift+select) selection, this will be especially nessesory when the list items are very big.

Gretam11:
This PR attempts to close #3212
I saw that this feature was postponed for quite long time, but it looks like several people at least need it, including me. I've tried to implement it as similar as possible to other features in these components, also it is almost identical in `picklist` and `orderlist`.
Please let me know if i need to cover some more cases or add tests, I would really like to have shift selection in picklist.

### DIFF
--- a/src/app/components/orderlist/orderlist.spec.ts
+++ b/src/app/components/orderlist/orderlist.spec.ts
@@ -159,7 +159,7 @@ describe('OrderList', () => {
       fixture.detectChanges();
       
       const ctrlClickEvent = {'ctrlKey':true};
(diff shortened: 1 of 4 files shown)


Cr3aHal0:
Any news on the probability to see this user-experience-mandatory feature in PrimeNG ?

konuch:
This is still a valid PR that should get an attention.

PR #8054, opened 2019-08-13
Update autocomplete.ts

vincent-lafargue:
Add some function that can be used by users to interacte

###Defect Fixes

###Feature Requests


cagataycivici:
We rather not have console logging

Issue #8057, opened 2019-08-15
Feat(ng-packagr): use ng-packagr to build library


leo6104:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**

primeng use gulp & ts compile for angular library build. But Angular recommend ng-packagr as library build system. 

**Expected behavior**

Please change primeng build system to ng-packagr build. It will make primeng compatible with Angular Ivy compile.

cagataycivici:
PrimeNG 9 uses ng-packagr.

Issue #7885, opened 2019-07-04
Got error TS-996002 when  build with Ivy - Angular 8.1


LillianZhang000:
I just update Angular 8.1 from 7.1, set enableIvy true in file tsconfig.app.json. Then run "ng build", and I got lots of errors as below. 

------------------------------------------------------------
ERROR in ../node_modules/primeng/components/button/button.d.ts:28:22 - error TS-996002: Appears in the NgModule.imports of AppModule, but could not be resolved to an NgModule class

28 export declare class ButtonModule {

../node_modules/primeng/components/inputtext/inputtext.d.ts:12:22 - error TS-996002: Appears in the NgModule.imports of AppModule, but could not be resolved to an NgModule class

12 export declare class InputTextModule {
                        ~~~~~~~~~~~~~~~
../node_modules/primeng/components/dropdown/dropdown.d.ts:153:22 - error TS-996002: Appears in the NgModule.imports of AppModule, but could not be resolved to an NgModule class

153 export declare class DropdownModule {
  
------------------------------------------------------------           


* **Angular version:** 8.1.0

    "@angular/animations": "~8.1.0",
    "@angular/cdk": "^8.0.2",
    "@angular/common": "~8.1.0",
    "@angular/compiler": "~8.1.0",
    "@angular/core": "~8.1.0",
    "@angular/forms": "~8.1.0",
    "@angular/platform-browser": "~8.1.0",
    "@angular/platform-browser-dynamic": "~8.1.0",
    "@angular/router": "~8.1.0",   
    "primeng": "8.0.0",
    "@angular/cli": "~8.1.0",
    "@angular/compiler-cli": "~8.1.0",
    "typescript": "3.4.5"

* **PrimeNG version:** 8.0.0


* **Browser:** Chrome XX 
 
* **Language:** [all | TypeScript 3.4.5 | ES5]

* **Node (for AoT issues):** `node --version` =   v11.5.0

blackholegalaxy:
It also fails under angular next 8.2.0.

JacobSiegle:
In the 8.0 blog they state PrimeNG currently doesn't work with ivy and they are working on making it compatible. I would recommend not using ivy yet if you plan on using PrimeNG - https://www.primefaces.org/primeng-8-0-0-released/

jmesa-sistel:
Ivy will not work at now. 
We have 2 options here, 
a) primeng team fix how the package is upload to npm: 
[...]

PR #8058, opened 2019-08-15
feat(ng-packagr): support ng-packagr library build

leo6104:
###Feature Requests

Please see #8057 https://github.com/primefaces/primeng/issues/7885

I follow https://github.com/ng-packagr/ng-packagr/blob/master/docs/secondary-entrypoints.md#secondary-entry-points 
This secondary entry points logic

You can try this version with command
```
ng build primeng-library
```

FilterKey static variable has problem with ng-packagr so i move it to global scope (but not export the variable)

Please consider it as acceptable PR. It will helpful to primeng compatible with Angular 9 and future release. I tried to minimize file changes to avoid conflicts between other issues.


cagataycivici:
Current codebase is based on ngpackagr, thank you.

Issue #8013, opened 2019-08-02
Ability to unsort Data aka Removable Sort


ranthonissen:
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

I have created a directive that is a possible solution to a previously reported bug (#6609) and feature request (#4493) to "undo" sorting on a third click on the table header.
It overrides the sort(event) method from the table to allow for this behavior.

This is what it looks like:

    import { Directive, Host, Self, Optional } from '@angular/core';
    import { Table } from 'primeng/table';

    @Directive({
        selector: '[withUnsort]'
    })
    export class TableUnsortDirective {

        defaultSortOrderInitialized = false;

        constructor(@Host() @Self() @Optional() public pTable: Table) {

            pTable.tableService.valueSource$.subscribe((val: any[]) => {
                if (val != null && val.length > 0 && !this.defaultSortOrderInitialized) {
                    let i = 0;
                    val.forEach(v => {
                        v['_defaultSortOrder'] = i++;
                    });
                    this.defaultSortOrderInitialized = true;
                }
            });

            pTable.sort = (event: any) => {
                const originalEvent = event.originalEvent;

                if (pTable.sortMode === 'single') {
                    if (pTable.sortField === event.field && pTable.sortOrder === -1) {
                        event.field = '_defaultSortOrder';
                    }

                    pTable._sortOrder = (pTable.sortField === event.field) ? pTable.sortOrder * -1 : pTable.defaultSortOrder;
                    pTable._sortField = event.field;
                    pTable.sortSingle();
                }
                if (pTable.sortMode === 'multiple') {
                    const metaKey = originalEvent.metaKey || originalEvent.ctrlKey;
                    const sortMeta = pTable.getSortMeta(event.field);

                    if (sortMeta) {
                        if (!metaKey) {
                            if (pTable._multiSortMeta.length > 1) {
                                pTable._multiSortMeta =  [{ field: event.field, order: pTable.defaultSortOrder }];
                            } else {
                                if (sortMeta.order === -1) {
                                    pTable._multiSortMeta = [];
                                } else {
                                    pTable._multiSortMeta =  [{ field: event.field, order: sortMeta.order * -1 }];
                                }
                            }
                        } else {
[...]

PR #8072, opened 2019-08-21
[Table] Third click to undo sorting

ranthonissen:
Implements #8013 to undo the sorting of a table on the third click on the column:
* works for both single as multiple sorting
* first 2 click sort ascending / descending
* 3rd click sorts the table data in its default order


cagataycivici:

loxy:
> One of those things that we prefer to be implemented on our side, thank you.

Because you are faster?

denversynchro:
How do you turn on unsort on 3rd click?

gianluca-moro:
So, has this been implemented yet?

ranthonissen:
@gianluca-moro No, they totally ignored it and never cared to implement it themselves

PR #8083, opened 2019-08-23
Fixing change detection on accordion animation

gferreira-keeps:
Added a verfication if the change detector is still active before detecting changes. If not, on setting the animation in a view alreay destroyed and detecting changes it will give an error.


cagataycivici:
170 files changed?

cagataycivici:
This is fixed already.

PR #8085, opened 2019-08-23
Add contains any filter

philjohnson789:
### Feature Requests
We're using a PrimeNG Table to display data with a column type that is an array. We display the values as a pipe-delimited string inside each cell. We would like to filter the grid using a multi-select and show only rows that contain any of the selected filters inside the array column. Right now, the `in` filter type only works for a single cell value. This pull request adds a new `containsAny` filter type that searches an array for any matching filter value.


cagataycivici:
Due to amount of changes, we're unable to merge however we'll consider it for 8.1.0.

cagataycivici:
Tracking at;

https://github.com/primefaces/primeng/issues/8213

PR #8103, opened 2019-08-29
Fixed #8102 hide calendar overlay in onpush components

### LINKED ISSUE #8102: Calendar overlay doesn't disappear in OnPush components.
```
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/github-zpt3km

**Current behavior**
Calendar overlay only disappears in ChangeDetectionStrategy.OnPush components, when another change detection runs because the markForCheck on calendar.ts:2077 is outside of zone and doesn't trigger change detection.

**Expected behavior**
Calendar should work like in ChangeDetectionStrategy.Default components.

The provided stackblitz shows a calendar in a default and an onpush component side by side.

* **Angular version:** >=8
* **PrimeNG version:** 8.0.2

Hapcy:


yigitfindikli:
Hi,
Thanks for contributing.
Best Regards.

PR #8105, opened 2019-08-30
fixes 6736: table checkbox selects disabled lines

yelhouti:
###Defect Fixes
fixes 6736 : table checkbox selects disabled lines


cagataycivici:
Not sure if this would be efficient when dealing with large datasets so performance could be hit, we're looking for alternatives.

yelhouti:
@cagataycivici , no worries, good luck!

Issue #8108, opened 2019-08-30
ListBox - Add selected option value to onClick and onDblClick


whizkee:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**What is the motivation / use case for changing the behavior?**
To know which value is being added to or removed from your selected values list (ListBox ). 

**Please tell us about your environment:**
Windows 10

* **Angular version:** 5.X
8.2.4

* **PrimeNG version:** 5.X
8.0.2

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
Chome, Firefox, IE
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
TypseScript 3.5.3

PR #8109, opened 2019-08-30
Add selected option.value and isSelected to onChange.emit

whizkee:
###Feature Requests
#8108


yigitfindikli:
Hi,
Added manually for onDblClick and onClick. Because toggle also uses onChange event and it can be create misunderstandings. Thanks for the pr.

PR #8121, opened 2019-09-05
Fixed  #6624 PrimeNG: 6.1.4 - Editor "onTextChange" is not triggered …

### LINKED ISSUE #6624: PrimeNG: 6.1.4 - Editor "onTextChange" is not triggered when Quill.insertText() is called with source 'api'
```
```

**Plunkr Case (Bug Reports)**

**Current behavior**
We are consuming PrimeNG editor in our project. We are calling quill API for inserting text into editor something like below.
`this.quill.insertText(index, text, format, "api");`

Earlier we were using PrimeNG:4.2.0 and above api call was resulting into "onTextChange" event. However, this has been broken after we updated to version 6.1.4.

**Expected behavior**
onTextChange() should be emitted is we call insertText() with source 'api'.

- use editor component in template something like below
```
      <p-editor [(ngModel)]="rich_text"
        (onTextChange)="onTextChange($event)"
        (onInit)="onEditorInit($event)">
```
- disable manual edit on editor within 'onEditorInit' function
`this.quill.enable(false);`
- call insertText() with source 'api'
`this.quill.insertText(index, text, format, "api");`
- check on 'onTextChange()' function. print something within this function
- Note that 'onTextChange' is not getting called

I'd like to describe the working of Quill API with respect to source input. Ref: https://quilljs.com/docs/api/#inserttext

> **insertText**
> Inserts text into the editor, optionally with a specified format or multiple formats. Returns a Delta representing the change. Source may be "user", "api", or "silent". Calls where the source is "user" when the editor is disabled are ignored.

In our case the editor is disabled (meant edit through keyboard is not allowed on editor) and we are supplying source as 'api'. 
Now let's take a look on PrimeNG code for Quill editor implementation https://github.com/primefaces/primeng/blob/master/src/app/components/editor/editor.ts
Please note below that "onTextChange" is emitted only when source is "user". And in our case when editor is disabled and source is 'api' the quill indeed emit the 'text-change' event but PrimeNG implementation does not do anything as it is expecting source as 'user'.
```
this.quill.on('text-change', (delta, oldContents, source) => {
            if (source === 'user') {
                let html = editorElement.children[0].innerHTML;
                let text = this.quill.getText().trim();
                if (text.length === 0) {
                    html = null;
                }

                this.onTextChange.emit({
                    htmlValue: html,
                    textValue: text,
                    delta: delta,
                    source: source
                });
                
                this.onModelChange(html);
                this.onModelTouched();
            }
        });
```

I hope I have explained the use case thoroughly. However, please let me know if you need elaboration at any step.

Operating system: windows
IDE: MS VS Code
package Manager : npm
HTTP Server: Does not matter

* **Angular version:** 5.X

Angular : 6.x
**Note:** this is not related to Angular update

* **PrimeNG version:** 5.X

PrimeNG version : 6.1.4
Earlier we were using PrimeNG:4.2.0 and above api call was resulting into "onTextChange" event. However, this has been broken after we updated to version 6.1.4.

yigitfindikli:
…when Quill.insertText() is called with source 'api'

### DIFF
--- a/src/app/components/editor/editor.ts
+++ b/src/app/components/editor/editor.ts
@@ -99,6 +99,8 @@ export class Editor implements AfterViewInit,ControlValueAccessor {
     onModelTouched: Function = () => {};
     
     quill: any;
+
+    isCreated:boolean = false;
     
     constructor(public el: ElementRef) {}
 
@@ -124,7 +126,7 @@ export class Editor implements AfterViewInit,ControlValueAccessor {
         }
         
         this.quill.on('text-change', (delta, oldContents, source) => {
-            if (source === 'user') {
+            if (this.isCreated) {
                 let html = editorElement.children[0].innerHTML;
                 let text = this.quill.getText().trim();
                 if (html === '<p><br></p>') {
@@ -141,6 +143,8 @@ export class Editor implements AfterViewInit,ControlValueAccessor {
                 this.onModelChange(html);
                 this.onModelTouched();
             }
+
+            this.isCreated = true;
         });
         
         this.quill.on('selection-change', (range, oldRange, source) => {


yigitfindikli:
fixes #6624

PR #8132, opened 2019-09-09
added busy message with spinner

UliPlabst:
I added a "busy" option to the messages component to give feedback about ongoing operations. It has the same style as the "info" message but contains a spinning spinner icon. I think this is a very common use case plus the changes are literally less than 10 lines so I would greatly appreciate if you merge this. Unfortunately my IDE removed a lot of unnecessary whitespace from the files therefore most of the changes below are removed whitespace. 
Btw I also updated the demo to have a "Busy" button to display this option.
Thanks.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,11351 @@
+{
+  "name": "primeng",
+  "version": "8.1.0-SNAPSHOT",
+  "lockfileVersion": 1,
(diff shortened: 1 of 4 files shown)


cagataycivici:
We'll reconsider if there is more demand, thank you.

PR #8173, opened 2019-09-26
Fixed #6743 Make leaf node not expandable in tree regardless of children

### LINKED ISSUE #6743: Make leaf node not expandable in tree regardless of children
```
```

**Current behavior**
When using the tree with nodes containing children, if the node has children, the expand caret will always show, regardless of "leaf" property value.

**Expected behavior**
If the node's "leaf" property is set to "true", then the node should not be expandable even if there are children on the node.

We've got a tree where there are a number of items a user can select by various properties and sub-properties (think a list of cars chosen by make and fuel-type). The users also have the ability to upload a custom list which wouldn't be arrived through the properties we supply (only 3 specific models of Ford). When the users load the new data, it is potentially in the hundreds of thousands of records and so we place it all under one node on the tree named "uploaded items" with a count of the items in that node but with no option to expand it (there is an option to export the data to a list should they need to review it). We expected to be able to set the "leaf" property for this node to "true" as we want this node to function as a leaf node. However the fact that this node has children is overriding the leaf property and displaying the node as expandable.

* **Angular version:** 5.2.9

* **PrimeNG version:** 5.2.4

* **Browser:** [all]

 * **Language:** [all]

yigitfindikli:
Fixed #6743.

### DIFF
--- a/src/app/components/tree/tree.ts
+++ b/src/app/components/tree/tree.ts
@@ -888,7 +888,7 @@ export class Tree implements OnInit,AfterContentInit,OnDestroy,BlockableUI {
     }
 
     isNodeLeaf(node) {
-        return node.leaf == false ? false : !(node.children&&node.children.length);
+        return !node.leaf ? !(node.children&&node.children.length) : true;
     }
 
     getRootNode() {


jmesa-sistel:
@yigitfindikli and @cagataycivici 
I think the right fix is just return the node.leaf value. The user can choice is he wants a leaf or not regardless children. 
See comment [541301134](https://github.com/primefaces/primeng/issues/6743#issuecomment-541301134) and issue [8237](https://github.com/primefaces/primeng/issues/8237)

Issue #8174, opened 2019-09-26
ConfirmDialog loses focus when navigating with tab key


csvinhal:
**I'm submitting a ...**  (check one with "x")
```
[ X ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
It's possible to simulate using the own PrimeNG showcase.
https://www.primefaces.org/primeng/#/confirmdialog

**Current behavior**
When ConfirmDialog is open, using "Tab" or "Shift + Tab" to navigate, the ConfirmDialog loses the focus and it goes to the body of the document.

**Expected behavior**
According to the WCAG the user should be able to navigate using keyboard without lose focus on a Dialog element. So that's the expected behavior.

**Minimal reproduction of the problem with instructions**
Go to showcase https://www.primefaces.org/primeng/#/confirmdialog, open a ConfirmDialog and keep pressing "Tab" or "Shit + Tab".

**What is the motivation / use case for changing the behavior?**
The component accesibility.

* **Angular version:** 5.X
8.2.0

* **PrimeNG version:** 5.X
8.0.3

* **Browser:** 
all

csvinhal:
Resolved with the commit https://github.com/primefaces/primeng/commit/903ece11a0478123fccf4630aeabd93147b43f6b#diff-f490f0e0b3c83cf3465419219de415b6

PR #8175, opened 2019-09-26
#8174 - trap ConfirmDialog focus

csvinhal:
Implemented cdkFocusTrap to keep the component focus when navigating with a keyboard.


cagataycivici:
We have our own trapFocus so will apply that.

PR #8191, opened 2019-10-02
Fixed  #8183 Node click script error

### LINKED ISSUE #8183: Script Error when user clicks on SVG element inside Tree node template
[X ] bug report
[ ] feature request
[ ] support request

**Current behavior**
If the template of a Tree Node contains an SVG, a script error is thrown from the Tree.prototype.onNodeClick method when the user clicks exactly on top of the SVG element. 

This happens because the onNodeClick function attempts to access the indexOf property of a SVGAnimatedString.

**Expected behavior**
Users should be able to use SVG icons inside Tree templates and click anywhere within the node without primeng throwing any errors

https://stackblitz.com/edit/angular-hlqzve

If you click exactly on the icon you can see a script error on the browser console.

AwadMaharoof:
###Defect Fixes
Added check for type of className property before trying to read the indexOf property

### DIFF
--- a/src/app/components/tree/tree.ts
+++ b/src/app/components/tree/tree.ts
@@ -625,7 +625,7 @@ export class Tree implements OnInit,AfterContentInit,OnDestroy,BlockableUI {
     onNodeClick(event, node: TreeNode) {
         let eventTarget = (<Element> event.target);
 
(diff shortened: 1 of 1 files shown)


cagataycivici:
Fixed using DomHandler hasClass instead

Issue #6808, opened 2018-11-01
Turbo Table cannot filter empty values with "in" Filter


JacobSiegle:
NOTE: PR exists.

**Current behavior**
It's not possible to filter the table by null or undefined values.

**Expected behavior**
It should be able to filter empty rows.

**Minimal reproduction of the problem with instructions**
https://stackblitz.com/edit/github-2ovumv

lwestfall:
Any chance that this will be looked at for near future release? I have a workaround in place for my app but it'd be nice to go back to the native "in" filter mode.

```ts
FilterUtils['in-fixed'] = (value, filters) => {
  if (filters === undefined || filters === null || filters.length === 0) {
      return true;
  }

  for (const filter of filters) {
      if (ObjectUtils.equals(value, filter)) {
          return true;
      }
  }

  return false;
};
```

cagataycivici:
Planned for 10.0.1.

cagataycivici:
We've discussed this internally as 10.0.1 should be released today, we postpone it to 10.0.2 as it requires more thinking. It has high priority so will be solved in the next one.

PR #8208, opened 2019-10-05
Filter nulls or undefined

JacobSiegle:
Issue: https://github.com/primefaces/primeng/issues/6808

Allows filtering of null of undefined rows in turbo table. You cannot filter both since one of the values is used to show all rows. Must be optionally turned on to keep default functionality that null or undefined shows all.


cagataycivici:
Planned for 10.0.1, thank you. Will be handled by the team, thank you for the feedback.

Issue #8158, opened 2019-09-22
Unable to clear a [readonlyInput]='true' Calendar Input element


davisb10:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

Visible from Showcase
https://www.primefaces.org/primeng/#/calendar

**Current behavior**
- Select a value in any of the elements where [readonlyInput]='true' (example Min-Max)
- Select the value in the input box
- Try to delete it using backspace / delete
- Value is unable to be removed from the input box

**Expected behavior**
Be able to delete a value from a readonlyInput Calendar

**What is the motivation / use case for changing the behavior?**
Should be able to have a readonlyInput Calendar but still be able to delete the value from the input box using the backspace / delete

JacobSiegle:
Just to be sure - your expecting the entire input to be removed when backspace/delete is hit correct?

davisb10:
With the readonlyInput set, you do not get a visible cursor inside the input element. So you are unable to backspace out one character at a time or not able to click and drag to select the whole date string (eg. 09/25/2019) and clear it.

The biggest motivation for this is to have a readonlyInput calendar so users are unable to put 'garbage' into the input. However in doing so, there is no way to clear the input.

mmlitvin:
[showButtonBar]="true" will give you the Clear button in the footer

JacobSiegle:
https://github.com/primefaces/primeng/pull/8210

cagataycivici:
This is the expected behavior, readonly inputs cannot be deleted.

PR #8210, opened 2019-10-05
calendar readonly allow delete

JacobSiegle:
https://github.com/primefaces/primeng/issues/8158


cagataycivici:
This is the expected behavior, readonly inputs cannot be deleted. It is against the HTML Spec.

PR #8222, opened 2019-10-08
Fixed #8218 panel min width

### LINKED ISSUE #8218: Dropdown panelStyle min width overridden when appendTo body is used
https://github.com/primefaces/primeng/blob/f5c11afcc22530de014e74c55462ce0000a8c601/src/app/components/dropdown/dropdown.ts#L634

When the user is using append to body and using panel style to set a min width, the min width will be overridden by primeng.

JacobSiegle:
Fixed https://github.com/primefaces/primeng/issues/8218.

Also panel style was in autocomplete but missing from the docs.


yigitfindikli:
Hi @JacobSiegle,
Thanks a lot for contribution!

PR #8223, opened 2019-10-08
Confirmation service subscribe to all events

JacobSiegle:
Right now it's not easy to reuse confirmation dialogs. This PR adds an event which allows you to subscribe to handle the events of the confirmation service from it's created dialogs where ever you would like. 

Example) I might want to use a generic piece of code to make a confirm dialog and then catch the callback myself in my component or a parent component. I don't want to have to duplicate the confirm code in many of many components.

The need for this is because requireConfirmation$ doesn't provide any callbacks for accept or reject, and you can't easily get the confirmation from the confirmation dialog component.

I didn't add any docs, wasn't sure the best spot to place the info on the feature. Also wasn't sure the best spot to put the models.

Sorry for the whitespace changes.


cagataycivici:

Issue #8226, opened 2019-10-10
Popup menu does not hide on document click when parent uses OnPush change detection


tasos-ch:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-xnuzdt

**Current behavior**
The popup does not close when clicking anywhere but its source button when the parent component uses the `OnPush` change detection strategy.

**Expected behavior**
The popup should hide when clicking anywhere on the page.

> **Note**: Removing `ChangeDetectionStrategy.OnPush` from `ParentComponent` achieves the expected behavior.

**Minimal reproduction of the problem with instructions**
Steps to reproduce:
1. In the above StackBlitz, click the `Show` button.
2. The popup appears.
3. Click anywhere in the page besides the `Show` button.

The popup remains visible.

Some notes that I hope will be of help:
1. `Menu.hide()` is called successfully by `Menu.bindDocumentClickListener()`.
2. I think there was a [similar issue](https://github.com/primefaces/primeng/issues/7904) with the `Dropdown` component. Its [solution](https://github.com/primefaces/primeng/commit/2944fce30e8beac5e8152e829bdb1be9a29a78db) was to emit an `onHide` event from within its `hide` function. Perhaps, something like that could be applied here, as well.
Or, calling `ChangeDetectorRef.markForCheck()` from within `hide()` might be enough.

* **Angular version:** 8.1.1

* **PrimeNG version:** 8.0.3

yigitfindikli:
Fixed in https://github.com/primefaces/primeng/commit/0dc99269b560329eea541762535825f4dab71647

PR #8228, opened 2019-10-10
Popup menu does not hide on document click when parent uses OnPush change detection #8226

avishon:
#8226


yigitfindikli:
Thanks for the contribution.

Issue #8112, opened 2019-09-03
Upgrade gulp to v4 (gulp 3 is no longer supported in node 12)


huineng:
when building components on node 12 './node_modules/.bin/gulp build-assets'

you get this error https://github.com/gulpjs/gulp/issues/2324

gulp 3 is no longer support from node 12.

Can we move primeng gulp to v4 ?

cagataycivici:
We haven't decided to upgrade yet, we'll do it on our side when decided, thank you.

PR #8239, opened 2019-10-15
[feature/issues/8112]: Upgrade to gulp 4

OskarsPakers:
Upgrade of gulp according to issue #8112


cagataycivici:
We haven't decided to upgrade yet, we'll do it on our side when decided, thank you.

PR #8289, opened 2019-10-31
Add Table ExportCSV option to ignore filters

erhickey:
Add another option to Table.exportCSV, ignoreFilters. When set to true
exportCSV will include data hidden due to filters.


cagataycivici:
We'll consider if there is more community feedback. Also the PR lacks documentation for merge.

Issue #8296, opened 2019-11-04
Timer running even after calendar closing and opening


mghextreme:
**I'm submitting a ...**
```
[x] bug report
[ ] feature request
[ ] support request
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/primeng-calendar-time

**Current behavior**
When opening the calendar component with a time picker, clicking (mouse down) on one of the arrows (up/down) for either hour or minute and "unclicking" (mouse up) outside the DOM element, the timer keeps running.
If you close and open the component again, the hour/minute is continuously changing, even though you haven't clicked anything.

**Expected behavior**
When "unclicking" it should stop 

**Minimal reproduction of the problem with instructions**
(as described in Current behavior)

**What is the motivation / use case for changing the behavior?**
Avoid weird behavior - component changing dates prior to user interaction.

**Please tell us about your environment:**
Windows 10, Mozilla Firefox

* **Angular version:** 8.2.11

* **PrimeNG version:** 8.1.1

* **Browser:** all - Tested in Chrome, Firefox and Edge
 
* **Language:** all

yigitfindikli:
fixed in https://github.com/primefaces/primeng/commit/722b46ac71e8e0ac82c68baf0ac2b601a12de676

acoroza:
Is there any hack to fix this issue in version 8.1.1? For some reasons we cannot use the latest version,

PR #8297, opened 2019-11-04
Avoid closing picker with running timer.

mghextreme:
About issue #8296 .

Solved adding MouseOut events and clearing timer in other methods.


yigitfindikli:
Hi,
Added manually. Because there is conflicts with branch.
Thanks for contributing.
Best Regards.

PR #8325, opened 2019-11-14
#8323 - adjusted tooltip alignment after text changed

### LINKED ISSUE #8323: Dynamic Tooltip rendering wrong
```
```

**Plunkr Case (Bug Reports)**
https://github-yibmjy.stackblitz.io (https://stackblitz.com/edit/github-yibmjy)

**Current behavior**
When changing dynamically the Tooltip text to one with a different size, it doesn't render correctly. The Tooltip size is correctly re-calculated but not it's position.

**Expected behavior**
When changing dynamically the Tooltip text to one with a different size, it should recalculate the Tooltip position correctly.

- Open the case (https://github-yibmjy.stackblitz.io)
- Get focus to the input field. (Notice the Tooltip at this point).
- Start typing. (Tooltip will change the message, change size and get in a wrong position)

There is some cases we need to change the Tooltip text dynamically and it's rendering in a wrong position making the user experience bad.

* **Angular version:** 8.X
* **PrimeNG version:** 8.X

csvinhal:
###Defect Fixes
#8323 - adjusted tooltip alignment after text changed.

### DIFF
--- a/src/app/components/tooltip/tooltip.ts
+++ b/src/app/components/tooltip/tooltip.ts
@@ -134,8 +134,10 @@ export class Tooltip implements AfterViewInit, OnDestroy {
         this._text = text;
         if (this.active) {
             if (this._text) {
-                if (this.container && this.container.offsetParent)
+                if (this.container && this.container.offsetParent) {
                     this.updateText();
+					this.align();
+				}
                 else
                     this.show();
             }


yigitfindikli:
Hi,
Thanks for contributing.
Best Regards.

Issue #8356, opened 2019-11-25
stateStorage does not work in p-table with rowsPerPageOptions


Cito:
**I'm submitting a bug report**

PrimeNG has a showcase for the [stateStorage feature of p-table](https://www.primefaces.org/primeng/#/table/state).

The problem appears when you add `rowsPerPageOptions` to the tables in the example.

For instance, edit `src/app/showcase/components/table/tablestatedemo.html` and add an `[rowsPerPageOptions]=[5,10,20]` attribute to the `<p-table #dt1 ...>` tag.

Now, when you `ng serve` and browse to http://localhost:4200/#/table/state, you will see the popup with values 5, 10, 20 in the first table. Change this from 10 to 5, and page forward to page 3. Now refresh the page in the browser.

**Current behavior**

Currently, the location in the table will be restored, but not the number of rows per page. It will be reset to 10, while the page is set to 2 instead of 3 (because the page size increased). Apparently, the `first` property is restored properly, but the `rows` property is not.

**Expected behavior**

After refreshing the page, the page size of 5 and page number of 3 should both be restored properly.

* **Angular version:**  8.2.3 and 9.0

* **PrimeNG version:** 8.1.1 and 9.0

Cito:
Closing this, since it seems this has be (accidentally?) fixed yesterday in 7098540d41ae786ec35e958b98c982f79fcffc1b.

However, I suggest to add a rowsPerPageOptions attribute to the TableStateDemo so that a regression can be easier tested/demonstrated, see PR #8357.

PR #8357, opened 2019-11-26
Add rowsPerPageOptions to TableStateDemo

Cito:
The [TableStateDemo](https://www.primefaces.org/primeng/#/table/state) shows how the current table page is restored when reloading the page. However, it does not show how the current table page size is restored as well. This PR adds a rowsPerPageOptions attribute so that we have a test/demo for this funtionality, which was broken until recently (see #8356).


yigitfindikli:
Hi,
Thanks for contributing.
Best Regards.

PR #8370, opened 2019-11-29
TreeTable | Added text property to allow clickable text along with icon.

satya-jugran:
TreeTable | Use case - As of now, we could not click on the row text shown on the cell and expand its child nodes. We can expand child nodes only on the icon click. With this change, it would be an option to show cell text as a part of anchor (clickable).

###Defect Fixes
N/A

###Feature Requests
TreeTable | Use case - As of now, we could not click on the row text shown on the cell and expand its child nodes. We can expand child nodes only on the icon click. With this change, it would be an option to show cell text as a part of anchor (clickable).


cagataycivici:
This should better be handled via templating instead. Thank you.

satya-jugran:
> This should better be handled via templating instead. Thank you.

@cagataycivici Could you please give an example?

Issue #8334, opened 2019-11-18
colorPicker - wrong color rendered


sylvain-pointud:
**I'm submitting a ...**  
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**StackBlitz Case (Bug Reports)**
https://stackblitz.com/edit/prime-ng-bug-0b16e2

**Current behavior**
the value of the colorPicker is not 0b16e2 => rgb(11,22,226) but rgb(11,21,226) 
![image](https://user-images.githubusercontent.com/21316308/69074065-f3022980-0a2e-11ea-9b6d-808d6226399f.png)

**Expected behavior**
rgb(11,22,226)

**Minimal reproduction of the problem with instructions**

* **Angular version:** 8.X

* **PrimeNG version:** 8.X

* **Browser:** [all]

yigitfindikli:
fixed in https://github.com/primefaces/primeng/commit/79e1718e2a84c3a5691d8b06c14e98c3258514aa

PR #8375, opened 2019-12-03
Fix HSBtoRGB from ColorPicker component

clementbreuzet:
#8334 colorPicker - wrong color rendered

**Issue**

Due to irrationnal numbers manipulation, method HSBtoRGB in ColorPicker component is not accurate as expected.

**Fix**
- Avoid Math.round in h,s,v values initialization.
- Use type number instead of var to have an accurate value needed in the process.


amranidriss:
Good job !

yigitfindikli:
Hi,
Thanks for contributing.
Best Regards.

PR #8380, opened 2019-12-04
Update messageservice.ts

steeu:
As i understand the PrimeNG MessageService is a nice example for a singleton service and should therefore use the `providedIn: 'root'` property beginning with Angular 6.0.

More details: https://angular.io/guide/singleton-services#singleton-services

###Defect Fixes
N/A

###Feature Requests


cagataycivici:
It may break existing apps if we make it default.

rafa-as:
> It may break existing apps if we make it default.

Why?

Pranen:
> It may break existing apps if we make it default.
Could we do this change in the next major Update?
Those updates may break existing apps.

PR #8382, opened 2019-12-04
[FileUpload] accept input to array

clementbreuzet:
**Issue:**
Input _accept_ from FileUpload component is a string type and it can be difficult to build a string like ".jpg,.png,.doc..."

**Fix:**
I propose to add a pipe to this input so that it will be possible to declare _accept_ input as an array.
Using a pipe string | string[] will not introduce a breaking change and let file upload component retro compatible.


cagataycivici:
We'll consider if there is more demand from the community, thank you.

PR #8391, opened 2019-12-11
fix: emit onEditComplete when change to cell

### LINKED ISSUE #8049: pTable cell editing: onEditComplete not triggered when clicking other cells.
**I'm submitting a ...** (check one with "x")

```
[x] bug report = Search github for a similar issue or PR before submitting
[ ] feature request = Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request = Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

The issue that seems to have been solved in #6349 has reappeared in 8.0.3 (or earlier).
I copied the problem description and added a new stackblitz which shows the events as they are being fired.

**Plunkr Case (Bug Reports)**
 
https://stackblitz.com/edit/github-rume1v
 
**Current behavior**
When using the event onEditComplete  the event isn't triggered by using the mouse to move to another field, when you press tab or esc the onEditCancel is triggered, when you press enter the onEditComplete is triggered, but there is no event when you change to another field through mouse.
 
 
**Expected behavior**
I think onEditComplete should be triggered when stopping the edit using the mouse to move to another field, 
 
The problem here is the inconsistency, because if you need something to happen everytime you finish editing, you'll never gonna use onEdit events because it will not apply to every case (the user will have the possibility to skip part of the process)
 
indifferent
 
* **Angular version:**
  8.0.3
* **PrimeNG version:**
  8.0.2
* **Browser:** all
* **Language:** all

sanchis:
###Defect Fixes
#8049

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -499,7 +499,8 @@ export class Table implements OnInit, AfterViewInit, AfterContentInit, Blockable
             if(this.virtualScroll && this.virtualScrollCallback) {
                 this.virtualScrollCallback();
             }
-    
+            this.updateEditingCell(null,null,null);
+            this.unbindDocumentEditListener();
             this.tableService.onValueChange(simpleChange.value.currentValue);
         }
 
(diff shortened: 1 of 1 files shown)


cagataycivici:
Will fix with a different approach, thanks.

PR #8424, opened 2019-12-27
optional checkbox icon in listbox

prizef:
###Feature Requests
Ability to use custom icons in p-listbox.


cagataycivici:
We'll reconsider if there is more demand for custom icons, thank you.

Currently you may even override it with css.

Issue #8377, opened 2019-12-04
Carousel  - Button inside carousel-item is not clickable in mobile phones


BissacM:
**Current behavior**
When adding buttons to a carousel item, on mobile it's not clickable.

**Expected behavior**
Should be able to emit click events when the buttons were clicked.

**Please tell us about your environment:**
Any

* **Angular version:** 
latest Angular version

* **PrimeNG version:** 
latest PrimeNG version

* **Browser:**
 In any **mobile** browsers.
 
* **Language:** 
All

suryamuppalla:
+ Same issue

suryamuppalla:
Same Issue

belialzxz:
Same here. The issue I see is that any click within the content area causes the "go to previous" event, as the carousel moves to previous item. I am able to replicate this on the showcase as well. I tried testing with Chrome dev console emulator, and no matter the viewport size, the bug was happening. I noticed it first on my Samsung Galaxy.

PR #8450, opened 2020-01-15
Add position check in carousel mobile onTouchEnd event comparing onTouchStart event

IhorPetr:
The main problem with OnTouchEnd event which prevent emit any nested event in ngTemplateContext. This pr add check on touchEnd event that the position click didn`t changed compare onTouchStarted event (see #8377)

### DIFF
--- a/src/app/components/carousel/carousel.ts
+++ b/src/app/components/carousel/carousel.ts
@@ -165,6 +165,8 @@ export class Carousel implements AfterContentInit {
 
 	startPos: any;
 
+	endPos: any;
+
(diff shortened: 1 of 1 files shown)


IhorPetr:
The same problem exist in [primereact](https://github.com/primefaces/primereact) library

luki215:
Fix works, please merge.
It would be even better to have configurable tolerance for touch radius.

cagataycivici:
We've done some fixes to this as well, please test with 9.0.5.

PR #8467, opened 2020-01-20
Fix bug when inputSwitch is inside a label

### LINKED ISSUE #8466: InputSwitch doesn't switch when clicked if it is inside a label
```
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/primeng-inputswitch-label-arround-bug (sorry, I'm not using the primeng template as I wrote the test before opening the issue and discovered a template existed).

**Current behavior**
When inputSwitch is used inside a `<label>`, a click on the inputSwitch doesn't switch it (in fact, it switches 2 times).

**Expected behavior**
Switch when clicked, even when inside a `<label>`

See stackblitz.

Bug.

* **Angular version:** 5.X
8.2.14

* **PrimeNG version:** 5.X
8.1.1

Chromium 79.0.3945.79 (probably any browser)
 
marob:
Fixes #8466

### DIFF
--- a/src/app/components/inputswitch/inputswitch.ts
+++ b/src/app/components/inputswitch/inputswitch.ts
@@ -53,7 +53,8 @@ export class InputSwitch implements ControlValueAccessor {
     constructor(private cd: ChangeDetectorRef) {}
 
     onClick(event: Event, cb: HTMLInputElement) {
-        if (!this.disabled && !this.readonly) {  
+        if (!this.disabled && !this.readonly) {
+            event.preventDefault();
             this.toggle(event);
             cb.focus();
         }


yigitfindikli:
Thanks for the contribution.

Xapuu:
It seems like this one resurfaced. I see that the `preventDefault` that fixed this was removed in the latest versions.

Andrei-NZ:
@Xapuu 

I've been able to work around this by disabling pointer-events on the switch when inside a label 

```
<label class="flex px-2 items-center justify-between w-full cursor-pointer">
  Images only
  <p-inputSwitch class="inline-flex pointer-events-none" [ngModel]="true" />
</label>
```

PR #8482, opened 2020-01-23
Fix multiselect filled state

### LINKED ISSUE #8279: ui-inputwrapper-filled is not removed when multi-select options are deseleced
```
```

**Plunkr Case (Bug Reports)**
When using `ui-inputwrapper-filled` to style inputs to highlight when a selection has been made, the class isn't removed when the input is no longer filled by the process of deselecting the option.

**Current behavior**
Select an option, then deselect. .ui-inputwrapper-filled still applies even though there is nothing "filled" in the input anymore

**Expected behavior**
.ui-inputwrapper-filled would no longer be present as the input is no longer filled out. 

https://github-jmrqwa.stackblitz.io

Select an option, then deselect that option. 

To help make the multiselect more usable for filtering, having a "highlight" state to indicate if an option is selected helps a user see that this input has a value.

andyg3:
Updated updateFilledState method to check for a non-empty value, instead of checking the label.
Since there is always a label (unless you add defaultLabel="") this was always returning true.

###Defect Fixes
#8279 

###Feature Requests

### DIFF
--- a/src/app/components/multiselect/multiselect.ts
+++ b/src/app/components/multiselect/multiselect.ts
@@ -376,7 +376,7 @@ export class MultiSelect implements OnInit,AfterViewInit,AfterContentInit,AfterV
     }
 
     updateFilledState() {
-        this.filled = (this.valuesAsString != null && this.valuesAsString.length > 0);
+        this.filled = (this.value && this.value.length > 0);
     }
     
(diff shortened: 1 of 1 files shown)


yigitfindikli:
Thanks for the contribution.

Issue #8486, opened 2020-01-24
'allSelected' does not exist on type 'MultiSelect'


Riscue:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Compile error when import the MultiSelectModule

**Expected behavior**
compilation with succesfully

**Minimal reproduction of the problem with instructions**
If module imported compilation fails.

**Please tell us about your environment:**
Angular CLI: 8.3.23
Node: 12.14.0
OS: linux x64
Angular: 8.2.14
 
**Language:** [all | TypeScript X.X | ES6/7 | ES5]
"typescript": "~3.5.3"

**Node (for AoT issues):** `node --version` =   
v12.14.0

Riscue:
Console output:
`ERROR in node_modules/primeng/multiselect/primeng-multiselect.d.ts.MultiSelect.html(26,109): Property 'allSelected' does not exist on type 'MultiSelect'.`

yigitfindikli:
fixed in https://github.com/primefaces/primeng/commit/b098062344716fbc86c84f9136cfbc581554599d

PR #8487, opened 2020-01-24
refactor allSelected -> isAllChecked()

Riscue:
`allSelected` field is changed to `isAllChecked()` method call because of error below:

https://github.com/primefaces/primeng/issues/8486


yigitfindikli:
Hi,
Thanks for the contribution.

PR #8489, opened 2020-01-25
Fix #8462 Multiple instances dynamicdialog

### LINKED ISSUE #8462: Multiple instances of the same Modal Dialog may be opened at a time
```
```

Behavior is visible from showcase website
https://www.primefaces.org/primeng/#/dynamicdialog

**Current behavior**

Click on Show button. Modal Dialog appears.
Click on the Space or Enter key on your keyboard. Another instance of the dialog appears.

**Expected behavior**

Keyboard presses should not be able to trigger the creation of new dialogs while one is currently active and set as modal.

tityus:
This closes issue #8462


meistaiwan:
I think this may have broken my application. I use one dialog to show some data, and a second pdialog to show a file. That broke very recently after an update from primeng 9 rc2 to primeng 9 rc4.
It's useful to show two different dialogs...

meistaiwan:
Verified. My main modal is through dialogservice, while a few components down inside I have a pdialog that shows something else (a file preview), that inner pdialog no longer works between rc2 and rc4, I suspect this code change is the issue

PR #8499, opened 2020-01-29
Fix #8498 Carousel - Invalid Array Length error

### LINKED ISSUE #8498: Carousel - Invalid Array Length error
```
```

**Plunkr Case (Bug Reports)**

Problem of issue still exists:
Carousel - Invalid Array Length error when array length = 0 #8300 

**Current behavior**
Throws error if data array has less elements as number of views in carousel is defined.

**Expected behavior**
No error independent of provided amount of data.

* **PrimeNG version:** 9.0rc4

* **Browser:** all

Language: Typescript | webpack

Rotzbua:
### Defect Fixes

Fix #8498 Carousel - Invalid Array Length error
Fix 6909edbb1f0d663f23d053e3ed1af7650ae49502

Current:
Just a case match if array length is 0 but not on negative numbers.

Fixed:
Now matches if array length is 0 or lower.

Edit:
This pr does not fix all problems of carousel if provided data contains less items 
than carousel should show.

### DIFF
--- a/src/app/components/carousel/carousel.ts
+++ b/src/app/components/carousel/carousel.ts
@@ -394,9 +394,10 @@ export class Carousel implements AfterContentInit {
 	totalDots() {
 		return this.value ? Math.ceil((this.value.length - this._numVisible) / this._numScroll) + 1 : 0;
 	}
+
 	totalDotsArray() {
-		let totalDots = this.totalDots();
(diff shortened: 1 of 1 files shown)


yigitfindikli:
Hi,
Thanks for the contribution.

PR #8503, opened 2020-01-30
Move saveState() method from call stack

kramar:
so saveColumnWidths() can use correct order of columns after reordering

The problem was in incorrect order of column DOM elements after reordering
so wrapping saveState() in timeout fix this problem

Fix for - https://pro.primefaces.org/browse/BMC-2


cagataycivici:
We'll change it a bit and make it run outside of the zone and merge, thank you.

kramar:
@cagataycivici Thank you!
Is there an ETA when fix will be merged?

PR #8505, opened 2020-01-31
Fixed #8335- Correct component flickering

### LINKED ISSUE #8335: Tooltip flickering
```
```

**Plunkr Case (Bug Reports)**
It can be simulated on PrimeNG showcase, so I didn't create a stackblitz.
tooltip_flickering

**Current behavior**
When mouse hover on input border the input starts to flicker not showing the tooltip. It only happens using tooltip with hover event and it can be simulated on PrimeNG showcase.

**Expected behavior**
When mouse hover on input border it should open tooltip and don't flicker.

It can be simulated on PrimeNG showcase doing the same as the gif above.

* **Angular version:** 8.X
* **PrimeNG version:** 8.1.1

* **Browser:** Chrome 78.0.3904.97

csvinhal:
PR to correct component flickering. Problem documented in issue #8335.

### DIFF
--- a/src/app/components/tooltip/tooltip.css
+++ b/src/app/components/tooltip/tooltip.css
@@ -3,6 +3,7 @@
     display:none;
     padding: .25em .5em;
     max-width: 12.5em;
+    pointer-events: none;
 }
 
 .ui-tooltip.ui-tooltip-right,


yigitfindikli:
Hi,
Thanks for the contribution.

PR #8510, opened 2020-02-02
[doc] Tooltip - fix switched row content

Rotzbua:
Documentation:

* fix switched row content
* add missing end points


yigitfindikli:
Thanks for the contribution @Rotzbua .

Issue #8511, opened 2020-02-03
Autocomplete doesn't clear input value if user tabs before suggestions are displayed


Primith:
**I'm submitting a ...**  (check one with "x")
```
[ X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/angular-8gfewa

**Current behavior**
Autocomplete with forceSelection option set on true emit onClear event only once completeMethod event has been triggered.
If user tabs before suggestions of autocomplete are displayed, the value typed is not erased.

**Expected behavior**
OnClear event should emit when value is cleared because of forceSelection option.

**Minimal reproduction of the problem with instructions**
See plunkr : 
- Type any value
- Tab before suggestions are displayed
- The value is not erased

This comportment is the same with an empty field or if a previous value has been selected. 

**What is the motivation / use case for changing the behavior?**
Users will not think the value is correct (the goal of the forceSelection property).

**Please tell us about your environment:**
Angular version: 8.2.14

PrimeNG version: 8.1.1

Browser: [all]

Language: [all]

souhailharrati:
@Primith  This problem is related to the forceSelection input
here is an example without forceSelection: https://stackblitz.com/edit/angular-9zfadc

souhailharrati:
I added pull request to resolve  this issue
https://github.com/primefaces/primeng/pull/8559

PR #8559, opened 2020-02-16
Fixed Autocomplete doesn't clear input value if user tabs before suggestions are displayed #8511

souhailharrati:
###Defect Fixes

###Feature Requests


cagataycivici:
Fixed with a different approach, thank you.

PR #8598, opened 2020-03-06
Fix 8597: click on header containing tags does not toggle sorter

### LINKED ISSUE #8597: Click header containing tags does not toggle sorter
```
[ x ] bug report => Search github for a similar issue or PR before submitting
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-7e8q34

```
    <ng-template pTemplate="header" let-columns>
        <tr>
            <th *ngFor="let col of columns" [pSortableColumn]="col.field">
                <div class="custom-table-header">{{col.header}}</div>
                <p-sortIcon [field]="col.field" ariaLabel="Activate to sort" ariaLabelDesc="Activate to sort in descending order" ariaLabelAsc="Activate to sort in ascending order"></p-sortIcon>
            </th>
        </tr>
    </ng-template>
```

**Current behavior**
In my custom table, I have col.header wrapped inside a div to apply a custom style.
However, when I click on this div, column sorter is not toggled.
This worked in primeNG v8.1.1 but is broken in v9.0.0, after https://github.com/primefaces/primeng/commit/9abd2e4aebf3c9a8c557a18fa6cbf0463a97dfc1.

**Expected behavior**
Click on column header (title) should toggle sorter.

* **Angular version:** 9.0.4
* **PrimeNG version:** 9.0.0

thaihacong:
Fix #8597.

### DIFF
--- a/src/app/components/dom/domhandler.ts
+++ b/src/app/components/dom/domhandler.ts
@@ -54,6 +54,14 @@ export class DomHandler {
             return new RegExp('(^| )' + className + '( |$)', 'gi').test(element.className);
     }
 
+    public static parentHasClass(element: any, className: string): boolean {
+        if (element.parentNode) {
+            return this.hasClass(element.parentNode, className);
+        } else {
+            return false;
+        }
+    }
+
     public static siblings(element: any): any {
         return Array.prototype.filter.call(element.parentNode.children, function (child) {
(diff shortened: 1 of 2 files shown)


cagataycivici:
Will revert the original change instead.

PR #8599, opened 2020-03-06
Fix 8591: resolve scrollFooter and scrollFooterBox @ViewChild after change detection

### LINKED ISSUE #8591: Table and TreeTable footer with vertical scroll might be misaligned with header and body
```
[ x ] bug report => Search github for a similar issue or PR before submitting
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-jkqfkq

**Current behavior**
TreeTable footer is always misaligned with header and body when vertical scroll is enabled.
This bug was introduced in primeng v9.0.0.
It worked before in primeng v8.1.1.

treetable-footer

**Expected behavior**
TreeTable footer is aligned with header and body when vertical scroll is enabled.
Column resize should be taken care as well.

The column in footer was misaligned from the beginning. Resizing columns did not fix the misalignment.

* **Angular version:** 9.0.4
* **PrimeNG version:** 9.0.0

thaihacong:
Set { static: false } to resolve scrollFooter and scrollFooterBox @ViewChild after change detection.
Otherwise, those fields are undefined. This makes alignScrollBar() does not work as expected and leads to the misalignment.
Fix #8591.

### DIFF
--- a/src/app/components/treetable/treetable.ts
+++ b/src/app/components/treetable/treetable.ts
@@ -1641,9 +1641,9 @@ export class TTScrollableView implements AfterViewInit, OnDestroy, AfterViewChec
 
     @ViewChild('loadingTable', { static: false }) scrollLoadingTableViewChild: ElementRef;
 
-    @ViewChild('scrollFooter', { static: true }) scrollFooterViewChild: ElementRef;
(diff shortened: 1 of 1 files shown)


cagataycivici:
Will also apply to Table.

Issue #8624, opened 2020-03-16
PTable - Paginator doesnot reset when data changes


kk1687:
Setup : PTable with paginator & resetPageOnSort set to true.

Issue recreation Step 1. data is updated on search, with results, lets say 100 rows are retrieved. Having rowCount set to 20, we will have pagelinks upto 5 and on page click the data is properly rendered.

Step 2. Move to any page other than Page#1.

Step 3. Search with a refined criteria The new results are loaded and data is rendered as it is the first page. But here the issue is that, the paginator just stays on the old state. Its not updated to page#1.

cagataycivici:
This is expected, when the data changes, the paginator only checks boundaries such as if current page still has data, if not it moves to last page. When data changes, first should not reset, assume you add a new row or delete one, it would be a bug if the paginator goes to first page in this case.

Built-in filtering resets it though since it is expected, in your case since you do the filtering, you need to reset it manually by a binding on first.

```xml
    <p-table [(first)]="first"
```

```ts
    filter() {
       //filter code
      //this.first = 0;
    }
```

kk1687:
I think on filter perspective your point is valid. I would like to raise a question specific to our scenario where we do expect some kind of fix.

Step 1: Load table with data, lets say pagesize is 20 & 100 rows in value.
here we will have 5 pages, lets say we move to page 3.

Step 2: Refine the search criteria and we get 3 records only. Now the paginator state will be an unknown one, though first page will be displayed. There is a strong possibility of stating it as a bug I believe. Kindly share your comments.

Simon-GHI:
Could you re-open this ticket as the issue is still there ?

"The paginator only checks boundaries such as if current page still has data, if not it moves to last page" 
It does not work as expected.
I reproduce the bug on stackblitz: https://stackblitz.com/edit/primeng-tabledoc-demo-vdi4uw?file=src/app/app.component.ts

Procedure :  
- Go on last page
- Hit filter button
- No data appears
- The paginator state get desynchronized. It shows only one page, the buttons are not disabled, the current page is not set to page 1. 

Regards,
Simon

PR #8626, opened 2020-03-16
Fix for paginator bug # 8624

kk1687:
https://github.com/primefaces/primeng/issues/8624

Setup : PTable with paginator & resetPageOnSort set to true.

Issue recreation Step 1. data is updated on search, with results, lets say 100 rows are retrieved. Having rowCount set to 20, we will have pagelinks upto 5 and on page click the data is properly rendered.

Step 2. Move to any page other than Page#1.

Step 3. Search with a refined criteria The new results are loaded and data is rendered as it is the first page. But here the issue is that, the paginator just stays on the old state. Its not updated to page#1.

Fix
A handler for paginator is required to update the paginator state and when resetPageOnSort flag is set, paginator state needs to be updated to Page#1


cagataycivici:
This is not a bug, see the issue please.

PR #8628, opened 2020-03-18
Fix sample source bug in Calendar

### LINKED ISSUE #8627: [doc] wrong demo code in Calendar doc
https://primefaces.org/primeng/showcase/#/calendar

In this example, there are two pieces are missing:

```
date14: Date;

tr: any
```

doggy8088:
This PR fixes #8627


cagataycivici:
Good catch, thank you.

PR #8658, opened 2020-03-25
Calendar.ts - just refactoring.

dpischalka:
Sorry if I had to create issue before, but I think about issue for refactoring like something weird.
Recently I faced with "calendar.ts" component, and by interest was investigating source code.
Except for other, I found some ideas for make code more clear. It's not ideal for me, but I think this code will looks better with my changes.

Component was tested.

I didn't make some architectural changes to code, just refactor for more clear view.
Thanks.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -1,154 +1,157 @@
 {
-    "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
-    "version": 1,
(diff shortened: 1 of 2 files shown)


cagataycivici:
Unable to accept due to large number of changes eg. 2000 lines of diff

Issue #8714, opened 2020-04-05
dismissableMask not working in dynamic dialog


IVmostafa:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
clicking in the modal background not closing the dialog

**Expected behavior**
if clicking the modal background should hide the dialog.

**Minimal reproduction of the problem with instructions**
Call a method to open dynamic dialog with `dismissableMask` set to true and try to close it by click on the background

**What is the motivation / use case for changing the behavior?**
We want to close the dialog when user click on the background

**Please tell us about your environment:**
Mac OS, VS Code but should be reproducible in all environments

* **Angular version:** 5.X
9.0.7

* **PrimeNG version:** 5.X
9.0.5

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
Chrome but should be reproducible in all browsers

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
TypeScript 3.7.5

* **Node (for AoT issues):** `node --version` =   
12.16.0

* **Note:** 
in `https://github.com/primefaces/primeng/blob/master/src/app/components/dynamicdialog/dynamicdialog.ts` the method `onMaskClick` was called in line 21 in previous version `8.1.1` now it isn't
change `<div #mask class="ui-widget-overlay ui-dialog-mask ui-dialog-visible ui-dialog-mask-scrollblocker">` 
to `<div #mask class="ui-widget-overlay ui-dialog-mask ui-dialog-visible ui-dialog-mask-scrollblocker" (click)="onMaskClick()">`
in line 21 in the mentioned link

i have created PR for the fix, tested it and it's working check my PR https://github.com/primefaces/primeng/pull/8715

Grauen:
@cagataycivici  this might be related to https://github.com/primefaces/primeng/issues/8720 :)

fredericojesus:
Having the same issue, dismissable mask is not working at all in dynamic dialog.

KissBalazs:
Having the same issue, but I am not using p-table in my project.

[...]

PR #8715, opened 2020-04-05
Fix dismissableMaks not closing the dialog issue

IVmostafa:
###Defect Fixes
https://github.com/primefaces/primeng/issues/8714


Khousheish:
🎉 thank you for fixing this!!!

yigitfindikli:
Thanks for the PR but we decided to the same implementation like dialog.

You can see https://github.com/primefaces/primeng/commit/58a694cd090405a8c3a0209c13884905a63f99dc

Issue #8455, opened 2020-01-15
Dropdowns and MultiSelects open inconsistently or not at all when using a wrapped-label layout


GregKosmo:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case**
https://stackblitz.com/edit/github-6wsadz

**Current behavior**
After surrounding a `p-dropdown` or `p-multiSelect` with a `label` tag, clicking on the dropdown icon or any portion of the multiSelect no longer expands their options.

**Expected behavior**
The components should behave the same regardless of whether they are wrapped in a `label` tag or not.

**Minimal reproduction of the problem with instructions**
1. Insert a `p-dropdown` or `p-multiSelect` in HTML and provide it an array of options.
2. Surround the component with a HTML `label` tag and add some arbitrary label.
3. Observe as the component no longer expands to show its options when clicked.

See example here: https://stackblitz.com/edit/github-6wsadz 

**What is the motivation / use case for changing the behavior?**
My organization would like to implement a wrapped-label form framework so labels and inputs are automatically associated with one another and we no longer have to provide "for" and "id" tags on our labels/inputs. It will save us a lot of headache and boilerplate code in the long run.

**Please tell us about your environment:**
* **OS:** Windows 10
* **IDE:** Visual Studio Code
* **Package Manager:** NPM
* **HTTP Server:** Apache Tomcat
* **Angular version:** 7.2.7, verified on 9.0.0-rc4
* **PrimeNG version:** 7.1.0, verified on 9.0.0-rc1
* **Browser:** verified on [Chrome 79.0.3945.117 | Firefox 72.0.1 | Edge 79.0.309.65 | IE 11.535.18362.0]
* **Language:** [TypeScript 3.2.4 | ES6/7]
* **Node (for AoT issues):** `node --version` =   v10.16.3

Cito:
I can confirm this problem. Was able to solve it via PR #8742.

cagataycivici:
Isn't it invalid to wrap with label, label has even no "for", why not use a div or similar instead.

Cito:
It is actually valid and there are good reasons to do so, see my comment in #8742.

GregKosmo:
[...]

PR #8742, opened 2020-04-15
Fix issue #8455 with clicking on dropdowns inside labels

Cito:
When the dropdown is wrapped inside a label, clicking on the dropdown (right next to the text) can open the dropdown twice, thereby effectively closing it again because the click event works as a toggle.

### DIFF
--- a/src/app/components/dropdown/dropdown.ts
+++ b/src/app/components/dropdown/dropdown.ts
@@ -513,6 +513,8 @@ export class Dropdown implements OnInit,AfterViewInit,AfterContentInit,AfterView
     }
     
     onMouseclick(event) {
+        event.preventDefault();
+
         if (this.disabled||this.readonly) {
             return;
         }


cagataycivici:
Isn't it invalid to wrap with label, label has even no "for", why not use a div or similar instead.

Cito:
Hi @cagataycivici - it *is* in fact valid to wrap a control with a label, and the fact that you don't need the "for" attribute in this case is the very reason why you often want to do that. If you have a component that is re-used multiple times on a page, it is often much easier and safer to wrap the component than inventing an id for every instance and making sure they are unique for the page.

The [HTML 5 spec](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) says: "The label represents a caption in a user interface. The caption can be associated with a specific form control, known as the label element’s labeled control, either using for attribute, *or by putting the form control inside the label element itself*."

Cito:
Just for the records: This has now been fixed in a different (and probably more proper) way by  @cagataycivici, and that fix will soon be available in PrimeNG 9.1.0.

Issue #8804, opened 2020-04-30
OverlayPanel doesn't disappear in OnPush components.


Hapcy:
**I'm submitting a ...**  (check one with "x")
```
[x ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-6dh6gj

**Current behavior**
OverlayPanel only disappears in ChangeDetectionStrategy.OnPush components, when another change detection runs because the markForCheck on https://github.com/primefaces/primeng/blob/469fcc05ae23e9b40eaa3efe41fa209b1e4a4e55/src/app/components/overlaypanel/overlaypanel.ts#L123 is outside of zone and doesn't trigger change detection.

**Expected behavior**
OverlayPanel should work like in ChangeDetectionStrategy.Default components.

* **Angular version:** 9.0.4
* **PrimeNG version:** 9.0.6

MaKCbIMKo:
Hi @Hapcy - I'm not sure, but it seems like PrimeNG team has it fixed slightly different.

I saw some commits to adjust `markForCheck` calls, but it wasn't released yet (from what I can see).

I also faced this problem with two different cases:
1. closing panel by clicking outside
2. window resize

But there definitely can be more affected use-cases...

cagataycivici:
Should be fixed in 10.0 as overlaypanel itself is also using push now.

PR #8805, opened 2020-04-30
hide overlaypanel in onpush components

Hapcy:
markForCheck was outside zone so it didn't trigger changedetection

###Defect Fixes
See #8804


cagataycivici:
Should be fixed in 10.0 as overlaypanel itself is also using push now.

PR #8834, opened 2020-05-09
Removed unclosed span in source tab inputswitch

Digender:
An unclosed span was part of source example for inputSwitch. Hence not creating and issue for this

###Defect Fixes

###Feature Requests


cagataycivici:
Seems, we've removed the value display from demo.

PR #8874, opened 2020-05-23
Fix import in the galleria component

Cito:
This is just a small fix of the import statement for the `UniqueComponentId` in the `Galleria` component.

Without this change, I got errors about "imports from outside of the rootDir" when running `npm run build-lib`. This fix also makes the import consistent with how other components such as `Carousel` import `UniqueComponentId`.


yigitfindikli:
Hi @Cito,
Thanks for the contribution but that was added in [this](https://github.com/primefaces/primeng/commit/1245c25d67dbc21608f5d75b878e3972fd9676ab) commit.

PR #8875, opened 2020-05-23
Add ES2016 as lib file to allow includes() on arrays

Cito:
This small change was necessary for me in order to succesfully execute `npm run build-lib`.

The problem is that `VirtualScroller.loadPage()` uses the `includes()` method on the `loadPage` array of numbers. That method only exists in ES2016, but the tsconfig had only ES2015 specified.

### DIFF
--- a/src/app/components/tsconfig.lib.json
+++ b/src/app/components/tsconfig.lib.json
@@ -12,6 +12,7 @@
       "rootDir": ".",
       "lib": [
         "es2015",
+        "es2016",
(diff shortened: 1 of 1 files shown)


yigitfindikli:
Hi @Cito,
Thanks for the contribution but that was added [this](https://github.com/primefaces/primeng/commit/b0d4c6dd093abe2e869da1ab49d574e40a39e882) commit.

Issue #8826, opened 2020-05-06
InputTextArea auto-resize moves scroll position


bjovanov:
I'm submitting a ...
[x] bug report => Bug fixed with https://github.com/primefaces/primeng/issues/7481 is re-introduced.

**Current behavior**
Auto resize moves scroll position. Issue still persists due to this line
https://github.com/primefaces/primeng/blob/master/src/app/components/inputtextarea/inputtextarea.ts#L64

**Expected behavior**
Setting auto resize shouldn't affect the scrolling. Issue is solved if the offending line is removed.
Current workaround is by overriding the resize method of textarea, but proper solution is needed.

This is broken in latest v8 (8.1.1) and also on master.

ginettev:
I am facing the same issue with v8.1.1
Could you show me show did you override the resize method for the textarea?

ginettev:
With some more fiddling I realised that the height = 'auto' is not the culprit.
The proper fix is as follows:

```
      **const scrollLeft = this._window.pageXOffset;
      const scrollTop = this._window.pageYOffset;**

        this.el.nativeElement.style.height = 'auto';
        this.el.nativeElement.style.height = this.el.nativeElement.scrollHeight + 'px';

        if (parseFloat(this.el.nativeElement.style.height) >= parseFloat(this.el.nativeElement.style.maxHeight)) {
            this.el.nativeElement.style.overflowY = "scroll";
            this.el.nativeElement.style.height = this.el.nativeElement.style.maxHeight;
        }
        else {
            this.el.nativeElement.style.overflow = "hidden";
        }

      **this._window.scrollTo(scrollLeft, scrollTop);**

        this.onResize.emit(event||{});


```

bjovanov:
I was referring to this commit https://github.com/primefaces/primeng/commit/d90db593c2c34931ec043cd9ee48514235150cde where removing one assignment of this.el.nativeElement.style.height somehow solves the problem and also with the explanation from here https://github.com/angular/material/issues/3070 I assumed that the 'auto' setting has something to do with it. Your solution makes sense, but still I cannot explain why having two assignments on same property one after the other causes scroll jumping.

ShelbyKelley:
I am experiencing this issue in my corporate application, is there a timeline for this bug to be fixed?

JulianFarhi:
Hi,
I'm facing the same problem but slightly different.
In my case, I use the inputTextarea inside a dialog (DynamicDialog).
[...]

PR #8908, opened 2020-06-02
Fix for InputTextArea auto-resize moves scroll position #8826

ginettev:
Remove the auto style for height as it was causing the scrollbar to jump for long text

###Defect Fixes

###Feature Requests


cagataycivici:
https://github.com/primefaces/primeng/issues/9129

PR #8910, opened 2020-06-02
dialog resize not scaling when centered

eberling:
Until container position has been changed (by dragging), container maintains centering on resize due to flexbox parent, causing it to resize in each direction equally, thus doubling size increase (as the right side is being resized, the left side also expands, maintaining centering). Therefore we add a second deltaX and deltaY distance, accounting for the other side.

Once container position has been changed, resize behavior changes and resizing causes only one side to resize. 
Because dragging causes the container to add css style "top" and "left", we can assess whether centered resizing behavior is active, storing it in "hasBeenDragged".

Behavior can be demoed here
https://stackblitz.com/github/qgnirxral
or naturally, here
https://www.primefaces.org/primeng/showcase/

### DIFF
--- a/src/app/components/dialog/dialog.ts
+++ b/src/app/components/dialog/dialog.ts
@@ -475,6 +475,12 @@ export class Dialog implements OnDestroy {
             let minHeight = this.container.style.minHeight;
             let offset = DomHandler.getOffset(this.container);
             let viewport = DomHandler.getViewport();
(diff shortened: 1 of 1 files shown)


eberling:
captured (1)

yigitfindikli:
Hi @eberling ,
Good catch! Thanks for your contribution.

PR #8946, opened 2020-06-13
Spinner always at center of screen for long tree table

ae-ae:
Spinner location fixed to be always at center of screen for scrollable tree table


cagataycivici:
Please create a ticket first and also PR needs an update as PrimeNG 10 has new styling called PrimeOne.

Issue #8974, opened 2020-06-24
Missing Table component method closeCellEdit


setfloat:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**

No public method by the name `closeCellEdit` exists on the table component; however, the docs indicate this method exists.

**Expected behavior**

The [table docs](https://github.com/primefaces/primeng/blob/6f1c6c30a67b78c2b5979a0656a917fb7be3a494/src/app/showcase/components/table/tabledemo.html#L3166) indicate that this method should exist and should be exposed.  Possible solutions are to amend the docs and remove the entry for `closeCellEdit` method or to add the method back onto the Table component, possibly under a different name if the behavior changes.

**Minimal reproduction of the problem with instructions**
The public method `closeCellEdit` was removed in this [commit](https://github.com/primefaces/primeng/commit/8f439c35de07abcad01a680c70a7a635583404ff#diff-0d572934b5355d23d4b13eb1f307f180L1417) from v7.0.0.

![image](https://user-images.githubusercontent.com/16740925/85502630-8b035b00-b59c-11ea-9e67-f5ebc40f13ff.png)

**What is the motivation / use case for changing the behavior?**
The docs don't accurately describe the table component's public methods.

* **Angular version:** 9.1.11

* **PrimeNG version:** 9.1.0

* **Browser:** all

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

PR #8975, opened 2020-06-24
Remove deprecated method from table docs

setfloat:
`closeCellEdit` is no longer a public method on the table component.

###Defect Fixes
Possible solution to https://github.com/primefaces/primeng/issues/8974 .  If there is no immediate intent to restore the `closeCellEdit` method to the table component, this fix would remove the confusion in the documentation.


yigitfindikli:
Hi @setfloat,
Thanks for your contribution.

PR #8992, opened 2020-07-01
Fix #8887: Ivy/aot compilation errors.

### LINKED ISSUE #8887: error on enableIvy false & fullTemplateTypeCheck true
**I'm submitting a
```
```
https://stackblitz.com/edit/github-xamncx is the attempt to reproduce it, but the strange part it's not reproducible on stackblitz while if you download source and run locally with lts node 12, issue is reproducible.

When using  "fullTemplateTypeCheck": true & enableIvy: false set in project's tsconfg.json, I get following errors

`ERROR in node_modules/primeng/dropdown/primeng-dropdown.d.ts.Dropdown.html(17,83): Expected 0 arguments, but got 1.
node_modules/primeng/table/primeng-table.d.ts.ScrollableView.html(18,63): This condition will always return 'true' since the types 'boolean' and 'string' have no overlap.
node_modules/primeng/table/primeng-table.d.ts.ScrollableView.html(27,75): This condition will always return 'true' since the types 'boolean' and 'string' have no overlap.
`
It should not give an error for using  "fullTemplateTypeCheck": true in tsconfig.json and enableIvy: false

Angular 9.1.0 and above & prime-ng 9.1.0

As per this https://github.com/primefaces/primeng/issues/6998 seems someone misses something.

Ubuntu 20.04

* **Node (for AoT issues):** `node --version` =   v12.16.3

kirks:
###Defect Fixes
Fix compilation errors for projects that have Ivy/AOT/fullTemplateTypeCheck options enabled.
#8887


yigitfindikli:
Hi @kirks,
Thanks for your contribution, but added manually.

PR #9040, opened 2020-07-14
A problem with using p-inputnumber  in incell editing

kltssll:
On the table-cell edit mode, enter p-inputnumber control directly after typing the value, the page is not worth updating, but stackblitz the environment is good. After adding enter key to trigger the event was resolved, now submit up to the modified code, hope to get a timely review and reply.

###Defect Fixes

###Feature Requests


cagataycivici:
I'm sorry but the files in the PR are not valid and cannot be merged.

Issue #9038, opened 2020-07-14
[bug] in  code merged in button.ts


jmesa-sistel:
@yigitfindikli 
There is a little bug in code merged [(PR9020)](https://github.com/primefaces/primeng/pull/9020) here:
https://github.com/primefaces/primeng/blob/63afa79e76ffa38a8dc203206af7d062406d89ef/src/app/components/button/button.ts#L72

It should be:
`this.el.nativeElement.className` = styleClass + ' ' + this._initialStyleClass;

behroozbc:
this bug fixed in https://github.com/primefaces/primeng/commit/6e37bcbd46bf0cc13916f9f358a02cde6869a34d

PR #9041, opened 2020-07-14
Fix bug concat css class

behroozbc:
###Defect Fixes
fix bugs when concat css class
https://github.com/primefaces/primeng/issues/9038


behroozbc:
this fixed

PR #9053, opened 2020-07-20
Fix tooltip accessibility issue

fhljys:
###Defect Fixes
https://github.com/primefaces/primeng/issues/9052

###Feature Requests


cagataycivici:
Regular html title property also behaves the same. You may consider focus event instead when there are interactive content inside tooltip.

PR #9057, opened 2020-07-20
Adding Cache on Auto Complete

VictorRChagas:
It's gonna improve user experience and save time and requests from server and client.


VictorRChagas:
I will edit the way the data is stored in the client, switching it from the local storage to local data base and edit the name of variables.

PR #9252, opened 2020-09-08
Added ability use separated frozen rows template for frozen columns

AzatKhalilov:
###Defect Fixes
If you want make different template for frozen rows in frozen columns you can't. This MR allow such functionallity


AzatKhalilov:
Found issue

yigitfindikli:

PR #9265, opened 2020-09-10
Fixed radiobutton with template-driven forms #9162

### LINKED ISSUE #9162: Model driven forms support for RadioButton
**I'm submitting a ...** 

```

**Plunkr Case (Bug Reports)**
Sorry, I'm not able to run  primeng 10.0.0-rc.3 on stackblitz. But the code is very simple: 
 <form [formGroup]="myFormGroup">
      <p-radioButton name="groupname" label="True" [value]="true" formControlName="radio"></p-radioButton>
      <p-radioButton name="groupname" label="False" [value]="false" formControlName="radio"></p-radioButton>
    </form>

**Current behavior**
I have a radio button group with 2 radio buttons inside a model driven form. If i select button one and than button 2, the right value is stored inside the form but in the user interface both radio buttons are selected. Also both have the class "p-radiobutton-checked"

* **Angular version:** 10.0.8
<!-- Check whether this is still an issue in the most recent Angular version -->

* **PrimeNG version:** 10.0.0-rc.3

sergeypayu:
###Defect Fixes
#9162 

Fix radiobutton for template-driven forms. Cause of the problem: writeValue is not called on controls other than control that is changed by user. If there are 3 radio buttons and user clicks on the first one, 2nd and 3rd controls will not receive updated value through writeValue call. It'is not really a bug, but rather made on purpose by angular team. To work around the problem we need a singleton service that will keep registry of all buttons in all groups. When user changes value of one of the button, corresponding control propagates this changes to all sibling buttons in the same form through registry.

The solution and code is borrowed from angular's RadioControlValueAccessor (https://github.com/angular/angular/blob/c0523fc3b45cdb5ba52109f7ae0e4dab555b7d72/packages/forms/src/directives/radio_control_value_accessor.ts)

### DIFF
--- a/src/app/components/radiobutton/radiobutton.ts
+++ b/src/app/components/radiobutton/radiobutton.ts
@@ -1,12 +1,47 @@
-import {NgModule,Component,Input,Output,ElementRef,EventEmitter,forwardRef,ViewChild,ChangeDetectorRef,ChangeDetectionStrategy} from '@angular/core';
+import {NgModule, Component, Input, Output, ElementRef, EventEmitter, forwardRef, ViewChild, ChangeDetectorRef, ChangeDetectionStrategy, Injectable, OnInit, OnDestroy, Injector} from '@angular/core';
 import {CommonModule} from '@angular/common';
-import {NG_VALUE_ACCESSOR, ControlValueAccessor} from '@angular/forms';
+import {NG_VALUE_ACCESSOR, ControlValueAccessor, NgControl} from '@angular/forms';
 
 export const RADIO_VALUE_ACCESSOR: any = {
     provide: NG_VALUE_ACCESSOR,
     useExisting: forwardRef(() => RadioButton),
     multi: true
 };
+
+/**
+ * Internal class used by RadioButton to uncheck radio buttons with the matching name.
+ */
+@Injectable({
+    providedIn: 'root',
+})
+export class RadioControlRegistry {
+    private accessors: any[] = [];
+
+    add(control: NgControl, accessor: RadioButton) {
+        this.accessors.push([control, accessor]);
+    }
+
+    remove(accessor: RadioButton) {
+        this.accessors = this.accessors.filter((c) => {
+            return c[1] !== accessor;
+        })
+    }
+
+    select(accessor: RadioButton) {
+        this.accessors.forEach((c) => {
+            if (this.isSameGroup(c, accessor) && c[1] !== accessor) {
+                c[1].writeValue(accessor.value);
+            }
+        });
+    }
+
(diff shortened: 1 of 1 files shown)


sergeypayu:
@cagataycivici please review. It affects our production code

huineng:
Yes indeed , we can now select all buttons

image

huineng:
any update on this ?

malbarmavi:
I have the same problem 😞😞

SimonCorellia:
Same problem
Is it planned for 10.1 ?

tiagomartins91:
Same here....

karthikdeva:
image

tiagomartins91:
Any update on this pr? Is it planning for 10.0.1?

cagataycivici:
Any way to solve it without the global service? I mean if angular could just call writeValue, that would solve it. I'll create a ticket at Angular to see if this can be solved by the core and ping the team.

sergeypayu:
> Any way to solve it without the global service? I mean if angular could just call writeValue, that would solve it. I'll create a ticket at Angular to see if this can be solved by the core and ping the team.

I don't think it's possible without the registry. I took the code from angular's RadioControlValueAccessor. It's their own approach. If they do it using special registry, chance are we will have to use it too. 

We also will much appreciate if you could still include the fix to 10.0.1. It's been 3 weeks

mjonas4356:
This could be working #9339

sergeypayu:
Hello? Also how is this an enhancement if this is clearly a bug? Radio buttons are unusable with Primeng 10 unless you use model-driven forms, which may simply be not an option.

huineng:
unless ?

EricWagnerJr:
I'm also waiting on this. Had to disable radio buttons in production because of it.

yigitfindikli:
Hi @sergeypayu,

Thanks for your report and contribution.

Best Regards.

huineng:
This has not yet been published right ?

SimonCorellia:
Sorry if obvious but same question.
This has not yet been published right ? Is it planned to be published for version 10 of PrimeNg ?

EricWagnerJr:
> Sorry if obvious but same question.
> This has not yet been published right ? Is it planned to be published for version 10 of PrimeNg ?

@SimonCorellia from what I can tell, they totally back tracked and skipped the milestone they had for 10.1 when they were supposed to release it. Now my group is put in the tough spot to request the business to upgrade to 11 for the sake of radio buttons or just replace it with something else. Frustrating it has come to this.

bendhanoa:
@EricWagnerJr, The workaround suggested by @karthikdeva temporality fixes the issue. You can use it to avoid upgrading.

EricWagnerJr:
@bendhanoa We had 2 members on our team try it out and it still misbehaved. Thank you for the input, though!

PR #9297, opened 2020-09-21
Fixed lint issues, Modified lint rules, Added Travis lint check

artemnih:
This PR fixes nearly 25000 existing lint issues without modifying the logic or unit tests.
- tslint rules have been updated to match the overall coding style 
- added automatic lint fix script `npm run lint:fix`
- added Travis lint check


cagataycivici:
Thank you, for a size of this change, we should do it on our side. Did you use ng lint --fix? to begin with?

artemnih:
@cagataycivici yes, used `ng lint --fix`, then manually fixed what's left

artemnih:
@cagataycivici my plan was to first fix lint, then add strong types everywhere. We are using this library in many of our projects and those types

cagataycivici:
Thanks, due to amount of changes, I think it would be better to do this on our end. thank you.

PR #9318, opened 2020-09-25
Added ability to set styleClass to separator

DrPennybags:
Hopefully this is a simple enough pull request. I implemented the styleClass property on the menubar for a separator. Currently with the public release it is not possible to style using a class.


cagataycivici:
If we add this to menubar, all other menu components should receive it as well.

DrPennybags:
> If we add this to menubar, all other menu components should receive it as well.

The MenuItem object has this property so it only makes sense that a separator would have the property as well. I don’t think adding this would affect all other objects.

yigitfindikli:
Hi,

There are conflicts about PR, so I'm gonna close the PR. You can create an issue ticket for this enhancement. We'll be glad to get community feedback.

PR #9356, opened 2020-10-02
Reset loadedPages on sort/filter for lazy loaded table

Jakob-em:
### Defect Fixes
Currently if a table is using lazy loading and virtual scroll there is a problem after sorting or filtering the table.
Because the internal list of the loaded pages is not reset when sorting or filtering, the previously loaded chunks won't be loaded again after sorting/filtering.
This PR fixes this problem.


dakotajd:
👍 for this. I had quite the time working around this issue in a recent app.

cagataycivici:
loadedPages are gone so closing, thank you.

Issue #7284, opened 2019-02-24
Lazy Loading support for Dropdown and MultiSelect Filtering


olegbabkevich:
**I'm submitting an enhancement**  (check one with "x")

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/lazy-load-filter

**Current behavior**
Now controls are not designed, that options could be loaded on filter input. There is client-side filtration happens on filter (input).

**Expected behavior**
I expect an option, that visible in filtration options might be loaded on filter input.

**What is the motivation / use case for changing the behavior?**
The case when we need to do such filtration on back-end.

**Please tell us about your environment:**

* **Angular version:** 5.X
 7.2.2

* **PrimeNG version:** 5.X
 7.0.5

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
* Chrome: 72.0.3626
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

* **Node (for AoT issues):** * `node --version` =  8.9.1

yelhouti:
I'm working on this on a local branch, but I don't think you can manage to convince the prime team to merge my code, any ways let's discuss it:
I think the options should resemble those of Table, ie:
```
 <p-multiSelect [lazy]="true" [lazyLoadOnInit]="false" (onLazyLoad)="onLazyLoadEvent($event)"
                               [options]="options" ....></p-multiSelect>
```
When lazy is true, infinit-scroll is set to true.

Implementation concerns:
For multiselect, the select/unselect all checkbox will not work as it should as not all elements are visible, it should be replaced with a clear button...

yelhouti:
You can find it implemented here in the meantime, I'll try to keep my branch up to date:
https://github.com/yelhouti/primeng/releases/tag/8.0.0-rc2

Walteann:
I also wanted to know. This PR would suit my scenario

hanyska:
+1

ashelest:
I Also wanted to know. Is it possible to extend MultiSelect with lazy loading functionality ?
All of us would appreciate you.

Xriuk:
Any news? This would be very useful

leandroheidemann:
+1

mertsincan:
Hi,

[...]

PR #9359, opened 2020-10-05
optionValue + lazyLoad select dropdown....

yelhouti:
###Defect Fixes
fixes: #7284
This a new PR for lazy dropdown... with no complicated infinite scroll..., it is supposed to be used with filter...


cagataycivici:
Thank you, however there are too many changes for us to consume, test and maintain. We'll be implementing it on our side however we'll use your PR as a source of feedback. Thank you very much.

PR #9370, opened 2020-10-07
[issues/9364]: Fix newValue slicing

### LINKED ISSUE #9364: InputNumber cannot set value when format property is false
```
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/github-t9sujt?devtoolsheight=33&file=src/app/app.component.html

**Current behavior**

An error is shown in the console when typing in a number:

`Error: newValue.slice is not a function`

**Expected behavior**

No error should be thrown.

If you set `format` to false, you get an error when typing in a number. The problem can be reproduced with this simple example:

```
<p-inputNumber [min]="8" [max]="20" [showButtons]="true" [ngModel]="9" [format]="false"></p-inputNumber>
```

* **Angular version:** 10.1.4

* **PrimeNG version:** 10.0.2

OskarsPakers:
###Defect Fixes

#9364

### DIFF
--- a/src/app/components/inputnumber/inputnumber.ts
+++ b/src/app/components/inputnumber/inputnumber.ts
@@ -843,7 +843,8 @@ export class InputNumber implements OnInit,ControlValueAccessor {
 
                 const tExpr = insertedValueStr.split('').join(`(${this.groupChar})?`);
                 const tRegex = new RegExp(tExpr, 'g');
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks a lot for the PR! But, this problem was solved in a different way.

Best Regards,

PR #9371, opened 2020-10-07
Fix incrementing value in inputNumber with spinner when input is empty

### LINKED ISSUE #9337: InputNumber with spinner mode throws a JS exception
```
[x] bug report 
[ ] feature request 
[ ] support request
```

**Current behavior**
TypeError gets thrown when text-field of input-number component is empty and the value is about to be incremented or decremented via buttons.

**Expected behavior**
TypeError does not get thrown. New value of the field should be 0 +/- [step] if no minimum or maximum boundaries are set.

- Go To PrimeNG Documentation InputSpinner Component
- In the example with buttons delete the value of the input field
- open Dev-Tools in browser
- try to increment or decrement the value

OskarsPakers:
###Defect Fixes
#9337


mertsincan:
Unfortunately, closed; https://github.com/primefaces/primeng/issues/9337#issuecomment-705220544

PR #9389, opened 2020-10-10
Inputmask - returning error 'begin' of undefined when mask is populated from chrome address

TinTrempetic:
If the mask was populated from an address saved in chrome, this.caret() will return null and error will occur. 
Happens only on android devices.


cmg-cchandler:
This error still occurs in version 16.4.1 , and the fix proposed here would fix it.  

image

Issue #9299, opened 2020-09-22
chips input: optional delete on backspace


capc0:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**What is the motivation / use case for changing the behavior?**

It would be nice if we could enable/disable the delete mechanism of chips with backspace.
Would be pretty simple to implement: add a new @Input property ```deleteOnBackspace``` and do not handle the backspace keypress in https://github.com/primefaces/primeng/blob/master/src/app/components/chips/chips.ts#L242

PR #9414, opened 2020-10-16
chips input: paremeter to disable delete on backspace

OskarsPakers:
###Feature Requests
#9299


yigitfindikli:

OskarsPakers:
"Nice"! Could you close issues that you are not accepting fixes to instead of waiting for contribution and then rejecting it?

PR #9432, opened 2020-10-21
Fix #9431: Customize p-sortIcon icon

### LINKED ISSUE #9431: Customize p-sortIcon icon
```
```

**Current behavior**

When using SortIcon component in a Datatable, the icon used to display sorting direction and capability is forced to pi-sort-alt family of icons. Sometimes it is desired that these icons vary (merely for visual purposes as this has no functional influence).
**Expected behavior**

One should be able to customize icons based on input parameters as such:
`<p-sortIcon sortAscIcon="pi-sort-up" sortDescIcon="pi-sort-down" sortNoneIcon="pi-sort"></p-sortIcon>`

Sometimes it is desired to use a different set of icons, like the non-alt sort icons, or even chevron icons (Depends entirely on the business).

OS: Windows 10 x64
IDE: PhpStorm 2020.2
Package manager: NPM v6.14.5
* **Browser:** [all]

 
* **Language:** [all]

scriptom:
Fixes #9431, allows de customization of the p-sortIcon icons used for displaying in sortable tables.

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -1331,7 +1331,7 @@ export class Table implements OnInit, AfterViewInit, AfterContentInit, Blockable
                             else {
                                 localMatch = this.executeLocalFilter(filterField, this.value[i], filterMeta);
                             }
-                            
+
                             if (!localMatch) {
                                 break;
                             }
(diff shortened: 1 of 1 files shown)


cagataycivici:
This should better be done from PrimeNGConfig globally or with css, it would be a pain to do this over and over again for every table with many columns.

Issue #9441, opened 2020-10-22
OverlayPanel target arrow points to the wrong position when target is on the right edge of the screen


dakotajd:
### 
### **FIXED IN**: https://github.com/primefaces/primeng/pull/9442
_________________________________________________________________

**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**StackBlitz Case (Bug Reports)**
Please demonstrate your case at stackblitz by using the issue template below. Issues without a test case have much less possibility to be reviewed in detail and assisted.

https://stackblitz.com/edit/primeng-overlaypanel-demo-tqua3z?file=src/styles.css

**Current behavior**
The OverlayPanel component's arrow points to the wrong place if the target is up against the edge of the right side of the screen. There is code to handle this in the align() method: https://github.com/primefaces/primeng/blob/a8551d4d2b29be823fbc1e74a2bb15c9a1932e5b/src/app/components/overlaypanel/overlaypanel.ts#L187-L199 but the CSS class: p-overlaypanel-shifted does not exist in https://github.com/primefaces/primeng/blob/master/src/app/components/overlaypanel/overlaypanel.css

The CSS looks to have (unintentionally?) been removed during the migration to PrimeOne (the switch from .ui- to .p- prefixes for class names) in this commit here: https://github.com/primefaces/primeng/commit/c7dd064da6e05d549a1a91f33ac3a9b5aa5dd267#diff-4166dfa8da6660ec4e195e891b761a9ce8f5e49db19ace3613ac9d66ad8e4f94L56

![image](https://user-images.githubusercontent.com/22322743/96893999-e8c38900-1450-11eb-8d72-dd15a349fcf1.png)

**Expected behavior**
The arrow should be shifted to the other side of the panel to point to the target if it is against the right edge of the screen.
![image](https://user-images.githubusercontent.com/22322743/96894806-7acb9180-1451-11eb-81fa-8fce7c063436.png)

**Minimal reproduction of the problem with instructions**
https://stackblitz.com/edit/primeng-overlaypanel-demo-tqua3z?file=src/styles.css

**What is the motivation / use case for changing the behavior?**
I want the arrow to point to the right place 🙂

[...]

PR #9442, opened 2020-10-22
Issue #9441: Re-add CSS for p-overlaypanel-shifted

dakotajd:
Re-added CSS for p-overlaypanel-shifted class so that the arrow points to the correct location when the panel is up against the right edge of the screen. This CSS was (unintentionally?) removed in commit: https://github.com/primefaces/primeng/commit/c7dd064da6e05d549a1a91f33ac3a9b5aa5dd267#diff-4166dfa8da6660ec4e195e891b761a9ce8f5e49db19ace3613ac9d66ad8e4f94L56 

Note this CSS was also removed (ui should be renamed to p):
```
.ui-overlaypanel-shifted:after {
    margin-right: -8px;
}

.ui-overlaypanel:before {
    margin-right: -10px;
}
```
But I chose not to add it back, because with it, the arrow overlaps the close button slightly. Without it though, the arrow is offset from the edge on the right side of the panel slightly more than it is when it appears on the left. Perhaps this could be made dynamic in the future to adjust when the close button is/isn't showing?

See issue for more information/context.

### Defect Fixes
https://github.com/primefaces/primeng/issues/9441

### Feature Requests

### DIFF
--- a/src/app/components/overlaypanel/overlaypanel.css
+++ b/src/app/components/overlaypanel/overlaypanel.css
@@ -36,6 +36,12 @@
 	margin-left: -10px;
 }
 
+.p-overlaypanel-shifted:after, .p-overlaypanel-shifted:before {
+    left: auto;
+    right: 1.25em;
+    margin-left: auto;
+}
+
 .p-overlaypanel-flipped:after, .p-overlaypanel-flipped:before {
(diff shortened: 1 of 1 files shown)


cagataycivici:
Thx!

PR #9472, opened 2020-10-29
Empty Filter Options Template

dancornilov:
This is simple adjustments to increase UX of your dropdown component. Just added new case in template extracting.
Screen Shot 2020-10-29 at 14 39 27
carbon


cagataycivici:
Thank you for the PR however for consistency, this has to be done on all components that support filtering so please monitor;

https://github.com/primefaces/primeng/issues/9532

PR #9577, opened 2020-11-25
modify event bindings to use ownerDocument where possible

akflaherty:
Allows column resize events to work correctly regardless of which document object is hosting the element.


yigitfindikli:

Issue #9482, opened 2020-11-02
Button | Renders empty label when the label is not defined


paulyipkh:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
When no label is assigned to pButton, text `&nbsp;` is rendered.

**Expected behavior**
When no label is assigned to pButton, the text should be empty, not space nor `&nbsp;`

**Minimal reproduction of the problem with instructions**
https://stackblitz.com/edit/github-7rxhpm?file=src/app/app.component.html

https://github.com/primefaces/primeng/blob/master/src/app/components/button/button.ts
e.g. line 45        `labelElement.appendChild(document.createTextNode(this.label||'&nbsp;'));`
the expression     `this.label||'&nbsp;'`    should be replaced with  `this.label === undefined || this.label === null ? '': this.label`

jonnomk:
I totally agree with this.

paulyipkh:
In PrimeNg 9, I could center align the content projected into the `<p-button>Content</p-button>`? Is there anybody know how this can be achieved in PrimeNg10 ?

![image](https://user-images.githubusercontent.com/12554293/97998355-f3a0f500-1e24-11eb-8c26-5eb2d89b073e.png)

rsilva1959:
workaround -> define label= "Edit", but include classes p-button-icon-only no-label where
::ng-deep .no-label>.p-button-label {
  display: none;
}
<button *ngIf="config.actions.edit" label="Edit" pButton pRipple icon="pi pi-pencil" class="p-button-rounded p-button-success p-button-icon-only no-label p-mr-2"  (click)="editItem(item)"></button>
ugly, but works!

jonnomk:
Yeah, you can solve most things by applying custom styles, but you shouldn't really have to.  The problem with a bunch of these components is that people have been using work-arounds to solve other issues too, and as a result the issues have been hanging around for years.

SternBreeze:
any updates?

h1ghland3r:
Any news?

MaKCbIMKo:
I'm sorry, but I don't think it solved.

You closed this issue, but it still inserts `&nbsp` if you have no label.

[...]

PR #9593, opened 2020-11-30
Remove &nbsp; from empty button label

JamesRArroyo:
### Defect Fixes
This fixes the issue where `&nbsp;` was displayed in plain text when implementing a `p-button` with no label. The issue for this PR is #9482

### DIFF
--- a/src/app/components/button/button.ts
+++ b/src/app/components/button/button.ts
@@ -69,7 +69,7 @@ export class ButtonDirective implements AfterViewInit, OnDestroy {
         this._label = val;
         
         if (this.initialized) {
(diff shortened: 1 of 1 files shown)


src/app/components/button/button.ts
```diff
@@ -69,7 +69,7 @@ export class ButtonDirective implements AfterViewInit, OnDestroy {
         this._label = val;
         
         if (this.initialized) {
-            DomHandler.findSingle(this.el.nativeElement, '.p-button-label').textContent = this._label || '&nbsp;';
+            DomHandler.findSingle(this.el.nativeElement, '.p-button-label').textContent = this._label === undefined || this._label === null ? '': this._label;
```
pmo1948, line 72: why not `this._label || ''`?

Issue #9631, opened 2020-12-09
Radio buttons do not get unselected on the same group, with Reactive Forms using formControl instead of formControlName


lisuizhe:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-f2wj15-yqzqu3?file=src/app/app.component.html

**Current behavior**
When using p-radiobutton element with ReactiveFroms, once a radio button gets selected, it never gets unselected. However the form value updates properly.

**Expected behavior**
Only the current value should be selected on screen

**Minimal reproduction of the problem with instructions**
Just create a reactive form, with a form control, and a group of radiobuttons. This is a minimal reproduction of the problem:
https://stackblitz.com/edit/github-f2wj15-yqzqu3?file=src/app/app.component.html

and fix in https://github.com/primefaces/primeng/pull/9265 (with https://github.com/primefaces/primeng/commit/b823983c5d12f1a2ea73255159d0ed85fd4ca06a) only work for formControlName is same with name: https://stackblitz.com/edit/github-f2wj15-vde3mw?file=src%2Fapp%2Fapp.component.html

**What is the motivation / use case for changing the behavior?**
As the original issue is closed without proper fix, I re-report the issue
(related: https://github.com/primefaces/primeng/issues/9440 , https://github.com/primefaces/primeng/issues/9162)
The fix in https://github.com/primefaces/primeng/pull/9265 (with https://github.com/primefaces/primeng/commit/b823983c5d12f1a2ea73255159d0ed85fd4ca06a) is quite limited to the case using formControlName. If formControl is used instead of formControlName, the fix will just not work.

It seems the purpose of https://github.com/primefaces/primeng/commit/b823983c5d12f1a2ea73255159d0ed85fd4ca06a  to add `if (this.formControlName)` condition in `ngOnInit /ngOnDestroy/select` is trying to not using registry for template-driven forms; but in that case should also take `formControl` under consideration.
Probably `if (this.formControlName)` condition in `ngOnInit /ngOnDestroy/select` should change to `if (this.formControlName || this.fromControl)` with `@Input() formControl: AbstractControl;`
.
[...]

PR #9632, opened 2020-12-09
[Issue #9631] fix radiobutton to support model-driven form with formControl

lisuizhe:
###Defect Fixes
Fixed: Radio buttons do not get unselected on the same group, with Reactive Forms using formControl instread of formControlName #9631

### DIFF
--- a/src/app/components/radiobutton/radiobutton.ts
+++ b/src/app/components/radiobutton/radiobutton.ts
@@ -1,6 +1,6 @@
 import {NgModule,Component,Input,Output,ElementRef,EventEmitter,forwardRef,ViewChild,ChangeDetectorRef,ChangeDetectionStrategy, Injectable, Injector, OnInit, OnDestroy} from '@angular/core';
 import {CommonModule} from '@angular/common';
-import {NG_VALUE_ACCESSOR, ControlValueAccessor, NgControl} from '@angular/forms';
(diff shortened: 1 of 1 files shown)


sergeypayu:
I wonder why the https://github.com/primefaces/primeng/commit/b823983c5d12f1a2ea73255159d0ed85fd4ca06a commit was needed. Angular uses internal registry for all radio buttons without separating model-driven from template-driven approaches. I don't see any downsides of always using the registry.

Also it's kinda sad that it took 5 weeks to merge my fix and in the end radio buttons are still broken because of the applied "refactoring".

loxy:
This is not really working for me (using Formly), because `this.formControl` is also `undefined`. It should be something like this:

```ts
ngOnInit() {
    const control = this.injector.get(NgControl);
    if (control) {
        this.control = control;
        this.checkName();
        this.registry.add(this.control, this);
    }
}
...
select(event) {
    if (!this.disabled) {
        this.inputViewChild.nativeElement.checked = true;
        this.checked = true;
        this.onModelChange(this.value);
        if (this.control) {
            this.registry.select(this);
        }
        this.onClick.emit(event);
    }
}
```

yigitfindikli:
Hi @lisuizhe,

 Thank you for your work. But will be added manually.
 
 Best Regards.

lisuizhe:
@yigitfindikli Thanks! Appreciate your effort to fix this issue!

PR #9719, opened 2021-01-06
Fixed #9718 - Expose InputMask OnKeyDown event

### LINKED ISSUE #9718: Expose InputMask OnKeyDown event
```
```

**Current behavior**
The (keydown) event on the input is not exposed.

**Expected behavior**
Expose the event as an Output EventEmitter, so it could be used outside the component.

We are currently writing a bill keying module that users quickly type in field from a paper form into the system for future processing. They need to be able to quickly tab around and possibly skip sections if that area is empty. Exposing the keydown would allow us to listen to the tab and if the input is empty, move to the next area.

davisb10:
###Defect Fixes

###Feature Requests


yigitfindikli:
Hi @davisb10,

Thanks for your contribution.

Best Regards.

PR #9739, opened 2021-01-09
Replace Fixed #9695 - Editor UI bug report 

### LINKED ISSUE #9695: Editor UI bug report
**I'm submitting a ...** 
```
```

**Plunkr Case (Bug Reports)**
The issue is visible in the demo pages https://www.primefaces.org/primeng/showcase/#/editor

**Current behavior**
In the `Editor` component, open the text align dropdown, the items are **not** visible

**Expected behavior**
text align dropdown items to be visible in the editor component

go to the https://www.primefaces.org/primeng/showcase/#/editor and click on the text align dropdown

bug fix

The problem is with too much padding for the span in `.p-editor-container .p-editor-toolbar.ql-snow .ql-picker.ql-expanded .ql-picker-options .ql-picker-item`  

I have only tested on Firefox

marhuero:
I think this is a better fix for issue #9695 . If we keep headerTemplate we need to declare two templates to change the template

###Defect Fixes

###Feature Requests

### DIFF
--- a/src/app/components/editor/editor.ts
+++ b/src/app/components/editor/editor.ts
@@ -17,7 +17,7 @@ export const EDITOR_VALUE_ACCESSOR: any = {
         <div [ngClass]="'p-editor-container'" [class]="styleClass">
             <div class="p-editor-toolbar" *ngIf="toolbar || toolbarTemplate">
                 <ng-content select="p-header"></ng-content>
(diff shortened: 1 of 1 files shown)


yigitfindikli:
Thanks for your PR but this change can be a problem for too many projects.

Issue #9660, opened 2020-12-16
optionValue backward compatibility issue


cagataycivici:
<img width="1946" alt="Screen Shot 2020-12-16 at 11 54 09" src="https://user-images.githubusercontent.com/686247/102326548-9658a180-3f95-11eb-96ff-135da6faa78e.png">

cagataycivici:
Reopen for review

Spaction:
Not fixed as commented on code review.
This prevent any type of binding data with ngModel with the most basic set-up of dropdown. This can be determined through setting NGModel to the non-default input. Additionally, if you change the dropdown value, the ngModel is always set to undefined
https://stackblitz.com/edit/primeng-dropdown-demo-pte6kg?file=src%2Fapp%2Fapp.component.html

PR #9743, opened 2021-01-10
Fixed #9742 - Implemented fix for #9660 prevents finding a value for simple array of options 

### LINKED ISSUE #9742: Implemented fix for #9660 prevents finding a value for simple array of options
```
```

https://stackblitz.com/edit/primeng-dropdown-demo-pte6kg?file=src%2Fapp%2Fapp.component.html

**Current behavior**
When binding an array of simple objects (string[], int[]), the internal value does not get set because of trying to get option.value which does not exist. This is caused by the recent change of looking if OptionLabel is defined and defaulting to option.value instead of option which is how it previously worked.

**Expected behavior**
Binding an array of simple elements should set the binding of ngModel or correctly update the value of the dropdown

Keep existing functionality for simple cases

* **Angular version:** 
11.0

* **PrimeNG version:** 
11.1.0-rc.1

* **Browser:** 
all

Spaction:
Adjusts the dropdown to allow for simple set-up should option.value not exist.

###Defect Fixes
#9742


yigitfindikli:
Hi @Spaction,

Thank you for your contribution.

Regards.

PR #9784, opened 2021-01-18
Docs: Replace p-header and p-footer with pTemplate

Cito:
I noticed that the documentation still uses the depracated `<p-header>` and `<p-footer>` tags. This PR replaces all of thse occurences with the new way of defining sections solely via `pTemplate`.


yigitfindikli:
Hi @Cito,

Thanks for the PR. But added manually.

Regards.

PR #9790, opened 2021-01-19
remove redundant header template for #8905

### LINKED ISSUE #8241: Resizable Columns enhancement with Frozen Columns
**I'm submitting a ...** 
```
```

**Plunkr Case (Bug Reports)**

[https://stackblitz.com/edit/angular-stmx3d](https://stackblitz.com/edit/angular-stmx3d)

**Current behavior**
When having frozenColumns enabled and resizableColumns set to true with the columnResizeMode to expand it wrongly sets the width of the `ui-table-scrollable-header` and the `ui-table-scrollable-body`. When resizing this will create a scrollbar on the page itself and not within the table, but when expanding a column even more, it sets the width to `auto` allowing the scroll to be within the table. Resize the column back to a smaller width and the scrollbar on the table is gone but back on the page.

**Expected behavior**
I expected the scrollbar to be in the table-body and not on the page.

[https://stackblitz.com/edit/angular-stmx3d](https://stackblitz.com/edit/angular-stmx3d)

Open the stackblitz link and make for example the year column first smaller, allowing the scrollbar to be visible on body level, afterwards make the year column wider allowing the scrollbar to be visible on table level. See example images below.

The scrollbar should never be on body level (x-axis), because this does not only scroll the table content, but the entire page content.

* **Angular version:** 
Angular 8.2.10

* **PrimeNG version:**
8.0.4

Tested this in FireFox 69.0, Edge 42.17134.1.0 and Chrome 77.0.3865.120

* **Quick fix**
`
.ui-table-scrollable-header.ui-widget-header,  
.ui-table-scrollable-body {  
  width: auto !important;  
}
` 

First load of application: looks correct.
start

Making the year column smaller, but the content is still bigger than the body element, showing a scrollbar on body level and not on table level.
part 1

Make the year column wider again and it is back to the correct overflow on the table.
part 2

### LINKED ISSUE #8905: Update Editor Custom Toolbar Documentation
**Current behavior**
Current editor / quilljs example in documentation for custom toolbar not working. the classes are old. but primereact has updated quill documentaion

**Expected behavior**
Documentation should provide new editor custom toolbar example

Documentation is old for editor

* **Angular version:** 
9.1.9

* **PrimeNG version:** 
9.1.0

MikeMatusz:
Fixing editor component header template to work as documented.

###Defect Fixes
 #8905

### DIFF
--- a/src/app/components/editor/editor.ts
+++ b/src/app/components/editor/editor.ts
@@ -15,11 +15,11 @@ export const EDITOR_VALUE_ACCESSOR: any = {
     selector: 'p-editor',
     template: `
         <div [ngClass]="'p-editor-container'" [class]="styleClass">
-            <div class="p-editor-toolbar" *ngIf="toolbar || toolbarTemplate">
+            <div class="p-editor-toolbar" *ngIf="toolbar || headerTemplate">
                 <ng-content select="p-header"></ng-content>
                 <ng-container *ngTemplateOutlet="headerTemplate"></ng-container>
             </div>
-            <div class="p-editor-toolbar" *ngIf="!toolbar && !toolbarTemplate">
+            <div class="p-editor-toolbar" *ngIf="!toolbar && !headerTemplate">
                 <span class="ql-formats">
                     <select class="ql-header">
                       <option value="1">Heading</option>
@@ -106,8 +106,6 @@ export class Editor implements AfterViewInit,AfterContentInit,ControlValueAccess
     
     quill: any;
 
-    toolbarTemplate: TemplateRef<any>;
-
     headerTemplate: TemplateRef<any>;
     
     constructor(public el: ElementRef) {}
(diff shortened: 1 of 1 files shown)


yigitfindikli:
Hi @MikeMatusz,

There are for backward compatibility. Thanks for the PR but I need to close this PR.

Regards.

MikeMatusz:
@yigitfindikli Please review this again in light of the discussion on #8905.  The showcase example is clearly broken, so there is a real issue here.  I can update the PR if you also want to support `pTemplate="toolbar"`, but since that also never worked, and based on the linked issues/migrations doesn't appear to ever have been the intent, I don't think keeping it for backwards compatibility is necessary.  If anyone *was* using the workaround I outlined of having both, that will continue to work.  And this PR does keep backwards compatibility with the deprecated `<p-header>` syntax, which is covered by the `toolbar` property, not `toolbarTemplate`.

PR #9801, opened 2021-01-21
Add Aria Label to button

### LINKED ISSUE #9760: Dropdown Component has WCAG2 Violations
```
```

**Current behavior**
Dropdown component has misplaced role attribute which causes failure of wcag2 test https://dequeuniversity.com/rules/axe/4.1/aria-command-name

**Expected behavior**
Dropdown should should pass wcag2 tests

Use Dropdown component and run accessibility tools such as axe

To pass automated accessibility audits

Angular version: 11

PrimeNG version: 11

Browser: Chrome

mbquinonez:
###Defect Fixes
Fixes #9760

### DIFF
--- a/src/app/components/dropdown/dropdown.spec.ts
+++ b/src/app/components/dropdown/dropdown.spec.ts
@@ -525,4 +525,19 @@ describe('Dropdown', () => {
 
 		expect(groupDropdown.selectedOption.label).toEqual("Mercedes");
 	});
+
+	it('should pass wcag2 specifications', () => {
+		dropdown.options = [
+			{label:'Select City', value:null},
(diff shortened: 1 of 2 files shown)


patricio-ezequiel-hondagneu-roig (approved):

mir4ef:
Can the `aria-label` be an `@Input` with a default value rather than hardcoding it? This will allow a custom `aria-label` value to be specified which will be especially important if we need to support languages other than English.

yigitfindikli:
Upcoming versions focus on accessibility where each component will get new sections for keyboard and screen reader support along with the improvements. We'll be checking with accessibility tools for 100% compliancy. And aria-label should be @Input for the users. Thanks for you're effort and time but I'm gonna close this PR.

PR #9808, opened 2021-01-22
fixed #9807  - Splitter property for initial size should be called panelSize on the showcase

### LINKED ISSUE #9807: Splitter property for initial size should be called panelSize on the showcase
```
```

**Plunkr Case (Bug Reports)**

**Current behavior**
The showcase calls the property to set the initial size of the panels in the splitter component `size` at some places, `panelSize` at others. Looking at the source code, the correct name should be `panelSize`.
When using `size`, angular complains, since it isn't a known property of `p-splitter`.

**Expected behavior**
Correct documentation/showcase and no error.

* visit https://primefaces.org/primeng/showcase/#/splitter
* checkout the example of Initial Sizes or the table of properties.

Correct the documentation. Help people not have angular errors when using the wrong property.

* **Angular version:** 11.0.6

* **PrimeNG version:** 11.1.0

* **Browser:** all

 
* **Language:** 

* **Node (for AoT issues):** `node --version` =   15.2.1

daniel-lang:
Showcase refers to the property for the initial size as 'sizes', which does not exist. Should be 'panelSize'.

### Defect Fixes

### Feature Requests

Fixes #9807

### DIFF
--- a/src/app/showcase/components/splitter/splitterdemo.html
+++ b/src/app/showcase/components/splitter/splitterdemo.html
@@ -108,9 +108,9 @@ import &#123;SplitterModule&#125; from 'primeng/splitter';
 </app-code>
 
             <h5>Initial Sizes</h5>
-            <p>When no sizes are defined, panels are split 50/50, use the <i>sizes</i> property to give relative widths e.g. [20,80].</p>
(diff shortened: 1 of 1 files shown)


yigitfindikli:
Hi @daniel-lang,
Thanks for your contribution.

PR #9828, opened 2021-01-27
Optimze label calculating #9829

billyan2018:
Only calcuate label when `this.value.length <= this.maxSelectedLabels`

### Defect Fixes
When submitting a PR, please also create an issue documenting the error: https://github.com/primefaces/primeng/issues/9829

### DIFF
--- a/src/app/components/multiselect/multiselect.ts
+++ b/src/app/components/multiselect/multiselect.ts
@@ -836,8 +836,9 @@ export class MultiSelect implements OnInit,AfterViewInit,AfterContentInit,AfterV
         if (this.value && this.options && this.value.length && this.displaySelectedLabel) {
             if (this.value.length <= this.maxSelectedLabels) {
                 let label = '';
-                for (let i = 0; i < this.value.length; i++) {
-                    let itemLabel = this.findLabelByValue(this.value[i]);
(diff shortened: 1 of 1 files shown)


src/app/components/multiselect/multiselect.ts
```diff
@@ -834,18 +834,17 @@ export class MultiSelect implements OnInit,AfterViewInit,AfterContentInit,AfterV
...
-            let label = '';
-            for (let i = 0; i < this.value.length; i++) {
-                let itemLabel = this.findLabelByValue(this.value[i]);
-                if (itemLabel) {
-                    if (label.length > 0) {
-                        label = label + ', ';
+            if (this.value.length <= this.maxSelectedLabels) {
+                let label = '';
+                for (let i = 0; i < this.value.length; i++) {
+                    let itemLabel = this.findLabelByValue(this.value[i]);
```
GavinWu1991, line 840: @billyan2018
Thanks for your changes, I think we can determine the range of labels to use at the time of splicing the strings, which will ensure higher efficiency. The range should be the smallest of `maxSelectedLabels` and `value.length`.
```suggestion
const labelScope = Math.min(this.value.length, this.maxSelectedLabels);
for (let i = 0; i < labelScope; i++) {
      let itemLabel = this.findLabelByValue(this.value[i]);
```

PR #9830, opened 2021-01-27
Add support for aria-describedby attribute in AutoComplete component.

patricio-ezequiel-hondagneu-roig:
Added support for the `aria-describedby` attribute in the AutoComplete component, to comply with WCAG accessibility standards.

There's no straightforward way to access the inner `input` element within the component and this addition behaves exactly like the existing binding for the `aria-labelledby` attribute.


Agusmazzeo:
This would be really useful. Good approach!

PR #9839, opened 2021-01-30
Adapt toast position after its width was increased

Cito:
When you go to the [Toast component demo page](https://primefaces.org/primeng/showcase/#/toast) and show a top center or bottom center toast message, you will notice that it is slightly off the center to the right. Make the browser window narrower if you don't see it.

The reason for this is that in 11e1ad7441473dbaaa2350ce0f98db62a8923791 the width of the toast message was increased from 20em to 25rem, but the classes for centering were not adapted accordingly.

### DIFF
--- a/src/app/components/toast/toast.css
+++ b/src/app/components/toast/toast.css
@@ -39,13 +39,13 @@
 .p-toast-top-center {
 	top: 20px;
     left: 50%;
-    margin-left: -10em;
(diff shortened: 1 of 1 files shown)


Cito:
On the new [toast demo page](https://primeng.org/toast) I can't reproduce the problem any more. Seems it was fixed in #10333 in June 2021.

So this PR can be closed.

Issue #8033, opened 2019-08-07
Autocomplete has wrong width when using appendTo body


plevko:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/angular-euhmsr

**Current behavior**
In case of long strings in suggestions, the autocomplete has wrong width and is out of expected bounds. This happens when using option appendTo="body".

**Expected behavior**
Autocomplete has maxWidth the same as minWidth

**Minimal reproduction of the problem with instructions**

* **Angular version:** 8.X

* **PrimeNG version:** 8.X

* **Browser:** [all]

 
* **Language:** [all]

JacobSiegle:
I'm not sure I follow what is incorrect here. Can you possibly post a screenshot of how you're expecting it to work?

plevko:
![image](https://user-images.githubusercontent.com/6647984/62886897-c14d3880-bd44-11e9-8bf3-f27c19814d25.png)

I am expecting the autocomplete dropdown to have max width the same as input (the green lines). Current behaviour you can see highlighted in red.

JacobSiegle:
If the word is too long it's going to go over the length of the popup no matter what, you can limit this by setting a width of 100% on .ui-autocomplete .ui-autocomplete-panel and it will put a scroll bar on the bottom of the popup. 

When you use appendTo you're moving the popup HTML element outside of the autocomplete element which causes it to lose context of what the width of the input element is. The only way to sync up the width of this currently would be to set the width manually.

I'm not sure a maintainer would change the way appendTo works to allow syncing the width state, I assume the way it's working now is the desired effect but i'm not entirely sure.

cagataycivici:
Unable to replicate with PrimeNG 9, if the issue persists please create a new ticket with a test case reproducing the issue e.g. stackblitz or a github repo and it will be reviewed by our team once again.

PR #9875, opened 2021-02-09
feat(autocomplete): add the prop overlayFitWidth

CyriacBr:
### Defect Fixes
When using `appendTo`, it becomes impossible to style the width of the overlay panel according to the input field. The panel can thus grow bigger than required, which can be a problem. I added an @input `overlayFitWidth` to perfectly fit the width of the overlay according to the input field when `appendTo` is used.

Related: https://github.com/primefaces/primeng/issues/8033

### DIFF
--- a/src/app/components/autocomplete/autocomplete.ts
+++ b/src/app/components/autocomplete/autocomplete.ts
@@ -118,6 +118,8 @@ export class AutoComplete implements AfterViewChecked,AfterContentInit,OnDestroy
 
     @Input() size: number;
 
+    @Input() overlayFitWidth: boolean;
+
     @Input() appendTo: any;
 
     @Input() autoHighlight: boolean;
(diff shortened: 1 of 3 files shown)


src/app/components/autocomplete/autocomplete.ts
```diff
@@ -496,6 +498,9 @@ export class AutoComplete implements AfterViewChecked,AfterContentInit,OnDestroy
 
             if (!this.overlay.style.minWidth) {
                 this.overlay.style.minWidth = DomHandler.getWidth(this.el.nativeElement.children[0]) + 'px';
+                if (this.overlayFitWidth) {
+                    this.overlay.style.maxWidth = this.overlay.style.minWidth;
```
CyriacBr, line 502: I wasn't really sure if I should have used simply `width` instead of a combination of both `minWidth` and `maxWidth`. But due to CSS weirdness sometimes, I opted for the latter.

src/assets/components/themes/arya-blue/theme.css
```diff
@@ -116,6 +116,7 @@ p-autocomplete.ng-dirty.ng-invalid > .p-autocomplete > .p-inputtext {
   background: transparent;
   transition: box-shadow 0.2s;
   border-radius: 0;
+  white-space: break-spaces;
```
CyriacBr, line 119: Also wasn't sure if this should be included in the base theme.

yigitfindikli:
Unable to replicate with PrimeNG 11 if the issue persists please create a new ticket with a test case reproducing the issue e.g. stackblitz or a github repo and it will be reviewed by our team once again.

PR #9881, opened 2021-02-10
Avoid an unsubscription error in the Steps component

Cito:
If the Steps component is created, but not initialized, then the unsubscription in the ngDestroy method fails, because the subscription property is not set.

This happens for instance in the default tests created by angular-cli which just test if a component can be created. If the component uses Steps, then you get an error in the console log and start to wonder what's happening here.

I think it's good practice to initialize the subscription property, see for instance the [code of the stepper component](https://github.com/angular/components/blob/master/src/material/stepper/stepper.ts) of Angular Material.


yigitfindikli:
Hi @Cito,

Thanks for the pull request but will be fixed manually.

Best Regards.

PR #9898, opened 2021-02-15
Fixed #9895 - Avatar-Images are overflowing circle-shaped parent-Div

### LINKED ISSUE #9895: Avatar Circle not working
```
```

Croping an Avatar-Image to the shape of an circle is not working.
p-avatar-circle, the css-class with border-radius 50% is only applied to the parent div-container.

If the containing image is in square format, the border-radius of the parent is ignored.

The demo is looking as if it is working because the used images are already in the shape of a circle

ProfEibe:
Fixes: https://github.com/primefaces/primeng/issues/9895

### DIFF
--- a/src/app/components/avatar/avatar.css
+++ b/src/app/components/avatar/avatar.css
@@ -13,13 +13,14 @@
 
 .p-avatar.p-avatar-circle {
     border-radius: 50%;
+    overflow: hidden;
 }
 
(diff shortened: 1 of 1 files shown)


yigitfindikli:
Hi @ProfEibe,

Thanks for your contribution.

Regards.

PR #9920, opened 2021-02-20
Allow hiding of sort icon for unsorted columns

SPKorhonen:
To reduce visual clutter it sometimes makes sense to hide the "sortable" icon from table columns headers. This PR allows such functionality. The defaults settings preserve previous behaviour.

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -2649,7 +2649,7 @@ export class SortableColumn implements OnInit, OnDestroy {
 @Component({
     selector: 'p-sortIcon',
     template: `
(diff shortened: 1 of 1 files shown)


src/app/components/table/table.ts
```diff
@@ -2649,19 +2649,24 @@ export class SortableColumn implements OnInit, OnDestroy {
 @Component({
     selector: 'p-sortIcon',
     template: `
-        <i class="p-sortable-column-icon pi pi-fw" [ngClass]="{'pi-sort-amount-up-alt': sortOrder === 1, 'pi-sort-amount-down': sortOrder === -1, 'pi-sort-alt': sortOrder === 0}"></i>
+        <i *ngIf="visible" class="p-sortable-column-icon pi pi-fw" [ngClass]="{'pi-sort-amount-up-alt': sortOrder === 1, 'pi-sort-amount-down': sortOrder === -1, 'pi-sort-alt': sortOrder === 0}"></i>
```
briancodes, line 2652: Would it be more performance to bind to a field rather than using a getter/method call for `visible`?

PR #9945, opened 2021-02-25
Fixed #9944 - Add style p-disabled to disabled selectable row.

### LINKED ISSUE #9944: SelectableRow when disabled provides no way to target styling for disabled row.
I'm submitting a feature request.
```
```

**Current behavior**
When the p-table component is used with the `[pSelectableRow]` directive there is a property to set the row to disabled. Currently there are no styles or indicators applied that the table row is disabled for selection.

**Expected behavior**
 In this case it would be helpful for external component users to be able to take action, style wise, on the row by using a new or existing css class applied based on the `isEnabled()` flag within the `SelectableRow` directive.

We've used `p-disabled-row` for our tweak to the `SelectableRow` directive. Just a suggestion. Naming is hard. :)

```
    host: {
        '[class.p-selectable-row]': 'isEnabled()',
        '[class.p-highlight]': 'selected',
+        '[class.p-disabled-row]': '!isEnabled()',
        '[attr.tabindex]': 'isEnabled() ? 0 : undefined'
    }
```

`<tr [pSelectableRow]="rowData" [pSelectableRowDisabled]="true">`

No way to easily apply a disabled styling to the contents of the p-table row when it is disabled for selection. 

* **Angular version:** 5.X
9.x

* **PrimeNG version:** 5.X
10.0.3

* **Browser:** 
all
 
TS / ES6

N/A

Coderrob:
### Feature Requests
https://github.com/primefaces/primeng/issues/9944

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -2696,9 +2696,10 @@ export class SortIcon implements OnInit, OnDestroy {
 
 @Directive({
     selector: '[pSelectableRow]',
-    host: {
+    host: {        
         '[class.p-selectable-row]': 'isEnabled()',
         '[class.p-highlight]': 'selected',
+        '[class.p-disabled]': '!isEnabled()',
         '[attr.tabindex]': 'isEnabled() ? 0 : undefined'
     }
 })
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -2696,7 +2696,7 @@ export class SortIcon implements OnInit, OnDestroy {
 
 @Directive({
     selector: '[pSelectableRow]',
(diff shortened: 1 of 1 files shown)


yigitfindikli:
Hi @Coderrob,

Thanks for your work. But p-disabled there are for for input components. But we're gonna discuss for 11.3.0-final this issue.

Regards.

Coderrob:
@yigitfindikli no worries. In our implementation I created a new style called `p-disabled-row` to not overlap with existing styles available. Tried to keep it in similar naming as `p-selectable-row`. 

Wanted to throw up an example implementation that went along with the issue I opened.

Thank you for considering the issue!

Issue #9924, opened 2021-02-23
Possibility to hide (unused) badge-directive


ProfEibe:
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

Using Badge as component I can simply hide it with a *ngIf or css display-none.
It would be great to have a feature to hide the badge used as directive it is isn't needed, for example the value is zero or empty.

A similar function would be the "tooltipDisabled" option from p-tooltip. 

Maybe like this?
```html
<i class="pi pi-bell" pBadge [value]="user.notifications" [badgeDisabled]="user.notifications == 0"></i>
```

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

PR #9959, opened 2021-03-02
Added Feature badgeDisabled

ProfEibe:
Added Feature for issue #9924


ProfEibe:
Okay, you have done it yourself, right? Would have been nice to know.

ProfEibe:
Your own implementation is not bugfree. When will it be solved?

PR #10036, opened 2021-03-24
Fixed dragdrop for picklist (using cdkDropListDisabled)

ramonvanbezouw:
Docs specify the `dragdrop` input for enabling / disabling drag and drop for the Picklist component:

dragdrop | boolean | false | Whether to enable dragdrop based reordering.
-- | -- | -- | --

This feature does not work.

**Expected behavior:**
Setting `[dragdrop]="false"` on `<p-picklist>` should disable drop behaviour for the items.

**Actual behavior:**
Drag and drop of items in `<p-picklist>` always works, regardless of the setting.

In the showcase change ` [dragdrop]="true"` to ` [dragdrop]="false"`, drag will still work.

**Proposed fix:**
The dragdrop Input is not being used by the component. Adding `[cdkDropListDisabled]="!dragdrop"` to the source and target `cdkDropList`, the dragdrop input works as expected.

Fix was tested in Windows Chromium 89, Windows Firefox 87, iOS Safari 14.

### DIFF
--- a/src/app/components/picklist/picklist.ts
+++ b/src/app/components/picklist/picklist.ts
@@ -27,8 +27,8 @@ import {UniqueComponentId} from 'primeng/utils';
                         <span class="p-picklist-filter-icon pi pi-search"></span>
                     </div>
                 </div>
-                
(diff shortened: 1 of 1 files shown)


yigitfindikli:
Hi @ramonvanbezouw,

Thanks for your contribution. But there are unnecessary changes in createStyle function. Will be fixed in 11.3.2.

Best Regards.

PR #10102, opened 2021-04-12
Fixed #10101 - TriStateCheckbox null state has false state icon 

### LINKED ISSUE #10101: TriStateCheckbox null state has false state icon
Since this [pull request](https://github.com/primefaces/primeng/pull/10072) the false icon is also used for the null state

luehmann:
Fixes #10101 
Now the null state should be empty again as expected


yigitfindikli:
Hi @luehmann,

Thanks for your contribution.

Issue #10125, opened 2021-04-16
Carousel swipe/drag support on desktop


vdbongard:
**I'm submitting a ...**
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**StackBlitz**
https://stackblitz.com/edit/feature-primeng-carousel-mouse-drag

**Description**
I have noticed that the Carousel component supports swiping through the slides on mobile but it does not work on desktop. It would be great if it would also work when dragging the slides via the mouse on desktop.

Another great improvement to this would be if the slide would stick to the pointer when holding down and moving. That would be awesome.

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

FlorianLudwig:
@mertsincan this feature was not implemented. Tested https://www.primefaces.org/primeng/carousel

On mobile swip works, on desktop unable to drag.

mehmetcetin01140:
Hi, 

Issue tracker is used for defects only as part of our commitment to quality and continuous improvement in all areas. Enhancements are collected as valuable community feedback and managed internally so moving this enhancement ticket to our internal project management backlog. For enhancement requests and other topics, you can open a new topic in [PrimeNG Discussions](https://github.com/orgs/primefaces/discussions/categories/primeng).

DAVIAMERICO242:
Still not implemented

PR #10141, opened 2021-04-20
Added pointer support to carousel 

vdbongard:
This enables dragging/swiping the carousel slides with more pointer devices than touch e.g. with mouse.

Issue #10125

(the formatting changes were done by my ide, i can remove them if wished)


FlorianLudwig:
@yigitfindikli Hi, what is the reasoning to not accept this MR?

Currently the carousel is broken on desktop this fixes the bug. Was looking forward to having this merged.

I don't see how the MR makes code more complex or any other reason why it would not be accepted. So I would be glad if you could provide some insight how it is determined if a mere request is considered or not.

Otherwise this is discouraging us to spend time on creating MRs / improving PrimeNG.

vdbongard:
I would also be interested in a more detailed reasoning behind closing this PR. @FlorianLudwig has some good points.

PR #10150, opened 2021-04-22
Added tool-tip for each Star in rating-component

NickSto139:
###Feature Requests
Allows to set a tool-tip and position for each "Star" in the rating-component
#10149


src/app/components/rating/rating.ts
```diff
@@ -31,6 +32,10 @@ export class Rating implements OnInit,ControlValueAccessor {
 
     @Input() stars: number = 5;
 
+    @Input() starTooltip: any;
```
misza1597, line 35: 1. you're assuming that everyone who want to use this component also want to provide tooltip for each star. Now if someone would not provide this property TypeError will be thrown.
2.  Name startToolitp in my opinion doesn't fit here.
3.  Type any doesn't sounds good.
4.  If you wan't to use toolip directive you must also add TooltipModule to imports array of RatingModule.
5. Where are test's ?

PR #10181, opened 2021-04-30
Change removeIconClass to removeIcon. 

seemme:
Change removeIconClass to removeIcon, because the showcase page does not fit with the actual implementation.


yigitfindikli:
Hi @seemme,

Thanks for your contrubution.

Regards.

PR #10300, opened 2021-06-03
Remove the redundant tag from the showcase menu

alwye:
The tag displayed in the menu was apparently added to indicate introduction of the new component.
However it displays 'Tag' instead of 'New' and is rather old component by now having been added 6 months ago.
It is apparently an error.


cetincakiroglu:
Hi @alwye,

It's by design. Closing the PR.

PR #10337, opened 2021-06-14
Selected filters from table returns an array

Subfly:
Filter applied to table returns a FilterMetadata array and needed to be declared in the interface in order to proceed.


himinat:
This pull request should be merged ASAP as this one fixes bug that lasts already for 2 years!

PR #10345, opened 2021-06-18
Added disabled property to OrderList

lokarithm:
Currently, there is no easy way to disable the OrderList component. I added a _disabled_ property in the OrderList component, which lets users disable the buttons and input. This implementation is exactly the same as the _disabled_ property in the PickList component.


danailvidev:
nice, thanks

PR #10403, opened 2021-07-10
fix: dont update `p-inputnumber` cursor when value is 0

Aaronzinhoo:
### What

This is a fix to the `p-inputnumber` component. The bug caused by the component is due to the fact that when an `p-inputnumber` component has a suffix & its value is 0. 

Example:
image

After typing some numbers in, the cursor goes behind the suffix like so and does not allow the user to backspace to delete any number any more.
image

### Why

Within the `updateInput` function, the case for inserting a value when `newLength===oldLength` & the current operation is `insert`, always moves the cursor position up 1 regardless of the value. In the case of zero this should not occur (the value zero should simple be replaced).

### Fix

Add a simple `if` to check if the value is zero, otherwise proceed as usual.

### DIFF
--- a/src/app/components/inputnumber/inputnumber.ts
+++ b/src/app/components/inputnumber/inputnumber.ts
@@ -910,7 +910,13 @@ export class InputNumber implements OnInit,ControlValueAccessor {
             }
             else if (newLength === currentLength) {
                 if (operation === 'insert' || operation === 'delete-back-single')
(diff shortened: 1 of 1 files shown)


Aaronzinhoo:
@onursenture @yigitfindikli  anyway we can get this bug fix merged in? It really affects using the `p-inputnumber` component

Aaronzinhoo:
This bug still exists I think

Issue #10246, opened 2021-05-17
Table | global filter doesn't reset on Table.clear()


Timmeeeey:
**I'm submitting a ...** 
```
[x] bug report
[ ] feature request
[ ] support request
```

**Current behavior**
The global filter doesn't reset when the clear function is called.

**Minimal reproduction of the problem with instructions**
This issue can be reproduced in the showcase:
https://primefaces.org/primeng/showcase/#/table/filter

Just enter a global filter and click the clear button.

![image](https://user-images.githubusercontent.com/46870283/118452900-e4193780-b6f6-11eb-9f27-3d829b53f4bf.png)

* **Angular version:** 11

* **PrimeNG version:** 11.4.2

ThoSap:
Also normal column filters are not cleared.
Caused by https://github.com/primefaces/primeng/issues/9570

nishant203:
Hi @yigitfindikli 
PR #10578 fixes this.

PR #10578, opened 2021-09-04
Fixes -  Table: global filter doesn't reset on Table.clear() #10246

nishant203:
Fixes - Table: global filter doesn't reset on Table.clear() #10246
Set default value of global filter in table.


yigitfindikli:
Hi @nishant203,

Will be added manually. Thanks for your effort and report.

sutoadam:
Still not working with angular 14 & primeng 14.1.1

Issue #10621, opened 2021-09-15
Dropdown filter has problem filtering when uses space bar.


nemeek2017:
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
In a dropdown, for version 12 of primeng, if you use the space bar in the filter, the dropdown will be closed.
For version 11, if you use the space bar in the filter, the dropdown will be shown "No results found"

**Expected behavior**
The dropdown will be shown "No results found"

**Minimal reproduction of the problem with instructions**
You can use the demo on the primeng documentation for the dropdown.
https://www.primefaces.org/primeng/showcase/#/dropdown
https://www.primefaces.org/primeng/v11/#/dropdown

**What is the motivation / use case for changing the behavior?**
We are trying to filter the data in the dropdown and the dropdown has several results with space in the words.

**Please tell us about your environment:**
VS Code, NPM, webpack, Angular 12, Azure devops..

* **Angular version:** 5.X
12.2.4

* **PrimeNG version:** 5.X
12.1.0

* **Browser:** all
 
* **Language:** all

* **Node (for AoT issues):** `node --version` =  v14.15.5

francowhyline:
Hi, news about this??? It is impossible to filter anything inside a dropdown if you can't have a space in the label.

mikela:
@francowhyline As far as I can tell it's fixed on the free (latest) version: https://www.primefaces.org/primeng/showcase/#/dropdown 
@nemeek2017 I think you can close this issue (I closed my PR since they fixed it internally).

PR #10626, opened 2021-09-16
fix(Dropdown filter has problem filtering when uses space bar)

mikela:
###Defect Fixes
Fixes https://github.com/primefaces/primeng/issues/10621 Dropdown filter has problem filtering when uses space bar


mikela:
Before: (dropdown closes when pressing the spacebar).
After: 
image

mikela:
Can't reproduce on version 13.0.1.

Issue #10729, opened 2021-10-08
Table Component Does Not Support Sorting of Immutable Arrays


johnathanradojevich:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Could not think of a test case to easily demonstrate this, but the linked demo shows issue quite clearly. https://stackblitz.com/edit/github-ak29zj?file=src%2Fapp%2Fapp.component.ts

**Current behavior**
Sorting a p-table unexpectedly tries to modify the underlying data array. In applications using immutable (frozen) arrays, this causes an error to be thrown, and sorting functionality to fail.

**Expected behavior**
The array that is passed in should not be modified.

**Minimal reproduction of the problem with instructions**
Set the value of a p-table to a frozen array, define some sort column, run everything and click the sort header. An exception will be thrown in the console, and the table will not sort.

**What is the motivation / use case for changing the behavior?**
Modifying the array that is passed in is unexpected behavior. Additionally, causes a runtime exception when using frozen arrays.

Issue was previously brought up multiple times, and fixed to provide the expected behavior. Why it now exists in current version is unknown.

https://github.com/primefaces/primeng/issues/2630
https://github.com/primefaces/primeng/issues/8488
https://github.com/primefaces/primeng/issues/2529

* **Angular version:** 12.1.0

* **PrimeNG version:** 10.0.3 - Issue still exists in most recent PrimeNG version as evidenced by linked stackblitz demo.

rueeggl:
is there any update to fixing this bug?

PR #10730, opened 2021-10-08
Table now creates its own copy of the array to mutate.

johnathanradojevich:
Fixes issue #10729.


rantiev:
Yep, it's relevant. Your datatable still doesn't work with immutable arrays.

PR #10899, opened 2021-11-30
Fix documentation bug

tomer953:
fix documentation bug

a small documentation bug in the avatar section.


yigitfindikli:
Hi @tomer953,

Thanks for your contrubution.

Issue #9320, opened 2020-09-26
p-tree onNodeDrop event dropIndex incorrect when dragging after final node   


LukeHartcher:
```
[ x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/primeng-treedragdrop-demo-wnssd9
(Fork of https://stackblitz.com/edit/primeng-treedragdrop-demo)

**Current behavior**
Drag "Documents" to AFTER "Movies"
Console: dropIndex = 2

**Expected behavior**
Drag "Documents" to AFTER "Movies"
Console: dropIndex = 3

**Minimal reproduction of the problem with instructions**
Run StackBLitz Demo
Drag "Documents" to between "Pictures" and "Movies"
Console: dropIndex = 2 (correct)
Restart StackBLitz Demo
Drag "Documents" to AFTER "Movies"
Console: dropIndex = 2 (INCORRECT)

**What is the motivation / use case for changing the behavior?**
Currently the dropIndex=2 when dropped either before or after the final item, so you are unable to determine the exact drop without investigation the new tree. 

**Please tell us about your environment:**
Your current stackblitz demo https://stackblitz.com/edit/primeng-treedragdrop-demo PrimeNG 10.0.0

* **Browser:** [all ]

 
* **Language:** [all]

* **Node (for AoT issues):** `node --version` =

hkjeffchan:
Please merge this ASAP. The inability to determine the exact drop position makes the tree drag and drop function basically unusable.

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

PR #11009, opened 2021-12-27
add possibility to distinguish tree drop targets

### LINKED ISSUE #11008: [p-tree] add possibility to distinguish drop target
```
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-rsyuaf

**Current behavior**

There is no possibility to distinguish whether a droppoint or node was the target of a drop. 

**Expected behavior**

It should be possible to distinguish if a drop occured before, on or after a tree node.

https://stackblitz.com/edit/github-rsyuaf
1. Drag node1 before node2 (droppoint)
2. Drag node1 on node2 (node)

The console will show index 1 in both cases.

In our case, we want to perform different API-requests for tree items depending on whether the drop occurred on or before/after a treenode.
If a node was dropped on another node, the API-Request would use the drop node's id as the parent id for the dragged node.
If a node was dropped before/after another node, the API-Requests would not change the parent id but the order instead.

* **Angular version:** 13.1.1

* **PrimeNG version:** 13.0.3

* **Browser:** all

 
* **Language:** all

ewokhias:
Fixes #11008 (and possibly #9320)

### DIFF
--- a/src/app/components/tree/tree.ts
+++ b/src/app/components/tree/tree.ts
@@ -210,6 +210,7 @@ export class UITreeNode implements OnInit {
                     dragNode: dragNode,
                     dropNode: this.node,
                     index: this.index,
+                    dropPointPosition: position,
                     accept: () => {
                         this.processPointDrop(dropParams);
                     }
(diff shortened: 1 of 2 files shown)


bence-balogh-intland:
Hi @ewokhias,
I'm sorry to see this PR has being closed.
This feature would be essential for us in order to determine the correct position of dropped node.
Can we expect re-open?
Also it would be really-really nice for us, if this could go out in a v14 release (if that's possible).

PR #11023, opened 2022-01-03
Fixed #11021 | Table ColResize Expand Mode not Compatible with Table …

### LINKED ISSUE #11021: Table | ColResize Expand Mode not Compatible with Table Scroll
Reported by PRO user.

cetincakiroglu:


dbelob:
@cetincakiroglu @yigitfindikli Maybe should be added `:not(.p-frozen-column)` in fix?
(not `:not(.p-forzen-column)`, misspelling in CSS class name)

Issue #4986, opened 2018-01-26
Accordion children elements receive keyboard focus even when the panel is collapsed


skhattry:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
The children of a collapsed panel receive keyboard focus.

**Expected behavior**
Items present inside a collapsed accordion panel should not receive keyboard focus.

**Please tell us about your environment:**

* **Angular version:** 2.0.X
4.1.3

* **PrimeNG version:** 2.0.X
4.2.1

* **Browser:** 
Tested on Microsoft Edge and Mozilla Firefox

peaverok:
As of 10.0.03 this is still not fixed. As a federal employee this is crucial for us because of accessibility reasons. This is not how accordions are supposed to function:

[https://www.w3.org/TR/2019/WD-wai-aria-practices-1.2-20191218/#accordion](https://www.w3.org/TR/2019/WD-wai-aria-practices-1.2-20191218/#accordion)

[https://www.w3.org/TR/2019/WD-wai-aria-practices-1.2-20191218/examples/accordion/accordion.html](https://www.w3.org/TR/2019/WD-wai-aria-practices-1.2-20191218/examples/accordion/accordion.html)

Use the example link above and you will see that it does not tab to items inside collapsed accordions.

PR #11033, opened 2022-01-05
Fix defect #4986 - Set visibility: hidden on collapsed accordion tab content

shaizel:
###Defect Fixes
Fixes defect #4986 - Setting `visibility: hidden` in the `hidden` state of the collapse animation of the accordion to prevent tab stop on the elements in the content


cetincakiroglu:
Hi @shaizel,

This issue does not exist anymore with the PrimeNG v15. 
Closing the PR.

PR #11058, opened 2022-01-12
Fixed #11036 - MultiSelect | Group ignores filterBy option

### LINKED ISSUE #11036: multi select with group ignores filterBy option
```
```

**Current behavior**

When using multiselect with groups and trying to apply a filterby, the option is ignored
**Expected behavior**

filterBy should not be ignored

It should function, or documentation should mention that it is not supported

* **Angular version:** 13.X

* **PrimeNG version:**13.X

* **Browser:** all

looking at the code : https://github.com/primefaces/primeng/blob/master/src/app/components/multiselect/multiselect.ts#L1045-L1049 it seems like line 1049 should not actually be there.

cetincakiroglu:


yigitfindikli:
Merged in https://github.com/primefaces/primeng/pull/11060

Issue #10792, opened 2021-10-29
TypeError: Cannot read properties of undefined (reading 'nativeElement') with scrollable and state enabled table


rdrrichards:
**I'm submitting a ...**
```
[X] bug report
```

**Plunkr Case (Bug Reports)**

I was unable to reproduce this issue with StackBlitz. I have code below to reproduce the issue in Chrome.

**Current behavior**

I added a table to my application that has sortable columns, is scrollable *and* persists state to `localStorage`. After a sort is set in state by clicking one of the table column headers, I start to see the error below when I refresh the browser to reload the application, or the application reloads after a rebuild when using `ng serve`.
```
core.js:6479 ERROR TypeError: Cannot read properties of undefined (reading 'nativeElement')
    at Table.scrollTo (primeng-table.js:1189)
    at Table.resetScrollTop (primeng-table.js:1153)
    at Table._filter (primeng-table.js:1028)
    at Table.sortSingle (primeng-table.js:508)
    at Table.ngOnChanges (primeng-table.js:258)
    at Table.rememberChangeHistoryAndInvokeOnChangesHook (core.js:1498)
    at callHook (core.js:2536)
    at callHooks (core.js:2495)
    at executeInitAndCheckHooks (core.js:2446)
    at refreshView (core.js:9480)
```

If I remove the `scrollable` and `scrollHeight` attributes, *no error is thrown*. If I remove the `stateKey` and `stateStorage` attributes, *no error is thrown*. If I have all 4 of these attributes on the table, *the error is thrown*. If I clear the state from `localStorage`, the error stops getting thrown until I sort the table again.

**Expected behavior**

I would not get any console error when loading the application. These attributes should work together without throwing an error.

**Minimal reproduction of the problem with instructions**

Create a new angular application using the CLI, install primeng, primeicons, and primeflex.

Create a new component named case-list, code below.

*case-list.component.ts*

```
import { Component, OnInit } from '@angular/core';

interface PendingCase {
    caseNo: string;
}

@Component({
    selector: 'app-case-list',
    templateUrl: './case-list.component.html'
})
export class CaseListComponent implements OnInit {
    cols: any[] = [];
    selectedCases: PendingCase[] = [];
    pendingCasesList: PendingCase[] = [];
    ngOnInit(): void {
        this.cols = [
        { field: 'caseNo', header: 'Case #', width: '0 0 12%' },
        { field: 'code', header: 'Code', filterType: 'custom', width: '0 0 10%' },
[...]

PR #11064, opened 2022-01-13
:bug: 10792 Resolve console console errors on table load

waratah:
###Defect Fixes
https://github.com/primefaces/primeng/issues/10792


waratah:
Note that the table reference could be replaced by changing a flag,  so I have changed that to a non static reference.

yigitfindikli:
Hi @waratah,

Thanks a lot for your contribution.

Regards.

Issue #11143, opened 2022-02-07
p-image does not allow src to be of type SafeUrl


ramonvanbezouw:
**I'm submitting a ...**
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
```html
<p-image [src]="sanitizer.bypassSecurityTrustUrl(base64EncodedImage)"></p-image>
```
Throws a type error `Type 'SafeUrl' is not assignable to type 'string'` due to the src Input Decorator only accepting values of type `string`.

**What is the motivation / use case for changing the behavior?**
Supporting the `SafeUrl` type would make the `p-image` component compatible with the underlying Angular `src` directive on the image element, which does support the `SafeUrl` type.

* **Angular version:** 13.1

* **PrimeNG version:** 13.1

* **Browser:** all
 
* **Language:** all

ramonvanbezouw:
PR here: #11144

PR #11144, opened 2022-02-07
fix: p-image does not allow src to be of type SafeUrl

ramonvanbezouw:
Fix for #11143


yigitfindikli:
Hi @ramonvanbezouw,

Thanks for your contribution.

PR #11208, opened 2022-02-16
Fixed #11191 - MultiSelect | overlay bug baseZIndex property is unused

### LINKED ISSUE #11191: Multiselect overlay bug - baseZIndex property is unused
```
```

**Current behavior**
It's basically the same Bug as [Dropdown overlay bug - baseZIndex property is unused in PrimeNG 12.2.0](https://github.com/primefaces/primeng/issues/10713), but for the MultiSelect component.

When you're in a modal/dialog that's above zIndex 1000, then the Multiselect overlay opens below the dialog, no matter what you set for the `baseZIndex` property, since, looking at the actual source code, it's not being used. (https://github.com/primefaces/primeng/blob/master/src/app/components/multiselect/multiselect.ts#L684)

**Expected behavior**
The Multiselect component honors the `baseZIndex` setting and therefore renders above modals/dialogs with higher zIndex values.

Have a modal with zIndex 2000 for instance and add a Multiselect with a baseZindex value that would render it above zIndex 2000.

The multiselect is unusable for containers with a high z-index since the overlay is hidden behind the container.

* **Angular version:** 13.1.0

* **PrimeNG version:** 13.1.1

* **Browser:** all
 
* **Language:** all

* **Node (for AoT issues):** `node --version` =   16.13.1

cetincakiroglu:


rinukkusu:
Duplicate PR, already open PR here: #11191

.browserslistrc
```diff
@@ -11,3 +11,7 @@ last 2 Edge major versions
 last 2 Safari major versions
 last 2 iOS major versions
 Firefox ESR
+not IE 9-10
+not IE 11
+not ios_saf 15.2-15.3
+not safari 15.2-15.3
```
rinukkusu, line 17: This has nothing to do with issue #11191 - it's probably better to separate these changes out into their own fix.

rinukkusu (changes requested):

rinukkusu:
Also what are the 4 merge commits for? A rebase would leave a cleaner commit history.

PR #11225, opened 2022-02-22
Fixed #11224 - TreeSelect | Fix path array algorithm in selection expansion

### LINKED ISSUE #11224: TreeSelect selection expands sibling nodes above the selection
```
```

**Current behavior**
When a node is selected in the TreeSelect component, and the node is not at the root of the list of components, the component will expand any siblings of the node above it. This is because while iterating through children, it appends each sibling to the `path` as it iterates through instead of creating a new array with only the current iterating item added to the path.

**Expected behavior**
Only the selected node and its parents should be expanded.

dcbroad3:
Fixes issue reported in #11224


cetincakiroglu:
Hi,

Issue #10914, opened 2021-12-02
Multiselect options missing role 'option'


v-colig:
If you have a PrimeNG PRO Support subscription please post your issue at;

https://pro.primefaces.org 

where our team will respond within 4 business hours.

If you do not have a PrimeNG PRO Support subscription, fill-in the report below. Please note that
your issue will be added to the waiting list of community issues and will be reviewed on a first-come first-serve basis, as a result, the support team is unable to guarantee a specific schedule on when it will be reviewed. Thank you for your understanding.

Current Queue Time for Review
Without PRO Support: ~8-12 weeks.
With PRO Support: 1 hour

**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Options in p-multiselect do not have 'option' role, even though parent element has role 'listbox', which requires child element to have 'option' role. This is not an issue with version of PrimeNG, as latest version examples at https://www.primefaces.org/primeng/showcase/#/multiselect have this issue.

**Expected behavior**
Each list item should have 'option' role in the multiselect component.

**What is the motivation / use case for changing the behavior?**
Need component to be compliant for accessibility. [MAS 1.3.1 - Info and Relationships](https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame2.aspx?sourcedoc=%7b54f28d1f-a2d1-4dcd-84e1-5c9b87e8aba4%7d)

* **Angular version:**
8.2.4

* **PrimeNG version:**
11.0.0

* **Browser:**
Microsoft Edge

yigitfindikli:
Hi @v-colig,

All accessibility issues will be fixed in 2022 Q1. We're gonna publish accessibility scores as well. Thanks for your report.

v-colig:
Can I please get an update on the status of this accessibility fix?

PR #11274, opened 2022-03-03
Update multiselect.ts

v-colig:
Need this change for accessibility compliance. Raised issue a few months ago and was informed this would be fixed in Q1 2022: https://github.com/primefaces/primeng/issues/10914

###Defect Fixes

###Feature Requests


fhljys:
This is not the right fix. It won't work.

PR #11306, opened 2022-03-15
Fix: #10630

### LINKED ISSUE #10630: Scrolling in a t-table that implements virtual scrolling and lazy loading bugs out the header of the t-table
```
```

**Plunkr Case (Bug Reports)**
https://primeng-tablevirtualscroll-demo-6scvvq.stackblitz.io 

**Current behavior**
While scrolling a basic p-table that implements virtual scrolling and lazy loading, it looks like the header doesn't stick to the top of the t-table. I did some investigating myself (in a work-related project) and think this bug is caused by the cdk-virtual-scroll.

The bug really shows itself when scrolling up and down really fast in the part of the t-table where the currently loaded rows collide with the empty (to be loaded by the lazyLoadEvent) rows, I added a screencapture of the stackblitz to make it as clear as possible.

**Expected behavior**
The header of the t-table does not move when scrolling.

I used the primeNG official demo fork to reproduce this problem, I also added some html/css and inline styling to set the height of the p-table. While running the stackblitz, scroll through the p-table and when you're about to hit a lazyLoadingEvent, start scrolling up and down really fast, the table header should start to move. Even easier to reproduce if you click in the middle of the scrollbar and then, before the records can load, start scrolling up and down really fast. Both cases were recorded in the following screencapture:
wTF3mQN34o

If I use the p-table component in my work-project, the behavior becomes much worse, to the degree where I cannot upgrade my project to version 12.1.1.

OS: Windows 10
Browser: Google Chrome
Can be reproduced in stackblitz, so all other environment information should be accessible there.

* **Angular version:**
Angular version: 12.X

* **PrimeNG version:**
PrimeNG version: 12.1.1

* **Browser:**
Chrome: Version 93.0.4577.82 (Official Build) (64-bit)
 
* **Language:** N/A

* **Node (for AoT issues):** `node --version` =   N/A

Laurent02:
###Defect Fixes
This PR will fix #10630


Laurent02:
There appeared to be an issue with the table header when virtual scrolling was enabled while scrolling horizontally due to my fix. Fixed that now as well.

Laurent02:
Fixed another issue I encountered with scroll bars. Would really like some feedback on it now because we really need this fix asap @cagataycivici @yigitfindikli  @cetincakiroglu

Laurent02:
I fixed another small issue related to the scrollbar behaviour on the cdk-virtual-scroll-viewport. See attached gif for the behaviour I fixed by adding the padding.

PrimeNG

iM4x:
You kind of reverted the component to its condition before PrimeNg v12.1, nearly the same way I expected the contributors of PrimgNg would do. Keeping the header and footer outside of CdkVirtualScrolling, as PrimeNg also did before in the TableBodyTemplate, is the correct way from my point of view. Your implementation looks good to me, well done :) Hopefully they will merge it soon.

cagataycivici:
We've decided to use our own virtual scrolling that we have in PrimeVue and PrimeReact to resolve this. CDK just doesn't work well for our cases. Target is PrimeNG 14 - Angular 14 likely to be back ported to v13 as well.

Laurent02:
> We've decided to use our own virtual scrolling that we have in PrimeVue and PrimeReact to resolve this. CDK just doesn't work well for our cases. Target is PrimeNG 14 - Angular 14 likely to be back ported to v13 as well.

That's great news, thanks guys. I agree that the main issue is with the CDK virtual scrolling and that this fix was only moving the issue from vertical to horizontal scrolling which made it less disruptive, so effectively no longer using CDK sounds great. Can't wait for the fix to arrive ;)

Issue #8217, opened 2019-10-07
Allow the ability to disable animated progress on determinate progress bars without resorting to CSS overrides


Phil-DS:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**What is the motivation / use case for changing the behavior?**
In some situations, the refresh/animation rate of the progress bar is a hinderance, and the value shown is not the value on the progress bar. While slowing down the rate at updating, there are some cases where it is too fast and the progress is finished by the time the next update is to be input. If the bar can be updated automatically, without waiting for the length to change, it would be a good help.

At the moment, to disable it, you need to override the CSS to disable the ".ui-progressbar-value-animate" animations. It would be better if there was an input (Say "disableChangeAnimation") to disable the update animations, by not adding the  ".ui-progressbar-value-animate" class to the update bar. It shouldn't affect the indeterminate bar.

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

PR #11327, opened 2022-03-22
Feature #8217 | Add option to disable change animation on ProgressBar Component

Fkscorpion:
Feature Request/Issue: #8217

- Adds an option to disable the change animation on the ProgressBar Component.
  -  This adds a new [ngClass] Rule to the component, wether or not to add the class "p-progressbar-value-animate"

This is especially useful when you have fast/often changing values for the progress bar, for example in "idle browsergames", that use bars to indicate progress on building production, cooldown of an attack etc.

Default behavior on fast changing values:
ProgressBar1

With [disableChangeAnimation]="true":
ProgressBar2


Fkscorpion:
@cetincakiroglu Any chance this will make it into one of the next versions?

I could really need it for some projects and don't want to work on a fork all the time :/

Fkscorpion:
@cetincakiroglu since you've assigned yourself a while ago to this I'm not giving up yet^^

Any chance, this will make it into one of your next versions?

PR #11344, opened 2022-03-27
fix #11343

### LINKED ISSUE #11343: dragdrop p-draggable-enter class not removed when dropped out of scope
```
[ X] bug report => Search github for a similar issue or PR before submitting
```

**Current behavior**
When pDraggable is dropped onto an out of scope pDropppable, .p-draggable-enter is not removed. 
Any styling applied to droppable elements remains in place.

**Expected behavior**
When dropped on an out of scope pDroppable, .p-draggable-enter is removed.

Dragging the div labelled DRAG1 onto DROP2 causes it to retain a red border from the .p-draggable-enter class

[Stackblitz Reproduction](https://primeng-dragdrop-bug-1.stackblitz.io)

N/A

N/A

* **Angular version:** 5.X

N/A

* **PrimeNG version:** 5.X

13.3.0

all

nicholasfoden:
Moved the removeClass logic out of the allowDrop condition

### DIFF
--- a/src/app/components/dragdrop/dragdrop.ts
+++ b/src/app/components/dragdrop/dragdrop.ts
@@ -200,10 +200,11 @@ export class Droppable implements AfterViewInit, OnDestroy {
     @HostListener('drop', ['$event'])
     drop(event) {
         if (this.allowDrop(event)) {
-            DomHandler.removeClass(this.el.nativeElement, 'p-draggable-enter');
             event.preventDefault();
             this.onDrop.emit(event);
         }
+        // Always remove the class from the dropped element
+        DomHandler.removeClass(this.el.nativeElement, 'p-draggable-enter');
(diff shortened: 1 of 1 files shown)


yigitfindikli:
Hi,

Thanks for your effort and your PR. But onDragLeave already removing the 'p-draggable-enter' class in case of drop outside.

nicholasfoden:
> Hi,
> 
> Thanks for your effort and your PR. But onDragLeave already removing the 'p-draggable-enter' class in case of drop outside.

Are you certain this is true? Please check my example at https://stackblitz.com/edit/primeng-dragdrop-bug-1 as an incidence of where this doesn't work as expected.

Alternatively, Are you suggesting that i should be using onDragLeave on every pDroppable element ?

Issue #11376, opened 2022-04-01
Set aria-disabled attribute to true when p-dropdown is disabled


ajsaraujo:
```
[ ] bug report => Search github for a similar issue or PR before submitting
[X] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
The `aria-disabled` attribute of p-dropdown is always null, no matter what the component's state is. Here is a [stackblitz](https://stackblitz.com/edit/github-baxtbl?file=src/app/app.component.html) demonstrating this behavior. 

**Expected behavior**
I expect `aria-disabled` to be `true` when dropdown is disabled and `false` when it's enabled. This is not only useful accessibility-wise but also from a testing perspective - it makes it easier to check what the dropdown state is without peeking into implementation details.

**Please tell us about your environment:**

* **Angular version:** 13
* **PrimeNG version:** 13
* **Browser:** Chrome

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

PR #11379, opened 2022-04-01
#11376 - Added aria-disabled accessibility attribute for dropdown.

MursalovAltun:
.


MursalovAltun:
Simply adding aria-disabled for internal inputs for accessibility requested in #11376.

Let me please now if something needs to be changed or if I need to create an issue in docs.

Issue #10964, opened 2021-12-13
Float Label always above dropdown when using reactive forms


likwidgames:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please demonstrate your case at stackblitz by using the issue template below. Issues without a test case have much less possibility to be reviewd in detail and assisted.

https://stackblitz.com/edit/primeng-floatlabel-demo-duwdjg?file=src/app/app.component.html

**Current behavior**
Float label appears above the dropdown input when using formControlName but inside the dropdown when using (ngmodel).

**Expected behavior**
The label should appear INSIDE the dropdown input until it is selected, similar to how it works with ngmodel.

**Minimal reproduction of the problem with instructions**
https://stackblitz.com/edit/primeng-floatlabel-demo-duwdjg?file=src/app/app.component.html

**What is the motivation / use case for changing the behavior?**
This issue prevents our UI from being consistent with other float label inputs.

**Please tell us about your environment:**

* **Angular version:** 5.X
13.1.0

* **PrimeNG version:** 5.X
13.0.2

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
all
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
* all

* **Node (for AoT issues):** `node --version` =

cetincakiroglu:
https://stackblitz.com/edit/primeng-floatlabel-demo-p1w56u?file=src%2Fapp%2Fapp.component.ts

PR #11380, opened 2022-04-02
#10964. Fix dropdown is filled when an empty text value is provided.

MursalovAltun:
As many Angular developers provide empty string to an input's value it's good to have check on empty string too to avoid visual issues.

new FormControl('') would lead to filled getter to be true.


MursalovAltun:
#10964

cetincakiroglu:
Hi @MursalovAltun,

Good catch, thanks for the effort and support!

PR #11392, opened 2022-04-07
Fix for #10178 - Range slider handles get stuck at maximum with non-default max value

### LINKED ISSUE #10178: Slider strange behaviour when min and max properties don't match with their default values
```
```

**Current behavior**
I have detected a bug when the **min** and **max** properties (in range mode) don't match with their default values.

If I move the start range control to the end (max value) the slider no longer works properly. That is, I'm not able to move again the start range control to another position.

image

**Expected behavior**

The slider control should work properly when min and max values aren't the default values and I move the start control to the end control. 

You can reproduce it very easy in your Slider control page, the StackBlitz area, changing the min and max values.

image

* **Angular version:** 11.2.3

* **PrimeNG version:** 11.4.0

* **Browser:** [all]
 
* **Language:** [all]

Phil-DS:
Sliders will now use the lower handle when both are set to max, irregardless of the value passed to max. 

Fixes #10178

### DIFF
--- a/src/app/components/slider/slider.spec.ts
+++ b/src/app/components/slider/slider.spec.ts
@@ -413,4 +413,25 @@ describe('Slider', () => {
 
         expect(slider.values[1]).toEqual(80);
     });
+
+    it('should use correct handle when both are at max',() => {
+        slider.range = true;
+        slider.handleValues = [0,100];
+        slider.values = [20,80];
+        slider.min = 20;
+        slider.max = 80;
+        slider.handleIndex = 0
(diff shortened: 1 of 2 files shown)


cetincakiroglu:
Hi,

That PR makes handles stuck at sibling's value, it makes the feature broken and handles cannot move after reaching out to other one's value.

Phil-DS:
Hi,
Probably should have closed this one down earlier myself, since I noticed that the solution to the original problem was solved as a side effect of #11428, while this was specific to the original problem.

PR #11398, opened 2022-04-08
fix(dropdown): Fix filter when options is array of strings

### LINKED ISSUE #10122: Dropdown Filter not working for simple arrays like string[]
**I'm submitting a ...**
```
[x] bug report
[ ] feature request
[ ] support request
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/primeng-dropdown-demo-f3wctg

**Current behavior**
Dropdown filter doesn't work when you use a string array instead of complex objects.

* **Angular version:** 11

* **PrimeNG version:** 11.3.2

lukasmatta:
Fixes #10122

### Defect Fixes
#10122

Basically, if the first `option` is of type `string`, we can assume that `options` is an array of strings. As `filter` function of the `FilterService` is general and already accepts a function for the `searchFields` parameter, we can provide simple identity function, that for string `X` returns it's identity, an `X`. This will make filtering work on simple string options.
**This PR fixes the problem without introducing any breaking change**.

### DIFF
--- a/src/app/components/dropdown/dropdown.spec.ts
+++ b/src/app/components/dropdown/dropdown.spec.ts
@@ -23,7 +23,7 @@ class TestDropdownComponent {
 
 	groupedCars = [
 		{
-			label: 'Germany', value: 'germany.png', 
+			label: 'Germany', value: 'germany.png',
 			items: [
 				{label: 'Audi', value: 'Audi'},
 				{label: 'BMW', value: 'BMW'},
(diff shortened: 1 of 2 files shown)


YemreGurses:
Is there any update on this pr?

mertsincan:
Hi @lukasmatta,

Sorry for the delayed response! Unfortunately, we made many changes to the Dropdown component and your PR is out of date. So it's closed. Can you create a new PR for us? I will merge it as soon as possible.

Best Regards,

Issue #11408, opened 2022-04-12
wrong in style Password Component


rezaokhravi:
#11356 
https://stackblitz.com/edit/primeng-sidebar-demo-if1s7u?file=src%2Fapp%2Fapp.component.ts

volvachev:
Hi, I  created a fix and an explanation of why the problem had been occurred.

cetincakiroglu:
Hi,

Invalid StackBlitz url, link is broken. Cannot replicate and closing the issue. Feel free to re-open with a VALID StackBlitz example.

Issue #11356, opened 2022-03-30
wrong in style Password Component


rezaokhravi:
hi 
bug in style Password Component , At the time of a click on Password Component, style will be correct .
if switch to PrimeNG version: <=12.2.0 style will be correct !!!
Please fix it .

![Screenshot (24)](https://user-images.githubusercontent.com/42868109/160749356-ba18e9d6-8e6d-4ed5-b580-6fa24ece24ed.png)

* **Angular version:12.0.0

* **PrimeNG version:>=12.2.1

yigitfindikli:
The issue is not clear, please feel free to reopen with a stackblitz example.

rezaokhravi:
> The issue is not clear, please feel free to reopen with a stackblitz example.

https://stackblitz.com/edit/primeng-sidebar-demo-if1s7u?file=src%2Fapp%2Fapp.component.ts

Hamidrezae:
By adding the following piece of code, the problem is solved

 styleClass=" p-password p-component p-inputwrapper p-input-icon-right"

Issue #10716, opened 2021-10-07
float-label pInputText problem


MatteoTrapasso:
when we preload the input data for the component pInputText, the label doesn't move up. 
This is due to the missing method ngAfterViewInit() in the component. 
In this link we can show a simulation error on stackblitz. Please don't interact with the application to display the error.

https://primeng-inputtext-demo-yetsfr.stackblitz.io/page

solution:
file inputtext.ts
```
export class InputText implements DoCheck, AfterViewInit {

    constructor(public el: ElementRef, @Optional() public ngModel: NgModel, private cd: ChangeDetectorRef) {}

    ngAfterViewInit() {
          this.cd.detectChanges();

   ...
    }
```

PR #11416, opened 2022-04-13
fix(primeng/p-password): incorrect class assignment for `true` value of `toggleMask` in the password component.

volvachev:
The problem appeared during fix of the bug #10716 (running `detectChanges` in `pInputText`). Setter has been ran by detector without running `doCheck` inside `ngClass`.

To avoid the problem, either do not run `detectChanges` inside `pInputText` or do not use methods inside the template, for example: replace them with `pipes`, or call the `detectChanges` method inside `ngAfterViewInit`(p-password) as well.

To solve the problem, I have decided to use `pipes` because this is the best solution from my point of view.
But if you want to replace the `detectChanges` call, I will do it

**Defect Fixes**
#11356, #11408.


JuanSeBestia:
Thanks @volvachev 

Please, I need this fix <3

mertsincan:
Thanks a lot for your contribution, @volvachev ;)

Issue #11425, opened 2022-04-14
when I select a year with a range, only the start year is highlighted.


kafelix496:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
Please demonstrate your case at stackblitz by using the issue template below. Issues without a test case have much less possibility to be reviewd in detail and assisted.

https://stackblitz.com/github/primefaces/primeng-issue-template

**Current behavior**
when I select a year with a range, only the start year is highlighted.

**Expected behavior**
when I select a year with a range, start year, end year and years between start year and end year should be highlighted

**Minimal reproduction of the problem with instructions**

**What is the motivation / use case for changing the behavior?**
because date range is highlighted start date, end date and dates between start and end

**Please tell us about your environment:**
MacOS Monterey v12.3.1, Neovim v0.6.0.1, npm

* **Angular version:** 13.2.3

* **PrimeNG version:** 13.3.0

* **Browser:** all
 
* **Language:** all

* **Node (for AoT issues):** `node --version` =   v16.14.2

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

PR #11426, opened 2022-04-14
year select with range

kafelix496:
#11425 when I select a year with a range, only the start year is highlighted.

### DIFF
--- a/src/app/components/calendar/calendar.ts
+++ b/src/app/components/calendar/calendar.ts
@@ -1243,9 +1243,18 @@ export class Calendar implements OnInit,OnDestroy,ControlValueAccessor {
 
     isYearSelected(year) {
         if (this.isComparable()) {
-            let value = this.isRangeSelection() ? this.value[0] : this.value;
-
-            return !this.isMultipleSelection() ? (value.getFullYear() === year) : false;
+            if (this.isMultipleSelection()) {
+                return false
+            }
+            else if (this.isRangeSelection()) {
+                if (this.value[1])
+                    return this.isYearEquals(this.value[0], year) || this.isYearEquals(this.value[1], year) || this.isYearBetween(this.value[0], this.value[1], year);
+                else
+                    return this.isYearEquals(this.value[0], year)
+            }
+            else {
+                return this.isYearEquals(this.value, year);
+            }
         }
 
         return false;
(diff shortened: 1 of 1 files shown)


kafelix496:
@cetincakiroglu Hey, I also added a feature to block selection of year if the year is less/greater than min/max year. Please take a look.

src/app/components/calendar/calendar.ts
```diff
@@ -1268,6 +1277,40 @@ export class Calendar implements OnInit,OnDestroy,ControlValueAccessor {
...
+        let between: boolean = false;
+        if (start && start instanceof Date && end && end instanceof Date) {
+            return start.getFullYear() <= year && end.getFullYear() >= year;
+        }
+
+        return between;
+    }
+
+    isLessThanMinYear(year) {
+        if (this.minDate) {
```
volvachev, line 1295: I think it's better.
```ts
return this.minDate && year < this.minDate.getFullYear();
```

src/app/components/calendar/calendar.ts
```diff
@@ -1268,6 +1277,40 @@ export class Calendar implements OnInit,OnDestroy,ControlValueAccessor {
...
+          if (year < this.minDate.getFullYear()) {
+            return true;
+          }
+        }
+
+        return false;
+    }
+
+    isGreaterThanMaxYear(year) {
+        if (this.maxDate) {
```
volvachev, line 1305: I think it's better.
```ts
return this.maxDate && this.maxDate.getFullYear() < year;
```

src/app/components/calendar/calendar.ts
kafelix496: Oh, yes. You are right. I'll update it. Thanks!

src/app/components/calendar/calendar.ts
kafelix496: Oh, yes. You are right. I'll update it. Thanks!

src/app/components/calendar/calendar.ts
```diff
@@ -1268,6 +1277,40 @@ export class Calendar implements OnInit,OnDestroy,ControlValueAccessor {
         return between;
     }
 
+    isYearEquals(value, year) {
+        if (value && value instanceof Date) return value.getFullYear() === year;
+        else return false;
+    }
+
+    isYearBetween(start, end, year) {
+        let between: boolean = false;
```
volvachev, line 1286: redundant variable. `between`
```ts
if () {
return ....
}

return false;
```

src/app/components/calendar/calendar.ts
```diff
@@ -1268,6 +1277,40 @@ export class Calendar implements OnInit,OnDestroy,ControlValueAccessor {
         return between;
     }
 
+    isYearEquals(value, year) {
+        if (value && value instanceof Date) return value.getFullYear() === year;
```
volvachev, line 1281: How about this variant code?
```ts
return value && value instanceof Date && value.getFullYear() === year;
```

sujeetjaiswara:
When is this important merge coming?

PR #11478, opened 2022-05-04
Update dropdown.ts

MatteoTrapasso:
issue #11481

add attr.aria-label for every item of dropdown list and 
avoid dispatching actions when I scroll through the elements of the dropdown

###Defect Fixes
scrolling the dropdown for each focused element the component sees a change and then dispatches an action.
furthermore, there is no possibility of inserting an aria-label, which is important for accessibility

### DIFF
--- a/src/app/components/dropdown/dropdown.ts
+++ b/src/app/components/dropdown/dropdown.ts
@@ -106,7 +106,7 @@ export class DropdownItem {
                         <ng-template #itemslist let-options let-selectedOption="selectedOption">
                             <ng-container *ngIf="!virtualScroll; else virtualScrollList">
                                 <ng-template ngFor let-option let-i="index" [ngForOf]="options">
-                                    <p-dropdownItem [option]="option" [selected]="selectedOption == option" [label]="getOptionLabel(option)" [disabled]="isOptionDisabled(option)"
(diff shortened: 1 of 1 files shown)


MatteoTrapasso:
issue link:
https://github.com/primefaces/primeng/issues/11481

MatteoTrapasso:
> issue #11481
> 
> add attr.aria-label for every item of dropdown list and avoid dispatching actions when I scroll through the elements of the dropdown
> 
> ###Defect Fixes scrolling the dropdown for each focused element the component sees a change and then dispatches an action. furthermore, there is no possibility of inserting an aria-label, which is important for accessibility

MatteoTrapasso:
the code has been formatted, I apologize for the oversight.

cetincakiroglu:
Hi,

We're fully focused on Accessibility. Our main goal is to consider all the components one by one and document the individual accessibility features. In doing so, it is to correct the deficiencies or faulty implementations of that component. For this reason, our team carries out a comprehensive and meticulous study. Of course, it will take some time to complete this process.

Currently, this PR will indeed be a temporary solution for us and we cannot guarantee that these solutions will not change. Because we examine a lot of accessibility documents and try to make the best decisions for users. At this stage, the best decision that can be made about such issues is to stop the actions to be taken on them for the time being. We will reactivate these issues after we complete our work on accessibility in Prime* component libraries. In this way, we will be able to make better decisions about issues.

Details; https://www.primefaces.org/primeng/accessibility

Best Regards,

PR #11513, opened 2022-05-12
Fix dropdown/item clicks causing the window to jump

### LINKED ISSUE #11512: Clicking a dropdown or its options when it is partially off the top of the screen causes the page to jump
```
```

**Current behavior**

The whole page jumps with you click a dropdown or one of its items when it is partially off the top of the screen.

**Expected behavior**

It should not do that.

[Minimal demo](https://stackblitz.com/edit/github-szx98z)

Click the dropdown.  The page will jump to place the dropdown approximately in the middle.  Without closing the list, scroll down so that the dropdown is slightly off the top.  Click one of the options.  The page will jump again.

Users find the current behavior disorienting, especially when working in a table that contains many dropdowns.

MacOS

* **Angular version:** 11.X - 13.X

This is probably an issue with versions before 11, but I have not tested that far back.

* **PrimeNG version:** 11.X - 13.X

This is probably an issue with versions before 11, but I have not tested that far back.

* **Browser:** Chrome
 
* **Language:** all

crnisc:
Fixes #11512

### DIFF
--- a/src/app/components/dropdown/dropdown.ts
+++ b/src/app/components/dropdown/dropdown.ts
@@ -510,7 +510,7 @@ export class Dropdown implements OnInit,AfterViewInit,AfterContentInit,AfterView
 
         if (!this.isOptionDisabled(option)) {
             this.selectItem(event.originalEvent, option);
-            this.accessibleViewChild.nativeElement.focus();
+            this.accessibleViewChild.nativeElement.focus({ preventScroll: true });
         }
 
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
> Which keyboard events?

Try to press tab and focus on dropdown with your solution

cetincakiroglu:
Hi @crnisc,

Thanks for your effort and support.

Regards

PR #11536, opened 2022-05-20
Fixed #11535 - pTooltip | focus event does not work with p-components

### LINKED ISSUE #11535: pTooltip | focus event does not work with p-components
Requested by PRO User.

cetincakiroglu:


src/app/components/tooltip/tooltip.ts
```diff
@@ -118,8 +118,10 @@ export class Tooltip implements AfterViewInit, OnDestroy {
             else if (this.getOption('tooltipEvent') === 'focus') {
                 this.focusListener = this.onFocus.bind(this);
                 this.blurListener = this.onBlur.bind(this);
-                this.el.nativeElement.addEventListener('focus', this.focusListener);
-                this.el.nativeElement.addEventListener('blur', this.blurListener);
+
+                let target = this.getTarget(this.el.nativeElement);
+                target.addEventListener('focus', this.focusListener);
```
volvachev, line 123: Hi, I think you should also update the `unbindEvents` method.

PR #11551, opened 2022-05-25
fix(primeng/table): invalid unsubscribe from events

### LINKED ISSUE #11414: Event listeners not removed for ReorderableRow and ReorderableColumn
```
```
Copied from https://github.com/primefaces/primeng/issues/11357 (which was partly solved and then closed)

The main problem is in the implementation of both the unbind methods for `ReorderableRow` and `ReorderableColumn`. To illustrate this, a small excerpt from one of the bind functions:

https://github.com/primefaces/primeng/blob/3fe4a807bdb62f8fde049db0da48280b1edd3c4a/src/app/components/table/table.ts#L3315-L3316

And the corresponding unbind:

https://github.com/primefaces/primeng/blob/3fe4a807bdb62f8fde049db0da48280b1edd3c4a/src/app/components/table/table.ts#L3333-L3336

As can be seen, the listeners are registered on the `nativeElement` but removed from the `document` (where the listener was never registered), leaving the event listeners on the nativeElement behind. One could argue that, as the element is removed also the listeners get cleaned up, but this does not work for every usecase. Also, it seems that the idea was to have unbind methods, so let's make sure that they're **correct** :)

volvachev:
Fixes #11414.

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -3296,16 +3296,6 @@ export class ReorderableColumn implements AfterViewInit, OnDestroy {
 
     @Input() pReorderableColumnDisabled: boolean;
 
-    dragStartListener: any;
-
-    dragOverListener: any;
-
-    dragEnterListener: any;
-
-    dragLeaveListener: any;
-
-    mouseDownListener: any;
-
     constructor(public dt: Table, public el: ElementRef, public zone: NgZone) { }
 
     ngAfterViewInit() {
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi @volvachev,

Thanks for your effort and contribution, however we cannot merge this pull request because it does not fit our code structure. So I'll close this one.

We appreciate your support.

volvachev:
Hi @cetincakiroglu , is it closed because of `arrow` function?
I can open pr without changing `bind` functions.
You need to properly unsubscribe from events in this component (not from document).

mertsincan:
Hi @volvachev,

Good catch! Could you please send this PR again after updating it? (Note: please don't use the arrow function in global methods)

Thanks a lot!

Issue #11435, opened 2022-04-19
Except maxmonth, the other months are not greyed out and remain clickable


shreyas-wgs:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```
**Plunkr Case (Bug Reports)**
Please demonstrate your case at stackblitz by using the issue template below. Issues without a test case have much less possibility to be reviewd in detail and assisted.

https://stackblitz.com/edit/github-2kpjgt-mnooft

**Current behavior**
Except maxmonth, the other months are not greyed out and remain clickable. All Months look like clickable.

**Expected behavior**
Upto maxmonth remain clickable and others are disabled with greyed out.

**Minimal reproduction of the problem with instructions**

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**
Common

* **Angular version:** 13.3.0

* **PrimeNG version:** 13.3.1
* **Browser:** all
 
* **Language:** all

* **Node (for AoT issues):** `node --version` =   v14.17.0

devildant:
+1 it's a bug with min and max

visually the other months outside the min/max range are not grayed out and are clickable

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

PR #11570, opened 2022-05-31
fix(primeng/calendar): add disabled class for month cell

### LINKED ISSUE #11264: p-calendar |  maxDate & minDate are broken in month-picker
basically, the p-calendar when using the month-picker, changes the years, and it has the maxDate-minDate boundaries, but it doesn't apply the p-disabled class to the months

StackBlitz with reproducing the issue: [https://stackblitz.com/edit/primeng-tablescroll-demo-wdclen?file=src%2Fapp%2Fapp.component.ts](url)

volvachev:
Fixes #11264, #11435.

### DIFF
--- a/src/app/components/calendar/calendar.ts
+++ b/src/app/components/calendar/calendar.ts
@@ -106,7 +106,7 @@ export interface LocaleSettings {
                         </div>
                     </div>
                     <div class="p-monthpicker" *ngIf="currentView === 'month'">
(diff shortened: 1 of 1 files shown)


cetincakiroglu (changes requested):
Hi @volvachev,

Thanks for your effort and contribution, this pr cannot be merged right now because it breaks the "Disabled Days" case. The month becomes unselectable. Could you please refactor it by considering all cases?

volvachev:
Hi @cetincakiroglu, I though about this moment, I think we shouldn't use `disabledDays` to disable cell in `monthView`

I have several questions how it should works:

- If the week is disabled in month view, how to the month part can be selected?

- What the week is initial for `monthView` (for each month)?

- Is the month disable when there are all weeks in `disableDays`?

PR #11583, opened 2022-06-03
fix(primeng/table): restore functionality

### LINKED ISSUE #11577: Table | State, sortField and resizableColumns result in filter loading error
```
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/github-dmq63c?file=src%2Fapp%2Fapp.component.html

**Current behavior**
When using a table with state, sortField and resizableColumns will result in error:
Firefox:

ERROR TypeError: this.containerViewChild is undefined
Chrome:

ERROR TypeError: Cannot read properties of undefined (reading 'nativeElement')

**Expected behavior**
No error

- Use a table with state, sortField and resizableColumns
- Set filters
- Refresh the website so that the filters are loaded and the error occurs

Correct behaviour, no error

* **Angular version:** 5.X
13.3.7

* **PrimeNG version:** 5.X
13.4.1

Chrome and Firefox (current versions)

volvachev:
Fixes #11577


cetincakiroglu:
Hi @volvachev,

I've reviewed and tested your pr, it works but there is a problem. The issue does not exist anymore in v15 I couldn't reproduce it with the PrimeNG v15. Because of that, closing the PR.

PR #11590, opened 2022-06-06
added badge inside menubar component

mustafa05deniz:
Only added ` <span class="p-menuitem-badge" *ngIf="child.badge" [ngClass]="child.badgeStyleClass">{{child.badge}}</span>`

### DIFF
--- a/src/app/components/menubar/menubar.ts
+++ b/src/app/components/menubar/menubar.ts
@@ -3,7 +3,7 @@ import { CommonModule } from '@angular/common';
 import { DomHandler } from 'primeng/dom';
 import { MenuItem, PrimeTemplate } from 'primeng/api';
 import { RouterModule } from '@angular/router';
-import { RippleModule } from 'primeng/ripple';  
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi @mustafa05deniz,

Thanks for your effort and support.

It's a good idea though, I cannot merge this pr because we need some base styles for ".p-menuitem-badge" class. Without styles, it's just a text element. 
In this case, we need to update our Theme Designer as well to generate styles for the badge in all themes. Since the Designer is not an open-source project you cannot contribute and it should be implemented by our dev team.

Thank you so much again!

cetincakiroglu:
Hi @mustafa05deniz,

Thanks for your effort and support.

It's a good idea though, I cannot merge this pr because we need some base styles for ".p-menuitem-badge" class. Without styles, it's just a text element. 
In this case, we need to update our Theme Designer as well to generate styles for the badge in all themes. Since the Designer is not an open-source project you cannot contribute and it should be implemented by our dev team.

Thank you so much again!

PR #11651, opened 2022-06-24
Accessibility issue for Paginator in Table

wasimdevapps:
###Defect Fixes

###Feature Requests


cetincakiroglu:
Hi,

We're fully focused on Accessibility. Our main goal is to consider all the components one by one and document the individual accessibility features. In doing so, it is to correct the deficiencies or faulty implementations of that component. For this reason, our team carries out a comprehensive and meticulous study. Of course, it will take some time to complete this process.

Currently, this PR will indeed be a temporary solution for us and we cannot guarantee that these solutions will not change. Because we examine a lot of accessibility documents and try to make the best decisions for users. At this stage, the best decision that can be made about such issues is to stop the actions to be taken on them for the time being. We will reactivate these issues after we complete our work on accessibility in Prime* component libraries. In this way, we will be able to make better decisions about issues.

Details; https://www.primefaces.org/primeng/accessibility

Best Regards,

Issue #11761, opened 2022-07-30
TreeSelect: style and type attribute of TreeNode API doesn't work


mortezabahrami:
### Describe the bug

While using TreeSelect, the style and type attributes of TreeNode don't work.

I have this in my HTML template:
`
<p-treeSelect >
    <ng-template let-node pTemplate="temp"> TEMP ITEM </ng-template>
</p-treeSelect>
`

and these attributes on each item:
`
"type": "temp",
"style": "color:red;",
`

and also [this is ](https://stackblitz.com/edit/primeng-treeselect-demo-2a1e1f?file=src/app/app.component.html) stack blitz sample.

It would be great if anyone can suggest me a workaround for this.

### Environment

Primeng 13.4.1

### Reproducer

https://stackblitz.com/edit/primeng-treeselect-demo-2a1e1f?file=src/app/app.component.html

### Angular version

13.3.2

### PrimeNG version

13.4.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

16.16.0

mortezabahrami:
I should mention that the "style" attribute neither doesn't work on the Tree component.
Like in [this link](https://stackblitz.com/edit/primeng-tree-demo-vg2wi9?file=src/assets/files.json).

mortezabahrami:
About the type attribute, I've found that the TreeSelect does not pass the templates attribute to the Tree component.

SHENGALTEC:
Issue still remain, type attribute is not working for tree select (p-treeSelect). Currently there are no way to make node template since type attribute does not work

PR #11765, opened 2022-08-01
fix for part of issue #11761

mortezabahrami:
I've added parentTemplates attribute to the Tree and added a templateMap to the TreeSelect to identify required types to pass to the Tree.

###Defect Fixes
It'll fix part of #11761 issue.

### DIFF
--- a/src/app/components/tree/tree.ts
+++ b/src/app/components/tree/tree.ts
@@ -653,6 +653,8 @@ export class Tree implements OnInit,AfterContentInit,OnChanges,OnDestroy,Blockab
     @Input() indentation: number = 1.5;
 
     @Input() trackBy: Function = (index: number, item: any) => item;
+    
+    @Input() parentTemplates: any;
 
     @Output() selectionChange: EventEmitter<any> = new EventEmitter();
 
@@ -764,7 +766,7 @@ export class Tree implements OnInit,AfterContentInit,OnChanges,OnDestroy,Blockab
     }
 
     ngAfterContentInit() {
-        if (this.templates.length) {
+        if (this.templates.length || this.parentTemplates != undefined) {
             this.templateMap = {};
         }
 
@@ -791,6 +793,10 @@ export class Tree implements OnInit,AfterContentInit,OnChanges,OnDestroy,Blockab
                 break;
             }
         });
+
+        for (let key in this.parentTemplates) {
+            this.templateMap[key] = this.parentTemplates[key];
+        }
     }
 
     updateSerializedValue() {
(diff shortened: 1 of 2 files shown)


src/app/components/tree/tree.ts
```diff
@@ -791,6 +793,10 @@ export class Tree implements OnInit,AfterContentInit,OnChanges,OnDestroy,Blockab
                 break;
             }
         });
+
+        for (let key in this.parentTemplates) {
+            this.templateMap[key] = this.parentTemplates[key];
```
bozzelliandrea, line 798: Hi @mortezabahrami , with this for loop you override all the template properties with the same name, is that correct? so if i specify some template only for the tree it will be overridden?

src/app/components/treeselect/treeselect.ts
```diff
@@ -60,7 +60,7 @@ export const TREESELECT_VALUE_ACCESSOR: any = {
             </div>
             <div class="p-treeselect-items-wrapper" [ngStyle]="{'max-height': scrollHeight}">
                 <p-tree #tree [value]="options" [propagateSelectionDown]="propagateSelectionDown" [propagateSelectionUp]="propagateSelectionUp" [selectionMode]="selectionMode" (selectionChange)="onSelectionChange($event)" [selection]="value"
-                    [metaKeySelection]="metaKeySelection" (onNodeExpand)="nodeExpand($event)" (onNodeCollapse)="nodeCollapse($event)"
+                    [metaKeySelection]="metaKeySelection" (onNodeExpand)="nodeExpand($event)" (onNodeCollapse)="nodeCollapse($event)" [parentTemplates]="getParentTemplates()"
```
bozzelliandrea, line 63: With Angular for the input binding is a best practice pass one property and not a function, to avoid the continuous updating of the DOM

src/app/components/tree/tree.ts
mortezabahrami: Hi @bozzelliandrea, yes and yes.
But I have another question: currently is it possible to pass templates to the Tree while using the Treeselect?

src/app/components/treeselect/treeselect.ts
mortezabahrami: You're right. I'll change it.
Thanks

mortezabahrami:
```suggestion
                    [metaKeySelection]="metaKeySelection" (onNodeExpand)="nodeExpand($event)" (onNodeCollapse)="nodeCollapse($event)" [parentTemplates]="templateMap"
```

src/app/components/tree/tree.ts
mortezabahrami: But, It's better to  change it this way:
```suggestion
            if (!['header', 'empty', 'footer', 'loader'].includes(key)) {
                    this.templateMap[key] = this.parentTemplates[key];
            }
```

bozzelliandrea: @mortezabahrami Yes i think you can customize the template, using styleClass or style bindings for the tree or add your custom template with projection like this
```
<p-treeSelect [(ngModel)]="selectedValue2" selectionMode="multiple" [options]="nodes" placeholder="Select Items">
    <ng-template pTemplate="value">
        Custom Content
    </ng-template>
</p-treeSelect>
```
but honestly I've never had to customize this component that much.

mortezabahrami: The main problem is, the style binding doesn't work at all. And the template only applies to the treeselect input area.
Look at [this](https://stackblitz.com/edit/primeng-treeselect-demo-2a1e1f?file=src/app/app.component.html).

cetincakiroglu:
Hi,

Issue is fixed, closing the PR

PR #11767, opened 2022-08-02
add p-inputtext class to cascadeSelect component

mohammadreza99:
All form components are affected by `p-inputtext-lg` css class and become larger except the `cascadeSelect`. Based on css styles, I add a `p-inputtext` class to `cascadeSelect` component.

### DIFF
--- a/src/app/components/cascadeselect/cascadeselect.ts
+++ b/src/app/components/cascadeselect/cascadeselect.ts
@@ -714,6 +714,7 @@ export class CascadeSelect implements OnInit, AfterContentInit, OnDestroy {
     labelClass() {
         return {
             'p-cascadeselect-label': true,
+            'p-inputtext': true,
             'p-placeholder': this.label() === this.placeholder,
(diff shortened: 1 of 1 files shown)


bozzelliandrea (commented):
The class `p-inputtext ` si not defined in the css file, it works? 
However, i think this isn't the best practice to set every `CascadeSelect ` larger, it should be possible to choose the size with one parameter.

mohammadreza99:
Yes `p-inputtext` is defined in every `theme.css` file. 
According to the style defined in [this file (line 1238)](https://github.com/primefaces/primeng/blob/master/src/assets/components/themes/arya-blue/theme.css#L1238), we can give the `p-inputtext-lg` class to the parent element wherever we need. As for one `inputText`, if you do this, you will make it larger, and naturally it will not be applied to other `inputText`s. 
This class exists in all form components except `cascadeSelect`.

bozzelliandrea (approved):

Issue #10232, opened 2021-05-12
Property [autofocus] not working on AutoComplete component


funregis:
Hi, 

When I use the Autocomplete stackblitz from primeng , and I set the property autofocus to true , the focus is not working

here an example with the autofocus property

https://stackblitz.com/edit/primeng-autocomplete-demo-faaeoh?file=src/app/app.component.html

**Current behavior**
The property [autofocus]="true" don't affect the focus of the autocomplete component

**Expected behavior**
When The property [autofocus]="true" then the focus is set on the autocomplete in the render view

**What is the motivation / use case for changing the behavior?**
Actually if you want to set the focus on the autocomplete component , you need to use a ViewChild and a "focusInput()" on the aucomplete retrieved by the viewchild , it makes difficult to share a common approach to handle the focus of every primeng input component

**Please tell us about your environment:**
IDE : Intellij , prineng 11.3.2 (also tested on 11.4.1)

* **Angular version:** 11.2.10

* **PrimeNG version:** 11.3.2 (also tested on 11.4.1)

* **Browser:** [all ]

 
* **Language:** [TypeScript 4.0.3 ]

quranc:
#11780

cetincakiroglu:
Hi,

Since this problem occurs across many components, I'm closing this issue and merging them into one https://github.com/primefaces/primeng/issues/11807. We'll fix them altogether.

Thanks for your support and report!

PR #11780, opened 2022-08-05
fix autofocus autocomplete

quranc:
This solves issue #10232.

Using the `[autofocus]` attribute on an `<p-autoComplete>` element was having no effect.
It did not change the behaviour, no matter of values.

Not working:
`<input [attr.autofocus]="true"/>`

Working:
`<input autofocus/>`

The element seems to only support that you have the attribute `autofocus` or not, which this PR now fixes.


cetincakiroglu:
Hi @quranc,

Thanks for your support and effort, great catch but this is not a valid implementation to solve this problem, so I have to reject your solution. 
Thanks to your pull request, we've realized that this issue exists on multiple components and is caused by Angular's logic. We'll fix all of them together in the near future.

Regards

PR #11904, opened 2022-09-07
Fixed #9842 - Button | Renders empty label when the label is not defined

### LINKED ISSUE #9842: p-autocomplete does not clears selected object when CTRL-Z is pressed in textbox
```
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/github-16c1yp

**Current behavior**
autocomplete with forceSelection=true
type 'first' into textbox and selected item
id 1 shows next to label id:
now press CTRL-Z (cursor moves to beginning of line) and then immediately press TAB
id is cleared but First remains in textbox so one has the idea an items was selected but no id is showing

**Expected behavior**

- CTRL-Z just moves cursor to beginning of line, it should delete the text
- When pressing TAB after CTRL-Z, the object (in this case 'item' on the component) should be null

https://stackblitz.com/edit/github-16c1yp

* **Angular version:** 11.1

* **PrimeNG version:** 11.2

* **Browser:** [chrome]

 
* **Language:** [TypeScript 4.x ]

cetincakiroglu:
.


cetincakiroglu:
Fixes #9482

cetincakiroglu:
Fixes #9482

xanderknol:
can this be implemented? This bug is still present and really annoying. Thanks!

PR #12004, opened 2022-10-03
Fixed #11918 - DataTable: Multiple sorting with Date field as first sort

### LINKED ISSUE #11918: DataTable: Multiple sorting with Date field as first sort
### Describe the bug

The same issue what Primereact had. [link to the issue](https://github.com/primefaces/primereact/issues/3284)
DataTable doesn't sort dates correctly. Two completely identical dates are different for table and this bug causes another bug, when you try to sort multiple columns and use date column as primary. Every next column with lower sort priority after date column will not be sorted correctly.
Added stackblitz with example. Just select Year column as primary sort and try to sort Color column as second column. You will not be able to sort cars with date "21.05.1995" by color alphabetically, because table see these dates as different.
As I see, issue in function
```
multisortField(data1, data2, multiSortMeta, index) {
...
if (value1 == value2)
}
```
Dates should be compared in another way

### Environment

Just new angular project with added primeng lib

### Reproducer

https://stackblitz.com/edit/github-zpj9hk-hkrmgz

### Angular version

14.2.0

### PrimeNG version

14.0.2

### Build / Runtime

Angular CLI App

### Language

ALL

16.13.1

### Browser(s)

_No response_

1. Create DataTable with multisort mode
2. Set data with date column and set some rows with the same date, but different any second column values
3. Select date column as primary and try to sort any another column

### Expected behavior

DataTable understand the same dates and correctly sort another column in multisort mode

cetincakiroglu:


andreduejon:
Hi guys!

I've been trying to figure out why my multi-column sort has been behaving strangely lately. Ever since I upgraded from an early version 14.x.x to the version 15.3.0. 

It seems that line 181 `value1.localeCompare(value2, locale, { numeric: true }` in `objectutils.ts` is no longer sorting our data properly. Setting the field `numeric` to `false` solved the issue for me, but I expect that this was implemented for a reason.

Example: When sorting in ascending order, the value _59A446_ (string) comes before _038898_ (string). The field contains unique IDs that can contain letters, but mostly consist of numbers.

I guess I could use a custom sort function, however, I do not believe that this behaviour is correct.

PR #12012, opened 2022-10-05
Fix primefaces#12011: align contextmenu properly when body is scrolled

### LINKED ISSUE #12011: Component: ContextMenu is not aligned properly when the document is scrolled
### Describe the bug

If the document is scrolled and we try to open a contextmenu which is too big to open at the cursor then the contextmenu gets aligned outside of the viewport.
See attached stackblitz.

### Environment

Latest Angular and Primeng in chrome or firefox.

### Reproducer

https://stackblitz.com/edit/github-qspcge?file=src%2Fapp%2Fapp.component.html,src%2Fstyles.scss

### Angular version

14.x.x

### PrimeNG version

14.x.x

### Build / Runtime

Angular CLI App

### Language

TypeScript

14.18.3

### Browser(s)

_No response_

Reproduction steps are available in stackblitz.

### Expected behavior

ContextMenu should appear as if the document wasn't scrolled or at least it should not get aligned outside of viewport.

Hapcy:
Fixes #12011 .

### DIFF
--- a/src/app/components/contextmenu/contextmenu.ts
+++ b/src/app/components/contextmenu/contextmenu.ts
@@ -303,23 +303,23 @@ export class ContextMenu implements AfterViewInit, OnDestroy {
             let viewport = DomHandler.getViewport();
 
             //flip
-            if (left + width - document.body.scrollLeft > viewport.width) {
+            if (left + width - document.scrollingElement.scrollLeft > viewport.width) {
                 left -= width;
             }
 
             //flip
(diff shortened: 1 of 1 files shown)


Hapcy:
I changed document.body.scrollLeft and document.body.scrollTop to document.scrollingElement.scrollLeft and document.scrollingElement.scrollTop respectively because these fields on document.body are always 0.

cetincakiroglu:
Hi @Hapcy,

PR #12030, opened 2022-10-12
Issue 11992

cetincakiroglu:
- Updates appendTo

### DIFF
--- a/src/app/components/api/primengconfig.ts
+++ b/src/app/components/api/primengconfig.ts
@@ -1,4 +1,4 @@
-import { Injectable } from '@angular/core';
+import { ElementRef, Injectable, TemplateRef } from '@angular/core';
 import { Subject } from 'rxjs';
 import { FilterMatchMode } from './filtermatchmode';
 import { Translation } from './translation';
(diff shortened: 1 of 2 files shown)


src/app/components/api/primengconfig.ts
```diff
@@ -8,6 +8,8 @@ export class PrimeNGConfig {
 
     ripple: boolean = false;
 
+    appendTo: any = "body" || HTMLElement || ElementRef || TemplateRef;
```
volvachev, line 11: Hi, I don't quite understand what to use the HtmlElement constructor for? mb only type is needed here?

```ts
appendTo: HtmlElement | ElementRef | TemplateRef | 'body' = 'body'
```

cetincakiroglu: It could be div, a "p-element" etc..

volvachev: Does not make sense, the string 'body' is always true when the HtmlElement constructor will be used?
```ts
const a = "body" || HTMLElement || ElementRef || TemplateRef;
a; // always 'body'
```

PR #12037, opened 2022-10-12
Feature: Custom buttons on dialogs

Tuxio:
###Feature Requests
This feature adds custom buttons to the top right header of a dialog.
grafik (see the question mark)

A user can add multiple buttons by declaring the ngTemplate _buttons_ to a dialog. To that template the user can add some button-tags which will be displayed on the top right of the dialog.

I have added some documentation, one example in the showcase and test cases.


src/app/components/dialog/dialog.ts
```diff
@@ -30,21 +30,38 @@ const hideAnimation = animation([
...
+                    <span [attr.id]="id + '-label'" class="p-dialog-title"
+                          *ngIf="!headerFacet && !headerTemplate">{{header}}</span>
                     <span [attr.id]="id + '-label'" class="p-dialog-title" *ngIf="headerFacet">
                         <ng-content select="p-header"></ng-content>
                     </span>
                     <ng-container *ngTemplateOutlet="headerTemplate"></ng-container>
                     <div class="p-dialog-header-icons">
-                        <button *ngIf="maximizable" type="button" [ngClass]="{'p-dialog-header-icon p-dialog-header-maximize p-link':true}" (click)="maximize()" (keydown.enter)="maximize()" tabindex="-1" pRipple>
-                            <span class="p-dialog-header-maximize-icon" [ngClass]="maximized ? minimizeIcon : maximizeIcon"></span>
+                        <button *ngFor="let btn of headerButtons" type="button"
```
volvachev, line 50: I think, need add trackBy for ngFor

src/app/components/dialog/dialog.ts
```diff
@@ -30,21 +30,38 @@ const hideAnimation = animation([
...
+                                [ngClass]="{'p-dialog-header-icon p-link':true}" (click)="btn.action()"
```
volvachev, line 51: Mb, instead of ngClass you should use class="p-dialog-header-icon p-link" here

src/app/components/dialog/dialog.ts
```diff
@@ -171,6 +188,8 @@ export class Dialog implements AfterContentInit,OnInit,OnDestroy {
 
     @Input() maximizeIcon: string = 'pi pi-window-maximize';
 
+    @Input() headerButtons: any[] = null;
```
volvachev, line 191: mb use `ng-content` for additional buttons inside `p-dialog-header-icons` instead of `@Input`?

Tuxio:
Thanks for your feedback @volvachev. I have added the trackBy function (but when I look at the DOM, also without the trackBy function a newly added button does not update all DOM-buttons-Elements of the list)

src/app/components/dialog/dialog.ts
Tuxio: I think it would be more flexible, but then you could also add other elements which could break the structure and formatting of the header. In my opinion it is better when you only have the option to add custom buttons to the top right part of the dialog as it is intended by the framework.

Tuxio:
I have removed the usage of the `@input` decorator and used the `ng-template ` element to add custom buttons to the header, like you suggested before @volvachev.

Tuxio:
Hi @volvachev, just wanted to ask if you have any new news to this pull request?

Issue #10472, opened 2021-08-03
In a Dynamic Dialog that is opened from another Dynamic Dialog, we can not press Tab to focus on the next field


phuctvt:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Stackblitz Case (Bug Reports)**
https://stackblitz.com/edit/primeng-dynamicdialog-demo-adhpi8

**Current behavior**
1. Open a Dynamic Dialog from another Dynamic Dialog.
2. There are some input fields in the dialog, when I press Tab, it will not focus on the next field.

You can try the issue by the link Stackblitz above.

**Expected behavior**
It should focus on the next input field.

**What is the motivation / use case for changing the behavior?**
If the dialog I mentioned is the dialog of filling some fields, the tab-focus-next-field behavior should works.

**Please tell us about your environment:** Stackblitz

* **Angular version:** 11.0

* **PrimeNG version:** 11.3.0

* **Browser:** I tested on Chrome and Firefox
 
* **Language:** TypeScript

tisys-parceiros:
was this problem resolved?

miguelarcjr:
I have the same problem

nev-21:
me too, same problem

mselerin:
This seems related to #9329 for which I have a [workaround](https://github.com/primefaces/primeng/issues/9329#issuecomment-1276432459) :-)

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

jayee:
This is still a problem in 17.3.3.

The tab-key does not cycle between focusable elements correctly. It gets stuck on the close-x when using tab-key. It can shift-tab one step.

PR #12039, opened 2022-10-12
fix: focus does not work on nested dialog 

### LINKED ISSUE #9329: #FocusTrap for nested DynamicDialogs still not working
Tab keys not working when i open dynamic dialog from another dynamic dialog, it stuck on close button so if i focus on control by using mouse keys and then i press tab key it focus again close button. I am using 9.1.5-lts version

mselerin:
### Defect Fixes  
Fix #9329 and #10472 

The fix use inject the parent `DynamicDialogComponent` inside the child dialog, disable the key listener, then restore it when the child dialog is destroyed.

Review and comments are more than welcome :-)


src/app/components/dynamicdialog/dialogservice.ts
```diff
@@ -17,6 +18,15 @@ export class DialogService {
 
         this.dialogComponentRefMap.get(dialogRef).instance.childComponentType = componentType;
 
+        const parentDialog = this.injector.get(DynamicDialogComponent, null);
+        if (parentDialog) {
+            parentDialog.unbindDocumentKeydownListener();
+            dialogRef.onDestroy.pipe(
+                take(1),
+                tap(() => parentDialog.bindDocumentKeydownListener())
```
volvachev, line 26: Hi, Why use tap instead of logic inside subscribe?

mselerin: Just a personnal preference, based on a lot of code reviewing :

*   easier to see the chain of treatment for the observable
*   if the subscribe need to be extracted (for example to be used with an async pipe), the refactor is easy
*   the subscribe accept in fact 3 arguments (next, error and complete handler), so always using it with no argument reduce a misplacement for those handlers

I could resume this way of writing my code like 'declare ; prepare ; execute' 😅

volvachev: The current project uses subscribe instead of `tap` more often. `Tap` also has 3 arguments)

mselerin: OK that make sense (consistency is key) ;-)

mselerin:
I've just update the code with a better solution : retrieve the parent `DynamicDialogComponent` inside the child `DynamicDialogComponent` directly, instead of modifying the `DialogService`.

It's cleaner since the service should not be aware of the key bindinds from the component.

mselerin:
> Hi,
> 
> Reviewed and it seems that does not work with #10472

It seems that the fix does not work when the dialog is opened a second time... I'll check that soon !

mselerin:
After some investigations, it seems that the `bindDocumentKeydownListener` was called twice. One time from my fix, the other time from `onAnimationStart`.

I've just add a guard to prevent a double keydown listener.

mselerin:
Hi @cetincakiroglu, did you have time to test again this PR with #10472 ?

Does it still need some works ?

cetincakiroglu:
Hi @mselerin 

I've just reviewed and tested your PR but it seems it still does not work. Here are the steps I follow, am I doing something wrong?

1- Click and open the first dialog
2- Click and open second dialog
3- Press "tab" to focus next input element

After the steps above, when I hit the tab after the second dialog is opened, only the close button receives focus instead of the next input element.

mselerin:
Hi @cetincakiroglu , thanks for your answer.

I have followed the exact same step as you and for me the problem is solved :/

Here is my modified `ProductListDemo` component to test this out :

```html
@Component({
    template: `
        <p-table [value]="products" responsiveLayout="scroll" [paginator]="true" [rows]="5" [responsive]="true">
            <ng-template pTemplate="header">
                <tr>
                    <th pSortableColumn="name">Name <p-sortIcon field="vin"></p-sortIcon></th>
                    <th pSortableColumn="year">Image</th>
                    <th pSortableColumn="price">Brand <p-sortIcon field="price"></p-sortIcon></th>
                    <th pSortableColumn="inventoryStatus">Status <p-sortIcon field="inventoryStatus"></p-sortIcon></th>
                    <th style="width:4em"></th>
                </tr>
            </ng-template>
            <ng-template pTemplate="body" let-product>
                <tr>
                    <td>{{ product.name }}</td>
                    <td><img src="assets/showcase/images/demo/product/{{ product.image }}" [alt]="product.image" class="product-image" /></td>
                    <td>{{ product.price }}</td>
                    <td>
                        <span [class]="'product-badge status-' + product.inventoryStatus.toLowerCase()">{{ product.inventoryStatus }}</span>
                    </td>
                    <td>
                        <button type="button" pButton icon="pi pi-search" (click)="selectProduct(product)"></button>
                    </td>
                </tr>
            </ng-template>
        </p-table>

        <p-button (onClick)="openNestedDialog()" label="Nested dialog"></p-button>
    `,
    styles: [
        `
            .product-image {
                width: 50px;
                box-shadow: 0 3px 6px rgba(0, 0, 0, 0.16), 0 3px 6px rgba(0, 0, 0, 0.23);
            }
        `
    ],
    providers: [DialogService]
})
export class ProductListDemo {
    products: Product[];

    constructor(
        private productService: ProductService,
        public ref: DynamicDialogRef,
        public config: DynamicDialogConfig,
        private dialogService: DialogService
    ) {}

    ngOnInit() {
        this.productService.getProductsSmall().then((products) => (this.products = products));
    }

    selectProduct(product: Product) {
        this.ref.close(product);
    }

    openNestedDialog() {
        this.dialogService.open(ProductListDemo, {
            header: 'Nested Dialog',
            width: '50%'
        });
    }
}
```

This demo open a nested dialog with the same component (so you can open infinite nested dialog if you want).

Without my fix, pressing tab in the nested dialog only focus on the close button.

With my fix, the focus switch on the table headers and selection buttons.

mselerin:
Hi, any news on this one ?

mselerin:
Hi @cetincakiroglu ,

I have just synchronize my branch with master.
Are there any chances for this fix to be merge soon ?

Do you need any extra informations or help for this ?

djay1977:
Hi @mertsincan, @cagataycivici
Would it be possible to merge this PR which works perfectly. We've been waiting for this fix for several years.
Thanks in advance

cetincakiroglu:
Hi @mselerin,

Sorry for the latency, great catch I've merged it after a tiny refactor.

yarikwest:
for me it still doesn't work :( (version 16.5.0)

PR #12085, opened 2022-10-22
fix: Menubar: Don't show icon on mobile if model has no items

### LINKED ISSUE #12084: MenuBar: Menu Icon in mobile view is shown even though there are no items in model
### Describe the bug

While the menu bar handles empty or NULL values for `[model]` gracefully in normal situations, once switched to the mobile view (`max-width <= 960px`) a menu-icon is shown no matter what.

It is confusing to a user if the menu bar shows this menu-icon and even displays a menu stub on click but no content is rendered.

### Environment

Default / standard

### Reproducer

https://stackblitz.com/edit/github-lcf9ws

### Angular version

14.2.1

### PrimeNG version

14.1.2

### Build / Runtime

Angular CLI App

### Language

ALL

16.16.0

### Browser(s)

_No response_

1. Use my example from https://stackblitz.com/edit/github-lcf9ws
2. Click menu icon (burger menu)
3. You see a menu sub popping up below the icon

### Expected behavior

When the model input on menubar is falsely or an empty array, the menu icon should never be visible.

tfohlmeister:
Fixes #12084

This MR tries to fix an issue for the MenuBar as described in #12084

Previously: Even though the model was empty or held no items, the menu icon was shown on mobile views.

With this MR: When the model input on menubar is falsely or an empty array, the menu icon is never visible.

### DIFF
--- a/src/app/components/menubar/menubar.css
+++ b/src/app/components/menubar/menubar.css
@@ -46,7 +46,7 @@
     z-index: 1;
 }
 
-.p-menubar .p-submenu-list > .p-menuitem-active > p-menubarsub > .p-submenu-list  {
+.p-menubar .p-submenu-list > .p-menuitem-active > p-menubarsub > .p-submenu-list {
     display: block;
     left: 100%;
     top: 0;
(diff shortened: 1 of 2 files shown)


cetincakiroglu:
Hi @tfohlmeister,

Good catch! We've appreciate your effort but this fix needs to be done by PrimeNG Dev Team.

Thanks for your support and effort!

Issue #12031, opened 2022-10-12
Component: Overlay


cetincakiroglu:
- [x] Implementation
- [x] Documentation

Affected Components for now;
- [x] Dropdown
- [x] MultiSelect
- [x]  TreeSelect
- [x]  CascadeSelect

cetincakiroglu:
(MultiSelect)[https://github.com/primefaces/primeng/pull/12111]

PR #12096, opened 2022-10-25
Refactor #12031 - Component: Responsive Overlay

mertsincan:
Refactor #12031


src/app/components/scroller/scroller.ts
```diff
@@ -473,31 +465,38 @@ export class Scroller implements OnInit, AfterContentInit, AfterViewChecked, OnD
...
+
+            this.defaultWidth = DomHandler.getWidth(this.elementViewChild.nativeElement);
+            this.defaultHeight = DomHandler.getHeight(this.elementViewChild.nativeElement);
+            this.initialized = true;
+        }
     }
 
     ngAfterViewChecked() {
+        if (!this.initialized) {
+            this.ngAfterViewInit();
```
volvachev, line 481: Hi, this looks weird. I think it's better to create a private method and call it instead of using the component's lifecycle method.

src/app/components/api/overlayoptions.ts
```diff
@@ -0,0 +1,43 @@
...
+    target?: HTMLElement | undefined;
+    mode?: OverlayModeType;
+}
+
+export interface OverlayOnHideEvent extends OverlayOnShowEvent {}
+
+export interface OverlayOptions {
+    style?: any;
+    styleClass?: string;
+    appendTo?: 'body' | HTMLElement | undefined;
```
volvachev, line 32: Hi, I think you should also add 'self' for type

```ts
 appendTo?: 'body' | 'self' | HTMLElement;
```

src/app/components/scroller/scroller.ts
mertsincan: Hi, thanks a lot for your review! I agree with you. I'll update this line.

src/app/components/api/overlayoptions.ts
mertsincan: Yes, this enhancement already exists in our roadmap. We will port a structure like in PrimeReact. https://www.primefaces.org/primereact/setup/

src/app/components/api/overlayoptions.ts
```diff
@@ -0,0 +1,43 @@
...
+    style?: any;
+    styleClass?: string;
+    breakpoint?: string;
+    media?: string;
+    direction?: ResponsiveOverlayDirectionType;
+}
+
+export interface OverlayOnShowEvent {
+    container?: HTMLElement | undefined;
+    target?: HTMLElement | undefined;
```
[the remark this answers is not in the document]
mertsincan: No, the `container` is an 'overlay' element like dropdown's overlay. The `target` is an element that manages the overlay.
Exp;
```
container: div.p-overlay.p-component.ng-star-inserted....

target: div.p-dropdown.p-component.p-dropdown-open.p-dropdown-clearable...
```

The container is an HTML element with a 'p-overlay'(mode: 'overlay') or 'p-overlay-modal'(mode: 'modal') class depending on the mode.

paulstelzer:
Any idea when this will be released?

src/app/components/api/overlayoptions.ts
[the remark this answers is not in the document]
mertsincan: Note: This line has been updated.

mertsincan:
@paulstelzer I'm working on virtualScroller issues now. We plan to release it next week.

Issue #9380, opened 2020-10-08
Tab key in p-dialog with p-InputNumber


hugograf:
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-dialog-tab?embed=1&file=src/app/app.component.html

**Current behavior**
* Click Button Show
* Set cursor on first input field
* Move cursor by pressing tab or shift+tab key
* Try to override or to correct values of the input fields

**Expected behavior**
The whole value of a input should be selected in a p-dialog, easier to override

* **Angular version:** 10.X 
* **PrimeNG version:** 10.0.2 (Possibly any version)
* **Browser:** [all]

dcbroad3:
I also see this issue in the latest version of PrimeNG (14.1.2)

dcbroad3:
Opened PR #12124 to resolve this issue, the root issue is that the `FocusTrap` directive used by the `Dialog` only focuses the chosen element, and does not `select` it if it is an `HTMLInputElement`.

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

dcbroad3:
@mertsincan @cetincakiroglu I can confirm that this issue still exists in PrimeNG/Angular version 15. 

PR #12124 still contains a viable fix.

cetincakiroglu:
> @mertsincan @cetincakiroglu I can confirm that this issue still exists in PrimeNG/Angular version 15.
> 
> PR #12124 still contains a viable fix.

Hi @dcbroad3,

I've reviewed your pr and requested some changes.

Thank you for your support and effort!

PR #12124, opened 2022-11-02
fix(primeng/focustrap): select the focused element when applicable

dcbroad3:
Resolves issue #9380

This PR slightly refactors the `FocusTrap` directive to use a new method on `DomHandler`, `getNextFocusableElement`, which is essentially a copy of the existing `FocusTrap` logic to choose which element to focus on.
Then, after calling `focus()` on the chosen element, it will also call the `select()` method, if it exists, so that the contents of `input` elements are selected as they would be normally.


cetincakiroglu (changes requested):
Hi @dcbroad3,

Thanks for your effort and support, could you please fulfill all pr checks, so it can be merged safely

dcbroad3:
> Hi @dcbroad3,
> 
> Thanks for your effort and support, could you please fulfill all pr checks, so it can be merged safely

A little unnerving because I didn't actually change anything, but I synced with master and it's now passed the PR checks.

cetincakiroglu:
Hi @dcbroad3,

Great PR! Thanks for your effort and support!

Issue #11848, opened 2022-08-19
Accordion Panel within an Accordion Panel is broken


rosenthalj:
### Describe the bug

A p-accordion within p-accordion  is broken in primeNG v14.0.1. 

The contents of all the p-accordionTab panels of an embedded p-accordion panel is  always visible when the  embedded p-accordion panel is visible

![AccordionWithAccordion](https://user-images.githubusercontent.com/45439491/185700645-5bb09772-6a75-4679-a6b1-df967161eae8.gif)

https://user-images.githubusercontent.com/45439491/185698133-f293ede4-4342-4bca-8fc2-6b2330a6805a.mov

I believe the bug was induced by #11617 commit 
 https://github.com/primefaces/primeng/commit/f849e9d23b0c2223b6fbb516d3b0c368c927cf22

![Fixed__11617_-_PrimeNG_component_animation_trigger_warnings_since_upd…_·_primefaces_primeng_f849e9d](https://user-images.githubusercontent.com/45439491/185705188-dce978e8-206b-4898-83d1-7d07149bd73d.gif)

### Environment

Happens in both stackblitz and local project

### Reproducer

https://stackblitz.com/edit/github-irxb1p?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.spec.ts

### Angular version

14.x

### PrimeNG version

14.0.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

14+

### Browser(s)

All Browsers

### Steps to reproduce the behavior

https://github-irxb1p.stackblitz.io

1. Expand all header tabs
2. Randomly open and close Header A and Header B tabbed panel.
- See that the contents of Header A and Header B are always visible

### Expected behavior

Embedded PrimeNG Accordion panels work the same as non-embedded PrimeNG Accordions.  

V14.x embedded p-accordion works the same as V13.x  embedded p-accordion

majkers:
The same as https://github.com/primefaces/primeng/issues/11736?

krtek4:
If a dropdown is bigger than the size of the accordion, its content isn't displayed entirely.

Instead of just removing the buggy css, something like that could be done :

```
.p-accordion-tab-active > .p-toggleable-content:not(.ng-animating) {
  overflow: visible;
}
```

kAgotsi:
Hello, I'm facing same issue, I tried fix :

[...]

Issue #11736, opened 2022-07-25
Toggleable fieldset inside another toggleable fieldset has its content visible when collapsed by deafult


majkers:
### Describe the bug

When one toggleable fieldset has another toggleable fieldset inside and thay are both collapsed by default, and when expanding the first one, then the content of the inner one is visible even though it is collapsed.

### Environment

Angular 14.1
PrimenNG 14
Windows

### Reproducer

https://stackblitz.com/edit/github-trkcts-tjm2ed?file=src/app/app.component.html

### Angular version

14.1.0

### PrimeNG version

14.0.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

14.17.3

### Browser(s)

Chrome  103.0.5060.134

### Steps to reproduce the behavior

1. go to https://github-trkcts-tjm2ed.stackblitz.io/ or https://stackblitz.com/edit/github-trkcts-tjm2ed?file=src/app/app.component.html
2. expand first fieldset
3. second fieldset is collapsed and its content is visible

### Expected behavior

The content of second fieldset should be hidden when fieldset is collapsed.

The problem is when toggleable fieldset are nested one inside another.

 This css is wrong:

```
.p-fieldset-toggleable.p-fieldset-expanded .p-toggleable-content:not(.ng-animating) {
    overflow: visible;
}
```

Smorgaz:
The same error applies to panels, accordions and panelmenus and was introduced in commit f849e9d23b0c2223b6fbb516d3b0c368c927cf22.

majkers:
Any news on that? I can't migrate to Angular 14 because of that.

Vernon-Accela:
Seeing this on nested p-panels. I've had to write ng-deep css to override it until a fix can be delivered

majkers:
Might be fixed in 14.0.2 as https://github.com/primefaces/primeng/issues/11848 got closed

krtek4:
The fix for #11848 broke dropdowns inside accordions.

If a dropdown is bigger than the size of the accordion, its content isn't displayed entirely.

Instead of just removing the buggy css, something like that could be done :

```
.p-accordion-tab-active > .p-toggleable-content:not(.ng-animating) {
  overflow: visible;
}
```

rosenthalj:
> The fix for #11848 broke dropdowns inside accordions.
> 
> If a dropdown is bigger than the size of the accordion, its content isn't displayed entirely.
> 
[...]

PR #12146, opened 2022-11-08
Fix issue 11736 - Toggleable fieldset inside another toggleable field…

gerardva:
Fix for issue #11736. Similar to issue #11848 (with corresponding PR #12145). The expanded fieldset should only set the visibility of its direct children, not of any sub fieldsets.


cetincakiroglu:
Hi @gerardva,

Issue #12202, opened 2022-11-13
Chips: Content is not displaying using object as input and an item-template


user4me:
### Describe the bug

Hello,

By using the p-chips component I've been experienced one issue while using a an array object as input value and an item template. When I write  a new text and press enter, the chip displays without its content. 

### Environment

```
    <p-chips [(ngModel)]="inputObject">
        <ng-template let-item pTemplate="item">
            <img class="p-chips-img" [src]="item.imgUrl">
            <span>{{item.name}}</span>
        </ng-template>
    </p-chips>
```

### Angular version

14.2.0

### PrimeNG version

14.2.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.2.1

### Browser(s)

Chromium 107.0.5304.91

### Expected behavior

The chip should be displayed with its content according to the template

cetincakiroglu:
Hi @user4me,

Could you please share a Stackblitz exampleS by using https://stackblitz.com/github/primefaces/primeng-issue-template

PR #12203, opened 2022-11-13
#12202 : Make chips work better with object input value and item template

user4me:
Hello,

By using this component I've been experienced one issue while using a an array object as value and a item template according to #12202. 

That's what I've analyzed :
- The value is not checked correctly into the addItem, so duplicate case is not managed 
- The text value is not displayed correctly into the chip since we add it straight as string without taking care of the type of the input object

I suggest to be able to handle thoses cases by specifying the field into the existing @input field and map the object correctly

I've not tested the code yet but I think it would be a good start. I'll test this if your feedback is positive about the idea.

### DIFF
--- a/src/app/components/chips/chips.ts
+++ b/src/app/components/chips/chips.ts
@@ -192,6 +192,31 @@ export class Chips implements AfterContentInit, ControlValueAccessor {
         this.disabled = val;
         this.cd.markForCheck();
     }
+    
+    isUnique(item: string): boolean {
+        if(this.field) {
+            return this.value.map((element: any) => this.resolveFieldData(element, this.field)).indexOf(item) === -1;
+        }
+        return this.value.indexOf(item) === -1
+    }
+        
+    createFieldData(item: string, field: string): any {
+         if (item && field) {
+            const indexOf = field.indexOf('.');
+            if (indexOf == -1) {
+                const data = {};
+                data[field] = item;
+                return data;
+            } else {
+               const data = {};
+                const currentField = field.substring(0, field.indexOf("."));
+                const nextField = field.substring(field.indexOf(".") + 1, field.length);      
+                data[currentField] = this.createFieldData(item, nextField);
+            }
+        } else {
+            return null;
+        }
+    }
 
     resolveFieldData(data: any, field: string): any {
         if (data && field) {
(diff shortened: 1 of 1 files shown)


src/app/components/chips/chips.ts
```diff
@@ -192,6 +192,31 @@ export class Chips implements AfterContentInit, ControlValueAccessor {
         this.disabled = val;
         this.cd.markForCheck();
     }
+    
+    isUnique(item: string): boolean {
+        if(this.field) {
+            return this.value.map((element: any) => this.resolveFieldData(element, this.field)).indexOf(item) === -1;
```
volvachev, line 198:
```ts
if () {
  return !this.value.some((element) => this.resolveFieldData(element, this.field) === item)
}

return !this.value.includes(item)
```

src/app/components/chips/chips.ts
```diff
@@ -192,6 +192,31 @@ export class Chips implements AfterContentInit, ControlValueAccessor {
...
+            const indexOf = field.indexOf('.');
+            if (indexOf == -1) {
+                const data = {};
+                data[field] = item;
+                return data;
+            } else {
+               const data = {};
+                const currentField = field.substring(0, field.indexOf("."));
+                const nextField = field.substring(field.indexOf(".") + 1, field.length);      
+                data[currentField] = this.createFieldData(item, nextField);
```
volvachev, line 214: Do you need to return `data`?

src/app/components/chips/chips.ts
```diff
@@ -192,6 +192,31 @@ export class Chips implements AfterContentInit, ControlValueAccessor {
...
+    isUnique(item: string): boolean {
+        if(this.field) {
+            return this.value.map((element: any) => this.resolveFieldData(element, this.field)).indexOf(item) === -1;
+        }
+        return this.value.indexOf(item) === -1
+    }
+        
+    createFieldData(item: string, field: string): any {
+         if (item && field) {
+            const indexOf = field.indexOf('.');
```
volvachev, line 205: field.includes

src/app/components/chips/chips.ts
```diff
@@ -192,6 +192,31 @@ export class Chips implements AfterContentInit, ControlValueAccessor {
...
+            if (indexOf == -1) {
+                const data = {};
+                data[field] = item;
+                return data;
```
volvachev, line 209:
```ts
return {
  field: item
}
```

src/app/components/chips/chips.ts
```diff
@@ -192,6 +192,31 @@ export class Chips implements AfterContentInit, ControlValueAccessor {
...
+            } else {
+               const data = {};
+                const currentField = field.substring(0, field.indexOf("."));
```
volvachev, line 212: why don't you reuse the `indexOf` variable?

mertsincan:
Hi @user4me, 

Could you please change your PR based on @volvachev's comment? Then, I'll merge it asap. 

Special thanks to @volvachev for the review!
Best Regards,

Issue #12240, opened 2022-11-17
Component: Password


ArumugamSiva:
### Describe the bug

Consider creating a password 

Admin@123

there is an eye icon over there in that field, when it is striked it is showing up the password and when it is unstriked it is listing the password in dots, have attached the screenshot of the same

![image](https://user-images.githubusercontent.com/118423874/202370480-c86177b8-e312-410d-a40c-8c555f87e96c.png)
![image](https://user-images.githubusercontent.com/118423874/202370542-b411ae0a-4637-4dee-84d0-6cd7913d5630.png)
![image](https://user-images.githubusercontent.com/118423874/202370614-fa5e534f-2702-4067-b205-7e81597258dc.png)

Expected Result:

Actual Result
It is behaving upside down currently

### Environment

Good environment

### Angular version

14.x

### PrimeNG version

14.2.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

16.15

### Expected behavior

When the eye icon is striked it should show the password in dots and when it is not striked it should show the password typed password

cetincakiroglu:
Hi @ArumugamSiva,

This is not a bug, both examples are correct and it was made in this way **by design**.
Closing the issue since it's not a bug report, enhancement, or new feature request.

We appreciate your support!

PR #12252, opened 2022-11-18
Fix:#12240

Muthulakshmisankar:
#12240  Password component bug fixed


cetincakiroglu:
Hi @Muthulakshmisankar ,

This is not a bug, both examples are correct and it was made in this way by design.
Closing the PR since the issue is not a bug report, enhancement, or new feature request.

We appreciate your support!

Issue #12205, opened 2022-11-13
Badge Attribute - Size option


ProfEibe:
### Describe the feature you would like to see added

For the Badge-Directive option, it would also be great if there was a size-Attribute like the new disabled option.

### Is your feature request related to a problem?

Using Badge as component I can set the size attribute. Using the attribute version you can't do this.

### Describe the solution you'd like

Maybe like this?

`<i class="pi pi-bell" pBadge [value]="user.notifications" [badgeSize]="badgeSize"></i>`

### Additional context

#12170

PR #12267, opened 2022-11-21
feat(primeng/p-badge): add size attribute to badge directive

volvachev:
### Feature Requests
#12205

### DIFF
--- a/src/app/components/badge/badge.ts
+++ b/src/app/components/badge/badge.ts
@@ -5,6 +5,7 @@ import { DomHandler } from 'primeng/dom';
 import { UniqueComponentId } from 'primeng/utils';
 
 type BadgeDirectiveIconPosition = 'left' | 'right' | 'top' | 'bottom';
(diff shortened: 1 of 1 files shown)


src/app/components/badge/badge.ts
```diff
@@ -22,13 +23,26 @@ export class BadgeDirective implements AfterViewInit, OnDestroy {
         this._disabled = val;
     }
 
+    @Input('badgeSize') public get size(): BadgeSize {
```
cetincakiroglu, line 26: In p-badge component, we use "size" as input could you please rename badgeSize as "size" to make it unify for both directive and component?

cetincakiroglu:
Hi @volvachev,

Great PR, I'll merge it after you rename the "badgeSize" as "size".

src/app/components/badge/badge.ts
volvachev: Hi, I updated the size `input`.

cetincakiroglu (approved):

PR #12311, opened 2022-11-30
Add output to TabMenu on item selection

### LINKED ISSUE #12310: TabMenu: Tab selection is not actionable due to missing output

The `TabMenu` component does not expose any events or observables. This makes the tab selection fully opaque from the outside. When the active tab is changed within the component, the parent component does not know this. The only way to handle tab changes currently is the usage of the `command` property on the `MenuItem` but this seems cumbersome. It would be great if the `TabMenu` could simply emit an event when a tab is selected.

When trying to wrap the tab menu it is impossible to keep track of the selected tab in the wrapper component without using complex change detection and patching the `command` property on the menu items. This is frustrating considering the only missing piece is a single event.

Whenever a tab is selected an event is emitted with the `MenuItem` that was selected. Ideally this event also allows for two-way binding on the `activeItem` input (https://angular.io/guide/two-way-binding).

_No response_

### Additional context

_No response_

sbarfurth:
The tab menu currently does not let outside components know when the tab inside has changed. While it is possible to bind commands to the menu items, it could still be very useful to have the ability to react to a different tab being selected. The event emitter is named to support two way binding on the activeItem property.

Fixes #12310

### DIFF
--- a/src/app/components/tabmenu/tabmenu.ts
+++ b/src/app/components/tabmenu/tabmenu.ts
@@ -1,4 +1,22 @@
-import { NgModule, Component, Input, ContentChildren, QueryList, AfterContentInit, AfterViewInit, AfterViewChecked, TemplateRef, ChangeDetectionStrategy, ViewEncapsulation, ViewChild, ElementRef, ChangeDetectorRef, OnDestroy } from '@angular/core';
+import {
+    NgModule,
+    Component,
+    Input,
+    Output,
+    ContentChildren,
+    QueryList,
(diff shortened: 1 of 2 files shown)


cetincakiroglu:
Hi @sbarfurth,

Good catch a nice-to-have feature. We appreciate your support and effort!

cetincakiroglu (approved):

sbarfurth:
I ran prettier on the file, hopefully that fixes the issue.

PR #12367, opened 2022-12-13
new directive MovableSelector added to TableModule

shayan-ghzl:
**as if now table selection has three mode (single, multiple, movable).**

### movable mode
If you select a row, there is no unselect but you can move the selector by arrow keys 
in this mode there is no 'focus then select', but focus means select

### Why did I add this?
When tables are used in accounting and important calculations, the user needs to move faster on the rows and not lose the row if the focus is off the row.


shayan-ghzl (commented):
Ffgg

mertsincan:
Hi @shayan-ghzl,

Thanks a lot for your contribution! But, can you create an issue to get feedback from other users? According to the incoming demand, we can decide whether to add this feature or not into the core.

Best Regards,

mertsincan:
Hi @shayan-ghzl,

No problem, go ahead ;) Thanks a lot for your contribution.

Best Regards,

PR #12424, opened 2022-12-27
Fixed #12351 - Tree with virtualScroll: Inconsistent focus

### LINKED ISSUE #12351: Tree with virtualScroll: Inconsistent focus
### Describe the bug

I understand that this may just be the result of default DOM `TAB` key behavior, but in terms how `TAB`/`Up`/`Down` keys work in conjunction with virtual scroll its definitely inconsistent.

If using `[virtualScroll]="true"` having node focused on, and changing the rendered range via mouse `scroll` or `pgUp`/`pgDown` keys breaks focus away from the node. This is understandable. The focused node is no longer in the rendered range.
However, from user's perspective this is extremely confusing and illogical, that once the scroll passed some point, pressing `TAB` or `Down` key, behaves differently then before. Same problems observed with `TAB` + `SHIFT` and `Left` and `Right` keys

Open the link to the example and follow the steps I have posted in the "Steps to reproduce"

### Environment

Nothing special (check the linked example)

### Reproducer

https://stackblitz.com/edit/angular-ivy-bf4ajq

### Angular version

14.0.2

### PrimeNG version

14.0.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

not relevant

### Browser(s)

Tested in latest Firefox & Chrome (on MacOS 12.4)

1. Start without having a focus on the tree and pressing the `TAB` key:
-> `p-tree` gets focus
-> `up`/`down`/`pgUp`/`pgDown` keys work as expected: scrolling content by 1-2ish items or by page up/down

2. Press `TAB` again:
-> first item in the tree gets the focus
-> `down` key now moves the focus to next item in tree, up does the opposite, `pgUp`/`phDown` still shift visible area by page

   2.b. Clicking the node also puts focus on node, but this is not really relevant

3. Press `TAB`/`down` keys repeatedly:
-> focus will move to next node and so on...
-> **when focus is on last visible node**, pressing `TAB`/`down` key will **move focus to next** but also **shift visible area so newly focused node is visible**

4. 
     a. using `pgDown` / mouse `scroll` down  at least 3x the visible area away from last focused node, then press `TAB`:
    -> **it will not focus on the node next to previously focused node**
    -> instead it will focus on the (probably) first "rendered" node in range, which may be very far down the tree from the previously focused node
    = This is somewhat understandable, as if node is no longer in DOM, we cannot focus on it by calling `el.focus()`
<< It would however be much more convenient that on `TAB` key rendered range should be reset to previous one so the next node in tree can be focused -- for better user experience

    b. Attempt the same scenario as 4.a, just instead of pressing `TAB`, press `Down` key:
    -> The tree will shift down the visible area, losing the focus on previously focused node and not focusing on any other node
    = This I fail to understand
    << Using keys `Up`/`Down` should not lose the focus - for better user experience

Using keys `Left`/`Right` also works inconsistently, its just my example cannot show this (it was built for other purpose, sorry).

### Expected behavior

Using `TAB`, `TAB`+`SHIFT`, `Down` or `Up` key when having focus on a node, should always focus on the nearest node down/up, even in the case of virtual scroll enabled.

Using `Left` or `Right`` key when having focus on a node, should always expand/collapse that node (or collapse parent node if node already collapsed), even in case of virtual scroll enabled.

cetincakiroglu:
Fixes #12351

### DIFF
--- a/src/app/components/tree/tree.ts
+++ b/src/app/components/tree/tree.ts
@@ -667,8 +667,6 @@ export class UITreeNode implements OnInit {
     }
 })
 export class Tree implements OnInit, AfterContentInit, OnChanges, OnDestroy, BlockableUI {
-    @ViewChild('treeNode') treeNode: UITreeNode;
-
     @Input() value: TreeNode[];
 
     @Input() selectionMode: string;
@@ -1375,10 +1373,6 @@ export class Tree implements OnInit, AfterContentInit, OnChanges, OnDestroy, Blo
 
         if (this.dragStopSubscription) {
             this.dragStopSubscription.unsubscribe();
-        }
-
-        if (this.treeNode.timeout) {
-            clearTimeout(this.treeNode.timeout);
         }
     }
 }
--- a/src/app/components/tree/tree.ts
+++ b/src/app/components/tree/tree.ts
@@ -1,36 +1,33 @@
-import {
-    NgModule,
-    Component,
-    Input,
-    AfterContentInit,
-    OnDestroy,
-    Output,
-    EventEmitter,
-    OnInit,
-    OnChanges,
-    ContentChildren,
-    QueryList,
-    TemplateRef,
-    Inject,
-    ElementRef,
-    forwardRef,
-    ChangeDetectionStrategy,
-    SimpleChanges,
-    ViewEncapsulation,
-    ViewChild
-} from '@angular/core';
-import { Optional } from '@angular/core';
 import { CommonModule } from '@angular/common';
-import { PrimeNGConfig, TranslationKeys, TreeNode } from 'primeng/api';
-import { SharedModule } from 'primeng/api';
-import { PrimeTemplate } from 'primeng/api';
-import { TreeDragDropService } from 'primeng/api';
-import { Subscription } from 'rxjs';
-import { BlockableUI } from 'primeng/api';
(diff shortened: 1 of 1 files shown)


src/app/components/tree/tree.ts
```diff
@@ -67,6 +64,7 @@ import { Scroller, ScrollerModule, ScrollerOptions } from 'primeng/scroller';
                     [attr.aria-expanded]="this.node.expanded"
                     [attr.aria-selected]="isSelected()"
                     [attr.aria-label]="node.label"
+                    [attr.data-id]="node.data"
```
ivan-artkod, line 67: Unfortunately this is even according to your documentation a "no-go". `TreeNode.data` is of generic type `<T = any> `. You are here assuming its string.

In my projects my `T` is usually and object, and even then it is not Readonly object, but allows mutation.
This simply would not work for anything other that when data is unique string (ie. ID).

This is just not good approach for triggering focus on this element.
Much better solution would be to use `ElementRef` injector (host element) instance and target this `.p-treenode-content` div element via its `this.hostEl.nativeElement.querySelector('.p-treenode-content').focus()`

cetincakiroglu: refactored in https://github.com/primefaces/primeng/pull/12485

Issue #12467, opened 2023-01-07
Component: We cannot Able To Add Custom Icons And Custom Img


RAJEEV2510:
### Describe the feature you would like to see added

In Tree node There is no option for user can add change image-icon according to requirement .
I add a customImage property if user true that property user now can add custom Images too.

### Is your feature request related to a problem?

I have a probem ,we want use our own images to tree node

### Describe the solution you'd like

I added a customImage property in tree node , if user true that property user now can add custom Images too.
in node model just true this property and do node.icon=image-path

nodeModel={
....
,
customImage:true

}
and  and give node.icon =image-path;
![customImageProperty](https://user-images.githubusercontent.com/49456702/211133694-347192e4-7376-4ffa-8006-ce0daf7cceb3.png)

rayen-mhamdi:
we want this feature, can any one reopen the PR and merge it  ?

RAJEEV2510:
Hii primefaces team , 
i want this feature which was requested my pull request.
 please merge this feature and let others to add any icon or images

mehmetcetin01140:
Hi, 

Issue tracker is used for defects only as part of our commitment to quality and continuous improvement in all areas. Enhancements are collected as valuable community feedback and managed internally so moving this enhancement ticket to our internal project management backlog. For enhancement requests and other topics, you can open a new topic in [PrimeNG Discussions](https://github.com/orgs/primefaces/discussions/categories/primeng).

PR #12468, opened 2023-01-07
we can add now customimage too for tree node dynamically

RAJEEV2510:
### Defect Fixes
we cannot able to add custom image icons 
issue link [issue no 12467](https://github.com/primefaces/primeng/issues/12467)
### Feature Requests
I solve this issue by adding customImage Property now user can add image icon node by customImage property true and use it.
Steps to Add CustomImage Icon
1. customImage Icon :true in node Model
2. then just give image path to node-icon

customImageProperty


rayen-mhamdi:
we want to have this feature it's really useful.

PR #12500, opened 2023-01-14
Fix #12499 - Accordion Typescript improvements

### LINKED ISSUE #12499: Accordion: Low type checking and outdated docs
### Describe the bug

Some of the inputs, otputs and implementation of the Accordion are not properly typed with typescript or described in the docs. 

Example:
- activeIndex is not any, it is a number or number[]
- style is not any (as implemented) or string (as docs), it is an object
- selectedChange output is not any, it is a boolean

The docs does not describe all the outputs available in the components, like activeIndexChange.

Additionally, the accordion is using "which" to check the key pressed for accesibility, it is recommended to avoid which.

### Environment

Angular 14.x.x
Primeng 14.x.x

### Reproducer

_No response_

### Angular version

14.x.x

### PrimeNG version

14.x.x

### Build / Runtime

Angular CLI App

### Language

TypeScript

16.15.0

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

IvanPenyaHuguet:
Fixed #12499


src/app/components/accordion/accordion.ts
```diff
@@ -315,6 +313,10 @@ export class Accordion implements BlockableUI, AfterContentInit, OnDestroy {
             this.tabListSubscription.unsubscribe();
         }
     }
+
+    private isMultiple(selected: number | number[]): selected is number[] {
```
volvachev, line 317: Hi, is this type guard or what? I can't understand what this argument is for.

IvanPenyaHuguet: Yes, it is a type guard. With the actual implementation of the accordion the _activeIndex (or tabs selected) can be a number or number[] depending of the value of the multiple property. With this type guard typescript detects correctly the typing for the argument passed (the activeIndex).
Ideally typescript could detect conditionally the type of the activeIndex property depending of the multiple value in compiling time, but this not seems to be possible for now.

cetincakiroglu:
Hi @IvanPenyaHuguet,

Thanks a lot for your effort, this pr cannot be merged right now since we're updating documentation across all Prime UI Libraries. The mentioned update will also cover type fixes such as in this PR. I'll keep it open and label it as "documentation" so we can evaluate it when we update related components' documentation.

Issue #11955, opened 2022-09-20
p-table: scrollheight changed from setting height to max-height


ehaddan-jh:
### Describe the bug

The check-in for _'Fixed #16664 - Reimplement virtual scroll feature on Table'_ changed the div #wrapper from setting the '_height_', to setting the '_max-height_' with the _scrollHeight_ property, making my table heights shrink when there are not enough record entries.

### Environment

Windows 10
Chrome Version 105.0.5195.127

### Angular version

14.2.0

### PrimeNG version

14.0.2

### Build / Runtime

TypeScript

### Language

TypeScript

### Node version (for AoT issues node --version)

16.14.2

### Browser(s)

Chrome 105

thornidlo:
I think that's correct behaviour now. Height creates automatic fixed space even there is not enough rows. We don't want to see blank space, if there is only one row, when scrollHeight is set to bigger number. If you have another content below, e.g. another tables, buttons, then it's not so compact. But i understand, that someone want it. So, better choice then defect is improvement and create parametrized attribute scrollHeight or create new attribute maxScrollHeight. So scrollHeight could create height attribute and maxScrollHeight max-height attribute.

sergio-carlavilla:
Any workaround to fix this in modern versions?

ehaddan-jh:
The workaround I did was this in the ngAfterViewInit:

            let wrapperElements: HTMLCollection = document.getElementsByClassName('p-datatable-wrapper');
            Array.prototype.forEach.call(wrapperElements, function(wrapperElement) {
                let maxHeight = wrapperElement.style["max-height"];
                if (maxHeight) {
                    wrapperElement.style["height"] = maxHeight;
                }
            });

sergio-carlavilla:
> The workaround I did was this in the ngAfterViewInit:
> 
> ```
>         let wrapperElements: HTMLCollection = document.getElementsByClassName('p-datatable-wrapper');
>         Array.prototype.forEach.call(wrapperElements, function(wrapperElement) {
>             let maxHeight = wrapperElement.style["max-height"];
>             if (maxHeight) {
>                 wrapperElement.style["height"] = maxHeight;
>             }
>         });
> ```

Thanks, works!

I created this PR to try to solve the problem [12562](https://github.com/primefaces/primeng/pull/12562)

PR #12562, opened 2023-01-27
Add the posibility to add height to table

sergio-carlavilla:
This PR solves the problem of keeping a blank space when there are not enough rows in a table.

Fixed: [11955](https://github.com/primefaces/primeng/issues/11955)


sergio-carlavilla:
@cagataycivici any news with this? thanks

PR #12566, opened 2023-01-27
Fixed #12564 - Sidebar: onHide ist emitted twice

### LINKED ISSUE #12564: Sidebar: onHide ist emitted twice
### Describe the bug

When closing a sidebar the onHide EventEmitter is emitted twice:
1. Through the `close` function call
2. Through the `onAnimationEnd` function

### Environment

-

### Reproducer

https://github-2usv44.stackblitz.io

### Angular version

at least since 13

### PrimeNG version

at least since 13.0.4

### Build / Runtime

Angular CLI App

### Language

TypeScript

11.0.4

### Browser(s)

_No response_

1. Click on the button with the arrow
2. Close the sidebar by either clicking on on the background mask or the close button (x)

### Expected behavior

The onHide should only be emitted once.

cetincakiroglu:
Fixed #12564

### DIFF
--- a/src/app/components/sidebar/sidebar.ts
+++ b/src/app/components/sidebar/sidebar.ts
@@ -1,26 +1,26 @@
-import {
-    NgModule,
-    Component,
-    AfterViewInit,
-    OnDestroy,
-    Input,
-    Output,
-    EventEmitter,
-    ElementRef,
-    Renderer2,
-    ChangeDetectionStrategy,
-    ViewEncapsulation,
-    ContentChildren,
-    QueryList,
-    AfterContentInit,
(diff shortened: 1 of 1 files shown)


lf-novelt:
the "fix" seems to have broken the "onHide" output property.
it's not called anymore so nothing happens when not binding the visible property event listener

```
<p-sidebar [visible]="displayApprovalDialog !== null" position="right" [style]="{width: '30em'}"
  (onHide)="displayApprovalDialog  = null" [showCloseIcon]="true">
```

emit = false so it does not call it from the close event anymore. 
Image

The `onHide` action was the thing that was closing the sidebar for me

PR #12618, opened 2023-02-12
Unsubscribe subscriptions on component unmount to prevent memory leaks

Arooba-git:
Hello 👋 
As part of our project, we are using Facebook's new Memlab tool to detect memory leaks in SPA applications and their libraries. While running the tool and analyzing the code of primeng, we saw that your project does a very good job of ensuring that all async operations are cancelled when components destroy. However, as per Memlab execution results, we found some unsubscribed subscriptions, causing the memory to leak (screenshots below).

[before]
prime-before Screen Shot 2023-02-12 at 7 26 01 PM

Hence we added the fix by unsubscribing the subscriptions on component unload, and you can see the count of leaks reducing noticeably:
 <br />
prime-after-subscribe Screen Shot 2023-02-12 at 3 06 46 PM

Note. Host listener is used to ensure that ngOnDestroy is called not only when the class destroy, but also when the page refreshes, tab or browser closes or the user navigates away from the page [1]

We executed the unit test suite after the fixes and it passed TOTAL: 788 successfully (in and outside headless mode). 

You can analyze this and other potential leak sources, if you like, by running [Memlab](https://facebook.github.io/memlab/) with a scenario file covering maximum count of use cases.

Following is a sample of the scenario file we used (it needs to be a .js file but attaching here in .txt form):
[primeeng-scenario-memlab.txt](https://github.com/primefaces/primeng/files/10716015/primeeng-scenario-memlab.txt)

Note that some other reported leaks (in Memlab) originated from Angular's internal objects, and hence were ignored.

References:
[1] https://wesleygrimes.com/angular/2019/03/29/making-upgrades-to-angular-ngondestroy.html

### DIFF
--- a/src/app/components/confirmdialog/confirmdialog.ts
+++ b/src/app/components/confirmdialog/confirmdialog.ts
@@ -26,9 +26,11 @@ import { Footer, SharedModule, PrimeTemplate, PrimeNGConfig, TranslationKeys, Co
 import { ButtonModule } from 'primeng/button';
 import { Confirmation } from 'primeng/api';
 import { ConfirmationService } from 'primeng/api';
-import { Subscription } from 'rxjs';
+import { Subscription, Subject } from 'rxjs';
 import { UniqueComponentId, ZIndexUtils } from 'primeng/utils';
 import { RippleModule } from 'primeng/ripple';
+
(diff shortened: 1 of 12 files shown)


Arooba-git:
bump @cetincakiroglu

Issue #12624, opened 2023-02-13
AutoComplete: grouped options are not selected with enter key


erikhorlings:
### Describe the bug

When selecting an option in an option group in p-utoComplete using the enter key, the selection is lost upon tabbing out of the field. This only occurs with Enter key and with forceSelection is true.

### Environment

Angular 14.2.12, PrimeNg 14.2.2.

### Angular version

14.2.12

### PrimeNG version

14.2.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

v16.3.0

### Steps to reproduce the behavior

Create a p-autoComplete with forceSelection is true and at least one grouped option. Navigate to that selection by using the arrow buttons and select it with the enter key. Press tab to move focus to another field. The selected value is now cleared.

### Expected behavior

I expect the selected value to not be cleared.

erikhorlings:
The problem is in the method onInputChange of AutoComplete. I have a working implementation of that locally that proofs that it can made working by making this method deal with grouped options. I am willing to clean that implementation up and create a pull request for that, but that will take some time.

perotorun:
Still not fixed in 15.4.2

PR #12628, opened 2023-02-14
Take into account that options can be grouped when validating input.

erikhorlings:
### Defect Fixes
This fixes a problem with validation grouped options. AutoComplete.onInputChange now also looks at grouped options when deciding on validity of the selected option.

This fixes https://github.com/primefaces/primeng/issues/12624


erikhorlings:
This is still open. Problems continues to exist.

PR #12638, opened 2023-02-16
resolves #12637 -- uses proper group children property in dropdown autoselect

### LINKED ISSUE #12637: p-dropdown unable to auto-select first item if group used
### Describe the bug

When using p-dropdown with a grouped data set, inside of a reactive form, where `autoDisplayFirst` is true / omit, and `optionGroupChildren` is not the default value of `'items'`, the following error is logged in the browser console and no default item is selected:

`core.mjs:8506 ERROR TypeError: Cannot read properties of undefined (reading '0')
    at Dropdown.updateSelectedOption (primeng-dropdown.mjs:355:69)
    at Dropdown.ngOnInit (primeng-dropdown.mjs:215:14)
    at callHook (core.mjs:2488:22)
    at callHooks (core.mjs:2457:17)
    at executeInitAndCheckHooks (core.mjs:2408:9)
    at refreshView (core.mjs:10434:21)
    at refreshEmbeddedViews (core.mjs:11434:17)
    at refreshView (core.mjs:10443:9)
    at refreshEmbeddedViews (core.mjs:11434:17)
    at refreshView (core.mjs:10443:9)
core.mjs:8506 ERROR TypeError: Cannot read properties of undefined (reading '0')
    at Dropdown.updateSelectedOption (primeng-dropdown.mjs:355:69)
    at Dropdown.writeValue (primeng-dropdown.mjs:340:14)
    at setUpControl (forms.mjs:2941:23)
    at FormGroupDirective.addControl (forms.mjs:4745:9)
    at FormControlName._setUpControl (forms.mjs:5295:43)
    at FormControlName.ngOnChanges (forms.mjs:5240:18)
    at FormControlName.rememberChangeHistoryAndInvokeOnChangesHook (core.mjs:1574:14)
    at callHook (core.mjs:2498:18)
    at callHooks (core.mjs:2457:17)
    at executeInitAndCheckHooks (core.mjs:2408:9)
    `
    
it looks like the reason is the wrong property is accessed in this one spot in https://github.com/primefaces/primeng/blob/085a4e6b09e54ef6f1c5ae6f770bdc7fe724414c/src/app/components/dropdown/dropdown.ts#L776
    
all areas appear to use `getOptionGroupChildren()` to get the defined field except this one area is still hard coded to 'items'

I'm happy to provide a PR sometime later today to address if the issue can be confirmed. Stack blitz attached, changing the data from children -> items works but should not be required

### Environment

MacOS 12.0.1
Chrome 109

### Reproducer

https://stackblitz.com/edit/github-ixgaz2?file=src%2Fapp%2Fapp.component.html

### Angular version

15.1.0

### PrimeNG version

15.1.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.10.0

### Browser(s)

Chrome 109

1. create a simple form control or form group
2. create an options array with groups where the children property is not called 'items' but something like 'children'
3. create a p-dropdown and bind the options
4. set [group]="true"
5. set optionGroupChildren to the children field (again not called items)
6. set the optionGroupLabel and optionLabel if needed
7. bind the formControlName to the same one created in step 1
8. open the dropdown in the browser and see the error

### Expected behavior

No error should be logged when the component is using a default value for autoDisplayFirst and group=true and data does not use field `items` as the child property

The first selectable option within the first group should be selected probably, as works when it is called `items`

contra1337:
addresses https://github.com/primefaces/primeng/issues/12637
test added to cover this issue, confirmed it fails prior to fix

there may be a remaining bug if the first group does not have any children but it is out of scope for what i wanted to resolve here

### DIFF
--- a/src/app/components/dropdown/dropdown.spec.ts
+++ b/src/app/components/dropdown/dropdown.spec.ts
@@ -17,6 +17,7 @@ import { OverlayModule } from 'primeng/overlay';
         </p-dropdown>
         <p-dropdown [(ngModel)]="selectedCity"></p-dropdown>
         <button (click)="setValue()"></button>
+        <p-dropdown [(ngModel)]="selectedCity" [options]="groupedCarsAlternate" optionGroupChildren="children" [group]="true"></p-dropdown>
     `
 })
 class TestDropdownComponent {
@@ -52,6 +53,12 @@ class TestDropdownComponent {
         }
     ];
 
+    groupedCarsAlternate = this.groupedCars.map(city => ({
+      label: city.label,
+      value: city.value,
+      children: city.items
+    }));
+
     disabled: boolean;
 
     editable: boolean;
@@ -66,6 +73,7 @@ describe('Dropdown', () => {
     let dropdown: Dropdown;
     let testDropdown: Dropdown;
     let groupDropdown: Dropdown;
+    let alternateGroupDropdown: Dropdown;
     let fixture: ComponentFixture<Dropdown>;
     let groupFixture: ComponentFixture<TestDropdownComponent>;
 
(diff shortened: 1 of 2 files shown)


contra1337:
as there are no conflicts, I expect that this  is still relevant

cetincakiroglu (approved):

PR #12669, opened 2023-02-23
fix(primeng/p-tag): style binding not change dynamically

### LINKED ISSUE #12665: p-tag: style not change dynamically
### Describe the bug

Using p-tag style not change dynamically.

### Environment

Angular 14
Primeng 14

### Reproducer

https://stackblitz.com/edit/github-k27wqg?file=src/app/app.component.html

### Angular version

14

### PrimeNG version

14.0.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

18

### Browser(s)

_No response_

1. Select an option
2. Show correct color p-tag
3. Change option
4. P-tag Not change to new color

### Expected behavior

Update new color after change.

volvachev:
### Defect Fixes
Fixes #12665

### DIFF
--- a/src/app/components/tag/tag.ts
+++ b/src/app/components/tag/tag.ts
@@ -8,7 +8,7 @@ import { PrimeTemplate, SharedModule } from 'primeng/api';
 @Component({
     selector: 'p-tag',
     template: `
-        <span [ngClass]="containerClass()" [class]="styleClass" [ngStyle]="style">
+        <span [ngClass]="containerClass()" [class]="styleClass" [ngStyle]="tagStyle">
             <ng-content></ng-content>
             <ng-container *ngIf="!iconTemplate">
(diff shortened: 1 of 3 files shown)


mertsincan:
Hi @volvachev,

Why did you change the property name? I didn't fully understand it. In addition, when making such a change, instead of changing the old name directly, it would be more accurate to deprecate the old one and add a new property.

Best Regards,

volvachev:
> Hi @volvachev,
> 
> Why did you change the property name? I didn't fully understand it. In addition, when making such a change, instead of changing the old name directly, it would be more accurate to deprecate the old one and add a new property.
> 
> Best Regards,

Hi, because in this situation, there is a style binding process going on and `@Input` does not work.

cetincakiroglu:
Hi @volvachev,

Related issue is fixed in https://github.com/primefaces/primeng/commit/21efda75e61c0e436b747932303692efe9ba8ad7

PR #12730, opened 2023-03-09
Fixed #12706 - Added isPlaying method and fixed autoplay behaviour

### LINKED ISSUE #12706: Component: Carousel

At the moment the interval variable is still set after using clearInterval and therefor it is not that good to identify if the carousel is having a interval or not.

This ticket adds ability to get if the carousel is moving by the value "playing" or the method "isPlaying" updating its state by the startAutoplay and stopAutoplay Methods

_No response_

_No response_

_No response_

### Additional context

_No response_

EnricoMessall:
This merge requests add the isPlaying Method to the carousel and sets interval 100% to undefined when stopping carousel. Also when restarting the carousel or stopping it from other points outside the dots or pagings, the carousel had a broken behaviour due to wrong allowAutoplay States. Here the allowAutoplay states are now handled in these functions and so have always a correct autoplay behaviour.

### DIFF
--- a/src/app/components/button/button.ts
+++ b/src/app/components/button/button.ts
@@ -1,5 +1,5 @@
-import { CommonModule } from '@angular/common';
-import { AfterContentInit, AfterViewInit, ChangeDetectionStrategy, Component, ContentChildren, Directive, ElementRef, EventEmitter, Input, NgModule, OnDestroy, Output, QueryList, TemplateRef, ViewEncapsulation } from '@angular/core';
+import { CommonModule, DOCUMENT } from '@angular/common';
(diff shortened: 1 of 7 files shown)


src/app/components/carousel/carousel.ts
```diff
@@ -613,14 +609,23 @@ export class Carousel implements AfterContentInit {
...
             clearInterval(this.interval);
+            this.interval = undefined;
+            if(changeAllow){
+                this.allowAutoplay = false;
+            }
         }
     }
 
+    isPlaying(){
+        return this.interval;
```
volvachev, line 626: Hi, can you convert to `boolean`, because now it is a number or undefined, for understanding is not clear.

EnricoMessall: Done 👍

PR #12735, opened 2023-03-10
Fixes primefaces/primeng#12734 - Add Document Injection

### LINKED ISSUE #12734: Angular SSR Support - Multiple Components
### Describe the bug

Some components use the document variable. That results in some errors for Angular SSR Rendering and can be easily handled by using the @Inject for the Angular Document Variable what is already used in some other components.

### Environment

Basic Angular CLI App with PrimeNG and NG Universal

### Reproducer

_No response_

### Angular version

15.1.0

### PrimeNG version

15.1.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

16.16.0

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

EnricoMessall:
Fixed #12734


src/app/components/button/button.ts
```diff
@@ -81,7 +81,7 @@ export class ButtonDirective implements AfterViewInit, OnDestroy {
 
     private _internalClasses: string[] = Object.values(INTERNAL_BUTTON_CLASSES);
 
-    constructor(public el: ElementRef) {}
+    constructor(public el: ElementRef, @Inject(DOCUMENT) private document: Document) {}
```
volvachev, line 84: Hi, I don't understand how importing helps?
You must use a variable from the token, `this.document` instead of `document`(Inside the component and helpers instead of the global variable `document`)

EnricoMessall: Hey actually it should work because then Angular SSR then knows a variable named document from the constructor if no one is present with just using document. But your absolute right that using this.document is the more correct way. Updated it :D

cetincakiroglu:
Hi @EnricoMessall 

I've reviewed your pr, but can't merge due to conflicts, could you please resolve them and update the pr so we can merge safely?
Thanks for the effort and support to make PrimeNG better!

EnricoMessall:
Hey @cetincakiroglu 

i've merged the master branch into this ticket and also fixed a prettier warning on the landing page also. But there is one error in the Headless Test that shouldn't be from this ticket ^^

cetincakiroglu (approved):

PR #12827, opened 2023-03-28
fix: allow resolveFieldData to work with field names containing a '.'

### LINKED ISSUE #12609: Object Utils: resolveFieldData check if field name exists in data as key (even if contains a period)
### Describe the bug

We have a service that returns a flattened array (i.e. none of rows have any fields that are objects with nested properties).  However, this service in some scenarios will return a field key (or column name) with a period in it.  We don't really have control over this service in terms of what they return.

Could the `resolveFieldData` function below add a check above the ELSE of something like...

```
else if(data[field]) return data[field];
```

The splitting on the period is what is causing us to not return expected results.  We did try a workaround to replace periods with underscores after the data service returns but there are also cases where we do have nested object properties w/ other data services that this creates a mess.

https://github.com/primefaces/primeng/blob/95f68c4339c8ea5a73aa4adcfdf940075e63003e/src/app/components/utils/objectutils.ts#L54-L74

### Environment

Angular 14 app

### Reproducer

_No response_

### Angular version

14.1.1

### PrimeNG version

14.0.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

16.13.2

### Browser(s)

Edge

_No response_

### Expected behavior

If the data row contains a key that matches the propertyName (even if period in name) that it would return that result.

yelhouti:
fixes #12609 

### Defect Fixes
When submitting a PR, please also <ins>**create an issue**</ins> documenting the error and [manually link to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#manually-linking-a-pull-request-or-branch-to-an-issue-using-the-issue-sidebar) or mention it in the description using #<issue_id>.

### Feature Requests

### DIFF
--- a/src/app/components/utils/objectutils.spec.ts
+++ b/src/app/components/utils/objectutils.spec.ts
@@ -23,13 +23,15 @@ describe('ObjectUtils Suite', () => {
                     name: 'Italy'
                 },
                 city: 'Corleone'
-            }
+            },
+            'key.with.dot': 'value'
         };
 
         expect(ObjectUtils.resolveFieldData(obj, 'firstname')).toBe('Silvio');
         expect(ObjectUtils.resolveFieldData(obj, 'lastname')).toBe('Andolini');
         expect(ObjectUtils.resolveFieldData(obj, 'address.city')).toBe('Corleone');
         expect(ObjectUtils.resolveFieldData(obj, 'address.country.name')).toBe('Italy');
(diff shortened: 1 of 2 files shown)


yelhouti:
@cetincakiroglu hi there :) there is an issue with the test you changed on the 21st could you please fix it. All CIs are failing

Issue #12933, opened 2023-04-24
Component: OverlayPanel


Nawsen:
### Describe the bug

When an overlaypanel is open and then toggled there is a check to confirm if the event passed originated from a different target.
For example when you have a custom close button different from the button that has opened up the overlaypanel.

In this case the following code is executed:
`this.show(null, target || event.currentTarget || event.target);`

This all worked out until commit [f5129d8cc90475b239ae5e3412dbe9ed23de31f8](https://github.com/primefaces/primeng/commit/f5129d8cc90475b239ae5e3412dbe9ed23de31f8)

This commit adds `target && event.stopPropagation();` to the `show(event, target?)` function.

Looking back at the first line of code its clear this will always throw an exception. 
**ERROR Error: Cannot read properties of null (reading 'stopPropagation')**

### Environment

All

### Reproducer

https://stackblitz.com/edit/angular-urkuxt?file=src/main.ts

### Angular version

15.2.8

### PrimeNG version

15.4.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

16.19.0

### Browser(s)

all

### Steps to reproduce the behavior

In the stackblitz, open the popup with the button, then click the pink "custom toggle button". The overlay will close but can never be reopened. (see error in console)

### Expected behavior

Overlay should close without giving an error and it should be possible to re-open the same overlay.

cetincakiroglu:
Hi @Nawsen, 

Since there wasn't a pr linked to the issue, I've implemented your solution as a fix. Great catch!
Thank you for your effort and support!

PR #12934, opened 2023-04-24
Component: OverlayPanel #12933 - added null-check

Nawsen:


cetincakiroglu:
Hi,

Great catch! 
This solution is already applied, unfortunately, I didn't see your PR at the moment because it wasn't linked to the issue. Closing the PR.
Thanks for your support and effort!

Issue #12973, opened 2023-05-01
DropDown: open even if disabled


ikiliki:
### Describe the feature you would like to see added

I would like to be able to open the dropdown even if it is disabled.

### Describe the solution you'd like

Add a boolean input that allows this behavior

mbenrey:
The problem is even worse with the multiselect component in readonly mode when there are multiple selected options. 
Users may only see the beginning of the list because the dropdown might be too small, which prevents them from viewing their selections. 
To solve this issue, users should be able to open the dropdown in disable mode and view the options without making any changes.

![image](https://user-images.githubusercontent.com/40571116/235490741-e5707d1a-5ae3-4bab-8064-3b4196324aeb.png)

mertsincan:
Unfortunately, we can't add the disabled behavior @ikiliki requested. But, @mbenrey yes this may be a bug. Could you please create a Github issue? When HTML select element has readonly attribute, it opens its panel.

Best Regards,

PR #12975, opened 2023-05-01
Add input that allows the dropdown to be opened when disabled

ikiliki:
https://github.com/primefaces/primeng/issues/12973

This is the general idea.


mertsincan:
Unfortunately, we can't add this behavior. This does not match the behavior of the HTML select tag. You can control the behavior of the HTML select tag when disabled.

Best Regards,

PR #13016, opened 2023-05-10
Added type for defaultFocus:

zanella:
- To document the possible options
- To enforce at compile time the correct options

### DIFF
--- a/src/app/components/api/confirmation.ts
+++ b/src/app/components/api/confirmation.ts
@@ -1,5 +1,7 @@
 import { EventEmitter } from '@angular/core';
 
+export type ConfirmationDefaultFocus = 'accept' | 'reject' | 'close' | 'none';
+
 export interface Confirmation {
     message?: string;
     key?: string;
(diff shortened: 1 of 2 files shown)


mertsincan:
Hi @zanella,

Thanks a lot for your contribution! But, this PR is outdated. Could you please update it and create a Github issue? Then, I'll merge it asap.

Best Regards,

zanella:
Hey, upon looking at it again, I'm not sure the whole change makes sense:

- The `Confirmation` interface is used in more than one component, but the values these different components accept for `defaultFocus` vary
- One thing that seems wrong though is the [inline definition of values in confirmdialog](https://github.com/primefaces/primeng/blob/master/src/app/components/confirmdialog/confirmdialog.ts#L280)

It seems to [be missing 'none' ](https://github.com/primefaces/primeng/blob/master/src/app/components/confirmdialog/confirmdialog.ts#L516)

Assuming `backward compatibility default` is meant to be removed

So, all in all this MR would really be the addition of 'none' to the union on line 280, wdyt ?

mertsincan:
Yes, maybe you can add only 'none' value as a type defination to defaultFocus in a new PR. So, closed. A new PR will really be better. Thanks a lot for your contribution!

Best Regards,

PR #13035, opened 2023-05-12
Fix for confirmPopup setTimeout issues

### LINKED ISSUE #13034: p-confirmPopup errors when setTimeout runs after onContainerDestroy
### Describe the bug

In the p-confirmPopup, there are a few listeners that get set up in a setTimeout that gets called when the popup is first opened. If, for some reason, that setTimeout does not get called before the popup is closed, which sets `this.confirmation = null` on the component, the component throws a bunch of errors in places where it is expecting `this.confirmation` to be defined. 

The most common case we are seeing in our app is that `this.confirmation` is null when `bindScrollListener()` is called from the setTimeout.

### Environment

Versions below.

### Reproducer

_No response_

### Angular version

14.2.5

### PrimeNG version

14.1.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

16.14.0

### Browser(s)

Chrome and Firefox, didn't test the others

I was unable to reproduce this in a stackblitz, but I can reproduce in our app consistently. What happens is as follows:

1. User resizes a chart which causes it to redraw. 
2. User clicks on another chart, hits delete and the confirm popup comes up.
3. User hits enter and the confirm popup disappears.
4. When the first chart finishes redrawing, an error is thrown in `bindScrollListener()` in the confirm popup component, because `this.confirmation` is null. 

### Expected behavior

_No response_

jacks-bush:
Fix for errors thrown when the setTimeout queued in bindListeners() ends up running after the popup has been closed, and the confirmation on the confirmpopup component has already been set to null. Added null checks to the confirmation property.

Fixes #13034

### DIFF
--- a/src/app/components/confirmpopup/confirmpopup.ts
+++ b/src/app/components/confirmpopup/confirmpopup.ts
@@ -21,8 +21,8 @@ import { DomHandler, ConnectedOverlayScrollHandler } from 'primeng/dom';
             (@animation.done)="onAnimationEnd($event)"
         >
             <div #content class="p-confirm-popup-content">
-                <i [ngClass]="'p-confirm-popup-icon'" [class]="confirmation.icon" *ngIf="confirmation.icon"></i>
-                <span class="p-confirm-popup-message">{{ confirmation.message }}</span>
+                <i [ngClass]="'p-confirm-popup-icon'" [class]="confirmation?.icon" *ngIf="confirmation?.icon"></i>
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi @jacks-bush,

Could you please resolve conflicts so I can merge it safely?

Thanks for the support!

jacks-bush:
@cetincakiroglu,
Just resolved the conflicts. Thanks!

jacks-bush:
@cetincakiroglu,
Any update on when this will be merged?

Issue #13061, opened 2023-05-17
p-dialog: Provide a way to apply dynamic styles to p-dialog-mask


hekmatshak:
### Describe the bug

We need to have a way to specify a dynamic value for the top position for the p-dialog.

And I'm opening it as a bug because **positionTop** property was removed without providing any other alternatives.

Kindly find the proposed fix in the PR https://github.com/primefaces/primeng/pull/13060

**Related issues:**
https://github.com/primefaces/primeng/issues/9089
https://github.com/primefaces/primeng/issues/8937

### Environment

Angular version: 15.2.4

PrimeNG version: 15.2.0

### Angular version

15.2.4

### PrimeNG version

15.2.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

v16.18.0

### Browser(s)

Chrome

cetincakiroglu:
There is already an existing way to set p-dialog to the top https://primeng.org/dialog#position
Your pr contains a "maskStyle" value and it's not initialized or defined as an input property, so it'll be undefined no matter what and it does nothing.

If you're referring dynamicdialog, as in #8937 there is already an input property called "position" in dynamic dialog config.

```
export class DynamicDialogConfig<T = any> {
    data?: T;
    header?: string;
    footer?: string;
    width?: string;
    height?: string;
    closeOnEscape?: boolean;
    baseZIndex?: number;
    autoZIndex?: boolean;
    dismissableMask?: boolean;
    rtl?: boolean;
    style?: any;
    contentStyle?: any;
    styleClass?: string;
    transitionOptions?: string;
    closable?: boolean;
    showHeader?: boolean;
    modal?: boolean;
    maskStyleClass?: string;
    resizable?: boolean;
    draggable?: boolean;
    keepInViewport?: boolean;
    minX?: number;
    minY?: number;
    maximizable?: boolean;
    maximizeIcon?: string;
    minimizeIcon?: string;
    position?: string;
}

```
Closing the issue since the issue does not exist.

dmcclure-qcells:
Hello,

I am encountering an issue with the dialog position where it will not accept string values:
error TS2322: Type 'string' is not assignable to type '"center" | "top" | "bottom" | "left" | "right" | "topleft" | "toprigh
t" | "bottomleft" | "bottomright"'.

24               [position]="position">
                  ~~~~~~~~


Template code:
`    <p-dialog [(visible)]="isModalVisible"
              [modal]="true"
[...]

PR #13062, opened 2023-05-17
#13061 Provide a way to set p-dialog top position

hekmatshak:
https://github.com/primefaces/primeng/issues/13061


cetincakiroglu:
Hi,

Thanks for support and effort but it doesn't do anything since there is no input property named "maskStyle", it'll always be undefined and thus, it won't effect anything. Closing the PR.

Issue #12720, opened 2023-03-07
AutoFocus: Does not work inside dialog


Urtgard:
### Describe the bug

If AutoFocus is used inside a dialog, focus is not applied to the element.

### Environment

 Stackblitz

### Reproducer

https://stackblitz.com/edit/primeng-dialog-demo-u6rtzk?file=src%2Fapp%2Fapp.module.ts,src%2Fapp%2Fapp.component.spec.ts,src%2Fapp%2Fapp.component.html

### Angular version

15.2.1

### PrimeNG version

15.2.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

16.17.1

### Steps to reproduce the behavior

Use autofocus inside a dialog.

```html
<p-dialog [(visible)]="displayBasic">
  <input
    type="text"
    pInputText
    placeholder="Automatically focused"
    pAutoFocus
    [autofocus]="true"
  />
</p-dialog>
```

### Expected behavior

Dialog opens and focus is applied to the element.

nvkuznetsova:
I have the same issue inside dynamic dialog

jgaston5:
I believe that this issue is related to the fix for #12319. It nolonger looks for the autofocus element, but rather just grabs the first focusable item.

alvarosinmarca:
Same error with version primeng 16.0.2

cetincakiroglu:
This happens because autoFocus is applied on the initial load of the page rather than the component, that's how the autofocus attribute works in a webpage. That directive's sole purpose is to give autofocus property in the correct way, some html attributes need to be assigned in a different way because of virtual dom. In your example, autofocus does the job and focuses on the first element, as I mentioned it does not work in component-wise, component initialization is not the case with HTML autofocus attribute it only works on the initial load of the webpage.

Because of this, autofocus will not work when the component is initialized and that's why the input inside the dialog doesn't receive focus since autofocus is already applied at the initial load to a different element. Instead, you can implement it manually as below, please see the refactored example. 
Closing the issue since it's a misunderstanding of the autofocus attribute, thanks for reporting!

https://stackblitz.com/edit/primeng-dialog-demo-hruxst?file=src%2Fapp%2Fapp.module.ts,src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.html

alvarosinmarca:
[...]

Issue #12319, opened 2022-12-02
DynamicDialog need to steal focus


Foprta:
### Describe the bug

When you open DynamicDialog with button, focus remains on the button, so you can keep presing "Enter" to spam new modals.

### Environment

MacOS 13.0.1, MacBook Pro 2021 Late
Brave Browser 1.45.131

### Reproducer

https://www.primefaces.org/primeng/dynamicdialog

### Angular version

15

### PrimeNG version

15

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

16

### Browser(s)

Brave 1.45.131

### Steps to reproduce the behavior

1. Go to https://www.primefaces.org/primeng/dynamicdialog
2. Press "Show" button (it will open dialog)
3. Press "Enter" on keyboard (it will open dialog again)

### Expected behavior

No modals opened on "Enter" press.

cetincakiroglu:
Hi,

Good catch, thanks for reporting the issue!

mserege:
Hi,

This change breaks focus on first element with autofocus in dynamic dialog.

Would it be possible to get this fix without loosing autofocus ?

Thanks

PR #13113, opened 2023-05-29
autofocus by default dynamic dialog

brKlaus:
### Defect Fix
This PR is linked to issue #12720. This exists since the fix for #12319, as [autofocus] elements were no longer taken into account; only the first focusable element is considered now.


mertsincan:
Hi @brKlaus,

Best Regards,

Issue #13152, opened 2023-06-06
CascadeSelect: Wrong type definition for optionGroupChildren


VCourdy:
### Describe the bug

The type definition for the property "optionGroupChildren" of the component CascadeSelect is wrong on "16.0.0-rc.2".
Everything work just fine with "16.0.0-rc.1"

![image](https://github.com/primefaces/primeng/assets/15244542/f1ff5f73-1d70-4c71-8aea-12beaec4cda7)

it look like the issue is linked to tsconfig "importHelpers".

### Environment

WSL2 Debian 11
NodeJS 18.13.0

### Reproducer

https://stackblitz.com/edit/github-dmcdmy-zzaurf

### Angular version

16.0.0

### PrimeNG version

16.0.0-rc.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.13.0

cholasimmons:
yea I'm also trying to use a custom interface but I'm being told only `string[] | string | undefined` are allowed

emeryao:
+1

still have the same issue here

cholasimmons:
Until it's fixed, you can Ctrl+click (on Windows) on `options` and change `string[]` to `any`.

YMingF:
no one fix this issue????

PR #13153, opened 2023-06-06
fix type

VCourdy:
PR for #13152


chunfeng-october:
hi 16.0.2  The problem has not been solved yet, please tell me when this problem will be solved

PR #13160, opened 2023-06-07
Clarify how the FileUpload accept property works

Michael-Ziluck:
This is simply a documentation clarification so I did not create an issue. Let me know if one must be created regardless.


cetincakiroglu:
Hi,

Closing the request because the file does not exist anymore. All the manual API documentation files are removed after the types are added and the documentation is updated.
I've updated the description with yours in here https://github.com/primefaces/primeng/commit/213867464857800404248e20d1121be8e121f813, thanks for reporting the issue!

Issue #13165, opened 2023-06-09
asd


SoyDiego:
### Describe the feature you would like to see added

asd

### Is your feature request related to a problem?

asd

### Describe the solution you'd like

asd

### Describe alternatives you have considered

asd

### Additional context

dsf

SoyDiego:
I already proposed a pull request here: https://github.com/primefaces/primeng/pull/13167.

PR #13167, opened 2023-06-09
Add aria label on menu

SoyDiego:
This change was implemented to add an optional arial-label on Menu Component Items (for **Accessibility**).

It's neccessary to ensure that elements, labels and names correspond with accessible names like aria-label so that users using dictation software are able to access them.

More info here: https://www.w3.org/WAI/WCAG21/Techniques/failures/F96

The relevant issue is available here: https://github.com/primefaces/primeng/issues/13165

### DIFF
--- a/src/app/components/api/menuitem.ts
+++ b/src/app/components/api/menuitem.ts
@@ -9,6 +9,10 @@ export interface MenuItem {
      */
     label?: string;
     /**
+     * aria-label of the item.
+     */
+    ariaLabel?: string;
+    /**
      * Icon of the item.
      */
     icon?: string;
--- a/src/app/components/menu/menu.ts
+++ b/src/app/components/menu/menu.ts
@@ -280,7 +280,7 @@ export class Menu implements OnDestroy {
         private cd: ChangeDetectorRef,
         public config: PrimeNGConfig,
         public overlayService: OverlayService
-    ) { }
+    ) {}
 
     toggle(event: Event) {
         if (this.visible) this.hide();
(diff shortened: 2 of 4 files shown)


Marcosrvcks:
Solving this problem would save me a lot of work to add the aria-label in the MenuItems! thank you very much Diego!

cetincakiroglu:
Hi @SoyDiego,

Your PR is not wrong, but we're following the accessibility implementation in the PrimeVue, the aria-label and aria-labelledby properties should be added to list items instead of anchor tags because the anchor tags will have aria-hidden. I've reworked the menu component and added them here, it'll be available on the 16.0.3 release. https://github.com/primefaces/primeng/commit/cd1f6e5ca0a5e57618382806912ba23e95d07d7d

We appreciate your effort and support, but if you want to contribute to accessibility please check the PrimeVue as an example.

Closing the PR, thanks for your contribution!

SoyDiego:
> Hi @SoyDiego,
> 
> Your PR is not wrong, but we're following the accessibility implementation in the PrimeVue, the aria-label and aria-labelledby properties should be added to list items instead of anchor tags because the anchor tags will have aria-hidden. I've reworked the menu component and added them here, it'll be available on the 16.0.3 release. [cd1f6e5](https://github.com/primefaces/primeng/commit/cd1f6e5ca0a5e57618382806912ba23e95d07d7d)
> 
> We appreciate your effort and support, but if you want to contribute to accessibility please check the PrimeVue as an example.
> 
> Closing the PR, thanks for your contribution!

No problem @cetincakiroglu thanks for the explanation and I'm glad If I can help. Was not in this case but I continue learning a lot doing all this things.
Maybe is a good idea add a `CONTRUBUTING.md` file and explain some rules like you wrote me in your message about PrimeVue and others that your Team consider to know because can help to the users and your Team when someone create a PR or try to fix some issues.

Thanks again and have a nice day!

PR #13193, opened 2023-06-13
fix(scroller): track all visible pages

### LINKED ISSUE #13192: Scroller: Issues loading all visible rows when scrolling quickly to page boundaries
### Describe the bug

If the scrollbar is dragged quickly down with virtual scroll enabled and the viewport happens to land on a page boundary, the scroller will assume the previous/top page is already loaded and only emit a lazy load event with start/size for the lower page, and the rows above will remain loading.  This behavior can also be reproduced if you scroll to the bottom, then drag the scrollbar quickly up to a page boundary.  In some cases, you can even scroll up slowly and the previous page won't be triggered to load until you've scrolled all loaded rows off-screen.

### Environment

Windows (both native and WSL2) and Mac, all browsers. 

### Reproducer

https://stackblitz.com/edit/9uttel?file=src%2Fapp%2Fdemo%2Ftable-virtual-scroll-lazy-demo.ts,src%2Fapp%2Fdemo%2Ftable-virtual-scroll-lazy-demo.scss,src%2Fapp%2Fdemo%2Ftable-virtual-scroll-lazy-demo.html

### Angular version

15.2.8

### PrimeNG version

15.4.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

16.17.0

### Browser(s)

_No response_

The linked StackBlitz automates the scroll position in a timeout to ensure reproduction every time.  Just open the link, wait a second, and note that the top rows never load.  There's an additional `setTimeout` that can be uncommented to demonstrate this can also happen when scrolling up.

Manual repro steps:

1. Have a virtual scroll table with 3+ pages of data.  For the purpose of this example, let's say there are 10 pages, and page size is 100.
2. Drag the scroll bar quickly so that it positions row 300 exactly in the middle of the viewport.
3. Note that a lazy load emits with `{ start: 300, size 100 }`, and the bottom half of the viewport loads.
4. Note that the top half of the viewport still shows the loadingbody.

### Expected behavior

Instead of tracking a single page, the scroller needs to be aware that multiple pages can be in view at once and emit lazy load events accordingly.  So if I scroll down halfway and land on a page boundary, the lazy load event should indicate via start/size that both pages need to be loaded, e.g. by having `start` set to the first index of the upper page, and `size` doubled to account for both.

MikeMatusz:
Fixes #13192

Track all visible pages instead of just a single visible page, and emit events accordingly when two unloaded pages are scrolled into view at the same time and both need to be loaded.


MikeMatusz:
I have run `npm run format` and `npm run format:check` passes locally.  I haven't touched the file it's complaining about (and the former command changed a contextmenu file I also did not touch).  Can someone else run the formatter and push changes?  Without any details on what the violations are I can't tell what's going on.  I'm on a Mac, if that makes a difference.

Also, I have tested these changes extensively in our application via monkey-patching, so I have high confidence that this resolves the behavior in the linked issue.

mertsincan:
Hi @MikeMatusz,

Thanks a lot for your contribution! But, this PR has conflicts. Could you please update it? Then, I'll check and merge it asap.

Best Regards,

MikeMatusz:
@mertsincan Of course it was the file touched by running format that conflicted, the PR has been updated.

MikeMatusz:
@mertsincan Would it be possible for someone on your end to run the formatter?  Should only take a minute.  I've done it but I think it's broken on MacOS, and I feel like the failing check is why this has been de-prioritized.  We almost had to disable virtualization because of the issues this is addressing, we're currently running with this locally as a monkey patch, but that's potentially fragile when we go to upgrade.

MikeMatusz:
@mertsincan Any updates on this?

cetincakiroglu:
Hi,

Couldn't replicate the issue with latest version (17.12.0). Issue is probably fixed by another commit between mentioned version and the latest. Closing the PR since it doesn't exist anymore. Please open a new issue and PR with a stackblitz reproducer if the issue persists.

cravay:
Hi @mertsincan, @cetincakiroglu,

Please reconsider the closing of this PR and the corresponding issue #13192.
We're affected by this bug and we've been using [patch-package](https://www.npmjs.com/package/patch-package) to apply the fix provided in this PR.

Here's the original reproducer with Angular updated to version 17 and PrimeNG version 17.18.0 (notice how half of the table isn't loading after scrolling down): https://stackblitz.com/edit/9uttel-6pheqg?file=src%2Fapp%2Fdemo%2Ftable-virtual-scroll-lazy-demo.ts

Here's the same reproducer but with the fix from this PR applied (the table is fully loaded after scrolling): https://stackblitz.com/edit/9uttel-ik3wnd?file=patches%2Fprimeng%2B17.18.0.patch,src%2Fapp%2Fdemo%2Ftable-virtual-scroll-lazy-demo.ts

I'd *really* appreciate it if you could reopen and review this PR.

PR #13199, opened 2023-06-14
fix(primeng/p-button): classes for custom icon template, normalize height

### LINKED ISSUE #13139: p-button: style error, when using custom icon
### Describe the bug

I am doing a proof of concept and comparing primeng and primevue, I have created a project in vue 3 and the latest version of primevue, now i am replicating the same project in angular 16 and the latest version of primeng, I have discovered that there is an error when using the p-button component with a custom icon imported from material icons, This problem does not occur with primevue, and comparing the code generated from the two projects I have detected that the error is that primeng does not correctly apply the style p-button-icon -only

This is the vue code where you can see that I create buttons and a template is passed with the icon, to be rendered in the slot

image

The generated code shows how a button element is created that has the style p-button and p-button-icon-only assigned, which sets the width to auto

image

replicating the same component in angular using primeng you can see that nothing is used apart from the component and material icon

image

In the generated code you can see how a button is created without the style that I mentioned before, this causes the button to be bigger, leaving an empty space next to the icon, also it is observed that the button with a custom icon is much bigger than a button using prime incon

image

without the p-button-icon-only style the button width is set to 4 rem so there is a space next to the icon, with the added style the width is auto and it fits the icon, although it is still bigger than a button with prime icon

### Environment

local computer with windows 11 16gb ram, ryzen processor
Chrome browser latest version

### Reproducer

_No response_

### Angular version

16.0.3

### PrimeNG version

16.0.0-rc.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

20.2.0

### Browser(s)

chrome 113

_No response_

### Expected behavior

_No response_

volvachev:
### Defect Fixes
Fixes #13139

### DIFF
--- a/src/app/components/button/button.ts
+++ b/src/app/components/button/button.ts
@@ -368,11 +368,11 @@ export class Button implements AfterContentInit {
     buttonClass() {
         return {
             'p-button p-component': true,
-            'p-button-icon-only': this.icon && !this.label,
+            'p-button-icon-only': (this.icon || !!this.iconTemplate) && !this.label,
             'p-button-vertical': (this.iconPos === 'top' || this.iconPos === 'bottom') && this.label,
             'p-disabled': this.disabled || this.loading,
             'p-button-loading': this.loading,
-            'p-button-loading-label-only': this.loading && !this.icon && this.label
+            'p-button-loading-label-only': this.loading && !(this.icon && !this.iconTemplate) && this.label
         };
     }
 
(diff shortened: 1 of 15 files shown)


mertsincan:
Hi @volvachev,

Thanks a lot for your contribution! But, this PR has conflicts. Could you please update it? Also, we use [primeng-sass-theme](https://github.com/primefaces/primeng-sass-theme) to provide theme.css files. Could you please create a Github issue and PR in it?

Best Regards,

mertsincan:
Hi,

 This PR is outdated and there doesn't seem to be any interaction on it. So, closed. Could you please create a new PR for us? Thanks a lot for your understanding and contribution!

Best Regards,

Issue #13268, opened 2023-06-27
The `breakpoints`options can be ignored when the dialog is maximized.


joreyaesh:
### Describe the bug

There is a bug that makes `maximizable` dialogs unable to be to be maximized when `breakpoints` are set.

### Environment

MacOS

### Reproducer

https://stackblitz.com/edit/github-rkg6v1?file=src%2Fapp%2Fapp.component.html

### Angular version

15.2.9

### PrimeNG version

15.4.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

14.21.3

### Browser(s)

Firefox Developer Edition 115.0b9

### Steps to reproduce the behavior

1. Click maximize button
2. Note that it's not maximized (width should be 100vw, but will be 85vw from the breakpoints).

### Expected behavior

A maximized dialog

mertsincan:
Fixed in https://github.com/primefaces/primeng/pull/13269

PR #13269, opened 2023-06-27
fix(dialog): maximizable responsive dialog

joreyaesh:
Fixed an issue where dialogs with `breakpoints` set were unable to successfully be maximized using the maximize button when `[maximizable]="true"`.

 #13268


SoyDiego (approved):

mertsincan:

Best Regards,

Issue #12983, opened 2023-05-03
Autocomplete: selectedItem template only works if multiple="true"


Kev1n337:
### Describe the bug

According to the documentation, there should be a `selectedItem`-Template, that is used for displaying the selected object from an autocomplete input. Currently this template is ignored and instead the [field]-attribute is used. This does not allow to specify a complex template when the `suggestions` are complex objects.

This issue has also been reported in #2242 
@mertsincan said a new ticket should be opened if the issue still persists.

### Environment

Current version of angular and primeng, happening in all browsers.

### Reproducer

https://github-cnxtrn.stackblitz.io

### Angular version

15.2.8

### PrimeNG version

15.4.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.16.0

### Steps to reproduce the behavior

1. Open Stackblitz or use this snippet in your application:

```html
<p-autoComplete
  [(ngModel)]="customer"
  (completeMethod)="searchCustomers($event)"
  [suggestions]="customerSuggestions"
  [forceSelection]="true"
  field="last_name"
>
  <ng-template pTemplate="selectedItem" let-customer>
    <div>{{ customer.first_name }} {{ customer.last_name }}</div>
  </ng-template>

  <ng-template let-customer pTemplate="item">
    <div>{{ customer.first_name }} {{ customer.last_name }}</div>
  </ng-template>
</p-autoComplete>
```

2. See the input value of a selected item being "Doe" (from 'field') instead of "Jane Doe" (from 'selectedItem')

### Expected behavior

When the selectedItem is specified it should be used instead of the field-attribute.

denny99:
Yeah selected item only work in multiple mode...

In single mode the input is field with the selected field

ThoSap:
Duplicate of #2242 (opened in 2017) and #10360 (2021) but these issues were somehow closed.
@mertsincan @cetincakiroglu

straiforos:
This is very important to my project, and it is heartbreaking to see this has been ignored and closed out many times.
BUMPing the thread.

ThoSap:
@cetincakiroglu

bluesbroz:
You can use `optionLabel` as a function to make a custom value.
```
/**
 * Property name or getter function to use as the label of an option.
 * @group Props
 */
@Input() optionLabel: string | ((item: any) => string) | undefined;
```

`field` must be omitted in this case, otherwise it will be used instead.
```
[...]

PR #13284, opened 2023-07-01
Fix p-autoComplete: selected item template is not displayed

### LINKED ISSUE #2242: Template for selected item in autocomplete (single selection mode)
```
```

**Current behavior**

It's already possible to define an "item" template for the suggestions of the autocomplete, but this template is not used for the selected item. The selected item only looks at the "field" attribute, so it can only display 1 property of the object to which the ngModel is bound.

It's also possible to define a "selectedItem" template, but this template only seems to work in multiple selection mode (as stated in the documentation).

I've tried this example:

    <p-autoComplete [(ngModel)]="customer" [suggestions]="customerResults" (completeMethod)="searchCustomers($event)" field="name">
      <template let-customer pTemplate="item">
        {{customer.name}} {{customer.firstname}}
      </template>
      <template let-cusstomer pTemplate="selectedItem">
        {{customer.name}} {{customer.firstname}}
      </template>
    </p-autoComplete>

With this code my selected item only shows the "name" property of my customer-object. If I don't provide the field-property, my selected item is displayed as [object Object].

**Expected behavior**
I expect it to be possible to define a template for the selected item when the autocomplete is in single selection mode. It seems like I'm missing something as no one else seems to be asking this question, but I've tried everything I could find in the documentation, and nothing seemed to work.

* **Angular version:** ~2.4.0

* **PrimeNG version:** 2.0.1

* **Browser:** [all ]

Tuxio:
### Defect Fixes
Fixes  #2242 #12983


mertsincan:
@cetincakiroglu Could you please check this PR?

Best Regards,

pjlasl:
Any update on this?

LiolluryZ:
up

cetincakiroglu:
Hi @Tuxio ,

Thank you for taking the time to submit your pull request. We genuinely appreciate your effort and interest in improving our project.

After careful review and consideration, we have decided not to merge your pull request at this time. Please understand that this decision does not reflect negatively on the quality of your work. We value every contribution and understand the effort you've put into it.

However, upon review, we believe that the changes proposed in this pull request may not align with the current goals or direction of the project. Our team is continuously evaluating and prioritizing features and enhancements based on various factors, including project roadmap, user feedback, and technical considerations. selectedItem template support will be implemented for singular input mode also in the future.

We encourage you to continue contributing to the project and to stay engaged with our community. Your input and contributions are valuable, and we look forward to seeing more of your work in the future.

Once again, thank you for your contribution, and please don't hesitate to reach out if you have any questions or would like further feedback.

Best regards,

uebelmar:
@cetincakiroglu the selectedItem Template doesnt work since several major versions. Here someone does the work and fixes it and it is ignored.

cetincakiroglu:
@uebelmar,

selectedItem template is only supported for multiple mode. It's not broken feature, it's not supported in single mode. We've closed the PR because:

1-) Styling becomes broken, and unnecessary borders and outlines can be seen on focus.
2-) Styles should be applied in primeng-sass-theme not in component CSS. Padding and margin values change depending on the applied theme. In this way, they will always be the same.
3-) We can't apply min-width to the styles.
4-) Dom structure changes and it's a breaking change.
5-) Also even the example doesn't work correctly. See the image, you can also see that weird size difference in second image.
6-) It breaks accessibility, overlay shouldn't open by down arrow, also overlay cannot be opened by search, it's a weird behavior.
7-) In the third image, you can enter input simultaneously while the selection is also rendered, which is the default behavior of multiple mode, it doesn't function correctly.

To conclude, if it's the use case, then just use multiple mode with

So it's not ignored, it's reviewed and closed. We can't merge a feature that doesn't function correctly.

Screenshot 2024-03-28 at 09 21 25
Screenshot 2024-03-28 at 09 22 08
Screenshot 2024-03-28 at 09 25 53

cetincakiroglu:
> @cetincakiroglu
> 
> Thank you for taking the time to explain. That of course makes sense. I also saw in another issue that you have support for selectedItem in single mode on the feature list.
> 
> Another thing to think about is that the documentation for the selectedItem template mentions that it is only supported in single mode. In the last few years there have been a few Github issues and StackOverflow articles that have come across this.

Yes we'll update the docs to indicate that it's not supported in single mode

PR #13349, opened 2023-07-19
[#13348] - Fixed calendar component, added yearOptions onInit

### LINKED ISSUE #13348: p-calendar: Cannot read properties of undefined (reading 'length')
### Describe the bug

I've reduced this down to its core repro:

```
<p-calendar
view="month"
dateFormat="mm/yy"
[yearNavigator]="true"
></p-calendar>
```

Try changing the month using the upper left/right arrows.

### Environment

Angular / PrimeNG 16

### Reproducer

https://stackblitz.com/edit/github-wh3ibs

### Angular version

16.0.0

### PrimeNG version

16.0.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.16.0

### Browser(s)

_No response_

1. https://stackblitz.com/edit/github-wh3ibs
2. Change the month using either arrow at the top

### Expected behavior

The month should change without throwing a JavaScript error.

SoyDiego:
Fix #13348.

The problem was caused by the `yearOptions` property not being initialized. I added the initialization of `yearOptions` to resolve the error. Now the component works as expected without any console errors.

# # PROBLEM
problem calendar

## SOLVED
fixed calendar


src/app/components/calendar/calendar.ts
```diff
@@ -1026,6 +1026,7 @@ export class Calendar implements OnInit, OnDestroy, ControlValueAccessor {
         this.createResponsiveStyle();
         this.currentMonth = date.getMonth();
         this.currentYear = date.getFullYear();
+        this.yearOptions = [this.currentYear, this.currentYear + 1, this.currentYear + 2];
```
gucal, line 1029: Hi, What is the reason why you think the `yearOptions` value should start this way?

SoyDiego: > Hi, What is the reason why you think the `yearOptions` value should start this way? @SoyDiego

Hi @gucal, thanks for the double check.
I don't remember exactly why I did that because was 1 month ago, but I was checking now and I

gucal: Thank you for making the changes.

Thanks for contribution.

Issue #13358, opened 2023-07-21
Component: p-password has incorrect width


MAN-Sendance:
### Describe the bug

The password component has a wrong width due to missing width inheritance in the child input field.

![incorrect_password_width](https://github.com/primefaces/primeng/assets/129411192/e4ee7791-efdd-42e3-a4cc-c49ddfa5809a)

### Environment

Angular  (latest version)
PrimeNG (latest version)
Primeflex (latest version)

### Angular version

16.1.4

### PrimeNG version

16.0.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.16.0

### Steps to reproduce the behavior

```
<label for="email1" class="block text-900 font-medium mb-2"
  >Email</label
>
<input
  pInputText
  id="email1"
  type="text"
  placeholder="Email address"
  class="w-full mb-3"
/>

<label for="password1" class="block text-900 font-medium mb-2"
  >Password</label
>
<p-password
  [feedback]="true"
  [toggleMask]="true"
  id="password1"
  placeholder="Password"
  class="w-full mb-3 block"
/>
```

### Expected behavior

Applying the following styles fixed the issue for me

```
.p-password {
  width: inherit;
}
.p-password-input {
  width: inherit;
}
```

calumk:
This is still an issue for me 

<img width="634" alt="image" src="https://github.com/primefaces/primeng/assets/1183991/909ab346-870e-48fd-88c9-9adaaf8c90f3">

spezifant:
To make it work I had to set both: `style` and `inputStyle`:
```
[inputStyle]="{'width': '100%'}" [style]="{'width': '100%'}" 
```

AutoAwesome:
For me this is still an issue. The fixed above worked though.

molnarerwin:
I think you just need to use `fluid`. 

For me using `Vue` this works:

```vue
<Password
  id="password"
  name="password"
  placeholder="Password"
  class="mt-2"
  toggle-mask
  fluid
></Password>
```

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation and try the latest PrimeNG version(v19). If there is no improvement on this, can you open a new issue so we can include it in our roadmap?

Thanks a lot for your understanding!
Best Regards,

PR #13418, opened 2023-08-02
Added inherit width for password directive

shreyashrajiokalaskar:
Inherited width from parent for password directive

#13358

### DIFF
--- a/src/app/components/inputtext/inputtext.css
+++ b/src/app/components/inputtext/inputtext.css
@@ -1,5 +1,6 @@
 .p-inputtext {
     margin: 0;
+    width: inherit;
 }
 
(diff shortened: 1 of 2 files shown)


SoyDiego (commented):
All refered to styles should be in primeng-sass-theme project:
https://github.com/primefaces/primeng-sass-theme/

mertsincan:
Hi @shreyashrajiokalaskar,

Why do you use the p-password tag 'width: 100%;' you added. The password component is 'display: inline-flex'. When you add this, you break the structure. It would be better if you just work on the password and send a new PR. So, closed.

Thanks a lot for your understanding and effort! We'll wait for a new PR from you. 
Best Regards,

PR #13423, opened 2023-08-02
Bugfix/#13421 tabview accessibility fixes

### LINKED ISSUE #13421: Accessibility | TabView - TabIndex order doesn't follow the W3C recommended pattern
### Describe the bug

As stated in the [W3C Authoring Practices Guide](https://www.w3.org/WAI/ARIA/apg/), for a Tab interface the `tab` button should behave as follow: 
- When focus moves into the tab list, places focus on the active tab element.
- When the tab list contains the focus, moves focus to the next element in the page tab sequence outside the tablist, which is the tabpanel unless the first element containing meaningful content inside the tabpanel is focusable.

Check: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#keyboardinteraction for reference. 

To be compliant with WCAG, all the patterns should be applied.

### Environment

Windows

### Reproducer

https://primeng.org/tabview

### Angular version

16.X

### PrimeNG version

16.1.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

16.20

### Browser(s)

_No response_

- Set focus to a tab header in any of the examples
- Press `tab`
- The focus moves to the next tab header

### Expected behavior

Pressing `tab`, the focus should move inside the content of the selected tab, if focusable, or to the next focusable element of the page.

GiacomoDM:
Fixes #13421

I have set the `tabindex=-1` to restore correct arrows navigation. Setting `null` is not a valid value for `tabindex` and might not work with all assistive technologies: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex and https://html.spec.whatwg.org/multipage/interaction.html#attr-tabindex


cetincakiroglu:
Hi @GiacomoDM,

There is one problem with your PR, If I have 3 items and set `[scrollable]="true"`, the arrow button should be visible if I resize the screen to the point of scroll needed. With your solution, arrow button stays invisible and it breaks the scrollable feature. So could you please refactor the PR and remove that line, other changes are okay and I'll merge them after the refactor.

For the arrow issue, we're currently working on it to provide a solution without affecting performance.

GiacomoDM:
done, thanks

PR #12723, opened 2023-03-08
Multiselect switch between filter & items with arrows

umuero:
### Defect Fixes
When submitting a PR, please also <ins>**create an issue**</ins> documenting the error and [manually link to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#manually-linking-a-pull-request-or-branch-to-an-issue-using-the-issue-sidebar) or mention it in the description using #<issue_id>.

### Feature Requests
Due to company policy, we are unable to accept feature request PRs with significant changes as such cases has to be implemented by our team following our own processes.
Smaller scaled feature implementations such as adding a property to a component will be considered for merging.

PR #13427, opened 2023-08-03
Multiselect item select with arrow keys @filter

umuero:
### Feature Requests
It shoud be possible to select the items in the multiselect component, without using mouse

Added - arrowDown to filterInput to focus options
Changed - arrowUp to onOptionKeydown to focus filterInput if already at 0
Changed - 2 var usages to let

image

previously submitted at #12723


cetincakiroglu:
Hi,

Accessibility implementation of the input section is in progress and we'll implement full keyboard support for multiselect. Implementation will be the same as https://primevue.org/multiselect/

Closing the issue since it'll be implemented by the PrimeNG dev team. 

PR #13428, opened 2023-08-03
fixed test failure "Expected '133px' to be greater than '37px'"

nhnb:
This is a very quick fix for a broken test. I did not check whether this pattern is used in other tests because this is the only test that fails for me.


cetincakiroglu (approved):

cetincakiroglu:
Hi @nhnb 

PR #13432, opened 2023-08-04
Bugfix #13431 Fix accordion accessibility

### LINKED ISSUE #13431: Accordion | wrong aria roles and add headerAriaLevel property (accessibility)
### Describe the bug

As per W3C specification, the accordion should be a sequence of headers with buttons. 
In the [Accessibility section of PrimeNG's Accordion documentation](https://primeng.org/accordion#accessibility) is stated that 
> Accordion header elements have a button role

but actually the accordion still has `role="tablist"` and each header has `role="tab"`. In PrimeVue the accordion is not a `tablist` and the `a` tag has role `button`, correctly. 

Furthermore, it is recommended in the [WAI ARIA pattern for Accordions](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/#wai-ariaroles,states,andproperties:) that each header button should be wrapped in a container with `role="heading"`. Such role also requires `aria-level` to be defined or [it will assume the default value of `2` ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role#associated_wai-aria_roles_states_and_properties).

### Environment

Windows

### Reproducer

_No response_

### Angular version

16.X

### PrimeNG version

16.1.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

16.20

### Browser(s)

_No response_

Open https://primeng.org/accordion and check the aria roles of the accordion's HTML structure.

### Expected behavior

Same as PrimeVue's one, plus having the `div` that wraps the `a` tag to have `role="heading"` and a customizable `aria-level`.

GiacomoDM:
Fix #13431

* remove role tablist
* change header role from tab to button
* added heading role to accordion header and the corrisponding level customizable via input
* updated docs

### DIFF
--- a/src/app/components/accordion/accordion.ts
+++ b/src/app/components/accordion/accordion.ts
@@ -35,11 +35,11 @@ import { UniqueComponentId } from 'primeng/utils';
     selector: 'p-accordionTab',
     template: `
         <div class="p-accordion-tab" [class.p-accordion-tab-active]="selected" [ngClass]="tabStyleClass" [ngStyle]="tabStyle" [attr.data-pc-name]="'accordiontab'">
-            <div class="p-accordion-header" [class.p-highlight]="selected" [class.p-disabled]="disabled" [attr.data-p-disabled]="disabled" [attr.data-pc-section]="'header'">
+            <div class="p-accordion-header" role="heading" [attr.aria-level]="headerAriaLevel" [class.p-highlight]="selected" [class.p-disabled]="disabled" [attr.data-p-disabled]="disabled" [attr.data-pc-section]="'header'">
                 <a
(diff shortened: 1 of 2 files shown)


cetincakiroglu (approved):

cetincakiroglu:
Hi @GiacomoDM,

You don't need to manually update the index.json since we generate it by command `npm run apidoc` before every release. Could you please refactor the naming of`headersAriaLabel` to `headerAriaLabel`.  I'll merge it afterward.

GiacomoDM:
done, thanks

cetincakiroglu (approved):

cetincakiroglu:
Thank you for your effort and support!

PR #13436, opened 2023-08-04
Bugfix #13435 Fix PanelMenu tab navigation

### LINKED ISSUE #13435: Accessibility | PanelMenu - collapsed region are still reachable via `tab` navigation when at least an item has a link
### Describe the bug

For PanelMenus with nested items, when a parent node item is collapsed its children are still reachable via `tab` navigation, as long as one child has a routerlink (href) or target attribute. 
When a parent node is collapsed, pressing `tab` should move focus to the node's sibling (or outside the PanelMenu if it is the last element). 

### Environment

Windows

### Reproducer

https://stackblitz.com/edit/wqfsxf?file=src%2Fapp%2Fdemo%2Fpanel-menu-basic-demo.ts

### Angular version

16.X

### PrimeNG version

16.1.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

16.20

### Browser(s)

_No response_

1. Open the provided StackBlitz
2. Place focus to the first element of the PanelMenu
3. Press `tab`
4. The focus is inside the collapsed region
5. Press focus more times, until you reach the second top-level item ("Edit")
6. Press `tab`
7. The focus correctly moves to the third top-level item, because it doesn't have children with `href`

### Expected behavior

Step 3 should behave the same as step 6

GiacomoDM:
Fix #13435

Setting `display: none` to the collapsed regions will prevent moving focus inside them. 
Also, it will hide all the content to all assistive technologies.


cetincakiroglu:
Hi,

display:none rule breaks the transition of the collapse animation. Possibly another solution is to setting tabindex of anchor elements. Could you please refactor this PR and remove display: none rule and add [attr.tabindex]="-1" to the anchor elements inside the list item.

GiacomoDM:
Done, thanks. 

I also added `aria-hidden` to the panelMenuSub's `ul` because even if not visible and collapsed, the screen reader still highlight it and read it as "tree view" but empty. 
This way if the parent is collapsed, the `ul` will be skipped completely.

cetincakiroglu (approved):

PR #13440, opened 2023-08-05
Inputnumber: Keyboard events using event.key

### LINKED ISSUE #13439: p-inputNumber -- deprecated keyCode and which used when checking for keypress and keydown
### Describe the bug

for the keyboardEvent, which and keyCode is deprecated and it's recommended to use `key`.

For example
```typescript
onInputKeyDown(event) {
  if (this.readonly) {
    return;
  }
  this.lastValue = event.target.value;
  if (event.shiftKey || event.altKey) {
    this.isSpecialChar = true;
    return;
  }
  let selectionStart = event.target.selectionStart;
  let selectionEnd = event.target.selectionEnd;
  let inputValue = event.target.value;
  let newValueStr = null;
  if (event.altKey) {
    event.preventDefault();
  }

  // we should use "event.key" in here
  // original code uses "event.which"
  switch (event.key) {
    //up
    case 'ArrowUp':
      // ...
      break;
    //down
    case 'ArrowDown':
      // ...
      break;
    //left
    case 'ArrowLeft':
      // ...
      break;
    //right
    case 'ArrowRight':
      // ...
      break;
    //enter
    case 'Enter':
      // ...
      break;
    //backspace
    case 'Backspace': {
      // ...
      break;
    }
    // del
    case 'Delete':
      // ...
      break;
    default:
      break;
  }
  this.onKeyDown.emit(event);
}

onInputKeyPress(event) {
  if (this.readonly) {
    return;
  }
  event.preventDefault();
  // so we can just use "event.key"
  // and change "48 <= code && code <= 57" to "/^\d$/.test(char)"
  let code = event.which || event.keyCode;
  let char = String.fromCharCode(code);
  const isDecimalSign = this.isDecimalSign(char);
  const isMinusSign = this.isMinusSign(char);
  if ((48 <= code && code <= 57) || isMinusSign || isDecimalSign) {
    this.insert(event, char, { isDecimalSign, isMinusSign });
  }
}
```

### Environment

I was testing using @testing-library/dom and  @testing-library/user-event and tried to type on inputNumber tag but nothing works because @testing-library did not implement keyboard event for `which` and `keyCode`. Once I change them to `key`, my test cases pass like magic.

### Reproducer

_No response_

### Angular version

14.2.3

### PrimeNG version

14.0.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.16.0

### Browser(s)

ALL

_No response_

### Expected behavior

_No response_

indulakshmikr:
### Defect Fixes
Fixes #13439

### DIFF
--- a/src/app/components/inputnumber/inputnumber.ts
+++ b/src/app/components/inputnumber/inputnumber.ts
@@ -780,36 +780,35 @@ export class InputNumber implements OnInit, AfterContentInit, OnChanges, Control
         if (event.altKey) {
             event.preventDefault();
         }
-
-        switch (event.which) {
+        switch (event.key) {
             //up
-            case 38:
+            case 'ArrowUp':
                 this.spin(event, 1);
                 event.preventDefault();
                 break;
 
             //down
-            case 40:
+            case 'ArrowDown':
                 this.spin(event, -1);
                 event.preventDefault();
                 break;
 
             //left
-            case 37:
+            case 'ArrowLeft':
                 if (!this.isNumeralChar(inputValue.charAt(selectionStart - 1))) {
                     event.preventDefault();
                 }
                 break;
 
             //right
-            case 39:
+            case 'ArrowRight':
                 if (!this.isNumeralChar(inputValue.charAt(selectionStart))) {
                     event.preventDefault();
                 }
                 break;
 
             //enter
-            case 13:
+            case 'Enter':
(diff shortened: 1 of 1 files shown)


SoyDiego (approved):

cetincakiroglu:
Hi,

We're implementing accessibility for all components and we're also updating deprecated keyboard events. This issue will be fixed by the core team when it comes to inputnumber accessibility implementation. 

We appreciate your effort and support!

Issue #13433, opened 2023-08-04
TieredMenu: On mobile tieredmenu requires double tap to open submenu


r0754497:
### Describe the bug

TieredMenu requires 2 touches to open a submenu. One focusses, next one opens. Required fix: touch to have the same behaviour as hover on menu items with submenu.
@cetincakiroglu 

### Environment

Occurred on multiple phones (Android and iOS), also the case when using Chrome dev tools.

### Reproducer

https://primeng.org/tieredmenu

### Angular version

15.2.8

### PrimeNG version

15.2.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

8.19.2

### Steps to reproduce the behavior

Go to https://primeng.org/tieredmenu, open Chrome dev tools (F12), press CTRL+SHIFT+M to switch to mobile view and press the example menu.

### Expected behavior

Open submenu on the first tap.

SoyDiego:
Hi @r0754497 I have checked the issue and you are right.
I have created a PR fixing the problem. You can check the PR here: https://github.com/primefaces/primeng/pull/13443.
Now we should wait PrimeNG Team approve it if it's ok for them this solution.

Thanks

rosenthalj:
After quickly searching all PrimeNG menu components,  it appears that at a minimum MenuBar and MegaMenu have similar code to TieredMenu.  The TieredMenu is the only component that has ben fixed "fixed" 

MenuBar has the same problem

https://github.com/primefaces/primeng/assets/45439491/31c00ed6-109f-4a58-9e63-25717840b3fc

MegaMenu has the same problem

https://github.com/primefaces/primeng/assets/45439491/3cb6cdcc-6c6e-416b-83a6-4a8fa115a412

Note:  I quickly tested PanelMenu and SlidMenu and they appear to work correctly.

SoyDiego:
> After quickly searching all PrimeNG menu components, it appears that at a minimum MenuBar and MegaMenu have similar code to TieredMenu. The TieredMenu is the only component that has ben fixed "fixed"
> 
> MenuBar has the same problem
> MenuBarTouchBug.mov
> 
> MegaMenu has the same problem
> MegaMenuTouchBug.mov
> 
> Note: I quickly tested PanelMenu and SlidMenu and they appear to work correctly.

Thanks @rosenthalj If you can create different issues will be great and we will try to fix it too.

rosenthalj:
@SoyDiego 

[...]

PR #13450, opened 2023-08-07
Revert "[#13433] - Fixed double tap to open submenu in tieredmenu component"

cetincakiroglu:
Reverts primefaces/primeng#13443


afontanaroyo:
why revert? now the bug keeps happening..

PR #13457, opened 2023-08-07
Fixed #13455 - TieredMenu does not show sub menus when mouse is hovered over the parent menu item

### LINKED ISSUE #13455: TieredMenu: SubMenus are not uncovered on mouse hover anymore
### Describe the bug

Since the `TieredMenu` refactoring done in https://github.com/primefaces/primeng/commit/28989fd4cafe4ee5b7d805f5b059618f5fc1128b, it works a bit different now (16rc2 -> 16.0.0).
Before, all sub `MenuItem`s of a `TieredMenu` uncovered themself after hovering over the parent `MenuItem` when in `popup` mode.

### Environment

Doesn't matter in this case.

### Reproducer

See below

### Angular version

16.0.0

### PrimeNG version

16.1.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

16.16.0

### Browser(s)

All

This can be easily reproduced by checking out the showcase.
https://www.primefaces.org/primeng-v15-lts/tieredmenu
https://primeng.org/tieredmenu

1. In the `Popup` section, click on the `Show` Button.
2. If you hover over all the `MenuItem`s like `File` or `Edit`, all sub `MenuItem`s will be uncovered in Version 15. This does not happen anymore in Version16, you need to click on the `MenuItem` to uncover all sub `MenuItem`s.

### Expected behavior

Sub `MenuItem`s uncover themself when you hover over the parent `MenuItem`.

NOTE: `autoDisplay` is `true` but it seems to not affect anything at all.
Before the refactoring this behaviour worked when the `TieredMenu` is a `popup` or `autoDisplay` was turned on AFAIK.

Maran23:
Fixes: https://github.com/primefaces/primeng/issues/13455

To fix the problem the `isDirty` flag needs to be set to `true` when the popup is shown or an item is clicked. 
This will ensure that `onItemChange(..)` is called when a `MenuItem` is hovered by the mouse, which will mark the `MenuItem` as active and therefore uncover the sub `MenuItem`s (children).

### DIFF
--- a/src/app/components/tieredmenu/tieredmenu.ts
+++ b/src/app/components/tieredmenu/tieredmenu.ts
@@ -639,7 +639,7 @@ export class TieredMenu implements OnInit, AfterContentInit, OnDestroy {
             this.activeItemPath.set(this.activeItemPath().filter((p) => key !== p.key && key.startsWith(p.key)));
             this.focusedItemInfo.set({ index, level, parentKey });
 
-            this.dirty = !root;
+            this.dirty = true;
             DomHandler.focus(this.rootmenu.sublistViewChild.nativeElement);
         } else {
             if (grouped) {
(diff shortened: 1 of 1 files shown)


SoyDiego (approved):

cetincakiroglu:
Thanks for reporting to issue with the pr, I appreciate your effort 👍

PR #13477, opened 2023-08-10
Accessibility

cetincakiroglu:


cetincakiroglu:
Contains accessibility enhancements for

- Messages section
- Overlay section
- Button section

fbobbio:
@cetincakiroglu Hello and thanks for your efforts on the lib!

After upgrading to `primeng` version 16.2.0 (while bumping my project to Angular 16), all my components using `DynamicDialogComponent` have been throwing the infamous dev-env error `

In this commit: https://github.com/primefaces/primeng/commit/8ed19e1a0f2e965b54b759577e6ba63feab6f6f8 I see the binding of a template properties (`[attr.aria-labelledby]` for example) to a component's function.
Afaik this would mess with the Angular change-detection mechanism.

Not sure if this is a bug or not or if it was already reported. I couldn't find anything similar on the project's site, or I may be missing something on my side.

My current workaround is to downgrade to version `16.1.0` but ideally I would want to bump to `16.2.0`.

A slice of the stack trace:
```
ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value for 'attr.aria-labelledby': 'pn_id_38_header'. Current value: 'pn_id_40_header'. Expression location: DynamicDialogComponent component.
```

It is triggered first when opening the dialog and keeps triggering constantly (I assume every time the change detection runs).

Thanks in advance!

PR #13502, opened 2023-08-15
Dynamic Dialog: Fixed errors when is opened

### LINKED ISSUE #13497: DynamicDialog: ExpressionChangedAfterItHasBeenCheckedError for attr.aria-labelledby
### Describe the bug

ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value for 'attr.aria-labelledby': 'pn_id_157_header'. Current value: 'pn_id_159_header'. Expression location: DynamicDialogComponent component.

aria-labeledby is assign for div with class 'p-dynamic-dialog'

### Environment

Basic as possible, Angular + PrimeNG only

### Reproducer

https://stackblitz.com/edit/pfqjsp?file=package.json

### Angular version

16.2.0

### PrimeNG version

16.2.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.16.0

### Browser(s)

Firefox 116.0

1. Open dynamic dialog

### Expected behavior

_No response_

SoyDiego:
Fix #13497

Changed the changeDetection to `onPush` solves the problem. I don't know if it's the best solution. If you have another, let me know.

Note: Ignore the warning message `responsive property is deprecated as table is always responsive with scrollable behavior.`

## CURRENT BEHAVIOUR
problem dynamic dialog

## AFTER SOLUTION
fixed dynamic dialog

### DIFF
--- a/src/app/components/dynamicdialog/dynamicdialog.ts
+++ b/src/app/components/dynamicdialog/dynamicdialog.ts
@@ -94,7 +94,7 @@ const hideAnimation = animation([animate('{{transition}}', style({ transform: '{
         </div>
     `,
     animations: [trigger('animation', [transition('void => visible', [useAnimation(showAnimation)]), transition('visible => void', [useAnimation(hideAnimation)])])],
-    changeDetection: ChangeDetectionStrategy.Default,
+    changeDetection: ChangeDetectionStrategy.OnPush,
     encapsulation: ViewEncapsulation.None,
     styleUrls: ['../dialog/dialog.css'],
     host: {
@@ -213,7 +213,7 @@ export class DynamicDialogComponent implements AfterViewInit, OnDestroy {
         public zone: NgZone,
         public primeNGConfig: PrimeNGConfig,
         @SkipSelf() @Optional() private parentDialog: DynamicDialogComponent
-    ) {}
+    ) { }
 
     ngAfterViewInit() {
         this.loadChildComponent(this.childComponentType!);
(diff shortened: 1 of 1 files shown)


MarcinBorowski95:
I'm not sure if changing the change detection is the best solution here. It could bring some unforeseen consequences.
Wouldn't it be better to:
Instead of using a method like this
`[attr.aria-labelledby]="getAriaLabelledBy()"`
create a property that will be initialized in a constructor/onInit?
That way change detection will not cause a refresh of labelledBy and then an error.

SoyDiego:
> I'm not sure if changing the change detection is the best solution here. It could bring some unforeseen consequences. Wouldn't it be better to: Instead of using a method like this `[attr.aria-labelledby]="getAriaLabelledBy()"` create a property that will be initialized in a constructor/onInit? That way change detection will not cause a refresh of labelledBy and then an error.

Thanks for check @MarcinBorowski95. In the [issue post](https://github.com/primefaces/primeng/issues/13497#issuecomment-1679284903) I said I don't know if it the best solution:
image.

I will try your solution and generate a new PR if continue working good.

Thanks for review and comment a possible solution here!

This afternoon I will try to do it! :)

SoyDiego:
Hi @MarcinBorowski95 I'm doing your way:

image

image


And it's true, works and no generate the error but now I have a doubt about this:

image

With your way we have same values on init and in the way of PrimeNG were different.

I don't know if it's correct or not but the error don't show like you said.

SoyDiego:
> I think this is still better. And if different id's are necessary then I guess two values can be generated. It will still be more optimized and less prone to bad change detection issues.

Thanks again bro. I added '_title' to the second one, just in case need to be different

image.

I will do the PR doing a mention to you :)

MarcinBorowski95:
Happy to help 😁

Issue #12875, opened 2023-04-11
Tooltip: Global configuration object


konuch:
### Describe the feature you would like to see added

I think it would be useful to have global configuration object for `tooltip` directive similar to global config for `overlay`.

### Is your feature request related to a problem?
We are currently working hard to align our application with WCAG. We use primeng v15.3 and we would like to globally set tooltip behavior, eg. `autoHide: false`. This setting is necessary to comply with https://www.w3.org/TR/WCAG21/#content-on-hover-or-focus This is not possible at the moment and we need to edit a lot of code to align all tooltips through-out our application. Having a global configuration object would help out meeting this requirement. 

### Describe the solution you'd like

Add `tooltipOptions` property to `PrimeNgConfig` class which can be injected to root component and modify tooltip behavior globally.

SoyDiego:
It's a nice idea.

Adrii77:
Any news about this ? this would be so useful and remove many boilerplate code !

mehmetcetin01140:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you open a new issue so we can include it in our roadmap?

Thanks a lot for your understanding!
Best Regards,

PR #13521, opened 2023-08-21
feat: global tooltip configuration object 

konuch:
- adds basic tooltip configuration object that can be set using `PrimeNgConfig` class in app root component
- it allowed to globally set tooltip behavior
- addresses #12875


cetincakiroglu:
Hi,

 We're currently working on PassThrough to make this possible for every component and planning to release before Angular v18. PassThrough implementation will supersede this PR. Because of this, no need to implement such feature. Closing the PR. Thanks a lot for the effort and support!

Yuri-Lima:
> Hi,
> 
> We're currently working on PassThrough to make this possible for every component and planning to release before Angular v18. PassThrough implementation will supersede this PR. Because of this, no need to implement such feature. Closing the PR. Thanks a lot for the effort and support!

Hi, have we implemented that global config for tooltip already?

PR #13543, opened 2023-08-24
Remove duplicated code in selectRange method of Table component 

dylansyoung:
No issue reported but removing duplicated code in selectRange method.


mertsincan:
Good catch! Thanks a lot for your contribution!

Best Regards,

PR #13551, opened 2023-08-26
Fix #13544: OrderList: Does not sort correctly when data is refreshed while item is selected

### LINKED ISSUE #13544: OrderList: Does not sort correctly when data is refreshed while item is selected
### Describe the bug

Selecting an item in an OrderList Component and then refreshing the data causes the OrderList component to become non-functional.  Causes undefined references and the array in the component now has negative indexing and undefined values inserted.

image

### Environment

 dev

### Reproducer

https://stackblitz.com/edit/primeng-orderlist-demo-6qysau?file=src%2Fapp%2Fapp.component.tshttps://stackblitz.com/edit/primeng-orderlist-demo-6qysau?file=src%2Fapp%2Fapp.component.ts

### Angular version

14.0.1

### PrimeNG version

14.0.2

### Build / Runtime

TypeScript

### Language

TypeScript

18

### Browser(s)

_No response_

1. Load an OrderList Component with data.
2. Select an Item.
3. Refresh Data.
4. Click down arrow.
5. View console logs & OrderList view, errors and item data disappear.

Note: Logging out OrderList shows odd values in list and negative array indexes after these steps are taken.

### Expected behavior

Data refreshes and selects the previously selected items or the selections are removed.

FlipWarthog:
Fix #13544: OrderList: Does not sort correctly when data is refreshed while item is selected

### DIFF
--- a/src/app/components/orderlist/orderlist.ts
+++ b/src/app/components/orderlist/orderlist.ts
@@ -470,6 +470,10 @@ export class OrderList implements AfterViewChecked, AfterContentInit {
                 let selectedItem = this.selection[i];
                 let selectedItemIndex: number = ObjectUtils.findIndexInList(selectedItem, this.value);
 
+                if (selectedItemIndex == -1) {
+                    break;
+                }
+
                 if (selectedItemIndex != 0 && this.value instanceof Array) {
                     let movedItem = this.value[selectedItemIndex];
                     let temp = this.value[selectedItemIndex - 1];
@@ -493,6 +497,10 @@ export class OrderList implements AfterViewChecked, AfterContentInit {
                 let selectedItem = this.selection[i];
                 let selectedItemIndex: number = ObjectUtils.findIndexInList(selectedItem, this.value);
 
+                if (selectedItemIndex == -1) {
+                    break;
+                }
+
                 if (selectedItemIndex != 0 && this.value instanceof Array) {
                     let movedItem = this.value.splice(selectedItemIndex, 1)[0];
                     this.value.unshift(movedItem);
@@ -513,6 +521,10 @@ export class OrderList implements AfterViewChecked, AfterContentInit {
             for (let i = this.selection.length - 1; i >= 0; i--) {
                 let selectedItem = this.selection[i];
                 let selectedItemIndex: number = ObjectUtils.findIndexInList(selectedItem, this.value);
+
+                if (selectedItemIndex == -1) {
+                    break;
+                }
 
                 if (this.value instanceof Array && selectedItemIndex != this.value.length - 1) {
                     let movedItem = this.value[selectedItemIndex];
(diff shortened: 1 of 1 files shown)


SoyDiego (changes requested):
Maybe it's better to create a function so as not to repeat the same thing so many times, right?

melloware:
@SoyDiego agree it looks like this is a lot of repeated code.  Not just the new check that @FlipWarthog added but it looks like it could be refactored into a function.

FlipWarthog:
@melloware @SoyDiego I considered that, but if you compare the 4 functions there are slight differences between them in the for loops and how they actually move the entries (Note: it's the same on the PrimeReact side for the [same component](https://github.com/primefaces/primereact/blob/master/components/lib/orderlist/OrderListControls.js#L17-L127). If it's okay, I'll leave that for someone else. I just pushed one small commit to use `continue;` rather than `break;` so it doesn't stop processing on the first "invalid" entry. Testing shows this handles a couple more scenarios like selecting an item > reloading data > Shift selecting another item > moving item.

SoyDiego:
> @melloware @SoyDiego I considered that, but if you compare the 4 functions there are slight differences between them in the for loops and how they actually move the entries (Note: it's the same on the PrimeReact side for the [same component](https://github.com/primefaces/primereact/blob/master/components/lib/orderlist/OrderListControls.js#L17-L127)). If it's okay, I'll leave that for someone else. I just pushed one small commit to use `continue;` rather than `break;` so it doesn't stop processing on the first "invalid" entry. Testing shows this handles a couple more scenarios like selecting an item > reloading data > Shift selecting another item > moving item.

Thanks for the information. I wrote this in the issue where the user had the problem, 
I don't know if you saw it or has any sense here:

image

Just in case I put again because maybe is not problem of the code, but of the user (Again, just in case, maybe I'm wrong too)

If it is none of this, we should let someone from the PrimeNG team see what is most convenient

FlipWarthog:
@SoyDiego Agreed, I think the solution of binding and clearing/correcting `selection` if `value` is replaced (which you're effectively doing by reselecting after the Refresh) is the actual answer. This PR just stops the component failing if that isn't done so I'm fine waiting on PrimeTek to weigh in.

While I was researching, I did notice dataKey is implemented on the PrimeReact equivalent of this component, which would fix the equality issue without needing to bind/clear `selection`, but I think implementing that for PrimeNG would be a larger effort requiring a new feature request/PR that's beyond the scope of the original ticket.
image

mertsincan:
Hi @SoyDiego,

Were you able to replicate this issue?

SoyDiego:
> Hi @SoyDiego,
> 
> Were you able to replicate this issue?

Hi, I didn't replicate the problem.
Only I gave a solution for the user in the issue post.  I think the issue is not a problem of Orderlist but maybe the PR of @FlipWarthog is about another thing that you should consider

mertsincan:
Thanks @SoyDiego ;)

I couldn't replicate it either. So I'm closing it for now. If the problem persists after the next release, can you please explain the situation again with a new PR?

Best Regards,

Issue #13578, opened 2023-08-30
Autocomplete Force Selection case sensitive issue


ashikjs:
### Describe the bug

#Autocomplete Force Selection
If we type United Kingdom the clcik outside of input then it able to select but if type `united kingdom` it just clear this text can not select **United Kingdom** Option from suggestion.

[Video for details](https://www.awesomescreenshot.com/video/20345591?key=9562afc912d5318bd2db1ef023ece967) 

Expected behavior: it should able to select United Kingdom option from suggestion.   

### Environment

latest version of primeng 16.2.0 

### Angular version

16

### PrimeNG version

16.2

### Node version (for AoT issues node --version)

18

PR #13579, opened 2023-08-30
fix: #13578 || Autocomplete Force Selection case sensitive issue

ashikjs:
Fix: #13578 

## Code Explain 
Previously we check just two string that's why it was return not equal.
So know I am doing lowercase item and input value before check so both it ensure same case.


ashikjs:
@mertsincan   hi, can you please check this PR. Thanks in advance.

SoyDiego (approved):

mertsincan:

Issue #13007, opened 2023-05-08
AutoComplete control inappropriately fires onClear event when selecting an item with the mouse (correctly doesn't when using the keyboard)


AlejandroFlorin:
### Describe the bug

In the autocomplete control, if you use the keyboard (down arrow and enter key) to select an item, the control triggers only the onSelect event as expected.

However, if you use the mouse to select an item, the control triggers the onClear event prior to triggering the onSelect event.

This is an issue because I have handlers for both the onClear and onSelect events which both make calls to a server api. It then becomes a race situation between the onClear event and the onSelect event when the user clicks on an option (as opposed to using the keyboard to select an option). There is also no way to know from the onClear event if it was triggered by the selection of an item (so I can not do the call) or by an actual clearing of the control, other than setting a state variable in the onSelect event and using a timeout on the onClear event to check the variable before making its call.

### Environment

Dev Machine:
  Windows 11
  Visual Studio 2019

### Angular version

15.2.8

### PrimeNG version

15.4.1

### Build / Runtime

TypeScript

### Language

TypeScript

### Node version (for AoT issues node --version)

18.16.0

### Steps to reproduce the behavior

Add an Autocomplete control to a component. The properties I'm using are:
                  <p-autoComplete
                                    formControlName="recruiter"
                                    [suggestions]="availableRecruiters"
                                    field="fullName"
                                    dataKey="id"
                                    [forceSelection]="true"
                                    (onClear)="onRecruiterClear()"
                                    (onSelect)="onRecruiterChange()"
                                    (completeMethod)="onRecruiterSearch($event)">
                    </p-autoComplete>

Assign an array of values to the control's suggestions object
Search for an item
Use keyboard (down arrow and enter) to select an item (control triggers the onSelect event only: expected)
Search for the same item
With mouse, click on one of the displayed items (control triggers the onClear event as well as the onSelect event: not expected)

### Expected behavior

The onClear event should not be triggered when selecting an item, regardless of the method used to select the item.

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you open a new issue so we can include it in our roadmap?

Thanks a lot for your understanding!
Best Regards,

PR #13580, opened 2023-08-30
fix: #13007 || AutoComplete control inappropriately fires onClear eve…

ashikjs:
FIx: #13007 

## Explanation 

Issue was `onClearEvent` fire when `forceSelection` is ` true` and user select from suggestion list 

Excepted behavior is should not fir when user select from suggestion list.

Currently we check  **forceSelection** every time when input change and  it's true .

Main problem is option select event fire after input change event. 

So I write a function name **onForceSelectOrClearInput** with existing code. Then **onInputChange** I call this function if user don't select suggestion option. For this case i did wait 200ms to fire next event **selectItem** because when this **selectItem** fire I set value true of `isSuggestionOptionSelect`  variable. 

## Extra Code

Github PR Checker give me error when it run Code format. 
So I solve it by remove on space from `src/app/components/table/table.ts`

### DIFF
--- a/src/app/components/autocomplete/autocomplete.ts
+++ b/src/app/components/autocomplete/autocomplete.ts
@@ -659,6 +659,8 @@ export class AutoComplete implements AfterViewChecked, AfterContentInit, OnDestr
 
     overlayVisible: boolean = false;
 
+    isSuggestionOptionSelect: boolean = false;
+
     suggestionsUpdated: Nullable<boolean>;
 
     highlightOption: any;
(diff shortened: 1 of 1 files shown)


ashikjs:
@mertsincan and @gucal  hi, can you please check this PR. Thanks in advance.

cetincakiroglu:
Hi @ashikjs 

Could you please update the PR and resolve the conflicts so we can merge safely?

ashikjs:
@cetincakiroglu  I re-check this bug but can not reproduce it so i belive it's already solved.

cetincakiroglu:
> @cetincakiroglu I re-check this bug but can not reproduce it so i belive it's already solved.

Thanks for the feedback, probably solved at one point in the history. I'm closing this one.

Issue #13576, opened 2023-08-30
[Menu] Menu has weak z-index 


dalenguyen:
### Describe the bug

When opening menu item, the menu has an overlay issue due to the other element has a bigger z-index.

<img width="741" alt="image" src="https://github.com/primefaces/primeng/assets/14116156/3b28def7-4878-4399-ad46-5553b907935c">

Link: https://primeng.org/megamenua 

### Environment

Production

### Angular version

16.0.0

### PrimeNG version

16.2.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.16.0

### Steps to reproduce the behavior

Go to https://primeng.org/megamenu

Click on the menu to show the menu item
The demo links (stackblitz...) interfere with the menu item

### Expected behavior

The menu should have a bigger z-index than the other items. The current value is 1.

ashikjs:
This issue also existed in MenuBar

![image](https://github.com/primefaces/primeng/assets/13074235/08738e12-4626-48b8-ba64-b583446af1f4)

PR #13586, opened 2023-08-31
fix: #13576 || [Menu] Menu has weak z-index

ashikjs:
FIX: #13576

### DIFF
--- a/src/app/components/megamenu/megamenu.css
+++ b/src/app/components/megamenu/megamenu.css
@@ -25,7 +25,7 @@
     display: none;
     position: absolute;
     width: auto;
-    z-index: 1;
+    z-index: 10;
 }
 
 .p-megamenu-root-list > .p-menuitem-active > .p-megamenu-panel {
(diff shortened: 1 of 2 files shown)


ashikjs:

SoyDiego (approved):

cetincakiroglu:
Hi @ashikjs ,

Thanks for reporting the issue. Your solution is not wrong but we cannot accept it because the issue is only demo related and it only happens in the showcase. Thus, we can't change z-index from the component CSS with a random value to fix the demo. Instead of that, I've temporarily fixed the issue by adding some custom CSS to the related demo. See here: https://github.com/primefaces/primeng/commit/4b40014486b56aef7d1dc1d1fc293fa32461e478

However, users may have problems as you've mentioned in the issue and my fix is temporary and it only covers the demo. The absolute solution is adding "baseZIndex" support so the user can update the zIndex without custom CSS. Because of that, I'm closing this PR, you can create a new PR for the issue below.

Here I created a new issue by referencing yours: https://github.com/primefaces/primeng/issues/13710

We appreciate your report and support!

ashikjs:
@cetincakiroglu, 
Understand your point and thanks for your explanations.

Issue #13571, opened 2023-08-29
MultiSelect: Overlay not repositioned after filtering options


guillaume-tritsch:
### Describe the bug

The position of the overlay is not updated after the option is filtered.

I'm looking for the problem in the code of the MultiSelect, I think the _onFilterInputChange_ method needs to set the **filtered** attribute to **true**.

onFilterInputChange(event: KeyboardEvent) {
    this._filterValue = (<HTMLInputElement>event.target).value;
    this.activateFilter();
    **// ADD "this.filtered = true;" HERE FIX THE PROBLEM**
    this.onFilter.emit({ originalEvent: event, filter: this._filterValue });
    this.cd.detectChanges();
}

### Environment

Windows 10
WSL Ubuntu

### Reproducer

https://primeng-multiselect-bug-repositioning-reproduction.stackblitz.io

### Angular version

16.1.7

### PrimeNG version

16.2.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.17.0

### Browser(s)

Firefox 116.0.3 (64 bits) and Google Chrome 116.0.5845.111

### Steps to reproduce the behavior

Create multiselect with one very long option:
![image](https://github.com/primefaces/primeng/assets/80285644/3da19d36-4708-4898-b571-5cabcf81780d)

Filter long option:
![image](https://github.com/primefaces/primeng/assets/80285644/8f7bd737-0240-4522-8aa4-0e43e1475bb3)

### Expected behavior

The overlay position need to be update:
![image](https://github.com/primefaces/primeng/assets/80285644/71d6ca19-164e-4ac7-aa11-b8324262c79c)

guillaume-tritsch:
@

PR #13587, opened 2023-08-31
fix: #13571 | Multiselect not repositioned

guillaume-tritsch:
fix: #13571 

## Explanation

The code for MultiSelect and Dropdown components closely resembles each other. However, I observed that the bug was occurring exclusively in the MultiSelect component. Upon comparing their respective code, I identified that a particular section of the MultiSelect code was never executed due to the constant state of the `filtered` variable being set to `false`. To address this issue, I introduced a code modification that ensures `filtered` is set to 'true' whenever the `onFilterInputChange` method is invoked.

## In code
```ts
onFilterInputChange(event: KeyboardEvent) {
    this._filterValue = (<HTMLInputElement>event.target).value;
    this.activateFilter();
    this.filtered = true; // I add this line
    this.onFilter.emit({ originalEvent: event, filter: this._filterValue });
    this.cd.detectChanges();
}
```
`this.cd.detectChanges()` refresh the view and call ngAfterViewChecked but `filtered` attribut is on false, so the `alignOverlay` is never call.
```ts
ngAfterViewChecked() {
    if (this.filtered) {
        this.zone.runOutsideAngular(() => {
            setTimeout(() => {
                this.overlayViewChild?.alignOverlay();
            }, 1);
        });
        this.filtered = false;
    }
}
```

I hope my explanations are clear, and I haven't made any mistakes when reading the code.

### DIFF
--- a/src/app/components/multiselect/multiselect.ts
+++ b/src/app/components/multiselect/multiselect.ts
@@ -1471,6 +1471,7 @@ export class MultiSelect implements OnInit, AfterViewInit, AfterContentInit, Aft
     onFilterInputChange(event: KeyboardEvent) {
         this._filterValue = (<HTMLInputElement>event.target).value;
         this.activateFilter();
+        this.filtered = true;
         this.onFilter.emit({ originalEvent: event, filter: this._filterValue });
         this.cd.detectChanges();
     }


cetincakiroglu (approved):
Hi @guillaume-tritsch ,

Issue #13499, opened 2023-08-14
SpeedDial onBlur: NG0600: Writing to signals is not allowed in a `computed` or an `effect` by default. Use `allowSignalWrites` in the `CreateEffectOptions` to enable this inside effects.


gustavosb26:
### Describe the bug

exception-handler.service.ts:19 Error: NG0600: Writing to signals is not allowed in a `computed` or an `effect` by default. Use `allowSignalWrites` in the `CreateEffectOptions` to enable this inside effects.
    at core.mjs:27953:15
    at throwInvalidWriteToSignalError (core.mjs:2683:5)
    at WritableSignalImpl.set (core.mjs:2718:13)
    at SpeedDial.onBlur (primeng-speeddial.mjs:284:33)
    at _class_Template_ul_focusout_5_listener (primeng-speeddial.mjs:560:29)
    at executeListenerWithErrorHandling (core.mjs:16195:16)
    at wrapListenerIn_markDirtyAndPreventDefault (core.mjs:16228:22)
    at HTMLUListElement.<anonymous> (platform-browser.mjs:665:17)
    at _ZoneDelegate.invokeTask (zone.js:446:35)
    at core.mjs:27127:55

### Environment

Template Atlantis

### Angular version

16.2.0

### PrimeNG version

16.2.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.17.1

### Browser(s)

Chrome Version 115.0.5790.170 (Official Build) (x86_64)

### Steps to reproduce the behavior

Click on item button SpeedDial

SoyDiego:
Hi @gustavosb26 Can you share a stackblitz example with your problem.
I have check the speed dial component and I cannot reproduce here: https://primeng.org/speeddial.

Thanks

utraum:
same problem here

i can reproduce it at  https://primeng.org/speeddial
opening developer console
after selecting items you get the error (not full stack because of production-build)
in stackblitz i got the complete error (after opening console)

SoyDiego:
> same problem here
> 
> i can reproduce it at https://primeng.org/speeddial opening developer console after selecting items you get the error (not full stack because of production-build) in stackblitz i got the complete error (after opening console)

Please can you share a stackblitz?
I cannot reproduce the problem
Thanks

utraum:
Update: Error occurs on chrome and edge (116.0.5845.110)
Firefox (115.1.0esr) does not throw any error
[...]

PR #13589, opened 2023-08-31
fix: SpeedDial onBlur issue #13499

dalenguyen:
### Defect Fixes (#13499)

Error in the console log when the SpeedDial is onBlur

```
Error: NG0600: Writing to signals is not allowed in a computed or an effect by default. Use allowSignalWrites in the CreateEffectOptions to enable this inside effects.
```

### DIFF
--- a/src/app/components/speeddial/speeddial.ts
+++ b/src/app/components/speeddial/speeddial.ts
@@ -421,7 +421,6 @@ export class SpeedDial implements AfterViewInit, AfterContentInit, OnDestroy {
 
     onBlur(event) {
         this.focused = false;
-        // fix error: NG0600: Writing to signals is not allowed in a computed or an effect by default
         asapScheduler.schedule(() => this.focusedOptionIndex.set(-1));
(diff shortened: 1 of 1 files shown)


src/app/components/speeddial/speeddial.ts
```diff
@@ -419,7 +421,8 @@ export class SpeedDial implements AfterViewInit, AfterContentInit, OnDestroy {
 
     onBlur(event) {
         this.focused = false;
-        this.focusedOptionIndex.set(-1);
+        // fix error: NG0600: Writing to signals is not allowed in a computed or an effect by default
```
ashikjs, line 424: Can you please remove this comment, it's not needed

dalenguyen: Absolutely.

ashikjs (approved):

SoyDiego (approved):

mertsincan:

PR #13599, opened 2023-09-01
Issues primefaces#8647 primefaces#5663 remove message or toast by id

sandrotonon:
Add a method to remove specific Toast/Message via MessageService

See #13600


mertsincan:
Hi @sandrotonon,

Can't we use a more dynamic key instead of 'id'? For example;
```
// this.messageService.remove(<object_key>, <object_value>)

this.messageService.remove('id', 1);
this.messageService.remove('severity', 'success');
```

In addition, we have the `removeMessage` method that works with the index in the messages component. Maybe if we add this to toast, users can easily remove the message by giving the index of the message they want to delete to this method. This is +1 for me.

WDYT?

sandrotonon:
Hi @mertsincan,

i really like the more generic approach. This way users do have a lot more flexibility in managing their messages.

However the index-based idea doesn't seem to be that comfortable, because you'd have to keep track of all displayed messages to be able to know the indices of the ones you'd want to delete.

I've updated the PR to reflect the first approach:

`messageService.remove(<object_key>, <object_value>)`

mertsincan:
Hi @sandrotonon,

Thanks a lot for the update! Frankly, I think we always need to filter all messages;
https://github.com/primefaces/primeng/pull/13599/files#diff-f1849731aed7bc226757a287fad99756df6389da4a30637bb84cf807a840e13fR179

That's why more feedback might be better for us. There is a similar desire in our other projects. That's why we can apply the most efficient solution for all of them.

sandrotonon:
> Frankly, I think we always need to filter all messages; https://github.com/primefaces/primeng/pull/13599/files#diff-f1849731aed7bc226757a287fad99756df6389da4a30637bb84cf807a840e13fR179

Hi @mertsincan, thanks for your feedback. But, what exaclty do you mean?

martinkasa:
@cetincakiroglu can we revive this PR?

mustafadalga:
@cetincakiroglu Could you find a solution ?

PR #13604, opened 2023-09-01
Fix #13041

### LINKED ISSUE #13041: InputMask: Custom numeric&alphabetic mask definitions as @Input()

Add an @ Input as two types (default value for numeric characters = 9 as it is now, default value for alphabetic characters = "a").
Numeric characters type can be 9 | # (9 or #), Alphabetic characters type can be a | $ (a or $).

That way if Developer puts maskDefinition for alpha characters as $ and for numeric characters as #, and that way [mask] property of InputMask component **will be able to accept fixed values for: 9 and "a" as for others.** 

I know allowing everything as mask definition  for maskDefinitionChars or maskDefNums could be tricky, but please can you think about just having two @ Input() as union types (written above).

There could be a situation when someone wants to predefine some parts of input mask as custom User Setting
(always to provide fixed number values** on first, second, etc part of input mask for example) as a help to the User - so user does not have to write same number all of the time again and again (1st part of InputMask).

The problem is when that user Setting is being defined and bound to p-inputMask [mask], InputMask will take all of characters except "a" (it thinks: allow all of Alphabetic characters for user to input), same behavior will be applied for 9 number (it thinks: allow all of Numeric characters for user to input).

Default (without writing) Mask behavior for Numeric characters: This should be 12-12-**192**
image

**But 9 in [mask] value means: write Numeric characters (with having **"#"** for example as numeric definition this could be avoided).**

Placeholder of input mask:

image

-------

Default (without writing) Mask behavior for Alphabetic characters: This should be 12-12-**bbaabb**)

image

**But "a" in [mask] value means: write Alphabetic characters (with having **"$"** for example as numeric definition this could be avoided).**

Placeholder of input mask:

image

Add @ Input() maskDefinitionChars (union type: "a" | "$"), @ Input() maskDefNums (union type: 9 | "#") to p-inputMask component.

Can you guys please do this? Another option is, I can do it at your codebase and create Pull request.

### Additional context

There's already existing deprecated Feature request before one year that you guys have closed without doing it: https://github.com/primefaces/primeng/issues/11546.

I'm aware that PrimeTek has Pro Support that does what ever features paid customers want in shortest time possible, but we're still not paid customer, but we're planning to be a paid customer in the future and I personally think this could be an improvement for InputMask component.

Thanks in advance, you guys are crushing it as Best Angular UI component library.

Kind regards,
Robert Medved

eXpertise7:
Fix #13041

### DIFF
--- a/src/app/components/inputmask/inputmask.ts
+++ b/src/app/components/inputmask/inputmask.ts
@@ -135,6 +135,16 @@ export class InputMask implements OnInit, ControlValueAccessor {
      * Advisory information to display on input.
      * @group Props
      */
+    @Input() maskDefNumbers: '9' | '#' = '9';
+    /**
+     * Custom mask definition for characters.
+     * @group Props
+     */
+    @Input() maskDefCharacters: 'a' | '$' = 'a';
+    /**
+     * Advisory information to display on input.
+     * @group Props
+     */
     @Input() placeholder: string | undefined;
     /**
      * Size of the input field.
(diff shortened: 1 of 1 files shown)


SoyDiego:
> Issue description: (#13041)

Thanks for your contribution. 
Please edit your first post and write: `Fix #13041` to link this PR to the issue with the problem.

eXpertise7:
> > Issue description: (#13041)
> 
> Thanks for your contribution. Please edit your first post and write: `Fix #13041` to link this PR to the issue with the problem.

@SoyDiego Done, is it good now? Thanks for feedback.

SoyDiego:
> > > Issue description: (#13041)
> > 
> > 
> > Thanks for your contribution. Please edit your first post and write: `Fix #13041` to link this PR to the issue with the problem.
> 
> @SoyDiego Done, is it good now? Thanks for feedback.

It's OK now, thanks

Mohammad-Eid-2000:
waiting the pull request to be completed

PR #13627, opened 2023-09-07
Add gtag manager

gucal:
Add gtag manager. #13626

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,7 +35,6 @@
                 "@nguniversal/express-engine": "^16.1.3",
                 "@stackblitz/sdk": "1.9.0",
                 "@types/express": "^4.17.17",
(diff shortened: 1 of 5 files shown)


SoyDiego (commented):
Maybe I'm wrong, but I leave here the same comment of @melloware in PrimeReact.

> Should we not render this script and run this code in Development mode? No sense in collecting analytics on local dev right?

I think apply the same case, right?, creating an environment variable.

Example of PrimeReact:

image

melloware:
Agreed when running locally it is not necessary to track metrics 😀

cagataycivici:
Agreed

cagataycivici:
This should run in prod only, please create a new PR. @gucal @cetincakiroglu

PR #13640, opened 2023-09-09
Fixed curser position for negative numbers with suffix

### LINKED ISSUE #13651: p-inputNumber: With suffix and minDigits > 0, negative numbers lead to an unexpected cursor jump and frozen inputs
### Describe the bug

When a p-inputNumber element has minFractionDigits>1 and suffix defined, there is a bug in concatValues() for negative values.

For example having minFractionDigits=2 and suffix=" %":

After typing -0, concatValues() produces -00 instead of 0.00 %.  This is because the differentiation between an existing and no existing decimalCharacter is missing, this differentiation is only made if there is no suffix defined. 

After when updating the value, after concatValues() there is another call of formatValue() in handleOnInput(). The format is then corrected to -0.00 %, but the cursor's position is put to the end of the input: "0.00 %|" instead of "0|.00 %". Further this results in the user not being able to type any more

### Environment

-

### Reproducer

_No response_

### Angular version

16.2.1

### PrimeNG version

16.3.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

-

### Browser(s)

_No response_

1. Create p-inputNumber with minFractionDigits=2 and suffix (any)
2. Type '-'
3. Type '0'
4. Now you cant edit the number before leaving the input and entering again. 

### Expected behavior

_No response_

johanneszelger:
When a negative number is entered with a suffix and minimum fraction digits, the format was incorrect.

For example:
-suffix " %"
-minFractionDigits 2
-0 resulted in a -00 concat value. (correct would be 0.00 %)

This results in a new call of formatValue() in handleOnInput(). The format is then correct, but the cursorposition is put to the end of the input: "0.00 %|" instead of "0|.00 %"

Fix #13651

### DIFF
--- a/src/app/components/inputnumber/inputnumber.ts
+++ b/src/app/components/inputnumber/inputnumber.ts
@@ -1289,7 +1289,7 @@ export class InputNumber implements OnInit, AfterContentInit, OnChanges, Control
             this._decimal.lastIndex = 0;
 
             if (this.suffixChar) {
-                return val1.replace(this.suffixChar, '').split(this._decimal)[0] + val2.replace(this.suffixChar, '').slice(decimalCharIndex) + this.suffixChar;
+                return decimalCharIndex !== -1 ? val1 : val1.replace(this.suffixChar, '').split(this._decimal)[0] + val2.replace(this.suffixChar, '').slice(decimalCharIndex) + this.suffixChar;
             } else {
                 return decimalCharIndex !== -1 ? val1.split(this._decimal)[0] + val2.slice(decimalCharIndex) : val1;
             }


SoyDiego:
Thanks for your contribution. 
You shoud create a issue here https://github.com/primefaces/primeng/issues, add screenshots, stackblitz or something showing the problem and then link this PR adding `Fix #YourIssueNumber" in your first comment of this PR.

johanneszelger:
Added issue

SoyDiego:
> Added issue

Fix #13651 with # symbol  to link your issue to this PR

johanneszelger:
Anything i can do to speed this up?

PR #13644, opened 2023-09-09
dialog: Possible fix ExpressionChangedAfterItHasBeenCheckedError with attr.aria-labelledby

### LINKED ISSUE #13636: p-dialog: ExpressionChangedAfterItHasBeenCheckedError with attr.aria-labelledby
### Describe the bug

I know some work is being done here on accessability, and I saw a [related issue](https://github.com/primefaces/primeng/issues/13497) but not with the `p-dialog` component.

We are on the latest PrimeNG version (16.3.1).

The exception is:

> ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value for 'attr.aria-labelledby': 'pn_id_70_header'. Current value: 'pn_id_73_header'. Expression location: _Dialog component

Call Stack:

    at throwErrorIfNoChangesMode (core.mjs:11010:11)
    at bindingUpdated (core.mjs:14251:17)
    at ɵɵattribute (core.mjs:14295:9)
    at Dialog_div_0_div_1_Template (primeng-dialog.mjs:357:8)
    at ReactiveLViewConsumer.runInContext (core.mjs:11103:13)

Line in primeng-dialog.mjs triggering this:

image

Dialog defined as:

```
<p-dialog
  [(visible)]="showUpdateMultiPointerDialog"
  [modal]="true"
>

```
**Also** - Can we get an updated reproducer template that isn't on Angular 14 and PrimeNG 14?

I am unable to update the template in stackblitz to Angular 16/PrimeNG 16, no matter what I try [I end up with](https://stackblitz.com/edit/github-qfq83u?file=package.json,src%2Fapp%2Fapp.module.ts,src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts):

> Error in /turbo_modules/primeng@16.3.1/fesm2022/primeng-tieredmenu.mjs (137:252)
> Cannot access 'TieredMenu' before initialization

### Environment

Angular 16.2.2
PrimeNG 16.3.1

### Reproducer

_No response_

### Angular version

16.2.2

### PrimeNG version

16.3.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.16.0

### Browser(s)

_No response_

1. Create a p-dialog component
2. Set it's [(visible)] property to a boolean flag
3. Set the boolean to true to show the dialog

### Expected behavior

There is no ExpressionChangedAfterItHasBeenCheckedError

SoyDiego:
Possible Fix #13636.

I added the same logic used in https://github.com/primefaces/primeng/pull/13538.

I couldn't replicate the problem in this case. Only @psarno, did it. But we can try with the same solution.

Note: I only changed the `dialog.ts`, the rest of files were formatted.


ashikjs (approved):

src/app/components/dialog/dialog.ts
```diff
@@ -549,6 +551,7 @@ export class Dialog implements AfterContentInit, OnInit, OnDestroy {
     }
 
     ngOnInit() {
+        this.ariaLabelledBy = this.getAriaLabelledBy();
```
volvachev, line 554: The behavior is not the same, since the header can change and the string cannot be recalculated. You can write a pure pipe and it will be recalculated when the header changes.

SoyDiego: Thanks for check.
I couldn't replicate the error of the user, only I did this PR because was very similar as I wrote in the post.

In my opinion it's working perfectly and I cannot see the error but @psarno had the problem and was very similar to another that I did before.

If you have a better solution you can share here or send new commits to this PR, but it's difficult to check because I cannot replicate the error

volvachev: In this case, `header` is an `Input` property and can be changed at runtime, and `ariaLabelledBy` is evaluated only once in `OnInit`. This `pr` breaks the previous behavior. You can use for example `ngOnChanges` and remove in `ngOnInit` line.

```ts
public ngOnChanges({ header }: SimpleChanges): void {
  if (header) {
    this.ariaLabelledBy = this. getAriaLabelledBy();
  }
}
```

SoyDiego: Thanks again. I did a new commit with your suggestion.
I would like to replicate the problem but I can't, so I'm trusting in our work here haha.
Check again please! :)

ashikjs:
@SoyDiego  can you squash those three commit to a single commit so we can ensure it, one single commit for a solution.

SoyDiego:
> @SoyDiego can you squash those three commit to a single commit so we can ensure it, one single commit for a solution.

Hi, thanks for check.
Like I said in the note, `I only changed the dialog.ts, the rest of files were formatted`. 

Regards

jncalderon:
Hi!! some update? to be honest I really need it :(

Issue #12948, opened 2023-04-26
Overlay: p-overlay div not removed from DOM on hiding panel


RinoGaetano94:
### Describe the bug

When hiding the overlay content (clicking outside the panel or programmatically) the wrapper div with class 'p-overlay' will not be removed from the DOM. This will cause an invisible layer blocking interactions with the page. If you perform another click, the p-overlay div will be removed from DOM.

It seems to happen only if responsive params is passed.

It seems to work properly on 14.2.2 Official Documentation.

This is the div that should be removed from DOM (it will removed on second click)
![Schermata del 2023-04-26 10-43-23](https://user-images.githubusercontent.com/16806264/234521056-01b6f35b-fec6-413e-a211-4347e0c22f8a.png)

### Environment

PrimeNG Official Documentation

### Reproducer

https://primeng.org/overlay#basic

### Angular version

15.2.2

### PrimeNG version

15.1.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.13.0

### Steps to reproduce the behavior

0. Trigger the responsive breakpoint resizing the window.
1. Click the "Show Overlay" button in the basic example.
2. Click outside the overlay panel to make it disappear.

If you try to click the "Show Overlay" button again to reopen the overlay, it will fail. You will not see the cursor pointer too. 

Even if the click fails, the p-overlay wrapper will be correctly removed this time

### Expected behavior

When hiding overlay panel, the p-overlay wrapper should be removed from DOM to re-establish the page interaction

faisal6699:
Need immediate solve :|

PR #13659, opened 2023-09-12
fix: #12948 || Overlay: p-overlay div not removed from DOM on hiding …

ashikjs:
Fix: #12948 

Previously I did try to solve it by `z-index` for CSS animation but now I re-check hole process and identified that dome should remove when overlay is hide. 

There issue was after complete animation `this.modalVisible = false;`  can not event fire angular change detection life cycle so dome was not removed properly so we fire it manually by `this.cd.markForCheck()` then dome gone and solve it issues.

## Extra code: 
I removed `event` from `(click)="onOverlayClick($event)"` because `onOverlayClick` function don't have any parameter.

### DIFF
--- a/src/app/components/overlay/overlay.ts
+++ b/src/app/components/overlay/overlay.ts
@@ -3,6 +3,7 @@ import { CommonModule, DOCUMENT, isPlatformBrowser } from '@angular/common';
 import {
     AfterContentInit,
     ChangeDetectionStrategy,
+    ChangeDetectorRef,
     Component,
     ContentChildren,
     ElementRef,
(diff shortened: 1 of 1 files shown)


ashikjs:

SoyDiego:
I don't know if this is related with a previous PR that I did. Please check if it's possible https://github.com/primefaces/primeng/pull/13501/files

Thanks

SoyDiego (approved):

SoyDiego:
> @SoyDiego Thanks for your concern. I checked PR and Issues you mentions. those are not relevent to this PR.
> 
> Actually here I am ensuring that default overlay z-index is negative so it shouldn't do any side effect to others. Previously overlap was not any z-index.

Thanks for double check 👌

cetincakiroglu:
Hi @ashikjs,

We cannot set z-index manually in this case we've got a utility function called zIndexUtils in components/utils, could you please remove the manual CSS and use zIndexUtils? You can check the sidebar or dialog to see how we manage zIndex by using zIndexUtils.

ashikjs:
@cetincakiroglu  ok let me try.

ashikjs:
@cetincakiroglu can you check my PR now. I wrote details about my latest change in description box.

ashikjs:
@gucal and @cetincakiroglu  can you check my PR.

cetincakiroglu (approved):

Issue #10938, opened 2021-12-07
Feature Request: Make activate-on-Enter optional for Inplace


KeithGillette:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[X] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Current behavior**
Currently, the Inplace component always activates when `Enter` is pressed when the component has focus, even if its `preventClick` property is set to `true`.

**Expected behavior**
It would be preferable to make activation on `Enter` conditional to allow full programmatic control via the `activate` method, perhaps by changing `preventClick` to `activateManually` and applying it as an execution guard in the `onKeydown` method.

**What is the motivation / use case for changing the behavior?**
While [my previous feature request](https://github.com/primefaces/primeng/issues/8931) gratefully resulted in partial programmatic control by disabling activate-on-click, the remaining activate-on-enter behavior prevents full programmatic control via the `activate()` and `deactivate()` methods, as the component will always activate on `Enter` when it has focus, overriding manually selected activation and deactivation behavior. Having the option to disable activate-on-Enter using the `activate` method allows full programmatic control and completely custom semantics.

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you reopen the issue so we can include it in our roadmap? 
Please don't forget to add your feedback as a comment after reopening the issue. These will be taken into account by us and will contribute to the development of this feature. Thanks a lot for your understanding!

Best Regards,

KeithGillette:
[...]

PR #13665, opened 2023-09-13
feat: #10938 || Feature Request: Make activate-on-Enter optional for …

ashikjs:
Fix: #10938

### DIFF
--- a/src/app/components/inplace/inplace.ts
+++ b/src/app/components/inplace/inplace.ts
@@ -38,8 +38,8 @@ export class InplaceContent {}
                 <ng-container *ngTemplateOutlet="contentTemplate"></ng-container>
 
                 <ng-container *ngIf="closable">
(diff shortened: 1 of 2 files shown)


ashikjs:

SoyDiego (commented):
Thanks for your contribution @ashikjs :)
Maybe I'm wrong, but the new features generally PrimeNG Team is who decide them or code about them, because If they implement here, also they need to do the same in PrimeReact and PrimeVue and they decide all as a Team. Imagine If every user wants something for his use case, will be complicated

Anyway, I hope they accept your PR, just in case, I wanted to tell you :)

ashikjs:
@SoyDiego  Good point, Thanks for your explanation, I am understand. 
If I get positive reply form  primefaces teams then I am also interested to contribute `PrimeReact ` and ` PrimeVue` on this feature.

PR #13691, opened 2023-09-18
Image: Added srcSet and previewImageSrcSet

### LINKED ISSUE #13683: Image: Add srcset attribute

On the documentation site it says:" Image is used as the native img element and supports all properties that the native element has.", but it is missing **srcset** attribute.
It would be beneficial to support it as well.

Nope

Add srcset and perhaps previewImageSrcset inputs to the component.

I could probably "hack it" in some way by using *ngIf or media queries but it would be better to support srcset attribute as you already use native img element.

### Additional context

_No response_

SoyDiego:
Fix #13683
- Added `srcSet`, `previewImageSrcSet`, `sizes` and `previewImageSizes` properties in the `p-image`component.
- Added documentation about this new properties and improved some of the currently documentation to be clearer.

**Note:** In the next example, the image will change with differents sizes:  `300w`, `500w` and `800w` to show the fuctionality of `srcSet`. When I click on the image, you will see the image is not the same. I did it in this way to show the `previewImageSrcSet` is working too.

## After solution
srcSet image


drilko:
Should "sizes" be added as well? @tomer953 mentioned this in #13683

SoyDiego:
> Should "sizes" be added as well? @tomer953 mentioned this in #13683

Added `srcSet`, `previewImageSrcSet`, `sizes` and `previewImageSizes`.

Thanks!

PR #13720, opened 2023-09-23
fix(inputNumber): improve maxlength behaviour

### LINKED ISSUE #13704: Component: InputNumber with maxlength attribute
### Describe the bug

When p-inputnumber has maxlength attribute and the max number of digits is entered, I can't select the whole number to overwrite it.

My code:
`
<p-inputNumber [useGrouping]="false" inputId="code" formControlName="code" maxlength="4"></p-inputNumber>
`

When I enter 4 digits into the input, then select them all and try to input another number to overwrite everything, nothing happens. If less than 4 digits are entered, everything works fine.

### Environment

Angular application with PrimeNg.

### Reproducer

_No response_

### Angular version

16.2.3

### PrimeNG version

16.3.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.14.0

### Browser(s)

Chrome 116.0.5845.188

1. Create p-inputnumber input field.
2. Set maxlength attribute on the field to 4.
3. Enter 4 digits into the field.
4. Select all digits.
5. Try to enter a number to overwrite the digits in the field.

The digits should be overwritten by the new pressed number, but nothing happens.

### Expected behavior

I expect the selected digits to be overwritten by new input from keyboard.

dalenguyen:
### Defect Fixes
Fix #13704 

The solution is to check if the user currently selects all numbers with maxLength of not. If yes, replace it will a new entered number.


ashikjs:
@dalenguyen  can you squash those two commit to  a single commit.

cetincakiroglu:
Hi @dalenguyen,

PR #13726, opened 2023-09-24
Fixed #13725 - Speed Dial Items only visible when dial is visible

### LINKED ISSUE #13725: SpeedDial | Menu Items are only hidden but still in view on closed dial
### Describe the bug

At the moment the items of the Speed Dial are hidden but still in view and it is blocking clicks on the webpage at the location where the items would be displayed.
Screenshot 2023-09-24 at 19 38 27

with the fix the items only show when the speed dial is expanded and are totally invisible until the dial is opened
Screenshot 2023-09-24 at 19 39 00

### Environment

macOS Sonoma 14.0
Angular CLI 16.2.1
PrimeNG 16.3.1

### Reproducer

_No response_

### Angular version

16.2.1

### PrimeNG version

16.3.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

20.5.1

### Browser(s)

Chrome

Just see https://primeng.org/speeddial

### Expected behavior

_No response_

EnricoMessall:
Fixed #13725 - Speed Dial Items only visible when dial is visible

### DIFF
--- a/src/app/components/speeddial/speeddial.ts
+++ b/src/app/components/speeddial/speeddial.ts
@@ -86,7 +86,7 @@ import { asapScheduler } from 'rxjs';
                     [attr.data-pc-section]="'menuitem'"
                 >
                     <a
-                        *ngIf="isClickableRouterLink(item); else elseBlock"
+                        *ngIf="_visible && isClickableRouterLink(item); else elseBlock"
                         pRipple
                         [routerLink]="item.routerLink"
                         [queryParams]="item.queryParams"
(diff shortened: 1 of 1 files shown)


src/app/components/speeddial/speeddial.ts
```diff
@@ -86,7 +86,7 @@ import { asapScheduler } from 'rxjs';
                     [attr.data-pc-section]="'menuitem'"
                 >
                     <a
-                        *ngIf="isClickableRouterLink(item); else elseBlock"
+                        *ngIf="_visible && isClickableRouterLink(item); else elseBlock"
```
volvachev, line 89: Hi, mb use only one `ngIf`?

```html
<ng-container *ngIf="_visible">
  <a *ngIf="isClickableRouterLink(item); else elseBlock"></a>
  <ng-template #elseBlock><a> </a></ng-template>
</ng-container>
```

EnricoMessall: Hey,you right would be a cleaner solution. But as you are saying it, maybe its better to pack it into the isClickableRouterLink method isn't it? Then the hole method would return false if not visible and blocks also other points where the method might be used.

SoyDiego (approved):

cetincakiroglu (approved):

cetincakiroglu:
Hi @EnricoMessall,

Issue #13707, opened 2023-09-20
Toast: closeIcon on message is outside of button


ethanolree:
### Describe the bug

Adding a customIcon to toast messages causes the icon to appear outside of the button

![image](https://github.com/primefaces/primeng/assets/112996028/80245a6c-dd6a-4539-a5e1-faf7d37c3c57)

![image](https://github.com/primefaces/primeng/assets/112996028/ac8a4a58-19ca-4eda-b20b-1a97b537b22c)

### Environment

Angular with PrimeNG

### Angular version

16.2.5

### PrimeNG version

16.3.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18

### Steps to reproduce the behavior

Add a closeIcon to any toast message

### Expected behavior

Icon will appear inside of the toast button

ashikjs:
@ethanolree can you make any reproducible code and share  with us

ethanolree:
@ashikjs 

HTML:
<p-toast></p-toast>

TS:
this.messageService.add({severity: 'success', summary: 'Success', detail: 'Message Content', closeIcon: 'pi pi-heart'});

_**any**_ toast using closeIcon

ashikjs:
@ethanolree ya, got it. Thanks for your response. 
I solve it and make a PR for this issues and waiting to approved from Primeng Core team.

Btw: you should use only icon name like `pi-heart` or `pi-times`.

PR #13727, opened 2023-09-25
fix: #13707 || Toast: closeIcon on message is outside of button

ashikjs:
### Defect Fixes
Fix: #13707


cetincakiroglu:
Hi @ashikjs,

Thanks a lot for your support and effort, we're appreciate it!

PR #13745, opened 2023-09-26
CQI #13742 Step aria-controls showcase

### LINKED ISSUE #13742: Accessibility | Step component bugs and improvements
### Describe the bug

The Step component has some accessibility problems:

1. When navigated with a screen reader, this will read each step two times: one for the `li` element and one for the `a` inside, adding links information such as `visited` and so on. The correct behaviour is the same one as the `TabView` component, so the aria tags and role should be switched between the `a` (role `tab` and all the aria tags) and the `li` (role `presentation` and no aria tags).
2. Currently the steps will never receive focus when navigating with keyboard only. You have to select an element with the mouse to apply focus and then the arrows navigation is available.  The correct behaviour is the same one as the `TabView` component, so when tabbing in the Stepper the focus should go to the active step (the only one with `tabindex=0`), a consequent tab should leave the Stepper and go to the next focusable element on the page. Shift-tabbing back to the stepper should return focus to the active step. 
3. Currently if you move focus with the arrows il will correctly change the tabindex of the focused element and remove it from the previously focused one. When tabbing out and back again, the focus should go to the active tab, not the last focused tab with arrows. 
4. Disabled or readonly steps don't have `aria-disabled` as they should.
5. Disabled or readonly steps still have `aria-current='step'`, that makes the screen reader read them as they were interactable. The aria tag should be present only when not disabled or readonly. 

Also, since the Stepper has role `tablist` and each step role `tab`, they should also have `aria-controls` with the id of the controlled element, that in turn should have `aria-labelledby`. This is more complicated to handle, since the current implementation of the Step component doesn't have a concept of step content. 
This could be solved by adding a new optional property to the MenuItem: `overrideAriaControls` or something like that, that delegates to the user the responsibility to correctly link the MenuItem to the actual id of the controlled element. While this would be simple to handle on the Step level, it would be more complicated for all the other components that use a MenuItem, as they should be updated to use this property when present instead of the automatically set one (eg: see `TabView` that already correctly implements `aria-controls` etc.)

I will open two different PRs, one for all the 5 points above and one to showcase the implementation of the `aria-controls` property in the `MenuItems` for the Step component only. If accepted, it should be ported to all other components too.  

### Environment

Windows

### Reproducer

https://primeng.org/steps/

### Angular version

16.X

### PrimeNG version

16.3.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.13

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

GiacomoDM:
Fix #13742 (doesn't actually fix, it's just for correct liking with the issue)

This PR showcase the usage of the MenuItem's new property `overrideAriaControls`. This allows to pass the MenuItem the `id` of the element that the step controls, as shown by the routing example in the Step component doc. 

This is just an example and it's difficult to interact with it, because of the previous problems with the Step component that are to be fixed with https://github.com/primefaces/primeng/pull/13743. 

If this solution will be accepted, it will require a porting (or at least an evauation to decide if it's needed) to all other components that uses the MenuItem and have an aria-controls attribute.

### DIFF
--- a/src/app/components/api/public_api.ts
+++ b/src/app/components/api/public_api.ts
@@ -21,6 +21,7 @@ export * from './selectitemgroup';
 export * from './shared';
 export * from './sortevent';
 export * from './sortmeta';
+export * from './stepitem';
 export * from './tablestate';
 export * from './translation';
 export * from './translationkeys';
--- a/src/app/components/api/stepitem.ts
+++ b/src/app/components/api/stepitem.ts
@@ -1,0 +1,12 @@
+import { MenuItem } from './menuitem';
+
+/**
+ * StepItem provides all the properties from MenuItem, plus some custom ones.
+ * @group Interface
+ */
+export interface StepItem extends MenuItem {
+    /**
+     * Id of the controlled element. Allows to set the aria-controls attribute for the step. Note that the id of the controlled element must be set manually in the controlled element.
+     */
+    ariaControls?: string;
+}
--- a/src/app/components/steps/steps.ts
+++ b/src/app/components/steps/steps.ts
@@ -3,7 +3,7 @@ import { ChangeDetectionStrategy, ChangeDetectorRef, Component, ElementRef, Even
 import { ActivatedRoute, Router, RouterModule } from '@angular/router';
 import { DomHandler } from 'primeng/dom';
 import { Nullable } from 'primeng/ts-helpers';
-import { MenuItem } from 'primeng/api';
+import { StepItem } from 'primeng/api';
 import { TooltipModule } from 'primeng/tooltip';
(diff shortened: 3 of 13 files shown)


GiacomoDM:
Hi @cetincakiroglu, 
I have updated the PR. I extended MenuItem in a new interface StepItem, that only adds the aria property. This way it will always inherit any changes made to the MenuItem and work as expected. 
I also updated the docs. 

Let me know if this solution is acceptable for you.

cetincakiroglu:
Hi @GiacomoDM,

Could you please update the PR and resolve the conflicts so we can merge safely?

cetincakiroglu:
Closing due to inactivity.

PR #13771, opened 2023-09-28
Feature : Add virtual scroll on Tree Select Component

GaetanRouzies:
### Feature Requests
Add virtual scroll on Tree Select Component so we can use the Tree Select with large datasets.

Fixed https://github.com/primefaces/primeng/issues/13772


src/app/showcase/service/nodeservice.ts
```diff
@@ -763,6 +764,31 @@ export class NodeService {
         ];
     }
 
+    getDynamicTreeNodes(parentCount: number, childrenCount: number): TreeNode[] {
```
SoyDiego, line 767: Thanks for your contribution.
I would like to suggest this code. Similar to your code but maybe a little bit more clear IMO.
If you add it, test it first because I'm writting without try it. :)

```
getDynamicTreeNodes(parentCount: number, childrenCount: number): TreeNode[] {
    const nodes: TreeNode[] = [];

    for (let index = 0; index < parentCount; index++) {
        const children: TreeNode[] = [];

        for (let j = 0; j < childrenCount; j++) {
            children.push({
                key: `${index}-${j}`,
                label: `Child ${index}-${j}`,
                selectable: true,
            });
        }

        nodes.push({
            key: index.toString(),
            label: `Parent ${index}`,
            selectable: true,
            children: children,
        });
    }

    return nodes;
}
```

GaetanRouzies:
@SoyDiego Thanks for your feedback, I implemented your code in [my last commit].(https://github.com/primefaces/primeng/pull/13771/commits/0f8e514ee5974cf21b0b7edf854e39338da46a70)

src/app/components/treeselect/treeselect.css
```diff
@@ -94,4 +94,12 @@
     .p-treeselect-clearable {
         position: relative;
     }
+
+    .p-treeselect-virtual-scroll .p-tree.p-component {
+      padding: 0;
+    }
+
+    .p-treeselect-virtual-scroll .p-tree-container.p-scroller-content {
+      padding: 1.25rem;
```
SoyDiego, line 103: If I'm not wrong this styles themes should be in `primeng-sass-theme` project: https://github.com/primefaces/primeng-sass-theme/
Maybe @cetincakiroglu can confirm you

GaetanRouzies: Can i have a confirmation ?

GaetanRouzies:
Still waiting for confirmation

cetincakiroglu:
Hi @GaetanRouzies,

Huge thanks for the effort and support!

Issue #13789, opened 2023-09-30
Menu Components: Add icon template


ianvaernet:
### Describe the feature you would like to see added

A template to customize the icons in menu components (Menu, Menubar, MegaMenu, PanelMenu, SlideMenu and TieredMenu) so they can be used with icon libraries that require to render a custom component instead of using classes.

E.g: [FontAwesome Angular](https://fontawesome.com/docs/web/use-with/angular) where icons are used like `<fa-icon [icon]="faCoffee"></fa-icon>`

### Describe the solution you'd like

Be able to use custom icon templates like this

HTML:
```
<p-menu [model]="customIconItems">
    <ng-template pTemplate="icon" let-icon>
        <fa-icon style="margin-right: 6px" [icon]="icon"></fa-icon>
    </ng-template>
</p-menu>
```

TS:
```
import { faPlus, faTrash } from '@fortawesome/free-solid-svg-icons';

this.customIconItems = [
    {
        label: 'New',
        icon: faPlus
    },
    {
        label: 'Delete',
        icon: faTrash
    }
];
```

ant2400:
I am looking for this same feature. I notice in the documentation that it explains that you can use ng-templates in the components, using the p-dropdown as an example. But this doesn't work in the p-menu, and I checked through the source code, it doesn't look like it would support it.

cetincakiroglu:
Hi,

We've introduced item template for the menu components. It covers icon of the item too and makes this feature unnecessary. Closing the issue because of that. 

Thanks a lot for the report!

AStoker:
Hate to resurrect this, but before I open a new defect, there seems to be a shortcoming regarding adding icons to the group label. Adding icons to the MenuItems is great, but I need an icon for the label that groups multiple items together. Code wise it looks like it just spits out the label and that's it. Would be great if we had some templating options for the MegaMenu group label.

AntoninoBonanno:
I haven't found any other way than using `<ng-template #item let-item>`, but this has the disadvantage of having to implement the logic already included in the library.

Using the official `fa-icon` tag is very useful and optimized, especially when deciding which icons to load individually.

The ideal would be to centralize the ability to override icons, perhaps with an injectable component instead of the basic icon component.

An example might be something like this: [https://ng-bootstrap.github.io/#/components/datepicker/examples#adapter]( https://ng-bootstrap.github.io/#/components/datepicker/examples#adapter)

@cetincakiroglu Everyone could create their own icon override, which would definitely solve the problem.

PR #13790, opened 2023-09-30
Feature #13789 Add icon template for Menu components

ianvaernet:
Feature #13789 

This PR adds a template to customize the icons in menu components (Menu, Menubar, MegaMenu, PanelMenu, SlideMenu and TieredMenu) so they can be used with icon libraries that require to render a custom component instead of using classes.

E.g: [FontAwesome Angular](https://fontawesome.com/docs/web/use-with/angular) where icons are used like `<fa-icon [icon]="faCoffee"></fa-icon>`

I've added a working example of each component under "Basic" section in docs in https://github.com/ianvaernet/primeng/tree/menu-icon-template-examples but I haven't included those examples in this PR given that their use may not be that frequent and they require an external library.

image

### DIFF
--- a/src/app/components/megamenu/megamenu.interface.ts
+++ b/src/app/components/megamenu/megamenu.interface.ts
@@ -10,6 +10,16 @@ export interface MegaMenuTemplates {
      */
     start(): TemplateRef<any>;
     /**
+     * Custom template of icon.
+     * @param {Object} context - item data.
+     */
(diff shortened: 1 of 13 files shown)


cetincakiroglu:
Hi,

We've introduced item template for the menu components. It covers icon of the item too and makes this feature unnecessary. Closing the issue because of that. 

PR #13836, opened 2023-10-09
Fix #13816 | MultiSelect - Delete button is not visible on chips with long label text

### LINKED ISSUE #13816: Multiselect: Chips: Delete button not visible if label is too wide
### Describe the bug

Given:
* a multiselect with display="chip"
* a max-width

When:
* An option with a wide label is selected

Then:
*  the overflow is hidden 
* the delete button on the right side of the chip is no longer visible
image

### Environment

Angular 16 app created with the CLI.

### Reproducer

https://github-gwnhkb.stackblitz.io

### Angular version

16.2.8

### PrimeNG version

16.3.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

N/A

### Browser(s)

Chrome

Select an option with a wide label.

### Expected behavior

The delete button in the chip should always be visible. The text should be shortened/hidden, maybe with overflow ellipsis.

mehmetcetin01140:
Fixes https://github.com/primefaces/primeng/issues/13816

Add missing css.


src/app/components/multiselect/multiselect.css
```diff
@@ -42,7 +42,12 @@
     .p-multiselect-token-icon {
         cursor: pointer;
     }
-
+    .p-multiselect-token-label {
```
SoyDiego, line 45: You have to format the code. Try `npm run format` because is not like the other styles

PR #13911, opened 2023-10-18
Fixed #13910 - Menu Templating

### LINKED ISSUE #13910: Menu Templating
Menu templating

gucal:
fixed #13910

### DIFF
--- a/src/app/components/api/menuitem.ts
+++ b/src/app/components/api/menuitem.ts
@@ -1,6 +1,6 @@
 import { QueryParamsHandling } from '@angular/router';
-import { TooltipOptions } from './tooltipoptions';
 import { MegaMenuItem } from './megamenuitem';
+import { TooltipOptions } from './tooltipoptions';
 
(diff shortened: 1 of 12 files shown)


cetincakiroglu (approved):

Michael-Ziluck:
This breaks all existing usage of the template for tab menus. Previously We only needed to do the following:
```html
<p-tabMenu [(activeItem)]="currentTab" [model]="tabItems">
  <ng-template pTemplate="item" let-item>
    {{ item.label | translate: I18N_PRODUCT }}
  </ng-template>
</p-tabMenu>
```

Which would result in this display:
image

Now with the same markup, it results in this:
image

The change is because previously, the `<ng-container>` was within the `<a>` tag. I know you updated the docs to reflect this, but shouldn't breaking changes be constrained to major version changes? Seems wildly arbitrary and inconvenient for users to have to make the change. I'll be opening a bug issue as well.

Issue #13908, opened 2023-10-18
splitbutton: Accessibility does not appear to work as documented


ppc-harmoney:
### Describe the bug

Our application requires WCAG compliance. I am currently trying to add aria-labels to a splitButton component, for which I am following the [documentation](https://primeng.org/splitbutton#accessibility).

```
    <p-splitButton
        *ngIf="shouldShowSplit"
        [model]="menuItems"
        (onClick)="onAddClicked()"
        label="{{ 'BUTTON.add' | translate }}"
        buttonProps="{'aria-label': 'Add document'}"
        menuButtonProps="{'aria-label': 'More upload options'}"
        styleClass="split-button p-button-lg"
        menuStyleClass="split-button-menu"
        appendTo="body"
    />
```

However, I do not see these changes reflected in the DOM. Testing using a screenreader also indicates that these attributes were not set. Furthermore, when trying to bind with a template expression (i.e. square brackets) the compiler complains that these properties do not exist on the splitButton. 

### Environment

Angular 16.2.9

Cannot set up a test case because the issue template appears to be broken.

### Angular version

16.2.9

### PrimeNG version

16.5.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.18.2

### Steps to reproduce the behavior

Add buttonProps/menuButtonProps to p-splitButton

### Expected behavior

I would expect to see the aria-label properties on the corresponding elements.

PR #13921, opened 2023-10-18
fix(splitbutton): Accessibility does not appear to work as documented…

eece:
Fixes [13908](https://github.com/primefaces/primeng/issues/13908)


cetincakiroglu (approved):

PR #13922, opened 2023-10-18
fixed #13901 pButton | Loading icon problem

### LINKED ISSUE #13901: pButton | Loading icon problem
### Describe the bug

For buttons with the directive `pButton`, when the loading property changes and the loading icon is removed, an empty element take its place. This element has a margin that gives the button an empty space next to the label where the loading icon was: 
pbutton

image

### Environment

Windows

### Reproducer

https://stackblitz.com/edit/gmhqyu-xzl3yd?file=src%2Fapp%2Fdemo%2Fbutton-loading-demo.html

### Angular version

16.X

### PrimeNG version

16.5.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.16

### Browser(s)

_No response_

1. Open the provided stackblitz
2. The two buttons have the same spacing around the label
3. click one of the two buttons
4. wait for the loader to stop
5. the button with the directive `pButton` now has additional spacing where the loading icon was. Furthermore, the remaining `span` has a css class named: `.undefined'`

### Expected behavior

When the icon is removed, the `span` containing it should be removed as well.

eece:
Fixes #13901

### DIFF
--- a/src/app/components/button/button.ts
+++ b/src/app/components/button/button.ts
@@ -223,7 +223,7 @@ export class ButtonDirective implements AfterViewInit, OnDestroy {
     }
 
     getIconClass() {
-        return this.loading ? 'p-button-loading-icon ' + (this.loadingIcon ? this.loadingIcon : 'p-icon') : this.icon;
+        return this.loading ? 'p-button-loading-icon ' + (this.loadingIcon ? this.loadingIcon : 'p-icon') : this.icon || 'hidden';
     }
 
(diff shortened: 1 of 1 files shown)


SoyDiego (approved):

eece:
> Could you please use `p-hidden` class instead of `hidden`. Because hidden comes from primeflex and p-hidden comes from common.css included in primeng package.

I changed 'hidden' class to 'p-hidden'.

PR #13947, opened 2023-10-24
fix: fix galleria tabbing issue

### LINKED ISSUE #13945: Galleria: Cannot read properties of undefined when tabbing through component
### Describe the bug

When tabbing through a galleria component, the tabbing will cause an unhandled exception stating `Cannot read properties of undefined (reading: 'code')` which is occurring within the `onIndicatorKeyDown` method of the `GalleriaItem`.  This issue causes the tabbing to get stuck within the galleria item, and not progress to the next item on the page.

Proposed Fix:
https://github.com/primefaces/primeng/blob/21e0575123e550058729f6a0347dcede91771fb9/src/app/components/galleria/galleria.ts#L677

The line above should be changed to `$event` instead of `event` so that the event object passed to the `onIndicatorKeyDown` method isn't `undefined`

image

### Environment

PrimeNG 16.5.1 and Angular 16.0.5

### Reproducer

_No response_

### Angular version

16.0.5

### PrimeNG version

16.5.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.14.0

### Browser(s)

Chrome 118

1.) On a page utilizing an existing `p-galleria` component with multiple images, tab into the component.
2.) Once you tab through all the navigator "dots" on the component, the error mentioned in the description will be shown in the console, and you'll be unable to tab out of the component.

### Expected behavior

The console error shouldn't be thrown, and the user should be able to tab out of the `p-galleria` component.

navedqb:
Fixes  #13945

### DIFF
--- a/src/app/components/galleria/galleria.ts
+++ b/src/app/components/galleria/galleria.ts
@@ -674,7 +674,7 @@ export class GalleriaItemSlot {
                     tabindex="0"
                     (click)="onIndicatorClick(index)"
                     (mouseenter)="onIndicatorMouseEnter(index)"
-                    (keydown)="onIndicatorKeyDown(event, index)"
+                    (keydown)="onIndicatorKeyDown($event, index)"
                     [ngClass]="{ 'p-galleria-indicator': true, 'p-highlight': isIndicatorItemActive(index) }"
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi,

Thanks for the contribution!

Issue #13709, opened 2023-09-21
ColorPicker: Setting Hex / RGB / HSB code in component 


t4keo:
### Describe the feature you would like to see added

I would like to be able to paste the Hex Code in the component and have the color result being shown in the gradient before validating.

### Describe the solution you'd like

Have ability to hide/show inputs adding the ability to set HEX and/or RGB and/or HSB code in the component. I could add an input outside but the design would be better if its inside. 

### Describe alternatives you have considered

An alternative would be the ability to access template and be able to modify it.

### Additional context

Example from powerpoint
![image](https://github.com/primefaces/primeng/assets/6909701/ee452157-ef4e-4f4f-bb0d-e2d668711d59)

ondriiicek:
Hi, can you assign me to this issue?

PR #14090, opened 2023-11-11
#13709 ColorPicker: Setting Hex / RGB / HSB code in component

ondriiicek:
In this PR, the colorpicker component was modified. Select was added so that the user can choose which color model he wants to use. Inputs have been added for each color model (RGB, HEX, HSB) where the user can enter the color code directly in the component. Two buttons were also added for confirming the color change and resetting to the default color.

Issue: https://github.com/primefaces/primeng/issues/13709

### DIFF
--- a/src/app/components/colorpicker/colorpicker.css
+++ b/src/app/components/colorpicker/colorpicker.css
@@ -17,6 +17,76 @@
         height: 166px;
     }
 
+    .picker-inputs {
+        margin-bottom: 10px;
+    }
+
+    .picker-inputs input {
+        outline: none;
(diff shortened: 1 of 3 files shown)


SoyDiego:
Hi, can you provide a description, screenshot, issue or something to understand your PR?
Thanks :)

ondriiicek:
Hi, it is for school purposes, in one of our university course we should find github issue, solve it and create PR.
In primeng project i found this issue: https://github.com/primefaces/primeng/issues/13709 and solve it by creating something like this:

image

Issue #14087, opened 2023-11-10
[Calendar] issue when clicking on "today" button it doesn't navigate to the current date.


cdskill:
### Describe the bug

When the calendar is display in inline mode with showButtonBar set to true when I navigate through the navigate button in previous on month (for e.g) and then I click on today I'm excpecting to display the current date and get back to the current month, current day.

### Environment

all

### Reproducer

https://stackblitz.com/edit/reszae?file=src%2Fapp%2Fdemo%2Fcalendar-inline-demo.html

### Angular version

16.2.0

### PrimeNG version

16.7.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20

### Browser(s)

chrome

### Steps to reproduce the behavior

Stackblitz :)

### Expected behavior

I'm excpecting when I click on "today" the calendar should display the current date.

ashikjs:
@cdskill it's working as expected. I can see it show today date as selected with selected color.

PR #14098, opened 2023-11-13
fix: #14087 || [Calendar] issue when clicking on "today" button it do…

ashikjs:
fix: #14087


cetincakiroglu (approved):

Issue #14119, opened 2023-11-15
TypeError: this.focusedItemInfo.mutate is not a function


HarunDayan:
### Describe the bug

Hi, after I upgraded Angular to its latest version, I started getting this error. I think it's because the mutate function has been deprecated from WritableSignal.  

CascadeSelect - Menubar- PanelMenu- MegaMenu- SlideMenu- TieredMenu

### Environment

Angular: 17.0.2
PrimeNG: 16.7.2
Nx/Angular: 17.1.1 

### Angular version

17.0.2

### PrimeNG version

16.7.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.17.0

cetincakiroglu:
fixed in https://github.com/primefaces/primeng/commit/465482e08e98f01811c8062831ee14ddc85323b2

PR #14150, opened 2023-11-18
Fix issue #14119

antonionesta:
Changed mutate function to update function for WritableSignal updates as mutate function is going to be deprecated since angular 17.


SoyDiego (approved):

cetincakiroglu:
Hi,

We've already changed the all `mutate` functions to `set` in v17. Closing the PR since it's already done.

Issue #14309, opened 2023-12-08
p-calendar


jaggi510:
### Describe the bug

Issue with year picker. 
When you have min date and max date on the p-calendar within a year. It is not allowing to select 2023 and 2024. It is only allowing 2024. Year 2023 is disabled. Once we pick a date in 2024 and then try to change back to a date in 2023, Only way to do this is by clicking on month navigator as year 2023 is disabled on year selection.

### Environment

angular 16.2.11, with primeng 16.6.0 

### Reproducer

https://stackblitz.com/edit/github-mz6vwk?file=src%2Fapp%2Fapp.component.html

### Angular version

16.2.11

### PrimeNG version

16.6.0 

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

v20.2.0

### Browser(s)

Version 120.0.6099.71

### Steps to reproduce the behavior

set min date as today
max date as today + 1 year

on p-calendar set minDate and maxDate. 

### Expected behavior

We should be able to select year 2023 and 2024. But only 2024 is selectable. If you select any date in 2024, and then click on year, We should be able to see both 2023 and 2024 as selectable.

Draganlazarev90:
Still present in 17.4

cetincakiroglu:
Hi,

Thanks a lot for reporting the issue. This issue is duplicate of https://github.com/primefaces/primeng/issues/13919
Related issue is fixed and the fix will be available with upcoming release (17.12.0).

Closing this one.

PR #13920, opened 2023-10-18
fix(calendar): added checking for each month when checking the disabled year

alexandergu:
Fixes #13919 

When checking the year for disabled status, now each month is checked for disability. 
An optional year parameter has been added to the isMonthDisabled method

PR #14378, opened 2023-12-16
issue: 14309

MichaelG824:
### Defect Fixes
https://github.com/primefaces/primeng/issues/14309

Screenshot 2023-12-15 at 7 28 09 PM

Using new Date('2023-12-15T01:43:14.818Z') as the minDate and maxDate = new Date('2024-12-16T01:43:14.818Z')
results in the screenshot above. We can't use day as 1 on the year view because we may have a min date that has a day that is higher than 1 and is on the same month and year.


cetincakiroglu:
Hi,

Thanks for the effort and support. This issue has been fixed by https://github.com/primefaces/primeng/pull/13920

PR #14382, opened 2023-12-17
Issue 14304 - Improve check for selecting option template in Dropdown

### LINKED ISSUE #14304: Dropdown: option is empty when it has group property and dropdown has [group]="false"
### Describe the bug

I'm reopening the following issue #14276 because I think I didn't provide a clear explanation for my problem and the team member misunderstood me.

We're want to update from v16.6.0 to latest version of PrimeNG.

We have models in our app with a `group` property. Currently we using a lot of dropdowns that are using these models as options, some of these dropdowns using `group` property as a `optionLabel` or `optionValue`, BUT we don't use grouped options. After updating to latest version dropdowns render these options as empty:

image
image

It happens because internally PrimeNG decides which template to use based on existence of `group` property:

image

But I think not to break existing behavior this check should be handled the same way as in MultiSelect and Autocomplete components by checking multiple properties:

image
image

As I see Dropdown component even has the same function implementation, it should be possible just to replace current check with this function call. For me it seems only logical that if `@Input() group` is set to `false` then it shouldn't choose group template at all.
And to rename all models and their usages or come up with some other solution just to return the previous behavior is far from ideal.

### Environment

Any environment

### Reproducer

https://stackblitz.com/edit/ee5k5j?file=src%2Fapp%2Fdemo%2Fdropdown-basic-demo.html,src%2Fapp%2Fdemo%2Fdropdown-basic-demo.ts

### Angular version

17.0.0

### PrimeNG version

17.0.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.18.0

### Browser(s)

_No response_

1. Add a `group` property to any option.
2. Add `[group]="false"` to Dropdown component.

The option is rendered empty even though we're not interested in grouped options.

https://stackblitz.com/edit/ee5k5j?file=src%2Fapp%2Fdemo%2Fdropdown-basic-demo.ts

And here's an example from v16.6.0
https://stackblitz.com/edit/dc7jhu?file=src%2Fapp%2Fdemo%2Fdropdown-basic-demo.ts,package.json

### Expected behavior

The items should be rendered OR the docs should warn that group cannot be used as a property.

tsinevik:
### Defect Fixes
Fix #14304

Use `isOptionGroup` method when selecting template to render option. Otherwise options may be rendered empty if they have `group` property, but Dropdown component has `[group]="false"`. It is the same logic as in MultiSelect and Autocomplete components.


SoyDiego (approved):

tsinevik:
@cetincakiroglu hi, I would really appreciate if you could check out this PR and provide some feedback if there's something wrong with it. We can't update to the Angular 17 for over a month because of that issue and we really don't won't to refactor part of the app only for dropdown to remain the same behavior on a new version.

PR #14395, opened 2023-12-19
SplitButton | Flag for hiding menu on default button click.

### LINKED ISSUE #13962: splitButton: close dropdown when default button is clicked

When opening the dropdown on a splitbutton, and then clicking the default the dropdown stays open. 

_No response_

In my opinion this menu should close automatically as soon as any button is clicked (not just a menu item). If this interferes with other expectation then I would like to see a config option or a function that we can call on the split button component.

Not tested yet, but I presume we can use viewChild and call isExpanded.set(); ourselves.

### Additional context

_No response_

kkoripl:
Continue on fix #13962 from #13983.

So right now, when someone would like to have same behaviour on default button as it was for the menu button, it was impossible due to hiding menu after event handling. 

One can see that here: https://stackblitz.com/edit/8tkckf?file=src%2Fapp%2Fdemo%2Fsplit-button-basic-demo.html

After the change, with setting the new flag to true, the same behaviour on default and menu button will be possible.
Probably there can be a flag to have same behaviour for both buttons, but... with onClick still possible to be set, both setups would be against each other.

**BEFORE:**
default-behaviour-without-fix

**AFTER:**
default-behaviour-fix

### DIFF
--- a/src/app/components/splitbutton/splitbutton.ts
+++ b/src/app/components/splitbutton/splitbutton.ts
@@ -166,6 +166,10 @@ export class SplitButton {
      */
     @Input() menuButtonProps: MenuButtonProps | undefined;
     /**
+     * When true menu is hidden on default button click, otherwise just event is handled
+     */
+    @Input() hideMenuOnDefaultButtonClick = true;
+    /**
      * Callback to invoke when default command button is clicked.
      * @param {MouseEvent} event - Mouse event.
      * @group Emits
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi,

This feature is not necessary, splitbutton should only open the menu when clicking the menu trigger button. Otherwise, it would be the same as popup menu and splitbutton would be unnecessary. This behavior also can be done by utilizing popup feature of the menu component.

PR #14433, opened 2023-12-26
Fixed #14432

### LINKED ISSUE #14432: Error loading dropdown menu after performing a filter on the field programmatically
### Describe the bug

Error loading dropdown menu after performing a filter on the field programmatically, the test was reproduced on versions '16.9.2' and '17.2.0'

### Environment

indifferent

### Reproducer

https://stackblitz.com/edit/github-8nu6jf?file=src%2Fapp%2Fapp.component.html

### Angular version

17.x.x

### PrimeNG version

17.2.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

x.x.x

### Browser(s)

_No response_

how to reproduce the error:

1. load the list:
image

2. performs the filter using

`  filterDate() {
    this.dataTableComponent.filter(
      new Date('2023-11-14'),
      'date',
      FilterMatchMode.DATE_IS
    );
  }`

or

`  filterMessage() {
    this.dataTableComponent.filter('test 1', 'message', FilterMatchMode.EQUALS);
  }`

3. opens the field filter menu and the error appears when the filter options are not loaded, we can also see the problem in the console

image

### Expected behavior

load the filled filter options
image

FabioLucenaRibas:
Performing the fix for the filters and maintaining the global filtering behavior

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -2033,7 +2033,7 @@ export class Table implements OnInit, AfterViewInit, AfterContentInit, Blockable
             clearTimeout(this.filterTimeout);
         }
         if (!this.isFilterBlank(value)) {
-            this.filters[field] = { value: value, matchMode: matchMode };
+            this.filters[field] = field == 'global'? { value: value, matchMode: matchMode } : [{ value: value, matchMode: matchMode }];
         } else if (this.filters[field]) {
             delete this.filters[field];
         }


cetincakiroglu (approved):

cetincakiroglu:
Hi,

You should use filterGlobal method instead of modifying filter method, it's not for global filtering.

I've reverted this PR in https://github.com/primefaces/primeng/pull/15053 because it introduced new breaking changes in #14588 and #14903

fabiolucenaribas:
@cetincakiroglu 

This is a filter-specific correction, solving the problem when there are multiple filters for the same field (complex), however I understand that it has generated an unintended impact on simple filters

this.filters[field] = _**field == 'global'?**_ { value: value, matchMode: matchMode } _**: [{ value: value, matchMode: matchMode }];**_

```typescript
    filter(value: any, field: string, matchMode: string) {
        if (this.filterTimeout) {
            clearTimeout(this.filterTimeout);
        }
        if (!this.isFilterBlank(value)) {
            this.filters[field] = { value: value, matchMode: matchMode };
            this.filters[field] = field == 'global'? { value: value, matchMode: matchMode } : [{ value: value, matchMode: matchMode }];
        } else if (this.filters[field]) {
            delete this.filters[field];
        }
        this.filterTimeout = setTimeout(() => {
            this._filter();
            this.filterTimeout = null;
        }, this.filterDelay);
        this.anchorRowIndex = null;
    }
    filterGlobal(value: any, matchMode: string) {
        this.filter(value, 'global', matchMode);
    }


```
The mention of the '**global**' type within the filter function is to maintain the existing behavior given that the 'filterGlobal' function calls the 'filter'

```typescript
    filterGlobal(value: any, matchMode: string) {
        this.filter(value, 'global', matchMode);
    }

```

It is necessary to look for a solution that meets both scenarios of #14433 and #14903, 
#14433 In the example mentioned, the error occurred when the field in the interface has more than one filter

image

Expected outcome:

image

image


*I'm using the English translator, sorry if I don't express myself well

PR #14176, opened 2023-11-21
Fix issue #14163

antonionesta:
Fixes #14163

PR #14464, opened 2024-01-03
Fix: Issue #14163 - AutoComplete: ArrowDown key no longer opens up the dropdown

### LINKED ISSUE #14163: AutoComplete: ArrowDown key no longer opens up the suggestions panel
### Describe the bug

In PrimeNg 16.6.0, pressing the ArrowDown key in the AutoComplete control box would open up the suggestions panel with any previously loaded suggestions.

With 16.7.0 and above does not happen. You have to type some some actual text to get it to show. If you were to press ESC, the panel closes, and ArrowDown can no longer be used to show it again

### Environment

Angular and PrimeNG, running on a Windows 11 box

### Reproducer

https://stackblitz.com/edit/nrt6yx?file=package.json

### Angular version

16.2.12

### PrimeNG version

16.7.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.6.1

### Browser(s)

Edge

1. Go to the Stackblitz page shown
2. Click on the searchbox to focus on it
3. Click the Arrow Down
4. Type "un"
5. Press ESC (to close the suggestion panel)
6. Click the Arrow Down again

### Expected behavior

For steps 3 and 6, the suggestions panel should appear, with a list of countries (For step 3, all countries should be listed. For step 6, the 3 matching countries would be listed).

The expected behaviour works fine in 16.6.0

timandella:
This should fix #14163 

**Notes**

There is already a pull request to fix the issue - https://github.com/primefaces/primeng/pull/14176

However, that PR has not been approved, due to it containing fixes for other files, and the author has not responded for about a month now. As the original issue is blocking the company I work from upgrading to v17, I decided to do my own PR for it, taking only the fix for Auto-Complete.

### DIFF
--- a/src/app/components/autocomplete/autocomplete.ts
+++ b/src/app/components/autocomplete/autocomplete.ts
@@ -1245,7 +1245,16 @@ export class AutoComplete implements AfterViewChecked, AfterContentInit, OnDestr
 
     onArrowDownKey(event) {
         if (!this.overlayVisible) {
-            return;
+            if(this.focused) {
+                const el = this.inputEL.nativeElement as HTMLInputElement | undefined;
+                if(el){
+                    // Put cursor at the end as the preventDefault blocks this behaviour
+                    el.selectionStart = el.selectionEnd = el.value.length;
+                }
+                this.overlayVisible = true;
+            } else {
+                return;
+            }
         }
 
         const optionIndex = this.focusedOptionIndex() !== -1 ? this.findNextOptionIndex(this.focusedOptionIndex()) : this.findFirstFocusedOptionIndex();


src/app/components/autocomplete/autocomplete.ts
```diff
@@ -1245,7 +1245,16 @@ export class AutoComplete implements AfterViewChecked, AfterContentInit, OnDestr
 
     onArrowDownKey(event) {
         if (!this.overlayVisible) {
-            return;
+            if(this.focused) {
```
SoyDiego, line 1248: IMO It is more clear and escape of the function fast if not `this.overlayVisible || !this.focused`.
Try it if continue works if you want :)IMO It is more clear and escape of the function fast if not `this.overlayVisible || !this.focused`.
Try it if  works,if you want :)

onArrowDownKey(event) {
    if (this.overlayVisible || !this.focused) {
        return;
    }

    const el = this.inputEL.nativeElement as HTMLInputElement | undefined;

    if (el) {
        // Put cursor at the end as the preventDefault blocks this behaviour
        el.setSelectionRange(el.value.length, el.value.length);
    }

    this.overlayVisible = true;
}

timandella: Your suggestion has omitted some existing code, which should run when this.overlayVisible is true, so I don't think it would work as expected

cetincakiroglu:
Hi,

Thanks a lot for the effort and support but it's made by design. We've updated all Prime UI Libraries during the accessibility implementation. This behavior is intended. Closing the PR.

PR #14481, opened 2024-01-04
feat(multiple): use new control flow for better perf and DX

eneajaho:
- Improves code readability
- Improves perf


eneajaho:
Btw, I'm thinking of adding more commits to migrate most of the components.

cetincakiroglu:
Hi,

Thanks a lot for the effort and support! All of the components should be refactored in this way if we do in 2-3 components. Since this is a big change and, it also affects nearly every component file, this refactor should be done by the Core team. I'm closing this one.

PR #14520, opened 2024-01-10
Fix #14490 p-multiselect is not resetting when using with reactive form

### LINKED ISSUE #14490: p-multiselect is not resetting when using with reactive form
### Describe the bug

P-multiselect is not going back to its default label even after resetting it when using reactive form

### Environment

Latest Angular + PrimeNg, easily reproducible (see the StackBlitz link below)

Step 1 : 
MicrosoftTeams-image

Step 2: 

23e4af63-b033-446c-bde5-37d63573016a

Step 3: 
The checkboxes are unchecked but label is still present with those values.

7084ce1c-784a-46dc-9eff-d17de8b5f2c0

### Reproducer

https://stackblitz.com/edit/swzu6g?file=src%2Fapp%2Fdemo%2Fmulti-select-template-demo.ts

### Angular version

17.0.8

### PrimeNG version

17.3.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

20.9.0

### Browser(s)

Chrome

1. Select values from multiselect dropdown
2. Click on reset button
3.  Checkbox is getting unchecked but it is not reflecting multiselect label (i.e. it is not going back to its default label or placeholder)

### Expected behavior

1. Select values from multiselect dropdown
2. Click on reset button
3. Checkbox should be unchecked and it should be back to default label or placeholder

HeavyRainLQ:
Fix #14490

### DIFF
--- a/src/app/components/multiselect/multiselect.ts
+++ b/src/app/components/multiselect/multiselect.ts
@@ -1097,6 +1097,8 @@ export class MultiSelect implements OnInit, AfterViewInit, AfterContentInit, Aft
                     this.selectedOptions = [...modelValue];
                 }
                 this.cd.markForCheck();
+            } else {
+                this.selectedOptions = [];
             }
         });
     }


cetincakiroglu:
Hi,

selectedOptions should be equal to modelValue every time the value updated. Because of that, it's not correct to assign empty array to the selectedOption while the modelValue is actually null | undefined. Fixed in https://github.com/primefaces/primeng/issues/14490

PR #14527, opened 2024-01-10
Fix #14525 reading properties of undefined during EditableColumn changes

### LINKED ISSUE #14525: EditableColumn: error in ngOnChanges when input other than `data` changes
### Describe the bug

When value of attribute `pEditableColumnDisabled` changes during runtime, an error is thrown due to `change.data` being undefined.

I believe that error is in: https://github.com/primefaces/primeng/blob/8c9f3d05e2d52aa37a0ff54aeb856294d1277614/src/app/components/table/table.ts#L4067

Which assumes that property `data` of SimpleChanges object will be always defined, which is not true when only the `pEditableColumnDisabled` input changes (f.e. with slight delay).

### Environment

Windows 10, starting Angular app with yarn.

### Reproducer

https://stackblitz.com/edit/github-8qgwug-rbxqag?file=src%2Fapp%2Fapp.component.html

### Angular version

17.0.8

### PrimeNG version

17.3.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

21.1.0

### Browser(s)

_No response_

Open example, click enable editing.

### Expected behavior

Error is not thrown into console when value of input `pEditableColumnDisabled` is changed during runtime

jkremlacek:
SimpleChanges might not have `data` defined when only other input than `data` changes.

For details see reporting issue: https://github.com/primefaces/primeng/issues/14525

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -4063,8 +4063,8 @@ export class EditableColumn implements OnChanges, AfterViewInit, OnDestroy {
 
     constructor(public dt: Table, public el: ElementRef, public zone: NgZone) {}
 
-    public ngOnChanges({ data }: SimpleChanges): void {
-        if (this.el.nativeElement && !data.firstChange) {
+    public ngOnChanges(changes: SimpleChanges): void {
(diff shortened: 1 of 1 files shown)


mertsincan:

Issue #14541, opened 2024-01-12
Tree: Missing data-id attribute on treeitem


adyeardley:
### Describe the bug

We recently updated our application to the latest version of PrimeNG, but some of our automated tests began to fail as it couldn't find some elements within the Tree component anymore.

We found that the automated tests are looking for the data-id attribute on each node, which was populated previously with the content of the key attribute on the TreeNode.

Did some digging in the commit history and found that on this commit, the attributes and event were moved from a div to a li, but the data-id attribute did not carry across:

https://github.com/primefaces/primeng/commit/12d79f7c45e8061ab9c797ba1dc2d50311c7d8fb

![image](https://github.com/primefaces/primeng/assets/40686932/1dffcb42-0255-414a-bb8f-05bd47924418)

### Environment

Angular 17 Web App

### Angular version

17.0.6

### PrimeNG version

17.3.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.10.0

### Browser(s)

All Browsers

### Steps to reproduce the behavior

Open a tree component with nodes inside
Inspect Element of any of the nodes
data-id is not populated with the 'key' of the node

### Expected behavior

data-id should be present and populated with the content of the 'key' of the node

adyeardley:
Raised PR #14552 to resolve this

mertsincan:
Hi @adyeardley,

Thanks a lot for your PR! But, we plan to fix such problems with the new passthrough API. We are currently working on it and will give an update on it soon. Please see our PrimeVue passthrough API; https://primevue.org/tree/#pt.image

For now, closed. Thanks a lot for your understanding!
Best Regards,

adyeardley:
HI @mertsincan,

Thanks for the reply.

Sorry, I'm a little confused about the response, we don't use Vue.js within our application, we use the Angular framework, having the fix in the PrimeVue API will not resolve this problem for us?

luke6040:
Good Morning @mertsincan 

I have just been through this and have come to the same conclusion as @adyeardley, are we missing something? Or is @adyeardley correct and fixing PrimeVue will not help this issue? These IDs are actually pretty important for our implementation of automation testing. 

Thank you for your help. 
Luke

PR #14552, opened 2024-01-15
#14541 - Added data-id attribute back into Tree component

adyeardley:
Fixes issue #14541 

Have added the attribute back onto the Tree component


mertsincan:
Added comment to the original issue.

adyeardley:
HI @mertsincan ,

Thanks for the reply on the original issue.

Sorry, I'm a little confused about the response, we don't use Vue.js within our application, we use the Angular framework, having the fix in the PrimeVue API will not resolve this problem for us?

PR #14560, opened 2024-01-16
Chart component should runOutsideOfAngular

steverob2k:
Fixes an issue with angular change detection firing and causing the application to crash.


steverob2k:
Fixes the issue reported here #14561

mertsincan:

Issue #14600, opened 2024-01-22
SplitButton : No disbaled state for menubutton


pouyababaie:
### Describe the bug

there were no component inputs to set the menubutton disabled state so i added it.

### Environment

default Primeng development environment 

### Angular version

16.0.2

### PrimeNG version

16.5.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.9.0

### Browser(s)

Chrome 20.0.6099.225 (Official Build) (64-bit)

cetincakiroglu:
Fixed in https://github.com/primefaces/primeng/pull/14602

PR #14602, opened 2024-01-22
[FIX] #14600 Add Disabled state for icon button 

pouyababaie:
#14600 

i added this feature , I'm guessing a lot of people felt the need of this.

image


cdskill:
+1 should be merged this issue is here since 2019

cetincakiroglu:
Hi,

Instead of changing functionality of current `disabled` attribute, could you please add two new inputs so users can disable it totally or disable them by their choice. It should be like this:

1-) All disabled, if user sets disabled as true, component will be disabled completely (this is for backward compatibility).
```
<p-splitButton [disabled]="true"></p-splitButton>
```

2-) Add new `buttonDisabled` and `menuButtonDisabled` input properties.

```
// that one should only disable the button of the component.
<p-splitButton [buttonDisabled]="true"></p-splitButton>

// that one should only disable the menu button of the component.
<p-splitButton [menuButtonDisabled]="true"></p-splitButton>
```

3-) If they are all used together, the `disabled` property should supersede `menuButtonDisabled` and `buttonDisabled` properties.

I'll merge after these changes. In this way, we keep the functionality of `disabled` and introduce two new input props if anyone needs to use.

pouyababaie:
> Hi,
> 
> 
> 
> Instead of changing functionality of current `disabled` attribute, could you please add two new inputs so users can disable it totally or disable them by their choice. It should be like this:
> 
> 
> 
> 1-) All disabled, if user sets disabled as true, component will be disabled completely (this is for backward compatibility).
> 
> ```
> 
> <p-splitButton [disabled]="true"></p-splitButton>
> 
> ```
> 
> 
> 
> 2-) Add new `buttonDisabled` and `menuButtonDisabled` input properties.
> 
> 
> 
> ```
> 
> // that one should only disable the button of the component.
> 
> <p-splitButton [buttonDisabled]="true"></p-splitButton>
> 
> 
> 
> // that one should only disable the menu button of the component.
> 
> <p-splitButton [menuButtonDisabled]="true"></p-splitButton>
> 
> ```
> 
> 
> 
> 3-) If they are all used together, the `disabled` property should supersede `menuButtonDisabled` and `buttonDisabled` properties.
> 
> 
> 
> I'll merge after these changes. In this way, we keep the functionality of `disabled` and introduce two new input props if anyone needs to use.
> 
> 
> 
> 
> 
> **Please tag me after the changes so I don't miss the PR.**

I will work on it 👍🏻

pouyababaie:
@[cetincakiroglu](https://github.com/cetincakiroglu)

I added the requested state for disabled button but i'm not sure i did it the right way. i would be very happy to see your comments on this commit. if there is a better way to solve this issue , i'd like to hear it.

image

Issue #1009, opened 2016-10-03
Autosize datatable columns


MerlinDuChaos:
Hi,

I think this would be a great addition to the DataTable component to be able to specify that a column should resize either to its content's width or header width.

Thanks !

hhubik:
Yes, please. I want this too.  When I double-click a column header side border, I expect the column to re-size itself to fit the content (like in Excel).

heavydawson:
Great suggestion

cagataycivici:
```
[tableStyle]="{'table-layout':'auto'}"
```

or via style class;

```
.ui-datatable table {
    table-layout: auto
}
```

Note that auto layout is not supported for resize or scrollable tables.

cvtoro:
is there any way to get to enforce 'table-layout: auto' on scrollable tables?

Olddude:
You can achieve the same `table-layout: auto` behavior on scrollable tables when you use the following configuration.

1.  set the `<p-datatable [tableStyle]="{'table-layout':'auto'}" [scrollable]="false">`
2. set the inside the p-datatable css `.ui-datatable-tablewrapper { overflow-x: auto;  }`

borisbrodski:
@Olddude This is an interesting idea, but the table you will get, will scroll (vertically) together with the header. Any ideas, how to achieve the proper scrolling (so that the header of the table always stays above the table body) and still have the nice "table-layout: auto"-like behavior?
Thank you!

nasirpinnacle:
Greatest idea, easy workability:
https://sites.google.com/site/christophenoel/home/primefacesdatatabledynamiccolumnswidth

i coded as

<p:columns value="#{liveRangeService.tableHeaderNames}"
                               var="mycolHeader" width="#{colIndex==0?'10':
                                                          colIndex==1?'150':
                                                          colIndex==2?'10':
                                                          colIndex==3?'10':
                                                          '0'}" columnIndexVar="colIndex" selectRow="true">

rudi1602:
fixed for PrimeFaces by primefaces/primefaces#14387

PR #14620, opened 2024-01-23
Discussion #1009 -- Enhancement: Update chips.ts to have a clearInput method

blake-w:
Discussion: https://github.com/orgs/primefaces/discussions/1009

### Feature Requests
I am creating a discussion and a pull request to enhance the Chips component. Currently, there is no way to clear the input field. This is problematic when you display the Chips component in a dialog, then you hide the dialog, and upon reopening the dialog any unsubmitted Chip will still linger in the input field.

Ideally, this would be tied into the hide event as it would be my expected behavior. However, that does align with how Primeng usually handles hiding components (does not hard-clear/edit fields).

You are currently required to access the input field through the element query selector:
`chipsRef.el.nativeElement.querySelector("input");`

I believe this enhancement would prevent others from manually targeting and clearing the input field and it could be done with a public method.

```
public clearInput(): void {
    this.inputViewChild.nativeElement.value = '';
}
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "primeng",
-    "version": "17.3.0",
+    "version": "17.3.3",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "primeng",
-            "version": "17.3.0",
+            "version": "17.3.3",
             "license": "SEE LICENSE IN LICENSE.md",
             "devDependencies": {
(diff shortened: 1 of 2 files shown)


cetincakiroglu:
Hi,

This issue is fixed in here https://github.com/primefaces/primeng/commit/283c71b05d5d6cda9e55435eacb293f6f21f48c2

There is no need to reset input value since we update it depending on model value. Closing the PR, thanks for the effort and support!

PR #14692, opened 2024-02-01
Fix #14691 DomHandler.getFocusableElements returns visible elements

### LINKED ISSUE #14691: DomHandler: getFocusableElements returns elements with hidden parents
### Describe the bug

DomHandler.getFocusableElements checks the display and visibility of each element itself but doesn't account for ancestors with `display: none`. This results in unexpected behavior when tabbing in, for example, a dialog box with focus trap enabled.

### Environment

Angular 17.0.8

### Reproducer

_No response_

### Angular version

17.0.8

### PrimeNG version

17.3.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.19.0

### Browser(s)

_No response_

1. Hide the parent of an otherwise focusable element via `display: none`
2. Use focus trap and tab around
3. Observe how focus gets stuck or otherwise behaves unexpectedly.

### Expected behavior

An element that is not displayed because its parent has `display: none` shouldn't be eligible for focus and thus should be excluded from the results.

jasonverber:
Fixes [#14691](https://github.com/primefaces/primeng/issues/14691) by checking if each element is visible.

### DIFF
--- a/src/app/components/dom/domhandler.ts
+++ b/src/app/components/dom/domhandler.ts
@@ -630,7 +630,7 @@ export class DomHandler {
         let visibleFocusableElements = [];
 
         for (let focusableElement of focusableElements) {
-            if (getComputedStyle(focusableElement).display != 'none' && getComputedStyle(focusableElement).visibility != 'hidden') visibleFocusableElements.push(focusableElement);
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi,

Thanks a lot for the quick action, we appreciate the effort and support!

PR #14710, opened 2024-02-05
Fixed #14682 - Dropdown | TypeError: option.toLowerCase is not a func…

### LINKED ISSUE #14682: Component: Dropdown TypeError: option.toLowerCase is not a function
### Describe the bug

Same problem as in issue #14370:
Declaring a dropdown with a filter, but not specifying `optionlabel="label"` or `filterBy="label"` breaks the dropdown on filter.

### Environment

Chrome, Mozilla

### Reproducer

https://stackblitz.com/edit/github-3wjoqr?file=src%2Fapp%2Fapp.component.html

### Angular version

17.x.x

### PrimeNG version

17.x.x

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.x.x

### Browser(s)

_No response_

_No response_

### Expected behavior

Default filter should reference to label as before, not undefined

mehmetcetin01140:
Fixes #14682

### DIFF
--- a/src/app/components/dropdown/dropdown.ts
+++ b/src/app/components/dropdown/dropdown.ts
@@ -903,7 +903,7 @@ export class Dropdown implements OnInit, AfterViewInit, AfterContentInit, AfterV
             const filteredOptions =
             !_filterBy && !this.filterFields && !this.optionValue
                 ? this.options.filter((option) => {
-                      if (typeof option === 'object' && option.label) {
+                      if (option.label) {
                           return option.label.toLowerCase().indexOf(this._filterValue().toLowerCase()) !== -1;
                       }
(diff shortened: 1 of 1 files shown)


basvandenberg:
@mehmetcetin01140, @cetincakiroglu I noticed that the code still contains a `option.toLowerCase()`, even though `option` has type `any`, and I would expect most users to provide objects as options, not strings. This PR might prevent the error for one use case, but I would expect it to still occur in other cases. I have seen multiple issues that mention this same error.

I am trying to understand the logic, to see how this can be fixed, but to me it is not clear why there are two filters. Why not always use the `filterService.filter`? Could you please explain why the more basic filter is used if:

`!(this.filterBy || this.optionLabel) && !this.filterFields && !this.optionValue`

I have a hard time trying to understand this logic, it would be great if you could provide some insight.

Issue #1105, opened 2016-10-22
Improve p-dropdown focus visuals for Omega


kt215:
When a p-dropdown control has focus, the blue focus rectangle is missing:

![image](https://cloud.githubusercontent.com/assets/14261464/19620348/e1ecdc82-9838-11e6-8cf4-ae92cdb7c52a.png)

Other controls, such as input/pInputText and p-checkbox, show the blue focus rectangle correctly:

![image](https://cloud.githubusercontent.com/assets/14261464/19620351/006a37b8-9839-11e6-90d9-edb7d6a94e3e.png)

Using PrimeNG 1.0.0 Beta 19, Omega Theme, and Chrome 53.0.2785.143

cagataycivici:
It gets .ui-state-focus actually so focus visuals are there but not same as input, will do the same.

kt215:
Thanks for improving this. It looks a lot better now, but the focus visuals still disappear on the p-dropdown after selecting an item from the list using the mouse.

PR #14717, opened 2024-02-05
Feat discussion 1105: Expose Country Service Via API

Khousheish:
Implements suggestion requested here.
Discussion [#1105](https://github.com/orgs/primefaces/discussions/1105)


cetincakiroglu:
Hi,

It's a dummy service for demo purposes we do not need to make it public. Doesn't affect anything in the library.

Issue #14716, opened 2024-02-05
dropdown: selected, disabled option not rendered


capc0:
### Describe the bug

When a dropdown has a disabled option that is initially selected by code, it is not rendered as the current selection.
It is a common use case that a dropdown has a current selection, that the user should change but to any value different than the currently selected one. Therefore the current/initial selected option gets disabled, so that it cannot be re-selected.

This used to be the case in Prime v15 and v16

### Environment

see below

### Reproducer

https://stackblitz.com/edit/github-npixcr?file=src%2Fapp%2Fapp.component.ts

### Angular version

17.1.2

### PrimeNG version

17.5.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18

### Steps to reproduce the behavior

1. open the stackblitz (see above)
2. look at the following initial data in `app.component.ts`
```ts
  options: SelectItem[] = [
    { label: 'one', value: 1 },
    { label: 'two', value: 2, disabled: true },
    { label: 'three', value: 3 },
  ];
  selection = 2;
```
3. view the result --> option `two` is not rendered/selected

![image](https://github.com/primefaces/primeng/assets/27284008/198afb59-7fe1-4695-8124-a07e635827ab)

### Expected behavior

The selected option should be rendered, since it is already the current selection via model binding, even if it is disabled.

capc0:
Thanks for the quick fix. However there is a follow-up issue with 17.7 detailed in https://github.com/primefaces/primeng/issues/14809

rmaleryk:
Looks like the fix is not working for the custom `selectedItem` template.

MartinaAnt:
Same here in v17.18.4 - `selectedItem` template not working with (programmatically selected) disabled option

PR #14718, opened 2024-02-05
fix(14716): Allow disabled (but programmatically selected) option to be taken into account for building selected option label

Cr3aHal0:
See #14716


cetincakiroglu:
Hi,

The related issue has been fixed with another PR and the issue doesn't exist with v17.12.0. Because of this, I'm closing this one. Thanks a lot for the effort and support!

Issue #14715, opened 2024-02-05
Dropdown: "0" Value in optionValue, causes Label to be replaced by Placeholder 


eduardheller89:
### Describe the bug

When the value of the dropdown is 0, the placeholder value is still shown in the input. 

[This is the line causing the issue](https://github.com/primefaces/primeng/blob/20b5c1c28ed29fcf6daca0d1b662b82e30fb9476/src/app/components/dropdown/dropdown.ts#L934C9-L934C114)

`return this.modelValue() ? this.getOptionLabel(this.selectedOption) : this.placeholder || 'p-emptylabel';` 

this Line in component.ts causes '0' to be accepted as False. You should check for null or undefined. 

Thanks

### Environment

Win10/chrome 

### Angular version

17.1.2

### PrimeNG version

17.5.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.19.0

### Browser(s)

Chrome 121

### Steps to reproduce the behavior

1. Set Dropdown to a value in optionValue to 0. 
2. Set Dropdown placeholder 
3. Change the value of the dropdown in the ui to the corresponding value with Value 0
5. Placeholder still remains in the input

### Expected behavior

The optionLabel should show the Label of the optionValue with Value 0

Cr3aHal0:
Hi,

Not sure if this is by design or if more options should be available other than '0' such as true, false or empty string. If empty/null comparison is enough, the linked PR should be okay

rosenthalj:
This issue is the result of a condition that I did not consider when the code was changed for pull request #14665. 

I believe that 2 sets of changes need to be made to fix this issue.


Please see the annotated screenshot of pull request #14665 listed below

![Fix__14596_-_fixes_p-dropdown_logic_for_selectedOption_and_label_by_rosenthalj_·_Pull_Request__14665_·_primefaces_primeng](https://github.com/primefaces/primeng/assets/45439491/4c0a592f-a565-4ade-a45a-2c1dcbc865ca)

Cr3aHal0:
Hi rosenthalj,

I'll be investigating your suggestion, I did not encountered any incorrect behaviour while testing the change but i'll retry :)

rosenthalj:
Cr3aHal0,

When testing I would recommend that you try the following "filtering" cases:

1. A test case similar to  the last video in pull request #14665 comments. 
--- set options 0, 1, 2
--- select 0
--- enter 1 in the filter
[...]

PR #14665, opened 2024-01-30
Fix #14596 - fixes p-dropdown logic for selectedOption and label

rosenthalj:
Fix #14596

1) fixes logic for the dropdown's selectedOption
2) uses dropdown's selectedOption for computing label
-- note: issue with virtual scrolling with lazy loading (PrimeNG Demo) as a result keeping original logic for computing label when virtual load (no harm)

The following  video demonstrates  the fix working for the reproducer in  issue #The following video demonstrates the fix.

https://github.com/primefaces/primeng/assets/45439491/2bacb0a7-ff42-480c-82d2-bef95157e6da

PR #14719, opened 2024-02-05
fix(14715): selecting a falsy value (such as '0' or 0) should not show placeholder

Cr3aHal0:
See #14715

Not sure if this is by design or if a larger set of value should or should not be accepted (such as true/false), feel free to reject this PR. I did hesitate to use ObjectUtils.isNotEmpty but since it excludes an empty string as being an authorized option, I thought it may be too much.

### DIFF
--- a/src/app/components/dropdown/dropdown.ts
+++ b/src/app/components/dropdown/dropdown.ts
@@ -931,7 +931,8 @@ export class Dropdown implements OnInit, AfterViewInit, AfterContentInit, AfterV
             const selectedOptionIndex = this.findSelectedOptionIndex();
             return selectedOptionIndex !== -1 ? this.getOptionLabel(this.visibleOptions()[selectedOptionIndex]) : this.placeholder || 'p-emptylabel';
         }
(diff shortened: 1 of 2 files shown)


rosenthalj:
I  beleive this pull request is not complete.   This pull request  is an attempt to fix a condition that I did not consider  when the code was changed for pull request #14665.  Was this pull requested fully tested with a value of 0 and using  the dropdown's filtering? 

Please see the annotated screenshot  of pull request #14665 listed bellow

Fix__14596_-_fixes_p-dropdown_logic_for_selectedOption_and_label_by_rosenthalj_·_Pull_Request__14665_·_primefaces_primeng

Cr3aHal0:
[rosenthalj](https://github.com/rosenthalj) 

You may find in latest commit some more changes (including the one you mentioned) about strict type check of modelValue(). I did test it against the filterable dropdown use case and it works well, both from filter and from ngModel initialization. I tweaked a bit the template of the filter example to get rid of the unnecessary selectionCountry check since the template is bound only if the dropdown as a valid value.

Let me know if this sounds good to you

cetincakiroglu (approved):

PR #14726, opened 2024-02-06
fix(14706): restore addOnTab behavior - allow user to add chip on tab…

### LINKED ISSUE #14706: p-chips: [addOnTab] feature has no effect. It used to keep focus within the text box.
### Describe the bug

In the `<p-chips>` element, there is an @Input() property called `addOnTab`. This property does nothing when the <kbd>tab</kbd> key is pressed, and is not referenced anywhere in the PrimeNG `chips.ts` code.

### Environment

Windows 10

### Reproducer

https://stackblitz.com/edit/github-mxeong?file=src%2Fapp%2Fapp.component.html

### Angular version

17.1.1

### PrimeNG version

17.5.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.16.1

### Browser(s)

_No response_

1. Go to the StackBlitz provided.
2. In the first field, type a new value like "test" and press <kbd>tab</kbd>.
3. In the second field, type a new value like "test" and press <kbd>tab</kbd>.
4. Nothing happens in either case.

### Expected behavior

When `addOnTab="true"`, pressing the <kbd>tab</kbd> key should create a new chip with the entered value and keep the focus within the text field for the next chip to be created, just like how the <kbd>enter</kbd> key works.

This behavior works perfectly in PrimeNG 16.0.2.

Cr3aHal0:
Fixes #14706 

Kept the initial behavior which tabbed the user out of the field if addOnTab is not enabled, add new chip otherwise

### DIFF
--- a/src/app/components/chips/chips.ts
+++ b/src/app/components/chips/chips.ts
@@ -514,6 +514,13 @@ export class Chips implements AfterContentInit, ControlValueAccessor {
 
                 break;
 
+            case 'Tab':
+                if (this.addOnTab && inputValue && inputValue.trim().length && !this.isMaxedOut) {
+                    this.addItem(event, inputValue, true);
+                }
+
+                break;
+
             case 'ArrowLeft':
                 if (inputValue.length === 0 && this.value && this.value.length > 0) {
                     this.containerViewChild?.nativeElement.focus();


cetincakiroglu (approved):
You forgot to add `event.preventDefault()` inside the condition. I'll add it now to save time.

cetincakiroglu:
Thanks a lot for PR!

PR #14736, opened 2024-02-06
Fix #14735 - Password | Input hover not displaying blocked cursor.

### LINKED ISSUE #14735: Input Components: Input hover not displaying blocked cursor when input is disabled.
### Describe the bug

When Password input is disabled it is not showing the "not-allowed" type cursor when hovering, instead it shows only the default cursor.

### Environment

Angular

### Reproducer

_No response_

### Angular version

18

### PrimeNG version

17.5.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

Latest

### Browser(s)

_No response_

1. Go to Disabled Password
2. Hover on the Disabled Password
3. It shows the default cursor.

### Expected behavior

1. Go to Disabled Password
2. Hover on the Disabled Password
3. It should be showing the block cursor.

VA-Ansari:
Fix #14735 : Password - Input hover not displaying blocked cursor.

### DIFF
--- a/src/app/components/common/common.css
+++ b/src/app/components/common/common.css
@@ -49,6 +49,10 @@
         pointer-events: none;
     }
 
+    .p-disabled-overlay{
+        cursor : not-allowed !important;
+    }
+
     .p-component-overlay {
         position: fixed;
         top: 0;
--- a/src/app/components/password/password.ts
+++ b/src/app/components/password/password.ts
@@ -328,7 +328,7 @@ export const Password_VALUE_ACCESSOR: any = {
 @Component({
     selector: 'p-password',
     template: `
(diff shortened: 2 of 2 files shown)


src/app/components/common/common.css
```diff
@@ -49,6 +49,10 @@
         pointer-events: none;
     }
 
+    .p-disabled-overlay{
+        cursor : not-allowed !important;
```
Vatspra, line 53: We should avoid using  !important. It would be better to use CSS Specificity .

cetincakiroglu: We cannot add `!important` to the component styles. Could you please remove and update the code?

cetincakiroglu:
Hi,

Unfortunately, I have to close this PR because we need to update all input components and test them. This enhancement will be implemented by Core Team in v17.

VA-Ansari:
You are welcome

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #14746, opened 2024-02-07
Fix #14744: Autocomplete: Empty list showed, when no item fit condition

### LINKED ISSUE #14744: Autocomplete: Empty list showed, when no item fit condition
### Describe the bug

When entering value that can't be auto completed due to lack of items fitting given condition, there is an empty html list shown.

emptydropdown-without-fix

### Environment

Windows 10
Primeng: v16.9.2-lts + Angular: v16.2.12 - reproduced at Primeng: v17.5.0 + Angular: v17.0.5

### Reproducer

https://primeng.org/autocomplete

### Angular version

16.2.12

### PrimeNG version

16.9.2-lts

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.6.0+

### Browser(s)

_No response_

1. Go to https://primeng.org/autocomplete
2. Pick autocomplete, that have static items list - like countries at 'Reactive Forms' paragraph.
3. Type something that won't fit any country, like 'notexisting'
4. Empty list will be shown.

### Expected behavior

There is no empty list shown - there is just input.

kkoripl:
Fix: #14744

Added condition to actually have items to show in list to ngif on mentioned list.
I'm a bit unsure about the name of list, cause needed to change it.

**Before:**
emptydropdown-without-fix

**After:**
emptydropdown-with-fix

### DIFF
--- a/src/app/components/autocomplete/autocomplete.ts
+++ b/src/app/components/autocomplete/autocomplete.ts
@@ -219,7 +219,7 @@ export const AUTOCOMPLETE_VALUE_ACCESSOR: any = {
                     </ng-container>
 
                     <ng-template #buildInItems let-items let-scrollerOptions="options">
-                        <ul #items class="p-autocomplete-items" [ngClass]="scrollerOptions.contentStyleClass" [style]="scrollerOptions.contentStyle" role="listbox" [attr.id]="id + '_list'">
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi,

The root cause of issue is showEmptyMessage is not working when search resulted with empty. I've changed the showEmptyMessage default value from false to true and add related missing styles. Now it will display `No result found` message. 

Thanks for the effort and support! Closing the PR since it's fixed by Core Team.

PR #14754, opened 2024-02-07
Feat: calendar - enable minDate & maxDate to be nullable

Cr3aHal0:
See discussion #1136 https://github.com/orgs/primefaces/discussions/1136

For context : actually minDate & maxDate are 2 inputs of calendar that only accept non-nullable Date value. What it means is that you can have no value for any of them at startup but once you set any value (for example for search purpose) you can't reset it programmatically, i.e. providing undefined or null value. The only way to perform such reset is by using clear option (with [showClear]=true) or emptying the field manually (such leading in false value).

This change is all about enabling minDate & maxDate value to accept undefined value allowing free programmatical change, all checks about its truthyness being already handled. Let me know if null should be accepted instead/aswell.

### DIFF
--- a/src/app/components/calendar/calendar.ts
+++ b/src/app/components/calendar/calendar.ts
@@ -732,10 +732,10 @@ export class Calendar implements OnInit, OnDestroy, ControlValueAccessor {
      * The minimum selectable date.
      * @group Props
      */
-    @Input() get minDate(): Date {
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi,

You're referencing an issue related with the table, could you please link the correct issue so we can review

PR #14771, opened 2024-02-08
Fixed #14770 - api-generator problem

### LINKED ISSUE #14770: Input properties defined using the setter/getter are not properly documented
### Describe the bug

Due to a problem in the [build-apidoc.ts](https://github.com/primefaces/primeng/blob/master/api-generator/build-apidoc.ts) script, description and deprecated values are not generated for the properties that are defined using the setter/getter (those that have the `@Input` annotation before the setter). 

Example: `locale` property in the Calendar component. It has the description and the deprecated annotation, but in the docs, it doesn't have any description and is not shown in strikethrough text as it should be as a deprecated property.

image

image
https://primeng.org/calendar

### Environment

darwin arm64

### Reproducer

_No response_

### Angular version

17.0.5

### PrimeNG version

17.6.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

20.11.0

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

lukasmatta:
Fixes #14770, ~~also changed file extension of the `api-generator` file from `ts` to `cjs` so it is possible to run it despite the `"type": "module"` definition inside the `package.json` file.~~ (not needed since https://github.com/primefaces/primeng/commit/2e1f069c4899ddebaf3f6afaa7e601bbf8429560)

Example stated in the issue now looks like this:
image
(now it contains property description, property name is in the strikethrough text as it is deprecated and on hover it shows the deprecation description tooltip)


api-generator/build-apidoc.cjs
SoyDiego, line 1: If I'm not wrong, rename a file only changing the name can have  future problems because then with other people are downloading the project (`git pull`)  can generate conflicts. I had a lot of problems with this example in my work.

The correct way  is doing:

```
git mv old-file new-file
```

In this case: 

```
git mv build-apidoc.ts build-apidoc.cjs
```

And GitHub will rename it  in all the files of the project and with no inconvenients. 

Maybe you can confirm it @cetincakiroglu. 

Font: https://docs.github.com/en/repositories/working-with-files/managing-files/renaming-a-file#renaming-a-file-using-the-command-line

lukasmatta: `git mv` is just a helper to rename the file and update the index, the example you linked has to do with changing filename case on the case-insensitive OS such as macOS, in that case, running `git mv` might be useful to catch the change

lukasmatta:
Hi @cetincakiroglu, I've noticed you removed `"type": "module"` in https://github.com/primefaces/primeng/commit/2e1f069c4899ddebaf3f6afaa7e601bbf8429560, so I renamed the generator file back to `.ts` in this branch (even though `.ts` exetension is a bit confusing, as it doesn't contain any typed code and it is run using `node`).

cetincakiroglu:
Hi @lukasmatta,

Thanks a lot for the effort and support, could you please resolve the conflicts and let me know so I can merge it safely. Probably rebasing to master and running `npm run apidoc` command fixes the issue.

lukasmatta:
thanks for the feedback, conflicts should be resolved now

PR #14781, opened 2024-02-10
Fix #14590 and #12286 - Single dates not possible with calendar in range mode

### LINKED ISSUE #12286: p-calendar in range mode does not allow manually typing in same start and end date
### Describe the bug

Using p-calendar in range mode. Need to be able to type in a date range with the same start and end date. This can be selected from the UI picker by clicking the same day twice. However, if a date range of the same day is typed in manually then the input is rejected and the calendar reverts to its previous selection.

### Environment

OS: Windows 11
Browser: Google Chrome
Can be reproduced in stackblitz in the library documentation example, so all other environment information should be accessible there.

### Reproducer

https://stackblitz.com/edit/primeng-calendar-demo-kj1zcc

### Angular version

14.2.8

### PrimeNG version

14.1.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.11.7

### Browser(s)

Chrome 121

When the user enters a date range with the same start and end date and after that clicks outside of the p-calendar input, the entered value is rejected.

### Expected behavior

Since a date range with the same start and end date is valid from the UI picker, it should also be valid when typed in manually not only when the range is selected from the picker.

### LINKED ISSUE #14590: p-calendar in range mode does not allow manually typing single date
### Describe the bug

Using p-calendar in range mode. Need to be able to type in a single date. This can be selected from the UI picker by clicking a date and outside the calendar. However, if a single date is typed in manually then the input is rejected and the calendar reverts to an empty value.
The input is only accepted if a full range is manually typed in.

Also see issue #12286

### Environment

OS: macOS Sonoma 14.2.1
Browser: Firefox 121.0.1

### Reproducer

https://stackblitz.com/edit/ngnlme

### Angular version

17.1.0

### PrimeNG version

17.3.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

20.10.0

### Browser(s)

_No response_

1. Open demo
2. Klick Input field
3. Type one single Date (e.g. 01/01/2022)
4. Tab outside the field

### Expected behavior

When typing in a single Date into a calendar in range-mode, the Datepicker should switch to the typed date (according to the configured locale).

eduardheller89:
fixes #14590 
fixes #12286 

calender-fix

### DIFF
--- a/src/app/components/calendar/calendar.ts
+++ b/src/app/components/calendar/calendar.ts
@@ -2733,16 +2733,14 @@ export class Calendar implements OnInit, OnDestroy, ControlValueAccessor {
     }
 
     isValidSelection(value: any): boolean {
-        let isValid = true;
+        let isValid = false;
         if (this.isSingleSelection()) {
-            if (!this.isSelectable(value.getDate(), value.getMonth(), value.getFullYear(), false)) {
-                isValid = false;
-            }
-        } else if (value.every((v: any) => this.isSelectable(v.getDate(), v.getMonth(), v.getFullYear(), false))) {
-            if (this.isRangeSelection()) {
-                isValid = value.length === 1 || (value.length > 1 && value[1] >= value[0] ? true : false);
-            }
-        }
+            return this.isSelectable(value.getDate(), value.getMonth(), value.getFullYear(), false);
+        } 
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi,

PR #14787, opened 2024-02-12
fix: #14785 rounded scrollLeft for scaled displays

### LINKED ISSUE #14785: Tabview: Forward Button Visible with Fractions
### Describe the bug

For checking if the forward button needs to be displayed, it compares the current scrollstate with the actual width of the content.

The issue is that the scrollWidth is rounded and the scrollLeft Value is not.

In my case the scroll width actually is: 1671.5, but during runtime the value reads 1672. Where as the scroll left value is and stays 150.5.

### Environment

MacOS

### Reproducer

_No response_

### Angular version

17.1.0

### PrimeNG version

17.4.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

v20.x

### Browser(s)

Chrome 121

_No response_

### Expected behavior

updateButtonState(this: TabView): void {
  const content = this.content!.nativeElement;
  const {scrollLeft, scrollWidth} = content;
  const width = DomHandler.getWidth(content);
  this.backwardIsDisabled = scrollLeft === 0;
  this.forwardIsDisabled = Math.round(scrollLeft) === scrollWidth - width;
}

denny99:
### Defect Fixes

#14785

### DIFF
--- a/src/app/components/tabview/tabview.ts
+++ b/src/app/components/tabview/tabview.ts
@@ -809,7 +809,7 @@ export class TabView implements AfterContentInit, AfterViewChecked, OnDestroy, B
         const width = DomHandler.getWidth(content);
 
         this.backwardIsDisabled = scrollLeft === 0;
-        this.forwardIsDisabled = scrollLeft === scrollWidth - width;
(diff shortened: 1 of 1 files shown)


denny99:
Reason for the fix is described in the ticket.

PR #14874, opened 2024-02-26
fix isObject detection (closes #14835)

### LINKED ISSUE #14835: multiselect: filter not working
### Describe the bug

The multiselect filter is not working anymore when the options passed to the component are class based (e.g. instances of a class named `City`).

image

The bug was introducted with this commit https://github.com/primefaces/primeng/commit/ed2c8679ed7dfc37784a1e12f1ab0d320a818528

The check for  `ObjectUtils.isObject(options[0])` fails for class based options, since `options[0].constructor === Object` evaluates to false.

### Environment

see below

### Reproducer

https://stackblitz.com/edit/github-97dcah?file=src%2Fapp%2Fapp.component.ts

### Angular version

17.1.2

### PrimeNG version

17.7.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18

### Browser(s)

_No response_

1. open the stackblitz
2. open the multiselect overlay
3. type something into the filter
4. see the console output

### Expected behavior

filter should also work with class based objects

capc0:
since the Util function is only used within multiselect, it should have no side effects. Also uses kind of best practices according to SO to determine if something is an object https://stackoverflow.com/questions/8511281/check-if-a-value-is-an-object-in-javascript

Fixes https://github.com/primefaces/primeng/issues/14835

### DIFF
--- a/src/app/components/utils/objectutils.ts
+++ b/src/app/components/utils/objectutils.ts
@@ -4,7 +4,7 @@ export class ObjectUtils {
     }
 
     public static isObject(value, empty = true) {
-        return value instanceof Object && value.constructor === Object && (empty || Object.keys(value).length !== 0);
+        return typeof value === 'object' && !Array.isArray(value) && value != null && (empty || Object.keys(value).length !== 0);
     }
 
     public static equals(obj1: any, obj2: any, field?: string): boolean {


cetincakiroglu (approved):
Hi,

PR #14915, opened 2024-02-29
fix: primefaces#14914, Password: (accessiblility) cannot access toggle button using keyboard

akshayaqburst:
### Defect Fixes
fix: primefaces#14914, Password: (accessiblility) cannot access toggle button using keyboard


cetincakiroglu:
Hi,

It's made by design, the toggle icon will not receive a tab index. However, we're implementing PassThrough, enabling users to give whatever attribute they want to any element of the component. I'm closing the issue, PassThrough is in the implementation process right now, and we're planning to release PassThrough for all components within v17.

PR #14920, opened 2024-03-01
Fixed #14917 - Dropdown | Clicking on currently selected item no long…

### LINKED ISSUE #14917: Dropdown: Clicking on currently selected item no longer closes the list
### Describe the bug

Previously, clicking on any item in a dropdown list would close it even if the item was already selected. Recently, the behaviour has changed: clicking on an item in the list closes it except if it is the currently selected item.

### Environment

*

### Reproducer

_No response_

### Angular version

17.*

### PrimeNG version

17.9.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

*

### Browser(s)

*

Go to the documentation of the Dropdown component: https://primeng.org/dropdown
On the first dropdown, select "New York" => The item is selected and the list disappears. (OK)
Open the dropdown again and select "New York" again => The list remains visible. (not OK)

### Expected behavior

On previous versions (<= 17.4.0), the list would always disappear which is what a dropdown should do regardless of which is the currently selected item.
You can check the old (good) behaviour here: https://www.primefaces.org/primeng-v16-lts/dropdown

mehmetcetin01140:
Fixes #14917

### DIFF
--- a/src/app/components/dropdown/dropdown.ts
+++ b/src/app/components/dropdown/dropdown.ts
@@ -1122,7 +1122,9 @@ export class Dropdown implements OnInit, AfterViewInit, AfterContentInit, AfterV
             this.focusedOptionIndex.set(this.findSelectedOptionIndex());
             preventChange === false && this.onChange.emit({ originalEvent: event, value: value });
         }
-        isHide && setTimeout(() => this.hide(true), 1);
+        if (isHide) {
+            this.hide(true);
+        }
     }
 
     onOptionMouseEnter(event, index) {


src/app/components/dropdown/dropdown.ts
```diff
@@ -1120,9 +1120,9 @@ export class Dropdown implements OnInit, AfterViewInit, AfterContentInit, AfterV
             const value = this.getOptionValue(option);
             this.updateModel(value, event);
             this.focusedOptionIndex.set(this.findSelectedOptionIndex());
-            isHide && setTimeout(() => this.hide(true), 1);
             preventChange === false && this.onChange.emit({ originalEvent: event, value: value });
         }
+        isHide && setTimeout(() => this.hide(true), 1);
```
SoyDiego, line 1125: Hi @mehmetcetin01140 I think you can remove the `setTimeout` and the code will be cleaner.
Using this lines works perfectly:

```
if (isHide) {
   this.hide(true);
}
```

Here you have a repro:

fixed

PR #15025, opened 2024-03-13
fix #15024 table filter row aria-labels

### LINKED ISSUE #15024: Table / Filter Row: Form elements missing aria-label
### Describe the bug

Missing aria-labels. Add aria-label attributes to the Table Filter Row form elements: p-columnFilter / p-columnFilterFormElement and its child components (input, p-inputNumber, p-triStateCheckbox p-calendar).

### Environment

https://primeng.org/table#filter-row

### Reproducer

_No response_

### Angular version

17.0.0

### PrimeNG version

17.10.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.16.0

### Browser(s)

_No response_

https://primeng.org/table#filter-row

### Expected behavior

_No response_

TaneliTuomola:
fixes https://github.com/primefaces/primeng/issues/15024
Added aria-label support.

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -4945,6 +4945,7 @@ export class ReorderableRow implements AfterViewInit {
                 class="p-fluid"
                 [type]="type"
                 [field]="field"
+                [ariaLabel]="ariaLabel"
                 [filterConstraint]="dt.filters[field]"
                 [filterTemplate]="filterTemplate"
                 [placeholder]="placeholder"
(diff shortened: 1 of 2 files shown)


cetincakiroglu:
Hi,

Thanks for the effort and support, your fix is correct but we need to get ariaLabel from the translateService to make it compatible with i18n. Could you please update the PR?

It should look like this:

```
// taken from the table.ts line 5283
    get showMenuButtonLabel() {
        return this.config.getTranslation(TranslationKeys.SHOW_FILTER_MENU);
    }
```

TaneliTuomola:
I thought that it would be better to be able to set aria-label to each p-columnFilter element separately (via input) and let the client application handle translations if needed. In your suggestion there will be only one generic aria-label in every p-columnFilter element in every table? @cetincakiroglu

PR #15083, opened 2024-03-19
fix #14984 Replacing listbox with grid structure

### LINKED ISSUE #14984: Chips: Accessibility problems in the listbox structure
### Describe the bug

A chip container (ul.p-inputtext) with role listbox should have a descriptive text label.

An element with a role that makes its children presentational (li[role=option].p-chips-input-token) should not contain focusable elements (input).

[https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/option_role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/option_role)

I wonder if those roles are necessary in this context? I think the Grid structure works best here. The same structure is used in Angular Material's Chips example https://material.angular.io/components/chips/overview#chips-connected-to-an-input-field

https://www.w3.org/WAI/ARIA/apg/patterns/grid/

### Environment

https://primeng.org/chips

### Reproducer

https://primeng.org/chips

### Angular version

17.0.0

### PrimeNG version

17.10.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.6.0

### Browser(s)

_No response_

https://primeng.org/chips

### Expected behavior

https://material.angular.io/components/chips/overview#chips-connected-to-an-input-field

TaneliTuomola:
fixes #14984

Replaces:
`listbox > option`
with
`grid > row > cell`

### DIFF
--- a/src/app/components/chips/chips.ts
+++ b/src/app/components/chips/chips.ts
@@ -38,9 +38,7 @@ export const CHIPS_VALUE_ACCESSOR: any = {
                 #container
                 [ngClass]="{ 'p-inputtext p-chips-multiple-container': true }"
                 tabindex="-1"
-                role="listbox"
-                [attr.aria-labelledby]="ariaLabelledBy"
-                [attr.aria-label]="ariaLabel"
+                role="grid"
                 [attr.aria-activedescendant]="focused ? focusedOptionId : undefined"
                 [attr.aria-orientation]="'horizontal'"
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi @TaneliTuomola 

We've discussed this with the team and decided not to change the listbox component's structure. Thanks a lot for the effort and support, closing the PR.

PR #15098, opened 2024-03-20
fix #1469 Removed tablist roles

### LINKED ISSUE #1469: Checkbox support for Listbox
**I'm submitting a ...**
```
```

I would like to use the MultiSelect as a Listbox with checkboxes. The master checkbox/search capabilities make this component perfect for our needs.

The closest thing I could try is to hide the combo with `display:none`, and set the `overlayVisible` to true. The problem is that the overlay is hidden when the user clicks outside of it. It would be great if there was a way to fix it on the screen.

### LINKED ISSUE #15099: Steps: Invalid and unnecessary tablist role structure
### Discussed in https://github.com/orgs/primefaces/discussions/1436

https://primeng.org/steps

Step container element (ul[role="tablist"]) has children which are not allowed: li[role=presentation]. The tablist role seems to have been included since the first version of the component, so it may not been considered from the accessibility perspective. 

In the primeVue and primeReact `<ol><li><a>` structure without tablist role is used.

image

### Environment

Chrome

### Reproducer

https://primeng.org/steps/personal

### Angular version

17.0.0

### PrimeNG version

17.11.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.16.0

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

TaneliTuomola:
fixes #15099

### DIFF
--- a/src/app/components/steps/steps.ts
+++ b/src/app/components/steps/steps.ts
@@ -14,7 +14,7 @@ import { Subscription } from 'rxjs';
     selector: 'p-steps',
     template: `
         <nav [ngClass]="{ 'p-steps p-component': true, 'p-readonly': readonly }" [ngStyle]="style" [class]="styleClass" [attr.data-pc-name]="'steps'">
-            <ul #list role="tablist" [attr.data-pc-section]="'menu'">
+            <ul #list [attr.data-pc-section]="'menu'">
                 <li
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi @TaneliTuomola,

PR #15309, opened 2024-04-18
feature(autocomplete): Add optionValue input

### LINKED ISSUE #14599: Component: Title
### Describe the bug

Well the autocomplete with dataKey is broken, as it is no longer implemented at all.

Autocomplete.ts:
```javascript
    getOptionValue(option) {
        return option; // TODO: The 'optionValue' properties can be added.
    }
```

There is another line with a todo that the optionValue should be used (instead of this.dataKey).

### Environment

Does not matter

### Reproducer

_No response_

### Angular version

Does not matter

### PrimeNG version

17.3.3

### Build / Runtime

Angular CLI App

### Language

TypeScript

Does not matter

### Browser(s)

_No response_

_No response_

### Expected behavior

Should be properly implemented as stated in the TODO

sandrotonon:
### Defect Fixes
fixes #14599

### Feature Requests
Implements the ability to set an optionValue, so the autocomplete can also store object properties and not only the object itself

### DIFF
--- a/src/app/components/autocomplete/autocomplete.ts
+++ b/src/app/components/autocomplete/autocomplete.ts
@@ -559,6 +559,11 @@ export class AutoComplete implements AfterViewChecked, AfterContentInit, OnDestr
      */
     @Input() optionLabel: string | ((item: any) => string) | undefined;
     /**
+     * Property name or getter function to use as the value of an option.
+     * @group Props
+     */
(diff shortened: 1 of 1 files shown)


sandrotonon:
Feel free to merge or dismiss this. We could really need this feature, so i hope this pr is fulfilling your requirements.

Issue #15311, opened 2024-04-19
inputNumber with numeric prefix is not working as expected


Anirudh16:
### Describe the bug

unable to enter 0 if prefix is given as 0

For Example:if we provide prefix as 0 and if we want to enter value as '10' then after 1 unable to enter 0

Please find the stackblitz link below for reference:
[https://stackblitz.com/edit/1a8vuh?file=src%2Fapp%2Fdemo%2Finput-number-prefix-suffix-demo.ts](https://stackblitz.com/edit/1a8vuh?file=src%2Fapp%2Fdemo%2Finput-number-prefix-suffix-demo.ts)

### Environment

Windows 10

### Angular version

17.3.1

### PrimeNG version

17.13.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

16.16.0

### Browser(s)

Chrome

### Steps to reproduce the behavior

1.Go to stackblitz link
2. In the primeng input try to enter 0,it wont allow

### Expected behavior

it should allow 0 to be entered

maruthumj:
@Anirudh16 I don't think it's an issue for number 0 alone. the same issue occurs for any other prefix number.

gercobrandwijk:
This one is probably also related? https://github.com/primefaces/primeng/issues/13868

PR #15323, opened 2024-04-20
skip replacement if prefix is a number on an input number field

elian1780:
issue #15311


cetincakiroglu:
Hi,

It breaks the component.

Try with:

```
<p-inputNumber inputId="integeronly" [(ngModel)]="value1" prefix="+01"> </p-inputNumber>
```

elian1780:
sorry cant seem to replicate, with what value does it break?

Issue #15308, opened 2024-04-18
Dropdown: placeholder isn't shown when having null as a selectable option


rubenteck:
### Describe the bug

When adding null as a selectable option, the placeholder isn't shown on load, when selecting the null value (the 'empty' option) or when clearing the dropdown. In my example, the values are objects of type { label: 'label', value: 'value' } with the optionLabel and optionValue params set on the p-dropdown component. The null value is set as { label: null, value: null }.

### Environment

I use a windows 11 laptop with IntelliJ. Angular 17.3.4, PrimeNG 17.13.0 and node v18.20.1.

### Reproducer

https://stackblitz.com/edit/github-bjxtkz?file=src%2Fapp%2Fapp.component.html

### Angular version

17.3.4

### PrimeNG version

17.13.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

v18.20.1

### Steps to reproduce the behavior

1. create a dropdown component with null value as an option (see stackblitz)
2. start the app, the placeholder is not visible
3. select the empty option, the placeholder is not visible
4. click the clear button, the placeholder is not visible

### Expected behavior

the dropdowns value should be the placeholder when loaded, cleared or empty is selected.

rosenthalj:
PrimeNG's Dropdown  allows the developer to define both a label and value.     The problem you are experiencing is because you are not fully utilizing this capability.  I have updated your reproducer and the dropdown works as expected as long as you use the new getValue method to access the selected.  PrimeNG is designed for your use case and even more complicated use cases.

updated reproducer:  https://stackblitz.com/edit/github-bjxtkz-ueanuf?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.module.ts

The video below, that demonstrates your use case working properly with the current PrimeNG dropdown


https://github.com/primefaces/primeng/assets/45439491/1e4fa756-44e5-4b46-94e4-065a57d26ad1

rubenteck:
Thank you for your solution. While it's possible to create an empty value like this, I don't think it solves the problem of the empty value not showing the placeholder when selected. If I'm wrong about this, please explain it further :)

rosenthalj:
> Thank you for your solution. While it's possible to create an empty value like this, I don't think it solves the problem of the empty value not showing the placeholder when selected. If I'm wrong about this, please explain it further :)

I don't think I can explain further.

[...]

Issue #15324, opened 2024-04-20
Component: Editor isn't functional within dialogs


aaron-cardenas:
### Describe the bug

Since the EditorModule fix doesn't support SSR in v17.13.0, the editor isn't functional within dialogs. The same issue persists in 17.14.x. Reverting to the 17.12.x fix leaves you with a choice between this functionality and SSR compatibility.
https://github.com/primefaces/primeng/issues/8817

### Environment

Angular version 17.3.5
PrimeNG version 17.13.0 - 17.14.x
Node version 18.19.1

### Angular version

17.3.5

### PrimeNG version

17.13.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.19.1

### Browser(s)

All

### Steps to reproduce the behavior

Open a dialog panel (p-dialog) containing an Editor, it dosent work and don't have styles.

armingdev:
I hope this will be fixed soon. I'm blocked on the project because of this issue. #14931

nejat-njonjo:
You should put the editor inside if block, so it gets initialized on when the visible state changes, like below. I use it like this and it works jus fine.

```html
<p-dialog
  header="some header"
  [modal]="true"
  [(visible)]="visible"
  [style]="{ width: '50rem' }"
  [breakpoints]="{ '1199px': '75vw', '575px': '90vw' }"
  [maximizable]="true"
>
    @if (visible) {
     <p-editor></p-editor>
    }
</p-dialog>
```

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you open a new issue so we can include it in our roadmap?

Thanks a lot for your understanding!
Best Regards,

PR #15325, opened 2024-04-20
show 'empty' on dropdown list and empty label on placeholder

elian1780:
#15308


rosenthalj:
This PR attempts to fix issue #15308.  I do not believe its related to issue #15324.

The changes proposed will most likely break  many current developer's code that use the dropdown.  I believe (but have not confirmed) this will also break some of the PrimeNG demo examples.  Has this change been tested, at a minimum,  against all of the the PrimeNG dropdown demos?

Also,  as I commented in issue #15308,  I do not believe that issue #15308 is a valid bug.

elian1780:
pr on hold

Issue #15048, opened 2024-03-14
Component: Menubar with Screen Reader is not activating menuitems


mristov:
### Describe the bug

When Windows screen reader(Narattor) is turned on, 
on pressing 'Space' or 'Enter' **menuitem** is not receiving action, and is not opening.

But when Narattor is turned off, is working okay as expected

We can see this clearly in Basic example in official PrimeNg page
<img width="785" alt="image" src="https://github.com/primefaces/primeng/assets/150306240/8a35ee53-1157-4ad3-8759-91c8ef3f1849">
 

### Environment

Windows 11

### Angular version

17.0.5

### PrimeNG version

17.10

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.16.1

### Browser(s)

Chrome

PR #15359, opened 2024-04-25
Menubar: Enter key moves focus at first menuitem 

### LINKED ISSUE #15040: Menubar: Enter key moves focus at first menuitem
### Describe the bug

In terms of accessibility, user uses keyboard to navigate through the page. When there is a menubar without sub-menu items, only menu items at first level, pressing enter key at any of the menu items moves the focus to the first menuitem instead of keeping it on the clicked one.

### Environment

N/A

### Reproducer

_No response_

### Angular version

17.0.8

### PrimeNG version

17.7.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

20.10.0

### Browser(s)

All

For example, look at the primeng documentation, https://primeng.org/menubar#basic menubar basic:
   
image

If users navigates with keyboard until **Quit** menu item, and does click on it, you will see that focus is moved to **File** menu item. This behavior is pretty confusing for those users that use some Screen Readers, as it will announce that the focus has been move to another element into the menu.

### Expected behavior

What is expected for those menu items (first level) without sub-items is that focus keeps on the clicked menu item.

RogueTea:
Solves issue [#15040 ](https://github.com/primefaces/primeng/issues/15040)
Fixed #15040 
menu items now keeps the focus the clicked/Enterkey one and no longer focuses back to first menuitem

### DIFF
--- a/src/app/components/menubar/menubar.ts
+++ b/src/app/components/menubar/menubar.ts
@@ -1050,10 +1050,6 @@ export class Menubar implements AfterContentInit, OnDestroy, OnInit {
 
             anchorElement ? anchorElement.click() : element && element.click();
 
-            const processedItem = this.visibleItems[this.focusedItemInfo().index];
-            const grouped = this.isProccessedItemGroup(processedItem);
-
-            !grouped && (this.focusedItemInfo().index = this.findFirstFocusedItemIndex());
         }
 
         event.preventDefault();


src/app/components/menubar/menubar.ts
```diff
@@ -1050,10 +1050,6 @@ export class Menubar implements AfterContentInit, OnDestroy, OnInit {
 
             anchorElement ? anchorElement.click() : element && element.click();
 
-            const processedItem = this.visibleItems[this.focusedItemInfo().index];
```
SoyDiego, line 1053: mmm I'm not sure if this is ok for accessibility. 
@TaneliTuomola could you check it, please? :)

RogueTea: I am aware there is a known issue(#15048) for accessibility for keyboard navigation whilst using Microsoft narrator. That issue I believe still needs to be resolved.

PR #15385, opened 2024-04-27
Bug #15342 - Split Button: Aria Label not correctly working

### LINKED ISSUE #15342: Split Button: Aria Label not correctly working
### Describe the bug

The ButtonProps and MenuButtonProps are not correctly defined and cannot be used currently. 

They define a property 'ariaLabel', but in the code in is accessed as 'aria-label'. 
The same problem exists for the other properties of MenuButtonProps. 

For reference see: 
- https://github.com/primefaces/primeng/blob/17.14.1/src/app/components/splitbutton/splitbutton.ts#L74
- https://github.com/primefaces/primeng/blob/17.14.1/src/app/components/splitbutton/splitbutton.interface.ts

I tested it in v17.13, but as you see in the links above, it should still exist in 17.14.1.

 

### Environment

I am using angular 17

### Reproducer

_No response_

### Angular version

17

### PrimeNG version

17.13.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

10

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

maruthumj:
fixes #15342

### DIFF
--- a/src/app/components/splitbutton/splitbutton.ts
+++ b/src/app/components/splitbutton/splitbutton.ts
@@ -32,7 +32,7 @@ type SplitButtonIconPosition = 'left' | 'right';
                     (click)="onDefaultButtonClick($event)"
                     [disabled]="disabled"
                     [attr.tabindex]="tabindex"
-                    [attr.aria-label]="buttonProps?.['aria-label'] || label"
+                    [ariaLabel]="buttonProps?.['ariaLabel'] || label"
                     pAutoFocus
                     [autofocus]="autofocus"
                 >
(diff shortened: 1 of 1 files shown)


src/app/components/splitbutton/splitbutton.ts
```diff
@@ -32,7 +32,7 @@ type SplitButtonIconPosition = 'left' | 'right';
                     (click)="onDefaultButtonClick($event)"
                     [disabled]="disabled"
                     [attr.tabindex]="tabindex"
-                    [attr.aria-label]="buttonProps?.['aria-label'] || label"
+                    [ariaLabel]="buttonProps?.['ariaLabel'] || label"
```
SoyDiego, line 35: I think the correct way is `attr.aria-label` in Angular. Are you sure about this change?
Maybe @cetincakiroglu can confirm it

maruthumj: @SoyDiego Since we are using pbutton as a directive and pbutton has the property named "ariaLabel", we can pass the property name as "ariaLabel".

SoyDiego: Thanks for clarify 😊

SoyDiego (approved):

src/app/components/splitbutton/splitbutton.ts
maruthumj: > Thanks for clarify 😊

No problem 🙂

PR #15398, opened 2024-04-29
Fixed inputNumber with numeric prefix is not working as expected 

### LINKED ISSUE #15311: inputNumber with numeric prefix is not working as expected

nbrown-ScottLogic:
FIXED  #15311 

The issue was essentially that each time a value was entered a RegExp would be used to replace the prefix and suffix with '' however the RegExp had the /g flag and as such would replace all occurances of the pre/suffix including any occurances entered by the user. I also fixed some issues around the use of backspace and delete keys when the selector was around the prefix and suffix

### DIFF
--- a/src/app/components/inputnumber/inputnumber.ts
+++ b/src/app/components/inputnumber/inputnumber.ts
@@ -616,16 +616,13 @@ export class InputNumber implements OnInit, AfterContentInit, OnChanges, Control
     }
 
     getPrefixExpression(): RegExp {
-        console.log("prefix", this.prefix);
 
         if (this.prefix) {
             this.prefixChar = this.prefix;
-            console.log("prefix char", this.prefixChar);
 
         } else {
             const formatter = new Intl.NumberFormat(this.locale, { style: this.mode, currency: this.currency, currencyDisplay: this.currencyDisplay });
             this.prefixChar = formatter.format(1).split('1')[0];
-            console.log("prefix char", this.prefixChar);
 
         }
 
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi @nbrown-ScottLogic,

If I enter the number same as suffix in empty input, it doubles the number. Could you please also take a look at it? Also, there are accordion related code in your PR, could you please remove them. It should be one file updated: inputnumber.

Screenshot 2024-05-03 at 13 06 01

nbrown-ScottLogic:
@cetincakiroglu Changes you've requested should be present now, thanks for the feedback.

cetincakiroglu:
Hi,

PR #15407, opened 2024-04-30
fix broken megamenu keyboard naviagation

### LINKED ISSUE #15110: MegaMenu | Keyboard navigation is broken
Keyboard navigation doesn't work as expected.

To replicate:

1-) Focus on the menu
2-) Open a root item
3-) Try to navigate by using arrow keys.

dobanisola-scottlogic:
### Defect Fixes
Fixes #15110 
The megamenu keyboard navigation was not working because the visibleItems kept returning the default processedItem rather than the items from the current processedItem. I removed the logic that was comparing the key with the focusedItem's parentKey. I don't know what that was doing but I manually tested the menu and also added a unit test. 

It would be great to have all the unit test fixed so it'll be easy to tell if a change has broken something or not.

### DIFF
--- a/src/app/components/megamenu/megamenu.spec.ts
+++ b/src/app/components/megamenu/megamenu.spec.ts
@@ -211,7 +211,7 @@ describe('MegaMenu', () => {
         fixture.detectChanges();
 
         expect(itemClickSpy).toHaveBeenCalled();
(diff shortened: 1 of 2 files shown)


cetincakiroglu:
Hi,

We appreciate the effort and support! Thanks a lot for helping us to make PrimeNG better!

Issue #15332, opened 2024-04-22
Fileupload and autocomplete: problems with input transform function


Kue2:
### Describe the bug

Since PrimeNG 17.13 we have been encountering unexpected behaviors with some components that now use the numberAttribute transform function for some inputs, probably because it is not defined a fallback value. The problem arise when the input is passed to the prime component, but with an undefined value (which is evaluated as NaN after being processed by the numberAttribute function)

Two examples:
In **p-fileupload** the input "**filelimit**" as undefined prevents the upload button from becoming active when a file is selected
https://stackblitz.com/edit/czauw4-y22yf5?file=src%2Fapp%2Fdemo%2Ffile-upload-template-demo.ts

In **p-autocomplete** the input "**maxlength**" as undefined renders the component unusable because the maxlength is ultimately evaluated as 0.
https://stackblitz.com/edit/mcz31v?file=src%2Fapp%2Fdemo%2Fautocomplete-basic-demo.html

### Environment

Angular

### Reproducer

https://stackblitz.com/edit/czauw4-y22yf5?file=src%2Fapp%2Fdemo%2Ffile-upload-template-demo.ts
https://stackblitz.com/edit/mcz31v?file=src%2Fapp%2Fdemo%2Fautocomplete-basic-demo.html

### Angular version

17

### PrimeNG version

17.13.0 - 17.14.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18

### Steps to reproduce the behavior

- p-fileupload: set the input "filelimit" as undefined
- p-autocomplete: set the input "maxlength" as undefined

PR #15436, opened 2024-05-02
Fix: Fileupload and autocomplete problems with input transform function

Kue2:
Fixes https://github.com/primefaces/primeng/issues/15332


cetincakiroglu:
Hi,

Issue #13929, opened 2023-10-19
TieredMenu Submenus do not display within viewport 


shawnss:
### Describe the bug

When tiered menu is on the far right side of page the submenus do not automatically display on the left. This was working correctly in Prime NG version 15 (https://stackblitz.com/edit/a2runv?file=src%2Fapp%2Fdemo%2Ftiered-menu-popup-demo.html). 

![image](https://github.com/primefaces/primeng/assets/7524790/10608e14-3067-4e3a-96ac-bca724198717)

### Environment

Windows

### Reproducer

https://stackblitz.com/edit/oje4xw?file=src%2Fapp%2Fdemo%2Ftiered-menu-popup-demo.html

### Angular version

16.x

### PrimeNG version

16.x

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

16.14.2

stity:
Related issue : https://github.com/primefaces/primeng/issues/12541

shawnss:
This was working in version 15!
https://stackblitz.com/edit/a2runv?file=package.json

AliHamrouni:
Salut @shawnss comment vous avez résolut le problème, la class que vous avez implimentez s.v.p

TaylorHo:
Hey @AliHamrouni, in my case, I have added globally the following CSS style:

```
.p-tieredmenu .p-submenu-list {
  left: -100%;
}
```

It works fine as a temporary troubleshooting since this CSS inverts the side where the submenu appears.

AliHamrouni:
merci !! @shawnss

WienemannCC:
We are experiencing the same problem. This should be fixed, the workaround is just that.

Yuri-Lima:
I confirm also that issue still happening...

![Image](https://github.com/user-attachments/assets/2b9dfbd2-ec4f-4f5c-b78a-83c07edbd3e3)

DevCeN:
Same, still bugged in the latest v19

PR #15447, opened 2024-05-03
fix: tieredMenuSub improve self-positioning in viewPort when viewPort…

f94e59:
Follow up of [13929](https://github.com/primefaces/primeng/issues/13929)

Fixes edge case: When viewPort size has changed between activations remove flipped class again if there is enough space. Also this MR cleans up the redundand code I added before regarding afterViewInit.


cetincakiroglu:
Hi,

Could you please clean up the conflicts so we can merge safely?

PR #15474, opened 2024-05-07
chore(tieredmenu): remove aria level attribute from list item

### LINKED ISSUE #15130: SplitButton: menu item is using 'aria-level' attribute incorrectly
### Describe the bug

We are using prime ng in an application that is legally required to be WCAG compliant. To check compliance we are using [Axe devtools](https://www.deque.com/axe/browser-extensions/#:~:text=Overview-,Overview,your%20time%20with%20false%20positives.). 

It appears that the menu items for the split button are not setting the correct aria attributes, as we are seeing this error: 
`ARIA attribute is not allowed: aria-level="1"`

image

### Environment

Windows 11 Pro

### Reproducer

_No response_

### Angular version

16.2.9

### PrimeNG version

16.9.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.18.2

### Browser(s)

chrome 123

Open the menu of a splitbutton and run a scan using axe devtools.

### Expected behavior

I'm not entirely sure if the attribute should be corrected or removed, but axe devtool should not indicate it as a problem.

ppc-harmoney:
Resolves #15130 

This is my first contribution. Forgive me if I have failed to follow S.O.P.

### Defect Fixes
This PR removes the `aria-level` attribute from the `<li>` items in the tiered menu component. The reason being that this attribute was flagged as incorrect by axe devtools (presumably due to the role attribute being incompatible with aria-level?)

### DIFF
--- a/src/app/components/tieredmenu/tieredmenu.ts
+++ b/src/app/components/tieredmenu/tieredmenu.ts
@@ -77,7 +77,6 @@ import { ObjectUtils, UniqueComponentId, ZIndexUtils } from 'primeng/utils';
                     [attr.aria-disabled]="isItemDisabled(processedItem) || undefined"
                     [attr.aria-haspopup]="isItemGroup(processedItem) && !getItemProp(processedItem, 'to') ? 'menu' : undefined"
                     [attr.aria-expanded]="isItemGroup(processedItem) ? isItemActive(processedItem) : undefined"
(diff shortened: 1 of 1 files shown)


cetincakiroglu:
Hi,

ppc-harmoney:
You're very welcome :)

PR #15532, opened 2024-05-10
Update calendar props types

zachary-svoboda-accesso:
Update calendar documentation to clarify the specific values that are allowed

### DIFF
--- a/src/app/showcase/doc/calendar/propsdoc.ts
+++ b/src/app/showcase/doc/calendar/propsdoc.ts
@@ -223,7 +223,7 @@ import { Component, Input } from '@angular/core';
                     </tr>
                     <tr>
                         <td>dataType</td>
-                        <td>string</td>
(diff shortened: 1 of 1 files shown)


src/app/showcase/doc/calendar/propsdoc.ts
```diff
@@ -223,7 +223,7 @@ import { Component, Input } from '@angular/core';
                     </tr>
                     <tr>
                         <td>dataType</td>
-                        <td>string</td>
+                        <td>'date' | 'string'</td>
```
SoyDiego, line 226: This not should be edit here, should be in `calendar.ts` and then use the command: `npm run apidoc` and this file will be generated automatically

cetincakiroglu:
Hi,

propsdoc.ts is not used in the project and will be deleted. Closing the PR.

PR #15563, opened 2024-05-14
#15512 Table: Numeric Filter - issue fixed

### LINKED ISSUE #15512: Table: Numeric Filter
### Describe the bug

The Numeric Filter is broken on the ColumnFilter as the currency input was changed to a boolean. However, the currency input on the InputNumber component is still a string and is treated as such. This breaks all numeric filtering on the table.

image

### Environment

PrimeNG

### Reproducer

_No response_

### Angular version

17.13.1

### PrimeNG version

17.15.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

20

### Browser(s)

_No response_

_No response_

### Expected behavior

Should be able to use the numeric filter type on the ColumnFilter

F3KTOR:
#15512 Table: Numeric Filter issue fixed based on the discussion with @ethanolre by converting the currency value in table.ts into string from boolean | undefined. The reasoning behind it is that it is the form that it is in the inputnumber.ts file and it works flawlessly there.

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -5243,7 +5243,7 @@ export class ColumnFilter implements AfterContentInit {
      * Enables currency input.
      * @group Props
      */
-    @Input({ transform: booleanAttribute }) currency: boolean | undefined;
+    @Input({ transform: booleanAttribute }) currency: string;
     /**
      * Defines the display of the currency input.
      * @group Props


src/app/components/table/table.ts
```diff
@@ -5243,7 +5243,7 @@ export class ColumnFilter implements AfterContentInit {
      * Enables currency input.
      * @group Props
      */
-    @Input({ transform: booleanAttribute }) currency: boolean | undefined;
+    @Input({ transform: booleanAttribute }) currency: string;
```
ethanolree, line 5246: I don't think this should use the booleanAttribute transform if it's no longer a boolean

F3KTOR: If you say so. I went with the idea of it having a few booleanAttributes for non boolean inputs. That's why I've left it as is. But I'll change it if you think it's the right call.

cetincakiroglu:
Hi,

We are planning to release a new version today. I have fixed this issue to speed up the process and you can ignore my review feedback to include it.

Fixed in https://github.com/primefaces/primeng/pull/15592

F3KTOR:
Hey man.

Superb. Thanks for the notification.

Issue #15557, opened 2024-05-14
Primeng Table: Clear the filters which are not applied when filter menu closed by clicking outside the menu


gopikrishna-manifest:
### Describe the bug

Hi Team,
We are using Primeng Tables filter option. We need to clear the filters which are selected but not applied when we close filter menu by clicking outside of the menu. 
**Steps to reproduce:**
1) Open the Primeng filter menu example. [https://primeng.org/table#filter-menu](https://primeng.org/table#filter-menu)
2) Click on Filter icon beside Name / Country column and open the filter menu.
3) Enter some text in the text box 
![Screenshot_4](https://github.com/primefaces/community/assets/85510959/4ecf0c6b-eba1-4217-8a39-45836e8d1477)
4) Click out side the filter menu. Now the filter menu will close without applying the filter
5) But you can see the filter icon turns to blue indicating that some filter applied to that column
![image](https://github.com/primefaces/community/assets/85510959/08e1944b-0971-4ebe-81af-92419a19e917)

6) Again open the Filter menu for the same column
7) You can see the same text in the filter.
![image](https://github.com/primefaces/community/assets/85510959/a4a2f8a4-51ad-4266-937e-f70c09bba187)

**Expected Result**
1) It should clear the filters which are not applied 
2) The Filter icon shouldn't turn to blue

Actual Result
1) It is not clearing the filters which are not applied when we click outside the menu
2) The icon turning to blue even no filter applied to that column.

PrimeNG Discussion Link [https://github.com/orgs/primefaces/discussions/1718](https://github.com/orgs/primefaces/discussions/1718)

### Environment

Angular And Windows

### Angular version

17.3.1

### PrimeNG version

17.16.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.13.0

### Browser(s)

Chrome

Vlad66M:
Have you solved the problem?

PR #15594, opened 2024-05-16
Table: clear filter when dirty but not applied 

RogueTea:
Fixes [#15557]( https://github.com/primefaces/primeng/issues/15557)


cetincakiroglu:
Hi @RogueTea 

PR #15617, opened 2024-05-20
fix: handle autocomplete input change when forceSelection is enabled

### LINKED ISSUE #15488: Component: p-autoComplete multiple=true sometimes causes adding different/multiple values if one of them is a prefix of another
### Describe the bug

Hello,

I've noticed that `<p-autoComplete [multiple]="true" ... />` sometimes causes adding multiple values if one of them is prefix of another, and sometimes causes adding prefix value only rather than selected:

https://github.com/primefaces/primeng/assets/23403545/efecf374-db07-4224-899c-6bdf2cadb6e8

(notice "test" was added rather than "test-1", and later both "test" & "test-1" were added)

To anyone that may have the same issue, the temporary workaround I've came up with is to use suggestions as an objects with prefixed `label` & custom `ng-template`:

https://stackblitz.com/edit/github-7hs33e-ciep28?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts

### Environment

https://stackblitz.com/edit/github-7hs33e-wudgma?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts

### Reproducer

https://stackblitz.com/edit/github-7hs33e-wudgma?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts

### Angular version

17.3.7

### PrimeNG version

17.16

### Build / Runtime

Angular CLI App

### Language

TypeScript

stackblitz so idk

### Browser(s)

_No response_

1. Open https://stackblitz.com/edit/github-7hs33e-wudgma?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts
2. Type "test"
3. Select "test-1"
4. Sometimes both "test" and "test-1" were added and sometimes "test" was wrongly added even if user clicked on "test-1".

### Expected behavior

Only single selected item is added.

Sinan997:
Resolves #15488

### DIFF
--- a/src/app/components/autocomplete/autocomplete.ts
+++ b/src/app/components/autocomplete/autocomplete.ts
@@ -1113,7 +1113,11 @@ export class AutoComplete implements AfterViewChecked, AfterContentInit, OnDestr
     }
 
     onInputChange(event) {
-        if (this.forceSelection) {
+        setTimeout(() => {
+            if (this.focused) return;
+
+            if (!this.forceSelection) return;
+
             let valid = false;
 
             if (this.visibleOptions()) {
@@ -1129,7 +1133,7 @@ export class AutoComplete implements AfterViewChecked, AfterContentInit, OnDestr
                 this.inputEL.nativeElement.value = '';
                 !this.multiple && this.updateModel(null);
             }
-        }
+        }, 150);
     }
 
     onInputFocus(event) {


src/app/components/autocomplete/autocomplete.ts
```diff
@@ -1113,7 +1113,11 @@ export class AutoComplete implements AfterViewChecked, AfterContentInit, OnDestr
     }
 
     onInputChange(event) {
-        if (this.forceSelection) {
+        setTimeout(() => {
+            if (this.focused) return;
```
cetincakiroglu, line 1117: why should we return if focused is true?

src/app/components/autocomplete/autocomplete.ts
```diff
@@ -1113,7 +1113,11 @@ export class AutoComplete implements AfterViewChecked, AfterContentInit, OnDestr
     }
 
     onInputChange(event) {
-        if (this.forceSelection) {
+        setTimeout(() => {
```
cetincakiroglu, line 1116: why setTimeout used in here?

cetincakiroglu (changes requested):

cetincakiroglu:
It also works if I only return if focused = true and remove everything else. It seems not a valid solution.

Sinan997:
hi @cetincakiroglu thanks for response. Somehow it seems it works fine in new release. I am closing this pull request.

PR #15658, opened 2024-05-22
Fix #15657 reading properties undefined Editable Column changes

### LINKED ISSUE #15657: EditableColumn: error in ngOnChanges when input other than data changes
### Describe the bug

In Reference to this issue https://github.com/primefaces/primeng/issues/14525 which was fixed for 17 version. I am facing same issue on version 16 now

When value of any input attribute such as pEditableColumnDisabled changes during runtime, an error is thrown due to change.data being undefined.
I believe that error is in:

[primeng/src/app/components/table/table.ts](https://github.com/primefaces/primeng/blob/8c9f3d05e2d52aa37a0ff54aeb856294d1277614/src/app/components/table/table.ts#L4067)
Line 4067 in [8c9f3d0](https://github.com/primefaces/primeng/commit/8c9f3d05e2d52aa37a0ff54aeb856294d1277614)
` if (this.el.nativeElement && !data.firstChange) { `

Which assumes that property data of SimpleChanges object will be always defined, which is not true when only the pEditableColumnDisabled input changes (f.e. with slight delay).

### Environment

unrelated

### Reproducer

https://stackblitz.com/edit/github-8qgwug-rbxqag?file=src%2Fapp%2Fapp.component.html

### Angular version

16.2.12

### PrimeNG version

16.9.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.10.0

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

ps-2512:
SimpleChanges might not have data defined when only other input than data changes.

For reference see reporting issue: https://github.com/primefaces/primeng/issues/14525 which was fixed for version 17
Changes intended are for https://github.com/primefaces/primeng/issues/15657 which are required for version 16.9.1 too


cetincakiroglu:
Hi,

Thanks for reporting the issue, we don't support v16 and v16 will not receive an update, please use newer versions.

Issue #15656, opened 2024-05-22
Table: Reorderable column false still gives cursor move


Emiliorth:
### Describe the bug

Table: Reorderable column false and pReorderableColumnDisabled true still gives cursor move

### Environment

Windows 10 PrimeNG 17.17.0 Angular 17

### Angular version

17.3.0

### PrimeNG version

17.17.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

21.6.0

### Browser(s)

Chrome

RogueTea:
Hello, I'm wondering if you're still using pReorderableColumn in your code like: 
```
  <p-table [value]="products" [columns]="cols" [reorderableColumns]="false" [tableStyle]="{ 'min-width': '50rem' }">
                    <ng-template pTemplate="header" let-columns>
                        <tr>
                            <th style="width:3rem"></th>
                            <th *ngFor="let col of columns" pReorderableColumn>
                                {{ col.header }}
                            </th>
                        </tr>
                    </ng-template>
                    <ng-template pTemplate="body" let-rowData let-columns="columns" let-index="rowIndex">
                        <tr [pReorderableRow]="index" [pReorderableRowDisabled]="true">
                            <td>
                                <span class="pi pi-bars" pReorderableRowHandle></span>
                            </td>
                            <td *ngFor="let col of columns">
                                {{ rowData[col.field] }}
                            </td>
                        </tr>
                    </ng-template>
                </p-table>
```

Emiliorth:
Yes, I do use [reorderableColumns] binding on p-table and this is my th:
`
 <th [pReorderableColumnDisabled]="someReordCondition"
        [pSortableColumn]="sortField"
        [pSortableColumnDisabled]="someSortCondition"
        pReorderableColumn
        [id]="id">
...
    </th>
`
I am looping over my @Input() columns instead of using let-columns
I do not use row reordering

PR #15667, opened 2024-05-23
Table: Reorderable column false still gives cursor move

RogueTea:
Fixes [#15656](https://github.com/primefaces/primeng/issues/15656)


src/app/components/table/table.ts
```diff
@@ -4025,7 +4025,8 @@ export class ResizableColumn implements AfterViewInit, OnDestroy {
 @Directive({
     selector: '[pReorderableColumn]',
     host: {
-        class: 'p-element'
+        class: 'p-element',
+        '[class.p-reorder-column-disabled]': '!isEnabled()'
```
cetincakiroglu, line 4029: Instead of adding class, could you please use style? 

```
    host: {
        class: 'p-element',
        '[style.cursor]':'isEnabled() ? "move" : "default"'
    }
```

RogueTea: Changed to style.

cetincakiroglu:
Thanks a lot for the quick action! @RogueTea

PR #15681, opened 2024-05-23
Update `prettier` version for new control flow

### LINKED ISSUE #15653: Old prettier version that does not support new control-flow syntax
### Describe the bug

Not a bug report. While implementing my recent PR and using the new control-flow syntax I noticed issues in formatting.

It seems that an old version of `prettier` is installed, which does not correctly support indentation in control-flow syntax which can be seen [here](https://github.com/primefaces/primeng/pull/15644/files#diff-31bf3441e64a90020c49d3c045a49e5c2c15faa8bd589cd5407cac59bfdcd381R22).

Sinan997:
Resolves #15653 

Angular new control flow syntax support was added to prettier after the `3.1` release. 

Updating like this `"prettier": "^3.0.0"` will support the new control flow syntax in PrimeNG. I have added the `caret symbol` to get the latest minor version.


SoyDiego (approved):

alexciesielski:
Not formatting the whole repository after this will lead to a lot of diffs...
But then again this would also mean a lot of merge conflicts in all open PRs...

Issue #15706, opened 2024-05-27
SplitButton: rotate chevron while TieredMenu is open


tri-harmoney:
### Describe the bug

The chevron always points downwards.
We would like to have the ability based on a boolean input to have the chevron rotate while the menu is open.

### Environment

MacOS 14.5 (23F79)

### Angular version

17.3.7

### PrimeNG version

17.16.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.13.0

### Steps to reproduce the behavior

Open the TieredMenu of a SplitButton. The chevron never rotates.

### Expected behavior

If the TieredMenu is open, based on a boolean input have the chevron rotate.

PR #15709, opened 2024-05-27
feature(splitbutton): 15706 add ability to have rotation of chevron icon

tri-harmoney:
Resolves https://github.com/primefaces/primeng/issues/15706

### DIFF
--- a/src/app/components/common/common.css
+++ b/src/app/components/common/common.css
@@ -155,6 +155,11 @@
         -webkit-animation: p-icon-spin 2s infinite linear;
         animation: p-icon-spin 2s infinite linear;
     }
+
+    .p-icon-rotate-180 {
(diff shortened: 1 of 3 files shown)


SoyDiego (commented):
I dont understand why is necessary the functionality rotate in the splitbutton 😕

tri-harmoney:
@SoyDiego I changed the proposal. We want to be able to rotate the chevron based on if the split button is expanded or not.
While not doing too much I added a css class `is-expanded`. This is a general solution and we can now rotate the chevron in our css based on that class. Do you agree more with this solution?

cetincakiroglu:
Hi,

Rotating the chevron is not in our design system. You can use `dropdownicon` template to customize the icon. Thanks a lot for the effort, closing the PR.

Issue #15778, opened 2024-06-03
Component: Table (Advanced Filter)


mastersys-giovanni:
### Describe the bug

In the advanced filter popup, when you press enter, the filter is applied, but does not save, that is, when you click outside to hide the popup, you lose the filter.

### Environment

Windows 12

### Angular version

18.0.1

### PrimeNG version

17.8.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.14

### Browser(s)

Chrome

### Steps to reproduce the behavior

in the advanced filter example itself

### Expected behavior

When you press enter, it should have the same behavior as clicking apply

rosenthalj:
The video listed below recreates the bug described in the issue.  I followed the steps provided in the description:  "when you press enter, the filter is applied, but does not save, that is, when you click outside to hide the popup, you lose the filter."


https://github.com/primefaces/primeng/assets/45439491/e32da37f-627f-42dd-8eb2-38216e680ea3

cetincakiroglu:
Fixed in https://github.com/primefaces/primeng/pull/15808

PR #15808, opened 2024-06-07
Fixed #15804 - Advance Filtering is broken (revert PR #15594 changes)

### LINKED ISSUE #15804: Table: Advance Filtering is broken
### Describe the bug

Advance table filters that have been "previously" applied are automatically cleared when a user starts a filter dialog overlay and then closes the filter dialog by clicking outside of the dialog overlay.

The Video listed below demonstrates the issue using the current PrimeNG Advance Filter demo (https://primeng.org/table#filter-advanced): 

https://github.com/primefaces/primeng/assets/45439491/e46a6f6f-5996-4f38-b2a3-3df499512ff4

This issue was introduced with the closing of issue #15557 by PR #15594

The changes that were made for PR #15594  failed to take into account a filter may have already been previously defined before the filter dialog is "relaunched" and then closed without making any changes to the filter.   

With Advance Table Filters,  the actual "filter rule" is not visible.  The only way for a user to see how a column filter is set is to display the column filter dialog.  The user may not want to make any changes or "reapply" the current filter.    This is a very common use case in my application.

### Environment

max and windows

### Reproducer

https://primeng.org/table#filter-advanced

### Angular version

17.x

### PrimeNG version

17.18.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18

### Browser(s)

chrome, ff, safari

1. Start the PrimeNG  Table Advanced Filter Demo:  https://primeng.org/table#filter-advanced
2.  Filter the demo's table rows using any of the column filters  (via the column filter dialog) 
3.  Start the same filter dialog as step 2, but then immediately close the filter (without changing the filter) by clicking  outside of the filter dialog
4.  The filter that was set in step 2 will automatically be cleared and all rows will be displayed

### Expected behavior

After step 4,  the filter should not be cleared and the table rows should remain filtered

rosenthalj:
This PR reverts the changes made in PR #15594

This PR also fixes issue #15778

This PR fixes the functionality of the PrimeNG's Table Advance Filters which was  "Critically" broken when trying to fix issue #15557.   Note:  PR #15594 actually only partially fixed issue #15557.

This PR also breaks issue #15557.  Since the problem identified in issue #15557 has been in PrimeNG for multiple years without being fixed,  I believe it is better to fix the Advance Filter immediately and in the future develop a more holistic/complete fix for issue #15557 that doesn't "critically" break  Advance Table Filters

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -5395,8 +5395,6 @@ export class ColumnFilter implements AfterContentInit {
 
     overlayId: any;
 
-    applyHasBeenClicked: boolean = false;
-
     get fieldConstraints(): FilterMetadata[] | undefined | null {
         return this.dt.filters ? <FilterMetadata[]>this.dt.filters[<string>this.field] : null;
     }
@@ -5651,11 +5649,6 @@ export class ColumnFilter implements AfterContentInit {
     }
 
     onEscape() {
-        if (this.hasFilterNotBeenApplied()) {
-            this.clearFilter();
-        }
-        this.applyHasBeenClicked = false;
-        this.overlayVisible = false;
         this.icon?.nativeElement.focus();
     }
 
@@ -5755,10 +5748,6 @@ export class ColumnFilter implements AfterContentInit {
         return false;
     }
 
-    hasFilterNotBeenApplied(): boolean {
-        return this.hasFilter() && !this.applyHasBeenClicked;
-    }
-
     isOutsideClicked(event: any): boolean {
         return !(
             DomHandler.hasClass(this.overlay?.nextElementSibling, 'p-overlay') ||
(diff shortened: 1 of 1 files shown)


rosenthalj:
Over the weekend I have been thinking about the how to **FULL** fix this problem (issue https://github.com/primefaces/primeng/pull/15811, issue https://github.com/primefaces/primeng/issues/15778, issue https://github.com/primefaces/primeng/issues/15557). I believe that the best/proper solution is when the filter overlay/dialog is hidden (via the hide method or any other means) the overlay/dialog is completely destroyed. When the the filter/dialog is displayed, it is always created and filled with the current filter setting. This is obviously not a trivial solution or this bug would have fixed (or initially implemented) years ago with this approach.

I believe that this PR #15808  (reverting changes made by PR #15594),  is the most prudent near-term fix.

Issue #15400, opened 2024-04-29
Multiselect: selectionLimit not working if the value 0 is provided


m3allimm4a4:
### Describe the bug

if the [selectionLimit] property is set to zero, it's treated the same as null or undefined (because javascript). Instead I am having to work around that by disabling all options if the selectionLimit is zero.
This error is caused by the following code snipet

```
maxSelectionLimitReached() {
        return this.selectionLimit && this.modelValue() && this.modelValue().length === this.selectionLimit;
}
```
null or undefined check should be done on this.selectionLimit

### Environment

Angular 17, reactive forms

### Angular version

17

### PrimeNG version

17.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18

### Browser(s)

chrome. edge

### Steps to reproduce the behavior

HTML:
```
<p-multiSelect
            formControlName="centrales"
            display="chip"
            appendTo="body"
            [options]="options.centrales"
            [showClear]="false"
            [style]="{ width: '100%' }"
            [virtualScroll]="true"
            [virtualScrollItemSize]="43"
            [selectionLimit]="maxSelectedConnectors">
</p-multiSelect>
```
TS:
when the value of maxSelectedConnectors is changed to 0, the feature doesn't work anymore as discribed previously

### Expected behavior

the selectionLimit should be set to 0, which means all options should be disabled

PR #15810, opened 2024-06-07
Multiselect: selectionLimit not working if the value 0 is provided

trigiacbmt:
Multiselect: selectionLimit not working if the value 0 is provided
[#15400 ](https://github.com/primefaces/primeng/issues/15400)


trigiacbmt:
fixes #15400

PR #15845, opened 2024-06-14
Fixed #15779 - Closing nested dialog removes CSS class still needed f…

### LINKED ISSUE #15779: Closing nested dialog removes CSS class still needed for parent dialog
#### Behaviour
When opening a dialog primeng adds the `p-overflow-hidden` class to the body which hides the scroll bar and adds padding to prevent the content from shifting. When a user closes a dialog then this class is removed from the body.

#### Bug
When a user opens a *nested* dialog and then closes it, the `p-overflow-hidden` class is removed from the body despite the parent modal still being present.

[This is demonstrated on your website in the last example](https://primeng.org/dynamicdialog)

### Environment

I am running a web application on an Ubuntu OS, however this issue is present in your examples where you show nested dialogs [here](https://primeng.org/dynamicdialog)

Because of this I do not believe it is an issue with my environment.

### Reproducer

_No response_

### Angular version

16.2.12

### PrimeNG version

16.0.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.19.0

### Browser(s)

Brave v1.66.118, Chrome  v125.0.6422.113 (Official Build) (64-bit)

- Find any nested dialog example (final example on [this page](https://primeng.org/dynamicdialog))
- Open your browser developer tools so that you have a view of the body template.
- click on the *Open Dialog* button.
- notice the appropriate css classes being applied to body template
- *Open* nested dialog
- *Close* nested dialog
- notice that css class has been removed from body template

### Expected behavior

I would expect that the css class which is added when a modal is opened would persist on the body as long as there is a modal open. When the final modal is closed then this css class should be removed.

mehmetcetin01140:
Fixes #15779

### DIFF
--- a/src/app/components/dynamicdialog/dynamicdialog.ts
+++ b/src/app/components/dynamicdialog/dynamicdialog.ts
@@ -217,6 +217,7 @@ export class DynamicDialogComponent implements AfterViewInit, OnDestroy {
 
     get parent() {
         const domElements = Array.from(this.document.getElementsByClassName('p-dialog'));
+        
         if (domElements.length > 1) {
             return domElements.pop();
         }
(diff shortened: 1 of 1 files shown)


src/app/components/dynamicdialog/dynamicdialog.ts
```diff
@@ -266,6 +267,13 @@ export class DynamicDialogComponent implements AfterViewInit, OnDestroy {
         return this.config?.templates?.closeicon;
     }
 
+    get dynamicDialogCount() {
+        let dynamicDialogs = this.document.querySelectorAll('p-dynamicdialog');
```
SoyDiego, line 271: You can use `const` instead of `let` because the variables will not be reasigned

SoyDiego (approved):

Issue #15816, opened 2024-06-10
Page freezes on resize when using many pFrozenColumn


PronDmytro:
### Describe the bug

The page is freezing on resize when using many pFrozenColumns.
In my work project, the page never unfreezes. I [reproduced](https://stackblitz.com/edit/vxva11?file=src%2Fapp%2Ftable-frozen-columns-demo.html) a similar table, and here the page freezes for 1-2 seconds.

Upd: When there are more rows, the freeze lasts longer.

It looks like the main problem is in this listener, because a listener is created for each instance of pFrozenColumn.
https://github.com/primefaces/primeng/blob/e5f593ea1f856a398414830b5cb613a5705fae4e/src/app/components/table/table.ts#L3329-L3338

### Environment

Windows 11

### Angular version

17.3.0

### PrimeNG version

17.18.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.12.1

nikolaSim:
This is critical, it freezes the whole application.

lautarogfernandez:
Is there any news from this fix? it is also freezing the whole application when it tries to run this method updateStickyPosition.

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation and try the latest PrimeNG version(v19). If there is no improvement on this, can you open a new issue so we can include it in our roadmap?

Thanks a lot for your understanding!
Best Regards,

MrRebchik:
I tried it on version 19. the problem of freezing when resizing remains

Issue #15303, opened 2024-04-18
Component: Table #15302


Lolibai:
### Describe the bug

https://user-images.githubusercontent.com/11733376/208727545-67866f60-7db8-4256-ae2b-43a0a178753d.mov

### Environment

Frozen Columns

### Reproducer

https://stackblitz.com/edit/qnmajp?file=package.json

### Angular version

17.3.1

### PrimeNG version

17.13.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.12.2

### Browser(s)

Chrome 123

### Steps to reproduce the behavior

1. Freeze columns to the right
2. Have at least 10 rows of data
3. See the "stairs" of columns that should be freezed

### Expected behavior

All columns should be freezed as per attribute and directive assignment.

Lolibai:
This is the fix #15302

PR #15302, opened 2024-04-18
Fix Issue with frozen columns #12408

Lolibai:
### Defect Fixes
Fixes #15303 

### Video with the actual issue attached

https://user-images.githubusercontent.com/11733376/208727545-67866f60-7db8-4256-ae2b-43a0a178753d.mov

### Environment

Frozen Columns

### Reproducer

Lolibai

### Angular version

17.3.1

### PrimeNG version

17.13.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.12.2

### Browser(s)

Chrome 123.0.6312.123

### Steps to reproduce the behavior

1. Freeze columns to the right
2. Have at least 10 rows of data
3. See the "stairs" of columns that should be freezed

https://user-images.githubusercontent.com/11733376/208727545-67866f60-7db8-4256-ae2b-43a0a178753d.mov

### Expected behavior

All columns should be freezed as per attribute and directive assignment.

### Stackblitz

[Reproduction](https://stackblitz.com/edit/qnmajp?file=package.json)

PR #15854, opened 2024-06-17
Fixed #15852 - Table: Frozen columns are displaced

### LINKED ISSUE #15852: Table: Frozen columns are displaced
### Describe the bug

In a table with many columns, of which also a large number are frozen, the placement of the fixed columns is totally messed up.
Headers are placed out of order, and row cells are not frozen at all

### Environment

Chrome
Edge
Firefox

### Reproducer

https://github-8nkwy5.stackblitz.io

### Angular version

17.3.9

### PrimeNG version

17.18.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

22.2.0

### Browser(s)

_No response_

1. Open the app
2. Scroll the table horizontally
3. Frozen columns scroll with the table

### Expected behavior

Frozen columns should remain fixed at their position, no matter how many columns are frozen, and no matter how large or small the screen is.

ranthonissen:
Fixed #15852 - Table: Frozen columns are displaced


bartlomiej-dobosz:
Hi @ranthonissen 
I haven't run your code but I just wanted to mention that the changes you made mostly undo the changes from this #15302 issue. Please check if your changes have not reversed that bug.

If so here is my qickfix:
```ts
recalculateColumns() {
        const siblings = DomHandler.siblings(this.el.nativeElement);
        const index = DomHandler.index(this.el.nativeElement);
        const time = this.alignFrozen === 'right' ? (siblings.length - index + 1) * 50 : (index + 1) * 50; //this line

        setTimeout(() => {
            this.updateStickyPosition();
        }, time);
        this.updateStickyPosition();
}
```

ranthonissen:
Hi @bartlomiej-dobosz,

As far as I am familiar with the code, the fix provided in #15302 did not solve the problem, but only delayed the calculation of the column position, causing this issue #15852.
The underlying issue was that `DomHandler.getOuterWidth(next)` and `DomHandler.getOuterWidth(prev)` in the `updateStickyPosition()` function always returned 0, because of the table not being rendered/displayed at the time of calculation.

I think this fix resolves that issue, making the delayed execution of `updateStickyPosition()` unnecessary.

ranthonissen:
I haven't tested it myself, but i might also fix #15816, as it is the same code portion that seems to cause that issue

bartlomiej-dobosz:
I found bug. When you add more columns and start sorting, this happens:

image

ranthonissen:
@bartlomiej-dobosz
I'm unable to reproduce the behavior you show in your screenshot .
Can you provide me with code that shows this bug?

chrome_wnG8ynj2Mg

ranthonissen:
@bartlomiej-dobosz Any update on this PR?

The issues on the frozen columns are a showstopper for our team.
So if there is any way I can help and speed things up (with this and related frozen table column issues), don't hesitate to reach out to me, and I will see how I can help.

I think some more issues were introduced by #15302. So I think it might be a good idea to revert these changes, start over from that point with another approach, and closely keep an eye on the fix for the original issue #15303, and making sure no new issues get introduced.

bartlomiej-dobosz:
Sorry I forgot to reply here.
I think the error I described above was the result of my mistake. Ignore it.
In my project I had to hack sorting to add a third state (ASC, DESC and NONE). And this affects your solution.

Issue #15647, opened 2024-05-21
AutoComplete: Increased drop-down height on first search when using virtual scroll


can-oezkan:
### Describe the bug

When using the autocomplete component with virtual scroll, the height of the drop-down on first search is increased.

![image](https://github.com/primefaces/primeng/assets/161650114/88366a09-50b9-4bd8-a050-8ec91cdceee2)

### Environment

unrelated

### Reproducer

https://primeng.org/autocomplete#virtualscroll

### Angular version

unrelated

### PrimeNG version

v17.17.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

unrelated

### Steps to reproduce the behavior

1. enable virtualScroll
2. Search for an item that only has one or zero suggestions
3. dropdown lists suggestions with increased height
4. Search for an item that has more than one suggestion.
5. dropdown lists suggestions with increased height
6. Search for an item that only has one or zero suggestions
7. **dropdown now has correct height**

### Expected behavior

![image](https://github.com/primefaces/primeng/assets/161650114/811e7d41-a70e-47e6-b011-4d7271190756)

PR #15971, opened 2024-07-06
fix(autocomplete): dropdown height with virtual scroll

leonidasdel:
Fixes [ #15647](https://github.com/primefaces/primeng/issues/15647)


rosenthalj:
If Issue #15647 only occurs for virtual scroll,  why not call  calculateAutoSize() only when virtualscroll=true?   Wouldn't that be more efficient?

leonidasdel:
could be the case but maybe the condition inside the function is enough `if (this._autoSize && !this.d_loading)`

Issue #11411, opened 2022-04-12
Tabmenu with routerLink items doesn't show offscreen item when reloaded


Jonnyprof:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primeng/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=35
```

**Plunkr Case (Bug Reports)**
https://stackblitz.com/edit/github-faf8r7-qpm6nh

**Current behavior**
If you have a tabmenu component set as `[scrollable]="true"` and go to a "hidden" item scrolling. When you refresh the page the selected item is not shown, you have to scroll and you don't know the current page.

**Expected behavior**
You should see the selected item when reload page.

**Minimal reproduction of the problem with instructions**
Set your browser as mobile device and go to https://github-faf8r7-qpm6nh.stackblitz.io/about. This selects correctly the about menuItem, but is hidden by the scrolling.

**Possible solution**
Call `updateScrollBar` on `ngAfterViewInit` method.
https://github.com/primefaces/primeng/blob/ab48ddae57f74f8ad2c2a31ad5a2864863d34595/src/app/components/tabmenu/tabmenu.ts#L173-L176

**What is the motivation / use case for changing the behavior?**
Tabmenu doesn't show current page and it's confusing.

**Please tell us about your environment:**
Ubuntu 20.04.4 LTS

* **Angular version:** 5.X
13.3.2

* **PrimeNG version:** 5.X
13.3.2

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
All browsers
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
All

PR #15982, opened 2024-07-08
Fixed #15973 - TabMenu doesn't scroll to active item when it's set pr…

### LINKED ISSUE #15973: TabMenu doesn't scroll to active item when it's set programatically
### Describe the bug

When TabMenu's activeItem is set from the code and there are more items then can be displayed on the screen (scrolling is needed and enabled), and the activeItem is not currently visible on the screen, then scrolling doesn't occur.

### Environment

Tested on both PrimeNG v16.9.12-lts and 17.18.3 and the behaviour is the same.

### Reproducer

https://stackblitz.com/edit/pvzkh6?file=src%2Fapp%2Fdemo%2Ftab-menu-scrollable-demo.ts

### Angular version

16.2.0

### PrimeNG version

16.9.12-lts

### Build / Runtime

TypeScript

### Language

TypeScript

18.20.3

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

mehmetcetin01140:
Fixes #15973

### DIFF
--- a/src/app/components/tabmenu/tabmenu.ts
+++ b/src/app/components/tabmenu/tabmenu.ts
@@ -12,10 +12,10 @@ import {
     Inject,
     Input,
     NgModule,
-    OnDestroy,
     Output,
     PLATFORM_ID,
     QueryList,
+    SimpleChanges,
     TemplateRef,
     ViewChild,
     ViewChildren,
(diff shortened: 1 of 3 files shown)


cetincakiroglu:
@mehmetcetin01140 

It reverts the fix of https://github.com/primefaces/primeng/issues/11411, it should not revert the fix while solving the issue

PR #16012, opened 2024-07-12
fix(overlay): min & max width configurable

### LINKED ISSUE #16011: overlay: there is no size restrictions for overlays
### Describe the bug

Hi there!

Currently when you have a component that utilizes the overlay (eg dropdown or multiselect) the overlay width can grow indefinitely. So far it even overflows the viewport.

Example with dropdown:
Y0XAXxA - Imgur

I thought about multiple ways of how to fix this behavior in a simple and unified way for all overlays, but feared that all approaches that don't require user configuration, are too opinionated and will not work universally. I will create a PR with a proposed solution soon.

In my proposed solution, the overlay options are extended with `calculateMaxWidth`. A type that accepts a variety of value formats:

      - `true` will set the maximum width to the control's width.
      - `false` will not set a maximum width.
      - A number will set the maximum width to a multiple of the control's width.
      - A number followed by 'px' will set the maximum width to a fixed size in pixels.
      - A number followed by 'vw' will set the maximum width to a percentage of the viewport width.
      Also accepts an array of values. The smallest value will be used.

An example of how it looks with the value 2:
kUIR627 - Imgur

As you can see the dropdown is now a bit more manageable. :)

I hope you agree with my solution and I will await PR feedback eagerly.

### Environment

Tested in the latest master version of primeNG in the PrimeNG Playground

### Reproducer

_No response_

### Angular version

18.0.2

### PrimeNG version

master

### Build / Runtime

Angular CLI App

### Language

TypeScript

20.11.1

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

LoaderB0T:
Fixes #16011 

I hope this PR is not beyond the "small fix" and can be merged. I am happy for feedback and can of course adjust the PR to the maintainers' requirements.

This PR does not introduce a breaking change.

All other details are in the issue description.
Thanks!

### DIFF
--- a/src/app/components/api/overlayoptions.ts
+++ b/src/app/components/api/overlayoptions.ts
@@ -7,6 +7,18 @@ import { AnimationEvent } from '@angular/animations';
 export type OverlayModeType = 'modal' | 'overlay' | undefined;
 
 /**
+ * Represents the type of maximum width option, which can be a number, a number followed by 'px', or a number followed by 'vw'.
+ * @group Types
+ */
+export type MaxWidthOption = number | `${number}px` | `${number}vw`;
+
+/**
+ * Represents the type of maximum width options, which can be a boolean, a number, a number followed by 'px', a number followed by 'vw', or an array of values.
+ * @group Types
+ */
+export type MaxWidthOptions = boolean | MaxWidthOption | MaxWidthOption[];
+
+/**
  * Represents the type of direction for a responsive overlay, which can be one of the specified values or undefined.
  * @group Types
  */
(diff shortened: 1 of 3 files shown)


src/app/components/api/overlayoptions.ts
```diff
@@ -6,6 +6,18 @@ import { AnimationEvent } from '@angular/animations';
  */
 export type OverlayModeType = 'modal' | 'overlay' | undefined;
 
+/**
+ * Represents the type of maximum width option, which can be a number, a number followed by 'px', or a number followed by 'vw'.
+ * @group Types
+ */
+export type MaxWidthOption = number | `${number}px` | `${number}vw`;
```
volvachev, line 13: Why not just use a string? And you can set any value you want with `%`, `rem`, etc.

LoaderB0T: Because the popover element is not a child of the dropdown element. Using % would not scale it according to the control, but (probably) also to the whole document size. To not create a false hope, I went for only allowing vw, which should behave in the same way as %, while not confusing the user. 

If you want relative styling to the control width, you have to provide a number.

cetincakiroglu:
Hi,

I understand the need for this at first glance. However, we already have `overlayOptions` property which allows you to set max-width without any additional property.

You can use content style property of overlayOptions.

` <p-dropdown  [overlayOptions]="overlayOptions" />`

```
// ts
    overlayOptions = {
        contentStyle: {'max-width': '150px' }
    };
```

Thanks a lot for the effort and support, I'm closing the PR because there is actually no need to implement this feature.

Looking forward to get more contributions from you :)

Issue #15885, opened 2024-06-24
InputNumberModule: Variant "filled" is not applied


cdillmann:
### Describe the bug

`<p-inputNumber
    [maxFractionDigits]="0"
    [min]="1"
    [max]="10"
    id="inputId"
    mode="decimal"
    variant="filled"
  />`

variant=filled is not applied.

### Environment

composer.json: 
"primeng": "^17.18.1",

### Angular version

17.1.3"

### PrimeNG version

17.18.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.14.0

### Browser(s)

all

### Steps to reproduce the behavior

1. See your documentation: https://primeng.org/inputnumber#filled , where the bug is visible already.

### Expected behavior

The component is in "filled" style.

PR #16013, opened 2024-07-12
fix: #15885, InputNumberModule: Variant filled is not applied

akshayaqburst:
### Defect Fixes
fix: #15885, InputNumberModule: Variant filled is not applied

### DIFF
--- a/src/app/components/inputnumber/inputnumber.ts
+++ b/src/app/components/inputnumber/inputnumber.ts
@@ -98,6 +98,7 @@ export const INPUTNUMBER_VALUE_ACCESSOR: any = {
                 [attr.data-pc-section]="'input'"
                 pAutoFocus
                 [autofocus]="autofocus"
+                [class.p-variant-filled]="true"
             />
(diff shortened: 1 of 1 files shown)


src/app/components/inputnumber/inputnumber.ts
```diff
@@ -98,6 +98,7 @@ export const INPUTNUMBER_VALUE_ACCESSOR: any = {
                 [attr.data-pc-section]="'input'"
                 pAutoFocus
                 [autofocus]="autofocus"
+                [class.p-variant-filled]="variant == 'filled'"
```
cetincakiroglu, line 101: In here, could you please also check the inputStyle signal in config? As an example, you can check autocomplete component: https://github.com/primefaces/primeng/blob/8446c5b3f896e40d99f7b9a2a626284c2c107580/src/app/components/autocomplete/autocomplete.ts#L816

cetincakiroglu (changes requested):
Change requested.

PR #16019, opened 2024-07-13
fixes #15937 Table: columnFilter

### LINKED ISSUE #15937: Table: columnFilter
### Describe the bug

when showButtons is false <div class="p-column-filter-buttonbar"> is being rendered

### Environment

17.16.0

### Reproducer

_No response_

### Angular version

17

### PrimeNG version

17.16.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

v20.12.1

### Browser(s)

All

create custom footer template then there is padding event if showButtons  is false

### Expected behavior

do not render <div class="p-column-filter-buttonbar"> when showButtons is false

ymg2006:
#15937
create a filter base on p-columnFilter and set show buttons to false:
```
<p-columnFilter [showButtons]="false">
     ......
     <ng-template pTemplate="footer">
     </ng-template>
</p-columnFilter>
```
the output will render the div and will have extra padding based on styles.

image

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -5208,7 +5208,7 @@ export class ReorderableRow implements AfterViewInit {
                             <ng-template *ngTemplateOutlet="addRuleIconTemplate"></ng-template>
                         </button>
                     </div>
-                    <div class="p-column-filter-buttonbar">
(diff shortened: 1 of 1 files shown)


ymg2006:
#16019

PR #16021, opened 2024-07-13
fixes #16020 Table: column-filter outside click does not close when it is inside a dialog component

### LINKED ISSUE #16020: Table: column-filter outside click does not close when it is inside a dialog component
### Describe the bug

put a table and a column filter inside dialog
click to open filter menu
it will not hide event if you close the modal

### Environment

Windows

### Reproducer

https://stackblitz.com/edit/github-978vwo

### Angular version

17.3.8

### PrimeNG version

17.8.4

### Build / Runtime

Angular CLI App

### Language

TypeScript

v20.12.1

### Browser(s)

All

put a table and a column filter inside dialog
click to open filter menu
it will not hide event if you close the modal

### Expected behavior

close the modal on outside click

ymg2006:
#16020
Table: column-filter outside click does not close when it is inside a dialog component

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -5133,7 +5133,7 @@ export class ReorderableRow implements AfterViewInit {
                 [ngClass]="{ 'p-column-filter-overlay p-component p-fluid': true, 'p-column-filter-overlay-menu': display === 'menu' }"
                 [id]="overlayId"
                 [attr.aria-modal]="true"
-                role="filterDialog"
+                role="dialog"
                 (click)="onContentClick()"
                 [@overlayAnimation]="'visible'"
                 (@overlayAnimation.start)="onOverlayAnimationStart($event)"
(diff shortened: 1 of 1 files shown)


src/app/components/table/table.ts
```diff
@@ -5133,7 +5133,7 @@ export class ReorderableRow implements AfterViewInit {
                 [ngClass]="{ 'p-column-filter-overlay p-component p-fluid': true, 'p-column-filter-overlay-menu': display === 'menu' }"
                 [id]="overlayId"
                 [attr.aria-modal]="true"
-                role="dialog"
+                role="filterDialog"
```
cetincakiroglu, line 5136: filterDialog role does not exist as an HTML Role. It breaks the accessibility of dialog.

src/app/components/table/table.ts
```diff
@@ -5788,7 +5788,7 @@ export class ColumnFilter implements AfterContentInit {
             const documentTarget: any = this.el ? this.el.nativeElement.ownerDocument : 'document';
 
             this.documentClickListener = this.renderer.listen(documentTarget, 'mousedown', (event) => {
-                const dialogElements = document.querySelectorAll('[role="dialog"]');
+                const dialogElements = document.querySelectorAll('[role="filterDialog"]');
```
cetincakiroglu, line 5791: You can try a different approach instead of adding and checking a random role. I'm not sure but, checking the elements parent would work.

cetincakiroglu (changes requested):
Changes requested.

src/app/components/table/table.ts
ymg2006: Yes, it is more appropriate to do so, I will do as described here.

ymg2006:
@cetincakiroglu the issue is fixed with a search in dialogs where they have date-picker class.
the reason is the commit #15094. So, this will only find date picker dialog elements not all of them which causes all the dialogs not close when there is another dialog open in whole document.
the latest commit fixes the issue without breaking previous commit.

cetincakiroglu:
Hi @ymg2006 

PR #16041, opened 2024-07-17
fix(MultiSelect): Check for NaN values in isSelectionAllDisabled method

### LINKED ISSUE #16039: MultiSelect - showToggleAll not working as expected
### Describe the bug

When using MultiSelect, we have a use case where we need to pass the `selectionLimit` prop conditionally. Like this:

```
<p-multiSelect 
    [options]="cities" 
    [(ngModel)]="selectedCities" 
    optionLabel="name"
    [showToggleAll]="true"
    [selectionLimit]="someCondition ? '3' : null" 
    placeholder="Select Cities" />
```

[This is the most recommended way](https://stackoverflow.com/questions/36745734/how-to-add-conditional-attribute-in-angular-2/36745752#36745752) to pass props conditionally.

So, in the MultiSelect component, you [have a method](https://github.com/primefaces/primeng/blob/e142de683b9cbcf2342774e9390ee09b0934580f/src/app/components/multiselect/multiselect.ts#L1473-L1475) to show/hide the Toggle All checkbox. The problem comes [in this line](https://github.com/primefaces/primeng/blob/e142de683b9cbcf2342774e9390ee09b0934580f/src/app/components/multiselect/multiselect.ts#L574) where you are transforming the **selectionLimit** variable. When passing null or undefined, this transform is transforming the variable to `NaN`, so later your `isSelectionAllDisabled` method is not working as expected because `ObjectUtils.isEmpty(NaN)` is false.

At the end, when passing null or undefined to the `selectionLimit` prop, we end up not showing the Toggle All checkbox at all.

Do you have a workaround? Or other recommended way?

Thanks a lot

Last note:

Angular people told me that [this is the expected behaviour for the numberAttribute transform](https://github.com/angular/angular/issues/57017#issuecomment-2233107503)

### Reproducer

https://stackblitz.com/edit/pi4qot?file=src%2Fapp%2Fmulti-select-basic-demo.html

### Angular version

18.1.0

### PrimeNG version

17.18.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

v22.4.1

### Browser(s)

_No response_

_No response_

### Expected behavior

I should see the Toggle All checkbox when passing null or undefined to the `selectionLimit` prop

dfernandez-elastic:
Fixes #16039

Added a check for NaN values.

Note:

I couldn't add new tests because my unit tests were not working locally.

When running `npm test`, I've got a lot of errors with this message: `export 'async' (imported as 'async') was not found in '@angular/core/testing'`

### DIFF
--- a/src/app/components/multiselect/multiselect.ts
+++ b/src/app/components/multiselect/multiselect.ts
@@ -571,7 +571,7 @@ export class MultiSelect implements OnInit, AfterViewInit, AfterContentInit, Aft
      * Decides how many selected item labels to show at most.
      * @group Props
      */
-    @Input({ transform: numberAttribute }) selectionLimit: number | undefined;
+    @Input({ transform: (value: unknown) => numberAttribute(value, null) }) selectionLimit: number | null | undefined;
     /**
      * Label to display after exceeding max selected labels e.g. ({0} items selected), defaults "ellipsis" keyword to indicate a text-overflow.
      * @group Props
@@ -1471,10 +1471,6 @@ export class MultiSelect implements OnInit, AfterViewInit, AfterContentInit, Aft
     }
 
     isSelectionAllDisabled() {
-        if (this.showToggleAll && isNaN(this.selectionLimit)) {
-            return true;
-        }
-
         return this.showToggleAll && ObjectUtils.isEmpty(this.selectionLimit);
     }
 
--- a/src/app/components/multiselect/multiselect.ts
+++ b/src/app/components/multiselect/multiselect.ts
@@ -1471,6 +1471,10 @@ export class MultiSelect implements OnInit, AfterViewInit, AfterContentInit, Aft
     }
 
     isSelectionAllDisabled() {
+        if (this.showToggleAll && isNaN(this.selectionLimit)) {
+            return true;
+        }
+
         return this.showToggleAll && ObjectUtils.isEmpty(this.selectionLimit);
     }
 


rosenthalj:
Another way to fix this issue would be to add the following to multiselect.'s ngOnInit method: 
```
        if(isNaN(this.selectionLimit)) {
            this.selectionLimit = undefined
        }
```

Note:    `ObjectUtils.isNotEmpty(this.selectionLimit)` is call in multiple places throughout  multiselect.ts:

dfernandez-elastic:
Hi @rosenthalj I wanted to fix only the Toggle All behavior. I can implement it in the `ngOnInit` method too. Let's wait for a maintainer response.

Thanks for your suggestion!

src/app/components/multiselect/multiselect.ts
```diff
@@ -1471,6 +1471,10 @@ export class MultiSelect implements OnInit, AfterViewInit, AfterContentInit, Aft
     }
 
     isSelectionAllDisabled() {
+        if (this.showToggleAll && isNaN(this.selectionLimit)) {
```
cetincakiroglu, line 1474: Hi @dfernandez-elastic,

The best solution would be to update the input transformer instead of adding NaN checks to methods.

You can directly implement this solution to your PR.
`@Input({ transform: (value: unknown) => numberAttribute(value, null) }) selectionLimit: number | null | undefined;`

cetincakiroglu (changes requested):
Could you please try the mentioned solution?

src/app/components/multiselect/multiselect.ts
dfernandez-elastic: Hi @cetincakiroglu it worked! I tested manually for a possible regression, too. Everything works as expected. Thanks for the tip!

Issue #16071, opened 2024-07-21
Showcase: does not persist theme config on reload or navigation


jjs98:
### Describe the bug

When I configure darkmode or a special theme on primeng.org and then
- reload the page
- search for a component
- click on a reference link to another component
the theme is resetted.

### Environment

-

### Angular version

-

### PrimeNG version

17.18.5

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

-

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation. If there is no improvement on this, can you open a new issue so we can include it in our roadmap?

Thanks a lot for your understanding!
Best Regards,

PR #16072, opened 2024-07-21
add: theme persist via localstorage

jjs98:
Possible fix for issue #16071.

I use the localStorage to persist the theme config.

I had lags when switching the theme on large components like calendar. I think this is a local issue because of SSR. I had this issue prior to my changes. (Just wanted to mention it ^^)


cetincakiroglu:
Hi,

Thanks a lot for the effort and support, right now we're updating the theme engine for the v18. We'll provide this functionality with v18. Therefore, I'm closing the PR.

Issue #13512, opened 2023-08-18
Menubar: autoDisplay and autoHide broken


g1thuser:
### Describe the bug

The autoDisplay and autoHide properties don't work anymore. autoHide doesn't work at all, autoDisplay works only once after clicking on any root menu item. Clicking on a menu item leaves the last clicked root menu item in the selected status.

The error is already present in primeng 16.1.0, but didn't exist in primeng 16.0.2

### Environment

Ubuntu 22.04
Firefox 116.03

### Angular version

16.2.0

### PrimeNG version

16.2.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.5.1

### Steps to reproduce the behavior

see [here](https://stackblitz.com/edit/hemvbj-ldr7mf?file=src%2Fapp%2Fdemo%2Fmenubar-basic-demo.html,package.json) for primeng 16.1.0

see [here](https://stackblitz.com/edit/hemvbj-ldbfpp?file=src%2Fapp%2Fdemo%2Fmenubar-basic-demo.html,package.json) for primeng 16.2.0

### Expected behavior

see [here](https://stackblitz.com/edit/hemvbj?file=src%2Fapp%2Fdemo%2Fmenubar-basic-demo.html,package.json) for primeng 16.0.2

SoyDiego:
Hi @g1thuser, they know about the problem and will fix it.
Here you have  a reply of one of the members of PrimeNG Team:

![image](https://github.com/primefaces/primeng/assets/19764334/d800847d-6116-4cdb-8314-81b8b99b2a41)

The link is this: https://github.com/primefaces/primeng/issues/13455#issuecomment-1671416418

AndreasHae:
Still causing issues on primeng 17.2.0.

**Steps to reproduce:**
1. Add a `p-menubar` with `[autoHide]="true" [autoHideDelay]="0"` and submenus
2. Hover over a root menu item -> submenu opens
3. Hover away from menubar -> submenu should close immediately, but stays open

chimayko:
Still happening on 17.5

> Still causing issues on primeng 17.2.0.
> 
> **Steps to reproduce:**
> 
> 1. Add a `p-menubar` with `[autoHide]="true" [autoHideDelay]="0"` and submenus
> 2. Hover over a root menu item -> submenu opens (works as intended because `autoDisplay` default is true)
> 3. Hover away from menubar -> submenu should close immediately, but stays open

[...]

PR #16077, opened 2024-07-22
fix(tieredmenu): show submenu on mouse hover

### LINKED ISSUE #13861: TieredMenu: Submenus in popup do not appear on hover.
### Describe the bug

In past version of PrimeNG, submenus in popups were shown immediately when hovering over an item when using `autoDisplay`. This bug was introduced in PrimeNG 16.1.0. #13582 reported this bug, and a fix was made in #13593, but it does not appear to have fixed the problem.

This is due to to checking `this.dirty` in `onItemMouseEnter()`. When the popup is initially shown, `this.dirty` is `false`, so `onItemChange(event)` is not called until an item is clicked.

A potential solution for this bug is to update the `show()` function to set `this.dirty` to `true` if `popup` & `autoDisplay` are both `true`.

### Environment

N/A

### Reproducer

See below

### Angular version

16.2.0

### PrimeNG version

16.5.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

16.16.0

### Browser(s)

All

Use PrimeNG's showcase for the TieredMenu: https://primeng.org/tieredmenu

1. In the **Popup** section, click the **Show** button.
2. Hover over the items in the root menu, like "File" & "Edit" and note that their submenus do not show.

### Expected behavior

Submenus should immediately appear when hovering over an item when `popup` and `autoDisplay` are set to `true`.

bjsawyer:
Fixes #13861

Implemented @kubedan's suggestion [here](https://github.com/primefaces/primeng/issues/13512#issuecomment-2044409597)

### DIFF
--- a/src/app/components/tieredmenu/tieredmenu.ts
+++ b/src/app/components/tieredmenu/tieredmenu.ts
@@ -702,6 +702,9 @@ export class TieredMenu implements OnInit, AfterContentInit, OnDestroy {
 
     onItemMouseEnter(event: any) {
         if (!DomHandler.isTouchDevice()) {
+            if (this.autoDisplay) {
+                this.dirty = true; 
+            }
             if (this.dirty) {
                 this.onItemChange(event);
             }


bjsawyer:
Hello @cetincakiroglu able to take a look at this when you have a moment?

bjsawyer:
@mehmetcetin01140 able to review? My team just upgraded to v17 and this is a blocker for us, would love to get this fix in (or will make updates if needed)

cetincakiroglu:
Hi @bjsawyer,

Thanks a lot for the effort and support! We appreciate and hope to receive more PRs from you.

PR #16085, opened 2024-07-23
fix(keyfilter): regex test with the correct value onKeyPress

### LINKED ISSUE #16084: KeyFilter: Can't edit value at some cursor positions
### Describe the bug

When using strict regex on a `keyfilter`, editing a value at any cursor position except the end will not test regex correctly. This will prevent changes to the input.

### Environment

Windows 11

### Reproducer

https://github-1l73hr.stackblitz.io

### Angular version

17.3.12

### PrimeNG version

17.18.5

### Build / Runtime

Angular CLI App

### Language

TypeScript

20.12.1

### Browser(s)

_No response_

https://github-1l73hr.stackblitz.io

To Reproduce:
1. Click into the input on the value anywhere besides the end.
2. Try to type any number.
3. Input will be block.

To allow input at any cursor position:
1. Delete the cents side of the value to be shorter than `.00`.
2. Now you can change value anywhere the cursor is.

### Expected behavior

I should be able to type `1000.00` and then move my cursor position to one between `1` and the first `0` and type `2` to change the value to `12000.00`.

BGBRWR:
fixes #16084

### DIFF
--- a/src/app/components/keyfilter/keyfilter.ts
+++ b/src/app/components/keyfilter/keyfilter.ts
@@ -113,7 +113,11 @@ export class KeyFilter implements Validator {
 
     lastValue: any;
 
-    constructor(@Inject(DOCUMENT) private document: Document, @Inject(PLATFORM_ID) private platformId: any, public el: ElementRef) {
+    constructor(
+        @Inject(DOCUMENT) private document: Document,
+        @Inject(PLATFORM_ID) private platformId: any,
+        public el: ElementRef
+    ) {
         if (isPlatformBrowser(this.platformId)) {
             this.isAndroid = DomHandler.isAndroid();
         } else {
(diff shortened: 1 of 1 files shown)


src/app/components/keyfilter/keyfilter.ts
```diff
@@ -227,8 +231,9 @@ export class KeyFilter implements Validator {
         }
 
         let valueCheck = this.el.nativeElement.value || '';
-
-        let val = valueCheck + cc;
+        const selectionStart = (<HTMLInputElement>e.currentTarget).selectionStart || 0;
+        const selectionEnd = (<HTMLInputElement>e.currentTarget).selectionEnd || 0;
+        let val = valueCheck.substring(0, selectionStart) + cc + valueCheck.substring(selectionEnd);
```
rosenthalj, line 236: I have not verified this but should lines 234 and 235 
` || 0`
be  replaced with
`|| Number.MAX_SAFE_INTEGER`

PR #16155, opened 2024-08-01
fix(autocomplete): not opening on second click

### LINKED ISSUE #15281: P-autocomplete after second click doesn't show dropdown
### Describe the bug

Hi, everyone! In angular 16 in primeng ^16.0.0-rc.2 autocomplete worked but with last version angular and primeng, I got critical bug. When I opened dropdown first time api loaded, list showed, but in second time, dropdown doesn't show list and api loaded. List only showed, when you must be click to dropdown so very fast and catch cancelled api request. Can you help me with current bug, please. It's my code https://stackblitz.com/edit/ja1m96?file=package.json.
Current problem you can solve using Subject or subscribe, but create subscribes is a bad solution for current problem.
About subjects you must write more code.

### Environment

https://stackblitz.com/edit/ja1m96?file=package.json

### Reproducer

_No response_

### Angular version

17.3.4

### PrimeNG version

17.13.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

21.1.0

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

LoaderB0T:
Fixes #15281

This PR fixes the mentioned issue & additionally a bug in the overlay which would close the panel & never open it again (when opening & closing before the animation finishes) as the internal state of the overlay was half opened / half closed.

Details:
`modalVisible` was `false` while `visible` was true. This basically "deadlocks" the overlay.

Also fixes the minlength behavior for when "0" is used as a value.

### DIFF
--- a/src/app/components/autocomplete/autocomplete.ts
+++ b/src/app/components/autocomplete/autocomplete.ts
@@ -880,7 +880,14 @@ export class AutoComplete implements AfterViewChecked, AfterContentInit, OnDestr
         return typeof this.modelValue() === 'string' && this.optionValue;
     }
 
-    constructor(@Inject(DOCUMENT) private document: Document, public el: ElementRef, public renderer: Renderer2, public cd: ChangeDetectorRef, public config: PrimeNGConfig, public overlayService: OverlayService, private zone: NgZone) {
+    constructor(
+        public el: ElementRef,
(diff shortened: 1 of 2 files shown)


LoaderB0T:
Please let me know if you require further details!

PR #16157, opened 2024-08-01
Adds new feature - Avatar support image and label together

EnricoMessall:
Feature Requests
Support image and label together for avatar.

Matching discussion with more informations:
https://github.com/orgs/primefaces/discussions/2435

### DIFF
--- a/src/app/components/avatar/avatar.ts
+++ b/src/app/components/avatar/avatar.ts
@@ -58,11 +58,6 @@ export class Avatar {
      */
     @Input() styleClass: string | undefined;
     /**
-     * Class of the element.
-     * @group Props
-     */
-    @Input() labelClass: string | undefined;
-    /**
      * Establishes a string value that labels the component.
      * @group Props
      */
--- a/src/app/components/avatar/avatar.ts
+++ b/src/app/components/avatar/avatar.ts
@@ -18,7 +18,7 @@ import { ChangeDetectionStrategy, Component, EventEmitter, Input, NgModule, Outp
     encapsulation: ViewEncapsulation.None,
     styleUrls: ['./avatar.css'],
     host: {
-        class: 'flex p-element'
+        class: 'p-avatar-host p-element'
     }
 })
 export class Avatar {
--- a/src/app/components/avatar/avatar.ts
+++ b/src/app/components/avatar/avatar.ts
@@ -9,7 +9,7 @@ import { ChangeDetectionStrategy, Component, EventEmitter, Input, NgModule, Outp
     template: `
         <div [ngClass]="containerClass()" [class]="styleClass" [ngStyle]="style" [attr.aria-labelledby]="ariaLabelledBy" [attr.aria-label]="ariaLabel" [attr.data-pc-name]="'avatar'">
             <ng-content></ng-content>
-            <span class="p-avatar-text" [style.position]="image ? 'absolute' : ''" *ngIf="label; else iconTemplate">{{ label }}</span>
+            <span class="p-avatar-text" [ngStyle]="labelStyle" *ngIf="label; else iconTemplate">{{ label }}</span>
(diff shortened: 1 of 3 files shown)


src/app/components/avatar/avatar.ts
```diff
@@ -9,16 +9,16 @@ import { ChangeDetectionStrategy, Component, EventEmitter, Input, NgModule, Outp
     template: `
         <div [ngClass]="containerClass()" [class]="styleClass" [ngStyle]="style" [attr.aria-labelledby]="ariaLabelledBy" [attr.aria-label]="ariaLabel" [attr.data-pc-name]="'avatar'">
             <ng-content></ng-content>
-            <span class="p-avatar-text" *ngIf="label; else iconTemplate">{{ label }}</span>
-            <ng-template #iconTemplate><span [class]="icon" [ngClass]="'p-avatar-icon'" *ngIf="icon; else imageTemplate"></span></ng-template>
-            <ng-template #imageTemplate><img [src]="image" *ngIf="image" (error)="imageError($event)" [attr.aria-label]="ariaLabel" /></ng-template>
+            <span class="p-avatar-text" [style.position]="image ? 'absolute' : ''" *ngIf="label; else iconTemplate">{{ label }}</span>
```
SoyDiego, line 12: If i'm not wrong this will break the Content Security Policy. 
All the components should be use `ngStyle`. Check the PR of the @mehmetcetin01140  and @cetincakiroglu

https://github.com/primefaces/primeng/pull/16005

src/app/components/avatar/avatar.ts
```diff
@@ -18,7 +18,7 @@ import { ChangeDetectionStrategy, Component, EventEmitter, Input, NgModule, Outp
     encapsulation: ViewEncapsulation.None,
     styleUrls: ['./avatar.css'],
     host: {
-        class: 'p-element'
+        class: 'flex p-element'
```
SoyDiego, line 21: I don't know if this is ok because this class will not be applied if the project is not using primeflex.
Maybe it's working in PrimeNG web because they are using it, but in my project and if i'm not wrong,  this will add `flex` and will not do nothing. Check it, please

src/app/components/avatar/avatar.ts
EnricoMessall: That commit looks like its about XSS attacks (strict CSP) because it then might be possible to inject variables into style that shouldn't be there, since all of these changes had variables. The style here is only defining hard style values so it should be fine.

SoyDiego: I cannot replicate in my project where we have CSP  applied in this moment and the previous PR that I sent was reported by myself but I 90% sure your code will break the CSP.

I asked to chatGPT too and said the same and the suggestion was:

```
[ngStyle]="{'position': image ? 'absolute' : ''}"
```

but again sorry because I cannot check it properly

src/app/components/avatar/avatar.ts
EnricoMessall: Absolute right here, i've implemented a own p-avatar-host class for it

src/app/components/avatar/avatar.ts
EnricoMessall: I see, its more about adjusting style inline the html. I've adjusted it to match the other components.

SoyDiego (approved):

src/app/components/avatar/avatar.ts
SoyDiego: I think looks good now.
I hope this doesn't break anything and that the problem with CSP was terrible, very difficult to find and know what the correct solution was.

cetincakiroglu:
Hi @EnricoMessall,

Thanks for the effort and support! Could you please open an issue and tag me so I can add it to the v18.0.0-rc.1 milestone

PR #16167, opened 2024-08-04
Fixed primefaces#16074 - ButtonDirective changing severity does not update styles

EnricoMessall:
Fixed primefaces#16074 - ButtonDirective changing severity does not update styles

The issue was that the severity still has been there for all severities added to the button in the past. Even if the severity changed it has p-button-primary and p-button-secondary for example. The fix removes the severity class before appending the new severity class again:

https://github.com/user-attachments/assets/65fc29c3-ba63-47f3-9e21-3074f51f5845

### DIFF
--- a/src/app/components/api/constants.ts
+++ b/src/app/components/api/constants.ts
@@ -1,0 +1,1 @@
+export type ButtonSeverity = 'success' | 'info' | 'warn' | 'danger' | 'help' | 'primary' | 'secondary' | 'contrast' | null | undefined
--- a/src/app/components/api/public_api.ts
+++ b/src/app/components/api/public_api.ts
@@ -31,3 +31,4 @@ export * from './lazyloadmeta';
 export * from './tooltipoptions';
 export * from './scrolleroptions';
 export * from './treetablenode';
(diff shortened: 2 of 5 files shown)


cetincakiroglu (changes requested):
For button & button components `warning` => `warn`
```
 @Input() severity: 'success' | 'info' | 'warn' | 'danger' | 'help' | 'primary' | 'secondary' | 'contrast' | null | undefined; 
 ```
 
 For message, messages, toast, badge, avatar. `danger` => `error` and `warning` => `warn`
```
 @Input() severity: 'success' | 'info' | 'warn' | 'error' | 'help' | 'primary' | 'secondary' | 'contrast' | null | undefined; 
```

EnricoMessall:
@cetincakiroglu 

See my last commit, i introduced a global type for ButtonSeverity. For other severities i think its better to do that in another ticket since its out of scope of this issue :)

mertsincan:
Thanks a lot for your contribution! Could you please try it with v18? If the problem persists, please create a new PR for v18.

For now, closed.

PR #16174, opened 2024-08-06
Add a boolean which indicates if the drop is in a drop point or a node

### LINKED ISSUE #16170: [p-tree] Distinguish between a drop on the nodes and a drop between the nodes
### Describe the bug

It is impossible to distinguish a drop on a drop point between the nodes from a drop on a node. A specific drop restriction behavior can be defined by rewriting the allowDrop method, but this missing information is problematic.

### Environment

Windows 11, Google Chrome

### Reproducer

_No response_

### Angular version

17.1.0

### PrimeNG version

17.18.7

### Build / Runtime

TypeScript

### Language

TypeScript

20.13.1

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

Eiael:
Fixed #16170 : We can now implement different behavior based on drop element type

### DIFF
--- a/src/app/components/tree/tree.interface.ts
+++ b/src/app/components/tree/tree.interface.ts
@@ -1,6 +1,5 @@
 import { TemplateRef } from '@angular/core';
 import { TreeNode } from 'primeng/api';
-import { Tree } from './tree';
 import { ScrollerOptions } from 'primeng/api';
 
 /**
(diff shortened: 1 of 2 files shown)


src/app/components/tree/tree.interface.ts
```diff
@@ -64,6 +64,10 @@ export interface TreeNodeDropEvent {
      * Dropped node instance.
      */
     dropNode?: TreeNode<any> | null;
+    /**
+     * Indicates if the node is dropped in a drop point between nodes
+     */
+    dropPoint?: boolean;
```
blaze0004, line 70: @Eiael How about using a literate string type, example: type DropPoint = 'node' | 'between'

dropPoint: DropPoint

Eiael: Sounds better, fixed

mertsincan:

Issue #16180, opened 2024-08-07
ColumnFilter: Template names are wrong.


wS2Zth:
### Describe the bug

The ColumnFilter template names on the Table API page and the implementation do not match.

In Table API Page;
[https://primeng.org/table#api.columnfilter.templates](url)

filterTemplate
headerTemplate
footerTemplate
filterIconTemplate
removeRuleIconTemplate
addRuleIconTemplate
clearFilterIconTemplate

In code;

case 'header':
case 'filter':
case 'footer':
case 'filtericon':
case 'clearfiltericon':
case 'removeruleicon':
case 'addruleicon':
default:

*When I pasted it as code, I couldn't get a proper view, so I wrote the names of the cases.

### Environment

no need

### Angular version

no need

### PrimeNG version

no need

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

no need

PR #16181, opened 2024-08-07
#16180 is solved

wS2Zth:
#16180 is solved


cetincakiroglu (commented):
Hi,

Thanks a lot for the effort and support! This issue should be fixed from `columnfilter.interface.ts` file to maintain changes in every `npm run apidoc` execution. Command regenerates `index.json` file.

PR #16212, opened 2024-08-11
closes #16211 Table: breaks after one call to updateStyleElement()

### LINKED ISSUE #16211: Table: breaks after one call to updateStyleElement()
### Describe the bug

@cetincakiroglu @mehmetcetin01140 
Commit #16154  completely breaks table after resizeTableCells or onColumnDrop.

@aaronshim maybe reading [innerhtml-vs-innertext-vs-textcontent](https://www.freecodecamp.org/news/innerhtml-vs-innertext-vs-textcontent/) helps to figure out why the whole table is broken after 17.18.7 to 17.18.8 commit.

### Environment

Windows 11

### Reproducer

_No response_

### Angular version

17.3.6

### PrimeNG version

17.18.8

### Build / Runtime

Angular CLI App

### Language

TypeScript

20.12.1

### Browser(s)

All

Do anything in the table that makes a call to resizeTableCellsm then the whole css and ids and etc are broken.

### Expected behavior

_No response_

ymg2006:
closes #16211 Table: breaks after one call to updateStyleElement()
Added security bypass and html sanitizer for user inputs without breaking components!
This will stop injections to happen with angular built-in DomSanitizer.

### DIFF
--- a/src/app/components/calendar/calendar.ts
+++ b/src/app/components/calendar/calendar.ts
@@ -19,6 +19,7 @@ import {
     Output,
     QueryList,
     Renderer2,
+    SecurityContext,
     TemplateRef,
     ViewChild,
     ViewEncapsulation
@@ -39,6 +40,7 @@ import { CalendarIcon } from 'primeng/icons/calendar';
 import { Nullable, VoidListener } from 'primeng/ts-helpers';
 import { NavigationState, CalendarResponsiveOptions, CalendarTypeView, LocaleSettings, Month, CalendarMonthChangeEvent, CalendarYearChangeEvent } from './calendar.interface';
 import { AutoFocusModule } from 'primeng/autofocus';
+import { DomSanitizer } from '@angular/platform-browser';
 
 export const CALENDAR_VALUE_ACCESSOR: any = {
     provide: NG_VALUE_ACCESSOR,
@@ -1169,7 +1171,7 @@ export class Calendar implements OnInit, OnDestroy, ControlValueAccessor {
         return this.currentView === 'year' ? this.getTranslation('nextDecade') : this.currentView === 'month' ? this.getTranslation('nextYear') : this.getTranslation('nextMonth');
     }
 
-    constructor(@Inject(DOCUMENT) private document: Document, public el: ElementRef, public renderer: Renderer2, public cd: ChangeDetectorRef, private zone: NgZone, private config: PrimeNGConfig, public overlayService: OverlayService) {
+    constructor(@Inject(DOCUMENT) private document: Document, public el: ElementRef, public renderer: Renderer2, public cd: ChangeDetectorRef, private zone: NgZone, private config: PrimeNGConfig, public overlayService: OverlayService, private readonly domSanitizer: DomSanitizer) {
         this.window = this.document.defaultView as Window;
     }
 
@@ -3573,8 +3575,7 @@ export class Calendar implements OnInit, OnDestroy, ControlValueAccessor {
                     `;
                 }
             }
-
-            (<HTMLStyleElement>this.responsiveStyleElement).innerHTML = innerHTML;
+            this.responsiveStyleElement.innerHTML = this.domSanitizer.bypassSecurityTrustStyle(innerHTML) as string;
         }
     }
 
--- a/src/app/components/calendar/calendar.ts
+++ b/src/app/components/calendar/calendar.ts
@@ -1047,7 +1047,7 @@ export class Calendar implements OnInit, OnDestroy, ControlValueAccessor {
 
     overlay: Nullable<HTMLDivElement>;
 
-    responsiveStyleElement: HTMLStyleElement | undefined | null;
+    responsiveStyleElement: any;
 
     overlayVisible: Nullable<boolean>;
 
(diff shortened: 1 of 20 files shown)


src/app/components/calendar/calendar.ts
```diff
@@ -3573,8 +3575,7 @@ export class Calendar implements OnInit, OnDestroy, ControlValueAccessor {
                     `;
                 }
             }
-
-            (<HTMLStyleElement>this.responsiveStyleElement).innerHTML = innerHTML;
+            this.responsiveStyleElement.innerHTML = this.domSanitizer.bypassSecurityTrustStyle(innerHTML) as string;
```
aaronshim, line 3578: If the original PR with textContent was not working **because the assignment value contained markup that wasn’t just the CSS code**, how can we be sure that we can trust this assignment value as a style?

ymg2006: bypassSecurityTrustStyle will Bypass security and trust the given value to be safe style value (CSS).
abstract bypassSecurityTrustStyle(value: string): SafeStyle
since the type is assigned to responsiveStyleElement we cannot set innerHTML as SafeStyle.
so the suggestion here is to make the responsiveStyleElement as any and then we do not need the as string cast.
I already merged this suggestion into this pull request.

On the other hand, **these styles are created internally by primeng**, so they are trusted and there is no need for sanitization.
Moreover, **all the inputs from user are SANITIZED** then bypassed by SafeHtml pipe.
```
   constructor(@Inject(PLATFORM_ID) private readonly platformId: any, private readonly domSanitizer: DomSanitizer) { }

    public transform(value: string): SafeHtml {
        if (!value || !isPlatformBrowser(this.platformId)) {
            return value;
        }

+         const sanitizedValue = this.domSanitizer.sanitize(SecurityContext.HTML ,value);
+         return this.domSanitizer.bypassSecurityTrustHtml(sanitizedValue);

    }
```

rosenthalj:
I took a quick cursory look at the proposed changes  and a few comments:

1. A lot of file code changes.   Has the all the changes been tested? 
2. Maybe I'm missing something,  there are no code changes to table.ts?

Note:   I would recommend modifying the text in issue #16210 and remove the following (since you took a different approach in the PR):
 
```
 To sanitize html my suggestion is to use [DOMPurify](https://github.com/cure53/DOMPurify) package and replace all methods that uses inner html with:
import DOMPurify from 'dompurify';
el.innerHTML = DOMPurify.sanitize(innerHtml, {RETURN_TRUSTED_TYPE: true});
```

It's good that this PR doesn't add another dependency to PrimeNG.

ymg2006:
> I took a quick cursory look at the proposed changes and a few comments:
> 
> 1. A lot of file code changes.   Has the all the changes been tested?
> 2. Maybe I'm missing something,  there are no code changes to table.ts?
> 
> Note: I would recommend modifying the text in issue #16210 and remove the following (since you took a different approach in the PR):
> 
> ```
>  To sanitize html my suggestion is to use [DOMPurify](https://github.com/cure53/DOMPurify) package and replace all methods that uses inner html with:
> import DOMPurify from 'dompurify';
> el.innerHTML = DOMPurify.sanitize(innerHtml, {RETURN_TRUSTED_TYPE: true});
> ```
> 
> It's good that this PR doesn't add new OSS dependency to PrimeNG.



> I took a quick cursory look at the proposed changes and a few comments:
> 
> 1. A lot of file code changes.   Has the all the changes been tested?
> 2. Maybe I'm missing something,  there are no code changes to table.ts?
> 
> Note: I would recommend modifying the text in issue #16210 and remove the following (since you took a different approach in the PR):
> 
> ```
>  To sanitize html my suggestion is to use [DOMPurify](https://github.com/cure53/DOMPurify) package and replace all methods that uses inner html with:
> import DOMPurify from 'dompurify';
> el.innerHTML = DOMPurify.sanitize(innerHtml, {RETURN_TRUSTED_TYPE: true});
> ```
> 
> It's good that this PR doesn't add new OSS dependency to PrimeNG.

@rosenthalj, to response your first question, I have not tested all changes, I will really appreciate it if @aaronshim do this on my behalf.

For the second question, I have merged the needed changes, thanks for the review.

Finally, I read angular documentation and saw their standard approach for this so implemented as suggested.

ymg2006:
@rosenthalj issue #16210 has nothing to do with this PR.

cetincakiroglu:
Hi,

Thanks for all the effort guys, you're awesome! since 20 files have been changed, it'll take time to review this PR. During this time, we'll fix the issue for the table beforehand.

ymg2006:
> Also, could you please remove table fix (we've just fixed that), and create a separate issue for the rest of the components so that we can include it in next week's milestone?

@cetincakiroglu I really appreciate if you bear with me up to 29th of August that I test all the changes exhaustively and then make the aforementioned corrections.
Thanks for fixing the table issue.

ymg2006:
@cetincakiroglu merged with changes from table.ts

cetincakiroglu:
Hi @ymg2006,

Thanks a lot for the effort and support! Could you please resolve the conflicts and test again, files are updated and it might resolved already, if not, please tag me after the conflict fix, so I can merge it safely.

ymg2006:
@cetincakiroglu merge is done and build issue is solved too.

mertsincan:
Hi,

Could you please update your PR? Then, I'll review asap. Thank you!

ymg2006:
@mertsincan merge is done.

mertsincan:
Thanks a lot for your contribution! Unfortunately, this PR has conflicts, so I cannot merge it. Could you please create a new PR and open a corresponding issue for this PR? Then, I'll merge it asap.

PR #16232, opened 2024-08-14
Fix #14529 - Google Gboard and inputmask control

### LINKED ISSUE #14529: Google Gboard and inputmask control
### Describe the bug

On a pixel device; the device uses GBoard and issue occurs if you type like 01/01/1980 and then delete 01/01/yyyy, the cursor moves before the forward slash and when you types it messes up the year typed 01/01/9801. This happens with any mask that has seperators in it.

### Environment

Any mobile device with Gboard keyboard.

### Reproducer

_No response_

### Angular version

15.2.4

### PrimeNG version

17.3.2

### Build / Runtime

Angular CLI App

### Language

TypeScript

16.13.1

### Browser(s)

_No response_

1. Use a Pixel phone or install Gboard on Samsung phone.
2. Go to a site like https://primeng.org/inputmask.
3. Type into one of the masked fields.
4. Delete characters to the masked separator.
5. The cursor will move in front of the separator.
6. Type and now the characters will be out of position.

### Expected behavior

Cursor should move after the masked separator, this will allow the correct characters to be typed.

Ventura2:
Fixes #14529 

It also makes the gboard behaviour similar to the keyboard one.

### DIFF
--- a/src/app/components/inputmask/inputmask.ts
+++ b/src/app/components/inputmask/inputmask.ts
@@ -556,8 +556,8 @@ export class InputMask implements OnInit, ControlValueAccessor {
             }
 
             setTimeout(() => {
-                var begin = pos.begin;
-                var end = this.seekNext(begin)
+                const begin = pos.begin;
+                const end = this.seekNext(begin)
                 this.caret(begin, begin);
                 this.clearBuffer(begin, end-1);
                 if (this.keepBuffer) {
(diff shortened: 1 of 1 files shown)


src/app/components/inputmask/inputmask.ts
```diff
@@ -550,21 +550,29 @@ export class InputMask implements OnInit, ControlValueAccessor {
...
             this.checkVal(true);
-            while (pos.begin > 0 && !this.tests[pos.begin - 1]) pos.begin--;
+            while (pos.begin > 0 && !this.tests[pos.begin]) pos.begin--;
             if (pos.begin === 0) {
                 while (pos.begin < (this.firstNonMaskPos as number) && !this.tests[pos.begin]) pos.begin++;
             }
 
             setTimeout(() => {
-                this.caret(pos.begin, pos.begin);
+                var begin = pos.begin;
```
SoyDiego, line 559: You should use `const` instead of `var`

mertsincan:
Thanks a lot for your contribution! We'll review and get back to you.

mertsincan:

PR #16307, opened 2024-08-29
Fix #15192: Row grouping breaks after first page

### LINKED ISSUE #15192: Grouping Functionality in p-table Breaks After Pagination Navigation Beyond First Page
### Describe the bug

Issue Title:

Grouping Functionality in p-table Breaks After Pagination Navigation Beyond First Page

Issue Description:

When integrating paginator logic into the existing PrimeNG <p-table> component, a functionality inconsistency occurs. After navigating beyond the first page using the paginator controls, the grouping functionality ceases to operate correctly. Instead of maintaining the intended grouping structure, each row appears to be grouped individually.

Steps to Reproduce:

    Use the existing PrimeNG example of Row Group Subheader in StackBlitz (accessible via https://primeng.org/table).
    Integrate paginator logic into the PrimeNG <p-table> component.
    Navigate beyond the first page using the paginator controls.
    Observe the breakdown in grouping functionality, where each row seems to be grouped individually.

Expected Behavior:

After navigating to subsequent pages using the paginator, the grouping functionality should remain intact, preserving the original group structure as defined by the groupRowsBy attribute.

Additional Information:

    PrimeNG Version: 17.12.0
    Angular Version: 17.3.1

Note:

This issue significantly impacts user experience and functionality, especially when managing large datasets where pagination and grouping are vital features. Prompt attention and resolution of this matter would be greatly appreciated. Additionally, it's worth noting that this issue may be related to https://github.com/primefaces/primeng/pull/14672.

### Environment

stackblitz.com

### Reproducer

_No response_

### Angular version

17.3.1

### PrimeNG version

17.12.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

v18.18.0

### Browser(s)

_No response_

   1) Use the existing PrimeNG example of Row Group Subheader in StackBlitz (accessible via https://primeng.org/table).
   2) Integrate paginator logic into the PrimeNG <p-table> component.
   3) Navigate beyond the first page using the paginator controls.
   4) Observe the breakdown in grouping functionality, where each row seems to be grouped individually.

### Expected behavior

_No response_

marcoschndlr:
Fixes #15192

### What was the problem
`shouldRenderRowGroupHeader` expects the `rowIndex` to be the absolute index (i.e. `first + rowIndex`)  when paginated.
The problem was that the `rowIndex` passed starts from `0` for every page, and the index calculation in the method then resulted in a negative number.

### How was the problem solved
For expanded rows,  the method `getRowIndex` is used, which adds the starting offset (`first`) to the `rowIndex`.
This method is used here now as well.

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -3084,7 +3084,7 @@ export class Table implements OnInit, AfterViewInit, AfterContentInit, Blockable
     template: `
         <ng-container *ngIf="!dt.expandedRowTemplate">
             <ng-template ngFor let-rowData let-rowIndex="index" [ngForOf]="value" [ngForTrackBy]="dt.rowTrackBy">
-                <ng-container *ngIf="dt.groupHeaderTemplate && !dt.virtualScroll && dt.rowGroupMode === 'subheader' && shouldRenderRowGroupHeader(value, rowData, rowIndex)" role="row">
+                <ng-container *ngIf="dt.groupHeaderTemplate && !dt.virtualScroll && dt.rowGroupMode === 'subheader' && shouldRenderRowGroupHeader(value, rowData, getRowIndex(rowIndex))" role="row">
                     <ng-container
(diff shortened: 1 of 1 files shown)


nckirik:
this should be also applied to group footers

PR #16391, opened 2024-09-16
Fix indexing in the p-tree so drag and drop will work with virtual scroll.

### LINKED ISSUE #16296: p-tree drag and drop breaks with virtualScroll
### Describe the bug

When you use p-tree and set the virtualScroll property to "true" (and the virtualScrollItemSize to some integer value), then drag and drop feature breaks.

It duplicates the item and adds it to the end of the children for the parent within which you are dragging/dropping.

To be more specific: I have a parent node with 3 child nodes. Let's call the parent A and the nodes 1, 2, 3. If you want to move node 1 to be after node 2, then you drag it and drop it after 2 in the list. But, primeng appears to have a bug: when virtualScroll is on, the dragged and dropped node (1) gets duplicated and put on the end of the list (and the original node remains unmoved). So parent A's children look like: 1, 2, 3, 1. I can print the whole tree data after the move and see that it's duplicated at the end of the child nodes. When I turn virtualScroll off, vualá: it works fine (the node moves to the proper position and is not duplicated). This bug happens regardless of which node you move and where you move it. It's always duplicated and appended to the end of the list.

### Environment

package.json dependencies:
Angular: ^17.0.9
"primeng": "~17.12.0"
chrome 127.0.6533.100

### Reproducer

_No response_

### Angular version

17.0.9

### PrimeNG version

17.12.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

22.3.0

### Browser(s)

_No response_

1. Enable virtualScroll in p-tree.
2. Open p-tree component.
3. Drag node and drop it in new location.

### Expected behavior

Should move the node to the proper location but isn't (duplicating at the end of the list no matter what).

ethan-gerardot:
Find the tree node's index programmatically based on its position within its parent, instead of using the ngFor array's properties because with virtual scroll, the list is flat and the indexes lost relativity to the tree node's parent.

### Defect Fixes
#16296

### DIFF
--- a/src/app/components/api/treenode.ts
+++ b/src/app/components/api/treenode.ts
@@ -79,4 +79,12 @@ export interface TreeNode<T = any> {
      * Defines if the node is loading.
      */
     loading?: boolean;
+    /**
+     * Defines if the node is the last child.
+     */
+    lastChild?: boolean;
+    /**
+     * Index of the node relative to its level.
+     */
+    index?: number;
 }
--- a/src/app/components/tree/tree.ts
+++ b/src/app/components/tree/tree.ts
@@ -30,7 +30,7 @@ import { RippleModule } from 'primeng/ripple';
 import { Scroller, ScrollerModule } from 'primeng/scroller';
 import { ScrollerOptions } from 'primeng/api';
 import { ObjectUtils } from 'primeng/utils';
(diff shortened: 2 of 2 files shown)


mertsincan:

Issue #16411, opened 2024-09-20
Dropdown: Floating placeholder overlapping when value is `null` or `undefined`.


Elias-Graf:
### Describe the bug

Let's say you have a dropdown with the following options:

```ts
  options: { label: string; value: number | undefined }[] = [
    { label: 'None', value: undefined },
    { label: 'First', value: 0 },
    { label: 'Second', value: 1 },
  ];
```

If you have a floating label and select the `None` value, the floating label will overlap with the value like this:
![image](https://github.com/user-attachments/assets/33edbeb9-d3a9-4b4a-a9fc-0a26807e778f)

Most likely this is because of this check, that is comparing if the _value_ is something, instead of checking if an option is selected:
https://github.com/primefaces/primeng/blob/ae58a4948ac6225e053641e7462a5f787be28e48/src/app/components/dropdown/dropdown.ts#L1012C1-L1015C8

Most likely the issue would be solved, if the same code is used as for:
https://github.com/primefaces/primeng/blob/ae58a4948ac6225e053641e7462a5f787be28e48/src/app/components/dropdown/dropdown.ts#L1002C1-L1010C8

### Environment

Not applicable, otherwise please let me know.

### Reproducer

https://stackblitz.com/edit/github-epdzxk?file=src%2Fapp%2Fapp.component.html

### Angular version

17.x.x

### PrimeNG version

^17.18.10

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

v18.20.3

### Browser(s)

Firefox, Chrome

### Steps to reproduce the behavior

1. Open provided reproducer
2. Select none from the dropdown

### Expected behavior

Label should be floating when an option is selected. It should only not be floating, if the current selected value is not an option, e.g.: no option is selected.

PR #16421, opened 2024-09-24
Fix dropdown filled calculation regarding null or undefined values

Elias-Graf:
Fixes primefaces/primeng#16411


mertsincan:

PR #16465, opened 2024-10-01
fix: move all columns with CTRL+A in picklist

### LINKED ISSUE #16260: Picklist: CTRL+A Does Not Move All Columns When Using Move Icon
### Describe the bug

Steps:
1. Press Ctrl + A to select all source data.
2. Click the 'Move To Right' icon.
3. Expected result: All source data should be moved to the right. 
    Actual result: Only some source data is moved to the right.

The cause of this issue lies in the following code location:

https://github.com/primefaces/primeng/blame/a7ba4deefd89e26df34140ac6b0983a6a3e14109/src/app/components/picklist/picklist.ts#L930

Preliminary cause analysis:
When an element is removed from `this.source,` the corresponding elements in `this.selectedItemsSource` are also removed.
pickList_ctrl_A_1
pickList_ctrl_A_2
pickList_ctrl_A_3

### Environment

angular17
primeng16

### Reproducer

_No response_

### Angular version

17

### PrimeNG version

16

### Build / Runtime

Angular CLI App

### Language

TypeScript

v16.15.1

### Browser(s)

Chrome

1. Press Ctrl + A to select all source data.
2. Click the 'Move To Right' icon.

### Expected behavior

Expected result: All source data should be moved to the right. 
Actual result: Only some source data is moved to the right.

navedqb:
Fixes #16260

### DIFF
--- a/src/app/components/picklist/picklist.ts
+++ b/src/app/components/picklist/picklist.ts
@@ -1130,11 +1130,14 @@ export class PickList implements AfterViewChecked, AfterContentInit {
 
     moveRight() {
         if (this.selectedItemsSource && this.selectedItemsSource.length) {
-            for (let i = 0; i < this.selectedItemsSource.length; i++) {
-                let selectedItem = this.selectedItemsSource[i];
+            const selectedItemsCopy = [...this.selectedItemsSource];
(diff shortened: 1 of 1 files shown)


mertsincan:

PR #16477, opened 2024-10-02
chore: Updates news banner for v18-beta.3 release

JamesTheTerry:
`🎉 v18-beta.1 is out now with new theming` -> `🎉 v18-beta.3 is out now with new theming`


mertsincan:
Thanks a lot, we'll update with rc.1 :)

PR #16497, opened 2024-10-07
fix: fixed label of movebottom-button in picklist component

e-matthaey:
**### Issue for this PR: https://github.com/primefaces/primeng/issues/16498**

The Movebottom Button in the picklist component references the wrong aria label.
Instead of 'Move Down' it should be 'Move Bottom'.
picklist


mertsincan:

Issue #16543, opened 2024-10-14
Primeicons.ts icon missing and wrong icons key


Mundi93:
### Describe the bug

I noticed a bug in [primeiconts.ts](https://github.com/primefaces/primeng/blob/master/src/app/components/api/primeicons.ts)

- `ASTERIKS = 'pi pi-asteriks'` is wrong, should be `ASTERISK = 'pi pi-asterisk'` [here](https://github.com/primefaces/primeng/blob/bc6952c82743f824cb2099202a072a3c9ff49129/src/app/components/api/primeicons.ts#L36)

The next ones are not real bugs, but suggestions for a more intuitive use:

- `ARROW_H = 'pi pi-arrows-h'` the key would be better if it were `ARROWS_H'` to better fit the value [here](https://github.com/primefaces/primeng/blob/bc6952c82743f824cb2099202a072a3c9ff49129/src/app/components/api/primeicons.ts#L34)
- `ARROW_V = 'pi pi-arrows-v'` the key would be better if it were `ARROWS_V'` to better fit the value; [here](https://github.com/primefaces/primeng/blob/bc6952c82743f824cb2099202a072a3c9ff49129/src/app/components/api/primeicons.ts#L35)

### Environment

N/A

### Angular version

18.0.1

### PrimeNG version

17.18.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.9.0

PR #16544, opened 2024-10-14
Asterisk icon wrong key and value; arrow-h and arrow-v wrong key

Mundi93:
I started a related issue some minutes ago [here](https://github.com/primefaces/primeng/issues/16543)

Just to have a small recap here too, [primeicons.ts](https://github.com/primefaces/primeng/blob/master/src/app/components/api/primeicons.ts) has a bugged value/keys with 3 icons.

Read related issue to a complete description


mertsincan:

Issue #16553, opened 2024-10-15
Component: RadioButton


Sergiorezende22:
### Describe the bug

The RadioButton component is not functioning correctly with screen readers. When navigating using only the keyboard, the screen reader does not announce when a radio button is checked. 

I searched for similar issues but didn't find any addressing this specific problem. The closest match was issue #12950, but it doesn't seem to be the cause of this behavior.

### Environment

I tested the RadioButton component using the NVDA screen reader on the PrimeNG main page:
https://primeng.org/radiobutton.

### Angular version

17.0.0

### PrimeNG version

17.18.10

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.11.1

### Steps to reproduce the behavior

1. Visit the following link: https://primeng.org/radiobutton.
2. Using any screen reader, navigate to the radio buttons on the page and use the keyboard to check one of them. Pay close attention to what the screen reader announces when the radio button is checked.

### Expected behavior

The screen reader should accurately announce what is displayed on the screen, but this is not happening at the moment.

PR #16554, opened 2024-10-15
Fix: Removed the unnecessary aria-checked attribute and improved acce…

Sergiorezende22:
The `event.preventDefault()` was causing the screen reader to misinterpret the state of the radio button. I also removed the unnecessary aria-checked attribute, as it is redundant for a radio input element.

#16553

### DIFF
--- a/src/app/components/radiobutton/radiobutton.ts
+++ b/src/app/components/radiobutton/radiobutton.ts
@@ -1,11 +1,11 @@
 import { CommonModule } from '@angular/common';
 import { ChangeDetectionStrategy, ChangeDetectorRef, Component, ElementRef, EventEmitter, Injectable, Injector, Input, NgModule, OnDestroy, OnInit, Output, ViewChild, booleanAttribute, forwardRef, numberAttribute } from '@angular/core';
 import { ControlValueAccessor, NG_VALUE_ACCESSOR, NgControl } from '@angular/forms';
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks a lot!

r05014:
On Primeng Angular version 19.0.6
I have also the w3c error 
The aria-checked attribute must not be used on an input element which has a type attribute whose value is radio
code Html 
`    <div class="flex items-center gap-1">
      <p-checkbox
        [binary]="true"
        [id]="id+'-p-checkbox'"
        [inputId]="id"
        [formControl]="formControl"
        [formlyAttributes]="field"
        [class.is-invalid]="showError"
        [styleClass]="showError?'ng-invalid ng-dirty':null"

        [attr.aria-describedby]="showError?id + '-formly-validation-error':null"
        [attr.aria-invalid]="showError"
        (onChange)="props.change && props.change(field, $event)"
      >
      </p-checkbox>
      <label [for]="id" class="ml-2" [ngClass]="showError?'text-red-500':null">  {{ props.label |translate}}</label>
    </div>`

Browser html : 
`<input type="radio" class="p-radiobutton-input" value="true" id="formly_29__RadioType_memeAdresse_1_0" name="formly_29__RadioType_memeAdresse_1-p-radioButton" aria-checked="true" pc113="">`

PR #16558, opened 2024-10-16
fix: primefaces#16557 - V18 extend card theming support

EnricoMessall:
fix: primefaces#16557

- New padding tokens for more elements
- Tokens for card content and footer
- Realigned caption token to match with cardstyle.ts
- Support font adjustments also for subtitle

### DIFF
--- a/.github/workflows/pr-checker.yml
+++ b/.github/workflows/pr-checker.yml
@@ -27,7 +27,7 @@ jobs:
             - uses: actions/checkout@v3
             - uses: actions/setup-node@v3
               with:
-                  node-version: 16
+                  node-version: 18
                   cache: 'npm'
 
             - name: Install packages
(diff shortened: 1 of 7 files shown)


cagataycivici:
I'm afraid we can't add new tokens right now, the tokens are shared between all Prime UI libraries and the Figma UI Kit needs to be updated when a new token is added so it has its own process and timeline. We're currently focused on stability of the theming system not adding new tokens until v18 final. New tokens cannot be added v18.1.0.

For now you can extend it yourself with custom tokens, here is an example from Vue version, same applies to NG;

https://www.youtube.com/watch?v=galIbID_Ywo

However eventually, I think we should add the tokens you mentioned. We've bookmarked this PR when new card tokens arrive in v18.1.0.

PR #16584, opened 2024-10-18
Fixed primefaces#16577 - Select invalid state is wrong

EnricoMessall:
fix primefaces#16577

Using border-color instead of outline as focus state does it already


mertsincan:
Thanks a lot for your contribution! We'll review it and get back to you.

mertsincan:
Thanks a lot for your contribution! but, fixed in https://github.com/primefaces/primeng/commit/2ead7043d3278d03a7bb7f8de7c93cf85e057393

Issue #16576, opened 2024-10-17
v18 p-columnFilter: Column filtering indicator is missing


rosenthalj:
### Describe the bug

In PrimeNG v18 when a table column is filtered  using the p-columnFilter component there is no indication of filtering in the corresponding table column header.

**Note:  When viewing the screenshots listed below and the videos listed below,  you may want to use "full screen mode",  because the "highlighting" of the background colors of the v17 filter icons buttons can be hard to see.** 

The following screenshot compares the "**correct**" v17 filter table demo after filtering vs "**incorrect**" v18 filter table demo after filtering:

![Angular_Table_Component_and_Angular_Table_Component_and_Angular_Select_Component](https://github.com/user-attachments/assets/bf551750-af72-47dc-963e-e2c1e0c29297)

The following screenshot compares the style of the filtered name button element for v17 and v18

![Angular_Table_Component_and_Angular_Table_Component_and_History_for_src_app_components_megamenu_megamenu_ts_-_primefaces_primeng_and_Angular_Select_Component](https://github.com/user-attachments/assets/f32dac3f-6f10-47c9-bc84-bba67a75d808)

The following video shows the v17 Table Advanced Filter demo "**indicating**" that column filters are set

https://github.com/user-attachments/assets/c40b22d0-337d-4163-9d2e-d06295ed3bbe

The following video shows the v18 Table Advanced Filter demo  "**not indicating**" that column filters are set

https://github.com/user-attachments/assets/b18edccb-1927-4226-ae0a-770eea155033

### Environment

n/a

### Reproducer

https://v18.primeng.org/table#filter-advanced

### Angular version

18.x

### PrimeNG version

18.0.0.beta-3

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

n/a

### Browser(s)

chrome, firefox, safari

### Steps to reproduce the behavior

1. Run the v18 Table  Filter Advance Demo: https://v18.primeng.org/table#filter-advanced
2. Filter for one or more columns
--- After the table is filtered there will be no indications in the table that any of the  filters have been set  (bad)

### Expected behavior

When a column filter is set the corresponding column filter has a visual indication that it has been set.

I.e.  v18 column filter set indicating functionality should work similarly to the current v17 functionality

cagataycivici:
[...]

PR #16588, opened 2024-10-18
Fixed primefaces#16576 - Column Filter missing primary color

EnricoMessall:
fix: primefaces#16576

+ CX Function add support for flatting functions for styleClass (Similar to ngStyle doing it)
+ Added new theming for p-column filter active button color
+ Added p-datatable-column-filter-button-active class again

The cx function adjustment is important so parts using [styleClass] can work with cx having a function as return value. With that the cx functions flattens the function result as the ngStyle does and directly provides styleClass with the required string. Making styleClass do that directly would require a rework of all styleClass components so that was the best option from my point of view.


rosenthalj:
This PR #16588 doesn't fix Issue #16576.

I download the PR branch and tested the fix.  The video and annotate3d screenshot listed below shows that this PR doesn't fix Issue #16576:  **"When a table column is filtered using the p-columnFilter component there is no indication of filtering in the corresponding table column header"**.


https://github.com/user-attachments/assets/6d74bb8e-0837-4ca7-aade-d45b023b4ce7



Angular_Table_Component

EnricoMessall:
@rosenthalj Look at the video i send, i couldnt upload it here because its too big for github. Its also showing this MR, you can see how the active button is applied to the filter but the color is new in PrimeNG v18 and is also black but with another level of darkness. You can also see that i change the overall primary color token and then also the selected filter color changes to the adjusted primary color.

rosenthalj:
> @rosenthalj Look at the video i send, i couldnt upload it here because its too big for github. Its also showing this MR, you can see how the active button is applied to the filter but the color is new in PrimeNG v18 and is also black but with another level of darkness. You can also see that i change the overall primary color token and then also the selected filter color changes to the adjusted primary color.

@EnricoMessall  can you try saving your video at a lower resolution or provide screenshots? I can not verify the results of your PR.  Somehow/Somewhere something is obviously wrong or inconsistent?

EnricoMessall:
@rosenthalj Look into the primeng discord. I've messaged you there with the video.

rosenthalj:
@EnricoMessall 

Your video on **PrimeNG Discord** is nice and I now understand what you are trying to convey.

I confirmed that after your changes, the **p-datatable-column-filter-button-active** class is now being added to the filter button.  This is similar to v17.  PrimeNG v17 uses the **p-column-filter-menu-button-active** class for the filter button.  It appears that PrimeNG v18 is changing a lot of CSS class names.

Unfortunately,  even with your changes,  the PrimeNG Table Filter Advance Demo is still not working properly.  As a result,  any software upgraded from 17.x  to 18.x will still not have any visual indication when the  table has been filtered unless the developer realizes its now their responsibility to fix table filter visual indicators.


Note on your video: You were eventually able to change the filter button color by using the browser's "inspection" functionality.   In v17 both the filter button color and the filter button background colors are changed to indicate  a column has been filtered.

EnricoMessall:
Hi @mertsincan i've fixed Merge Conflicts. I would please you to not take that long with merging it due to the size of changes, i acutally dont want to resolve all conflicts in every file a fourth time :D

mertsincan:
Thanks a lot for your contribution! But, unfortunately, we cannot accept this PR. New tokens can only be added by the PrimeNG team. Because we have many products related to it and we have to make decisions with the design team to add new tokens.

PR #16592, opened 2024-10-19
fix: #16591, InputNumber: InputNumber with horizontal and vertical buttons are not accessible through keyboard.

akshayaqburst:
fix: #16591, InputNumber: InputNumber with horizontal and vertical buttons are not accessible through keyboard.


mertsincan:

PR #16593, opened 2024-10-19
Fix #16298

### LINKED ISSUE #16298: Menu closes itself immediately when the menu and its button toggle is inside ShadowDom
### Describe the bug

`p-menu` cannot be opened when the menu and its button toggle is inside ShadowDom encapsulation. The menu closes itself right away.

This seems to be caused by `bindDocumentClickListener`. Inside the shadow dom, `event.target` will be the shadow root even if the button itself is clicked, so `isOutsideContainer‎` and `isOutsideTarget` are always true. It might need to check if the event is composed and if the toggle element contains the first element of the event's composed path (`event.composed ? toggleElement.contains(event.composedPath()[0]) : toggleElement.contains(event.target)`).

### Environment

Angular app with PrimeNG

### Reproducer

https://stackblitz.com/edit/github-fhq1jp?file=src%2Fapp%2Fapp.component.html

### Angular version

17.3.7

### PrimeNG version

17.18.8

### Build / Runtime

Angular CLI App

### Language

TypeScript

21.7.1

### Browser(s)

_No response_

- Create a component with a popup p-menu and button to toggle the menu.
- Wrap that component around with another component that has ShadowDom encapsulation.
- In primeng-menu.mjs, put a debug point inside the `hide()` function.
- Click on the button. The debug point will pause inside `hide()`.
- The menu is visible when the debug point is paused. After the debug point resumes, the menu will disappear.

### Expected behavior

`p-menu` should stay opened.

kapnak:
I encounter the exact same issue as mention in #16298.

This pull request implement the fix proposed by @hyilai in the issue #16298.

### DIFF
--- a/src/app/components/menu/menu.ts
+++ b/src/app/components/menu/menu.ts
@@ -724,8 +724,9 @@ export class Menu implements OnDestroy {
             const documentTarget: any = this.el ? this.el.nativeElement.ownerDocument : 'document';
 
             this.documentClickListener = this.renderer.listen(documentTarget, 'click', (event) => {
-                const isOutsideContainer = this.containerViewChild?.nativeElement && !this.containerViewChild?.nativeElement.contains(event.target);
-                const isOutsideTarget = !(this.target && (this.target === event.target || this.target.contains(event.target)));
+                const eventTarget = event.composed ? event.composedPath()[0] : event.target;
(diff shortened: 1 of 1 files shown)


mertsincan:
Thank a lot for your contribution!

Issue #16567, opened 2024-10-17
ViewEncapsulation.ShadowDom / Missing styles


fxprunayre:
### Describe the bug

When using 
```ts
  encapsulation: ViewEncapsulation.ShadowDom,
```

![image](https://github.com/user-attachments/assets/b78283fd-7187-412f-807f-6fe4f4ae1fb4)

Styles are added in main body 
![image](https://github.com/user-attachments/assets/56b3fac5-4460-4af9-ac01-9e82d2a1681a)

and are not available in the shadowDOM part.

![image](https://github.com/user-attachments/assets/0cb7424b-32c9-4e2a-960b-20a47a671ac9)

### Environment

-

### Angular version

18.2.0

### PrimeNG version

18.0.0-beta.3

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.19.1

### Steps to reproduce the behavior

Create a component and use
```ts
  encapsulation: ViewEncapsulation.ShadowDom,
```

### Expected behavior

Insert style in ShadowDom instead of main document.

mertsincan:
I think we can try to inject styles using https://github.com/angular/angular/blob/main/packages/platform-browser/src/dom/dom_renderer.ts#L408

fxprunayre:
> I'll work on it and get back to you.

@mertsincan , I tried to investigate the issue in https://github.com/primefaces/primeng/issues/16567 but being new to primeng it is quite hard to figure out all the mechanism. Any help from a more experimented developer on the theming system would be appreciated! It you can have a look? and check if it is the right direction for a fix. Thanks.

maitrungduc1410:
what I realize is PrimeNG appends CSS to `<head>` at runtime (when a component is actually used)

<img width="480" alt="Image" src="https://github.com/user-attachments/assets/a5fae33e-8b12-48bc-84b7-120ebdb430f3" />

the logic to append CSS is here: [primeng/packages/primeng/src/usestyle/usestyle.ts at master · primefaces/primeng](https://github.com/primefaces/primeng/blob/master/packages/primeng/src/usestyle/usestyle.ts)

I modify the code to append styles to eh shadowRoot, but still some styles are missing:

![Image](https://github.com/user-attachments/assets/83d0d9a5-e655-4b65-b863-587908a20fa6)

Then I found that it's because the style here are under `:root`, while if it's inside ShadowDOM it should be `:host`, then I manually change, then it finally works:

[...]

PR #16596, opened 2024-10-20
Fixes primefaces#16595 - CSS Layers for example with Tailwind CSS fail when appending styles to header and SSR enabled

EnricoMessall:
fix primefaces#16595

Using custom implementation of prepend instead of the normal one that is not known by SSR


fxprunayre:
Would this also fix https://github.com/primefaces/primeng/issues/16567? At least it gives me a pointer where to look about styles.

Issue #16546, opened 2024-10-14
Modernize Components to Align with New Angular Syntax (Signals, Input/Output Signals, Computed, and Effect)


medbenmakhlouf:
In preparation for upcoming Angular versions, we need to update our components to leverage the new Angular syntax, which includes using Signals, Input/Output Signals, computed, and effect. This modernization aims to reduce the reliance on lifecycle hooks, improving change detection and overall performance. These changes will also future-proof the application, aligning it with Angular's evolving best practices.

I will take the initiative to convert components progressively, starting from the simplest to the most complex ones, working in collaboration with the PrimeNG team. Additionally, I will write more comprehensive test cases to ensure the correctness and reliability of the updated components.

This update will include:

Replacing lifecycle hooks with Signals for better state management.
Introducing Input and Output Signals for component communication.
Utilizing computed for derived state.
Applying effect to handle side effects efficiently.
By adopting these changes, we can ensure our components are more reactive, performant, and easier to maintain in the long term, with better test coverage to support the changes.

mertsincan:
Wowww, thanks a lot for your contribution! After the 18 release, we plan to work on tests and signal base structures. That's why I've set a 18.x milestone for now. I'll review and merge this PR before starting work. Thanks a lot for your understanding!

medbenmakhlouf:
@mertsincan You're very welcome! I'm glad to contribute and be part of this evolution. The 18.x milestone makes perfect sense, especially with the focus on tests and signal-based structures. Please feel free to share any feedback on the PR—I’ll be happy to make adjustments if needed. I’m also willing to help convert other components to modern Angularr. Looking forward to seeing the progress in this exciting phase. Thank you for your support!

medbenmakhlouf:
@cagataycivici Could you please explain why all my PRs are being closed? I understand that it might be better to address these changes once PrimeNG supports v19. I’m happy to resume working on them when v19 is ready.

jadgeagea:
Is primeNG 19 works with angular signals and how.
The documentation use ngModal not signal.
Please advise

PR #16602, opened 2024-10-20
feat (v18): Modernize floatlabel and expand test

medbenmakhlouf:
### Refactor:
Updated the ToggleSwitch component to leverage Angular's new signal-based inputs for enhanced reactivity and performance optimization.

### Test Cases: Added comprehensive test cases to cover:
Dynamic input handling (label, icon, image)
Inline styles and CSS class binding
Size variations (normal, large, xlarge)
Shape variations (square, circle)
Error handling for image loading
Accessibility (aria-label, aria-labelledby)
Custom styling and class inputs (styleClass, style)

Fix https://github.com/primefaces/primeng/issues/16546

### DIFF
--- a/src/app/components/floatlabel/floatlabel.spec.ts
+++ b/src/app/components/floatlabel/floatlabel.spec.ts
@@ -1,0 +1,62 @@
+import { ComponentRef } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { FloatLabel } from './floatlabel';
+
+describe('FloatLabel', () => {
+    let floatLabel: FloatLabel;
+    let fixture: ComponentFixture<FloatLabel>;
+    let floatLabelRef: ComponentRef<FloatLabel>;
+    let hostElement: HTMLElement;
+
+    beforeEach(() => {
+        TestBed.configureTestingModule({ imports: [FloatLabel] });
+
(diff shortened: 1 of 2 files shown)


vladovelichkovski:
@medbenmakhlouf 
An idea for all the PRs that you have for PrimeNG for v18.
mark all `inputs, outputs, viewChild, ciewChildren...` as `readonly` to prevent accidental mutations or assignments.
Also, you can mark all `injected` variables like you have here (`_componentStyle = inject(FloatLabelStyle);`) as `readonly` as well for the same reason. They can be also `private` or `#` prefixed.

Another idea is to start replacing all `private` methods and variables with `#` prefix. ([https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties))

A discussion is maybe needed with the PrimeNG team if they want to accept this code style and changes.

P.S: PR Checker is failing because the PR is not linked with any issue. Create an issue for every PR and link them.

mertsincan:
After the 18 release, we plan to work on tests and signal base structures. That's why I've set a 18.x milestone for now. I'll review and merge this PR before starting work. Thanks a lot for your understanding!

PR #16644, opened 2024-10-27
fix(docs): removes extra spacing and renames "warning" to "warn"

saloiofun:
- Removes extra spacing 

- Renames "warning" to "warn" to match the `severity` type defined in 
https://github.com/primefaces/primeng/blob/6f926f194c970794d8829db0dfa204a526c187bb/src/app/components/button/button.ts#L150

Issue https://github.com/primefaces/primeng/issues/16645


mertsincan:
Good catch! Could you please update PR? and I'll merge it asap.

Issue #16625, opened 2024-10-23
Component: Auto Complete - with Force Selection, selected option is overriding with the matched value 


dpeese:
### Describe the bug

When Force Selection is enabled, if we have options like below, on entering "Afghanistan" and selecting option "Afghanistanb" is still emitting "Afghanistan" instead of "Afghanistanb". It is only happening for the first time
Ex:
Afghanistan
Afghanistana
Afghanistanb

### Environment

primeng
nodejs
angular

### Reproducer

https://stackblitz.com/edit/632ydq?file=src%2Fapp%2Fautocomplete-reactive-forms-demo.ts,src%2Fservice%2Fcountryservice.ts,src%2Fapp%2Fautocomplete-reactive-forms-demo.html

### Angular version

18

### PrimeNG version

17.18.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18

### Browser(s)

chrome

mertsincan:
Fixed in https://github.com/primefaces/primeng/pull/16649

PR #16649, opened 2024-10-27
fix: In autocomplete dropdown, the item’s event has been changed from click to mousedown

P5175:
#16625
In an autocomplete input component, events can often lead to unintended order of execution, particularly when handling change and click events. When a user types in the input and then selects an item from the dropdown list, the change event fires first because it’s triggered as soon as the input field loses focus or updates its content. Then, the click event on the dropdown item is executed.

Problem:
change Event: This triggers first when the input value changes, such as when the user clicks outside the input or the dropdown, updating the internal value.
click Event on Dropdown Item: This triggers after the change event, potentially overriding or undoing the intended selection behavior.
Solution with mousedown Event:
To change the order and have the dropdown selection (click) execute before the change event, you can replace click with the mousedown event on the dropdown items. This way:

mousedown Event on Dropdown Item: Fires when the user presses the mouse button down on the dropdown item, executing before the input loses focus.
change Event: Now, the change event fires after mousedown, updating the input field only after the item is selected.


mertsincan:

Issue #16651, opened 2024-10-28
Multiselect: When opntionName is not set, getting value in template is failing


mperezd:
### Describe the bug

When optionValue is not set and we override selectedItems template, getting value is not working as expected when the object in options has a "value" field.

### Environment

Chrome and Linux

### Angular version

17.3.10

### PrimeNG version

17.18.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

21.6.0

PR #16652, opened 2024-10-28
Fix variable in getOptionValue in multiselect #16651

mperezd:
Fix for issue #16651


rosenthalj (commented):
I'm not sure if this proposed change is correct or not (I'm not trying to validate it),  but  the conditional ternary operator (this.optionValue ?) results in  (!this.optionValue) always being true and can be removed:

Fix_variable_in_getOptionValue_in_multiselect__16651_by_mperezd_·_Pull_Request__16652_·_primefaces_primeng

I then decided to review the original code change and now my question is was the original "!this.optionLabel" logic correct or not?  I don't know,  but ...

Blaming_primeng_src_app_components_multiselect_multiselect_ts_at_master_·_primefaces_primeng

Issue #16684, opened 2024-11-03
MultiSelect: Unable to select remove icon of chips through keyboard 


akshayaqburst:
### Describe the bug

MultiSelect: Unable to select remove icon of chips through keyboard 

### Environment

System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 1.61 GB / 7.47 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.20.2 - ~/.nvm/versions/node/v18.20.2/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v18.20.2/bin/yarn
    npm: 10.5.0 - ~/.nvm/versions/node/v18.20.2/bin/npm
    pnpm: 9.5.0 - ~/.nvm/versions/node/v18.20.2/bin/pnpm
  npmPackages:
    react: 18.3.1 => 18.3.1

### Angular version

17.x.x

### PrimeNG version

17.18.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.x.x

### Steps to reproduce the behavior

1. Go to https://primeng.org/multiselect#chips
2. Select one or two items, selected items appear as chips in the input box.
3. Try tab navigating to go the selected chips/items, the items are not focussible.

### Expected behavior

The selected chips/items should be focussible on tab navigation and on pressing enter key, we should be able to dismiss them.

PR #16685, opened 2024-11-03
fix: #16684, MultiSelect: Unable to select remove icon of chips through keyboard

akshayaqburst:
fix: #16684, MultiSelect: Unable to select remove icon of chips through keyboard


mertsincan:
Thanks a lot for your contribution! We'll check and get back to you.

PR #16735, opened 2024-11-13
fix(table): column resize does not work with sortable columns

### LINKED ISSUE #16734: Table: Column Resize broken with sortable columns
### Describe the bug

When resizing columns that are sortable, the resize event is triggered on the div with the `cbsortableheaderwrapper`. This element does not have the correct width to determine the new column width which results in strange resizes. Also later in the code when the index of the resized column is resolved, it always resolves to `0` so only the first column size is changed as observed here:

FDq0i6M - Imgur

### Environment

-

### Reproducer

_No response_

### Angular version

18.2.0

### PrimeNG version

17.18.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

20.18.0

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

LoaderB0T:
Fixes #16734 by finding the parent TH element and using it as the resizeElement

### DIFF
--- a/src/app/components/table/table.ts
+++ b/src/app/components/table/table.ts
@@ -2512,8 +2512,8 @@ export class Table implements OnInit, AfterViewInit, AfterContentInit, Blockable
     onColumnResizeBegin(event: any) {
         let containerLeft = DomHandler.getOffset(this.containerViewChild?.nativeElement).left;
         let resizeThElement = event.target.parentElement;
-        while(resizeThElement.tagName !== 'TH') {
-            if(resizeThElement.tagName === 'TABLE') {
+        while (resizeThElement.tagName !== 'TH') {
+            if (resizeThElement.tagName === 'TABLE') {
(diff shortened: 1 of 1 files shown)


src/app/components/table/table.ts
```diff
@@ -2511,7 +2511,14 @@ export class Table implements OnInit, AfterViewInit, AfterContentInit, Blockable
 
     onColumnResizeBegin(event: any) {
         let containerLeft = DomHandler.getOffset(this.containerViewChild?.nativeElement).left;
-        this.resizeColumnElement = event.target.parentElement;
+        let resizeThElement = event.target.parentElement;
+        while (resizeThElement.tagName !== 'TH') {
```
mertsincan, line 2515: Why do you use `event.target.closest('th')`?

LoaderB0T: You are absolutely right, I should have done that! :)
Fixed!

mertsincan:

Issue #15834, opened 2024-06-12
Calendar: showClear with showIcon & [iconDisplay]="'input'" makes the clear button float


vmmansilla93:
### Describe the bug

When adding `[showClear]="true"` and [showIcon]="true" [iconDisplay]="'input'"` the clear icon is floating.
![image](https://github.com/primefaces/primeng/assets/164879765/fc22b18a-a127-48eb-9487-11cea0155646)

### Environment

-

### Reproducer

https://stackblitz.com/edit/izdhsy?file=src%2Fapp%2Fcalendar-icon-demo.html

### Angular version

last version

### PrimeNG version

last version

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20

adnanjlassi:
You need to add this line .p-calendar .p-calendar-clear-icon { right: 1.2rem; top: 0.5rem; } in the global CSS file styles.scss.

PR #16741, opened 2024-11-14
Issue 15834 floating button fix in calendar component

PaterMajkel:
Fixes https://github.com/primefaces/primeng/issues/15834

This PR addresses a button layout issue when showClear is enabled on a non button type p-calendar component.

### DIFF
--- a/src/app/components/calendar/calendar.css
+++ b/src/app/components/calendar/calendar.css
@@ -167,6 +167,21 @@
         cursor: pointer;
     }
 
+    .p-calendar-clear-icon-template {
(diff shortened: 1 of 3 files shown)


cetincakiroglu:
Hi,

Thanks a lot for the effort and support! Indeed the provided solution fixes the issue, we cannot introduce new style classes or manually enter the random rem values to position the items since the new theming architecture does it by dynamic tokens and styles. Accepting this will break the unity between Prime UI Kit and other Prime UI Libraries. Closing the PR.

Issue #16733, opened 2024-11-13
Component: p-inputNumber


Makarion:
### Describe the bug

Value in an inputNumber component cannot be overwritten when maximum number of characters is reached (Firefox) 

### Environment

Linux, Firefox

### Angular version

18.2.8

### PrimeNG version

17.18.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

22.4.0

### Browser(s)

Firefox 131.0.3

### Steps to reproduce the behavior

1. Set maxLenth property in a p-inputNumber component.
2. Fill the input with maximum number of digits allowed.
3. Select all the digits and try to overwrite it by entering any digit.

### Expected behavior

I would like to select all the digits and overwrite them by entering any digit as it happens when maxLength is not set or reached (that how it works in Chrome).

PR #16742, opened 2024-11-14
Fix: InputNumber(16733): unsupported `getSelection()` causes input unwritable

mark7p:
Fix for #16733

The component uses `getSelection()` which [is not supported](https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection#related_objects) in Firefox and Edge (Legacy). Does making `onInputKeyPress()` read `null` on selected texts.

Updated `onInputKeyPress()` for more browser agnostic approach.


mertsincan:

Issue #16768, opened 2024-11-18
Tab: disabled input is not disabling tab


JamesTheTerry:
### Describe the bug

The v18 Tab component's "disabled" input is not disabling the tab.

This can be observed within the v18 docs https://v18.primeng.org/tabs#disabled

<img width="860" alt="Screenshot 2024-11-18 at 18 20 29" src="https://github.com/user-attachments/assets/ab4a76a1-543f-4c49-be3e-0d6ba1233ad9">

Neither `<p-tab disabled>` nor `<p-tab [disabled]="true">` work.

### Environment

PrimeNG Tab Disabled example

### Reproducer

https://v18.primeng.org/tabs#disabled

### Angular version

18

### PrimeNG version

18.0.0-beta.3

### Build / Runtime

TypeScript

### Language

TypeScript

### Node version (for AoT issues node --version)

Sample PrimeNG

### Expected behavior

The tab is supposed to be disabled and non-selectable, like "Header IV" is in v17

https://primeng.org/tabview#disabled

<img width="859" alt="Screenshot 2024-11-18 at 18 20 48" src="https://github.com/user-attachments/assets/37354adc-a258-4a4f-8ad2-6f6e256ecba3">

PR #16773, opened 2024-11-18
Fix: Tab[(16768): disabled input is not disabling tab

mark7p:
Fix for #16768 

**Solution Implemented**: added `p-disabled` to host class mapping.


mertsincan:
Thanks a lot!

PR #16783, opened 2024-11-20
feat(angular): allow/support Angular 19 with zone.js 0.15

ussu:
Hi :)
Locally, I built a version and link it to a project and found out the current state of the lib is not fundamentally incompatible with Angular 19.
So I'm opening this PR so we can, if possible allow Angular 19 as a peer dependency, or at least have a place to discuss the issues encountered by those who have tried using PrimeNG with the latest version of Angular :)

I also posted a [message on Discord](https://discord.com/channels/787967399105134612/787967806367989790/1308595968182194270) about this

The first place I tried to look for information about Angular 19 support was in the Issues, 2nd was open PRs so if you think it's appropriate I can open an issue as well

Cheers

### DIFF
--- a/src/app/components/package.json
+++ b/src/app/components/package.json
@@ -18,10 +18,10 @@
     "module": "primeng.js",
     "typings": "primeng.d.ts",
     "peerDependencies": {
-        "@angular/core": "^17.0.0 || ^18.0.0",
-        "@angular/common": "^17.0.0 || ^18.0.0",
-        "@angular/forms": "^17.0.0 || ^18.0.0",
+        "@angular/core": "^17.0.0 || ^18.0.0 || ^19.0.0",
+        "@angular/common": "^17.0.0 || ^18.0.0 || ^19.0.0",
+        "@angular/forms": "^17.0.0 || ^18.0.0 || ^19.0.0",
         "rxjs": "^6.0.0 || ^7.8.1",
-        "zone.js": "~0.14.0"
+        "zone.js": "~0.14.0 || ~0.15.0"
     }
 }


mertsincan:
Hi @ussu,

Did you encounter any problems after making these changes? Any feedback will help us :) Also, could you please create an issue for it?

garysb:
This is the issue I get when running with Angular 19.0.0.  The code works on Angular 18.x.x and used to work with earlier versions of Angular 19 beta and PrimeNg 18 beta releases. From what I can tell, it is trying to initFeatures with an undefined value at index 3 in the array.  I can report it as a new bug if wanted.



Screenshot 2024-11-20 at 18 21 51

ussu:
I see that https://github.com/djay1977 already created a discussion for Angular 19 support here https://github.com/orgs/primefaces/discussions/3065 :) @mertsincan 

I didn't encounter any problems but again I'm trying this on a fairly simple app 🙌🏼

ussu:
@garysb I didn't encounter the `fn is not a function` problem  
could you share where in primeng-icons-baseicon.mjs the issue occurs ?

ussu:
I'll also ask other colleagues who use angular/primeng to try it out and report issues/blocages if they see any on other projects

AStoker:
I'm running into the `fn` issue. Created a clean new Angular 19 app via the cli, then added PrimeNG. Tried to import the Toolbar component, and that's where I get the issue.
To expound on the part that seems to be failing around the init of features, it is explicitly trying to init the `StandaloneFeature`, but that is null (likely due to something in angular 19 where standalone is now the default?).

AStoker:
Related. Pulled down the repo and updated deps to use angular 19 to build (needs more than just those peer deps updated). I did this because it seems the build process from 18 to 19 has changed in angular, particularly around the standalone components. The standalone feature flag that was previously included in angular 18 was removed because standalone is now on by default. 
So when you build for angular 19, that feature isn't included. This is a problem with the shipped code for v18 of PrimeNG which was built with  version 18 of angular and expecting that feature to exist.
In the process of trying to update PrimeNG to use angular 19, I did notice that the dropdown component seemed to be bugged, as it was treating the Dropdown component incorrectly inside the DropdownModule. Looking into it more now, but wanted to give a bit of context for any other lost soul who stumbles over here.

AStoker:
Alright! Confirmed. After pulling down PrimeNG and building it with Angular 19, the fesm output correctly doesn't use the standalone feature flag that's removed in 19.
There were multiple components though that seemed to be "incomplete" in the standalone transition, and I had to manually set their `standalone` value to `true`, which seemed odd.
I can make a fork so @cetincakiroglu can see what I did (since he's marked as the reviewer here), but I wouldn't use that fork for any kind of production (it was very much a quick mashup to get prime to build with ng 19)

AStoker:
The dep for eslint-schematics is still being updated to support v19 though, so it's a bit behind and might cause some noise in the dep install: https://github.com/angular-eslint/angular-eslint/pull/2109

mertsincan:
Thank you so much for all the feedback! I've created an issue, and we are currently working on it.
https://github.com/primefaces/primeng/issues/16827

mertsincan:
Thanks a lot for your contribution! I'll add this change to v18 branch.

radarsu:
For reference I'm posting here workaround solution I've found: https://github.com/primefaces/primeng/issues/16827#issuecomment-2525468965

Issue #16781, opened 2024-11-19
Component: PanelMenu: Submenu displays angledownicon when submenu is collapsed


josefKiefer:
### Describe the bug

Hello,

We're seeing a couple issues with panel menu's submenus. The primary issue is if you have a submenu's item selected then you go to a different top-level menu and come back, it shows the chevron icon as a down arrow (indicating it's expanded) but the menu is collapsed. Then, when you then click that submenu item, it expands without an animation.

In addition it seems all submenu items collapse without an animation now. This was not the old behavior, before it would collapse with the same animation as the top-level menus. Please let me know if I should create a separate issue for this piece.

Thanks in advance!

### Environment

I noticed this when upgrading from Angular 13.3.0 to 18.2.8 and upgrading from PrimeNG 13.4.0 to 17.18.11.

We're using the panel menu in a controlled manner. On route changes we recursively loop through the panel menu and set the expanded property, then update the component's model through the use of a signal.

It appears to be an issue in all browsers.

### Reproducer

https://primeng.org/panelmenu#controlled

### Angular version

18.2.8

### PrimeNG version

17.18.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.12.2

### Steps to reproduce the behavior

The demo on your website seems to be on the same version and has the same issue: https://primeng.org/panelmenu#controlled

To repro at that link:

Click Users > New > Member
Click another top level menu (such as Calendar)
Click Users again and see that the New submenu is using the angledownicon instead of the anglerighticon. In addition, when clicking on the New submenu, there is no animation to open it up.
![image](https://github.com/user-attachments/assets/b9985ab0-5fa5-423b-aabe-78515d7b3aff)

### Expected behavior

1. The chevron should be pointed to the right, indicating the submenu item is collapsed.
2. Clicking that submenu item should expand it **with an animation**.
3. When all other submenu items collapse, they should have an animation just like the top-level menu items have.

PR #16790, opened 2024-11-20
Fix: PanelMenu(15403) - Submenu displays angledownicon when submenu is collapsed

### LINKED ISSUE #15403: PanelMenu nested subitems no longer get the expanded prop
### Describe the bug

In previous version 16.x.x 2nd & 3rd level nested sub menu items also had an expanded prop **in the model** when toggled open. In versions 17.x.x they are no longer getting the expanded prop. Only the 1st level parent items.

This makes it impossible to persist the expanded state of the panel menu beyond the first level. On a page refresh the menu will be reset to it's original state and user loses context.

```
{
    expanded: true,
    label: "Admin",
    items: [
         { 
            expanded: true,   < --- ! Not added here anymore. Same with deeper levels
            label: "Tasks",
            items: [...]
        },
        ...
    ]
 }
```

### Environment

Angular 18.0.1

### Reproducer
Toggle the menu a few items deep. The model will update below. Notice no nested items get an expanded attribute. Only top level parents. If you manually add expanded attribute to nested child (Documents) it won't update either. Notice how Files changes from false to true.

https://stackblitz.com/edit/s5kubx-5x7cem?file=src%2Fapp%2Fpanel-menu-basic-demo.html

### Angular version

18.0.1

### PrimeNG version

17.18.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

20.9.0

### Browser(s)

_No response_

Toggle a panel menu with 3rd or 4th level items. Inspect the model and you will see the nested parent items do not have the expanded: true property.

### Expected behavior

_No response_

mark7p:
Resolves #15403

**Problem:**: The logic from `ngOnChanges()` forces the component to create new `processedItems` with `expanded: undefined`.

**Solution**: Added check to keep `expanded` properties.

_This is a possible solution for #16781 too, but I created another PR(#16791) with the fix for the `hide` animation._

### DIFF
--- a/src/app/components/panelmenu/panelmenu.ts
+++ b/src/app/components/panelmenu/panelmenu.ts
@@ -357,7 +357,17 @@ export class PanelMenuList implements OnChanges {
     constructor(private el: ElementRef) {}
 
     ngOnChanges(changes: SimpleChanges) {
-        this.processedItems.set(this.createProcessedItems(changes?.items?.currentValue || this.items || []));
+        const hasItems = !!changes?.items?.currentValue;
+
+        if (hasItems) {
+            this.processedItems.set(this.createProcessedItems(changes?.items?.currentValue || this.items || []));
+            return;
+        }
+
+        // Update and keep `expanded` property from previous data
+        else {
+            this.processedItems.update(prev => prev.map(i => ({...i, expanded: i.expanded})))
+        }
     }
 
     getItemProp(processedItem, name) {
--- a/src/app/components/panelmenu/panelmenu.ts
+++ b/src/app/components/panelmenu/panelmenu.ts
@@ -366,7 +366,7 @@ export class PanelMenuList implements OnChanges {
 
         // Update and keep `expanded` property from previous data
         else {
-            this.processedItems.update(prev => prev.map(i => ({...i, expanded: i.expanded})))
+            this.processedItems.update((prev) => prev.map((i) => ({ ...i, expanded: i.expanded })));
         }
     }
 


mark7p:
Hi @josefKiefer , as for your issue #16781, the reason that the **_down icon_** still shows is because the submenu should retain its expanded state once its parent is opened again. If you toggled the submenu _(labelled "New")_ open, then toggled another top level menu _("Calendar")_ open, instead of the submenu _("New")_ collapsing, its parent menu _("Users")_ will collapse -- considered that the property `multiple` is not enabled.

I have tested it with the latest v17.18.13 [https://v17.primeng.org/panelmenu#controlled](url) on Chrome and Firefox and it behaves as it should following the steps you mentioned here:

>Click Users > New > Member
>Click another top level menu (such as Calendar)
>Click Users again and see that the New submenu is using the angledownicon instead of the anglerighticon. In addition, when >clicking on the New submenu, there is no animation to open it up.



The actual bug on #16781 is that the **submenu doesn't show its children when in expanded state**. The _angle down icon_ shows its right state, but the panel refuses to expand. I believe you misunderstood the right behavior of submenu because of this. Here is the expected behavior you provided:

> ### Expected behavior
>`1. The chevron should be pointed to the right, indicating the submenu item is collapsed.`
>2. Clicking that submenu item should expand it with an animation.
>3. When all other submenu items collapse, they should have an animation just like the top-level menu items have.

The first point in your **Expected Behavior** should be _1. The chevron should still be pointed down - indicating the submenu item is expanded and the children are visible._

You can observe this submenu behavior throughout the PrimeTek UI Suites [https://v3.primevue.org/panelmenu/](url) [https://primereact.org/panelmenu/](url) although by default, they are using `multiple = true`.

If I'm missing something regarding your issue, let me know.

Issue #16702, opened 2024-11-07
Docs: Scroller page sidebar disappears


Abodawoud:
### Describe the bug

When I search for 'Scroller' in the search bar and navigate to the Scroller page, the sidebar disappears.

### Environment

That is in the docs.

### Angular version

17

### PrimeNG version

17.18.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20

### Browser(s)

Chrome 130.0.6723.116

PR #16798, opened 2024-11-20
fix( #16702): disappearing sidebar

pointers-enjoyer:
fix for #16702 

This pull request removes unnecessary "init()" method and its usages, this results in a functioning sidebar.


mertsincan:
Thanks a lot for your contribution but, the issue is related to Doc search bar. I think this change will break our virtual scroller structure. So, closed.

pointers-enjoyer:
From my quick checkup. it seems to be working as it was, and this method (init) seems to be unecessary, are you sure this change will break your visual scroller sctructure?

PR #16807, opened 2024-11-21
Fix: Scroller(#16438 and #16564) - Scroll behavior bug

### LINKED ISSUE #16438: Component: Table Virtual Scroll does not correctly work with Row Expansion
### Describe the bug

Combining Virtual Scroll with Row Expansion does not always seem to work. 
When the content of the expanded area is quite high, you cannot scroll to the end of the table when there are only a few rows. When there are more rows (e.g. 100) You can correctly scroll to the end, but while scrolling down from the top you will encounter a point where the scroll jumps ahead a couple of rows more than it should. Probably due to swapping out the rows in the DOM.
I need to use lazy loading as there can be quite a number of rows to display. Also the expandable area can have a different height per row.

Here is a stackblitz that shows this: https://stackblitz.com/edit/stackblitz-starters-s7bkee

### Environment

Windows 10

### Reproducer

_No response_

### Angular version

18.0.0

### PrimeNG version

17.18.10

### Build / Runtime

Angular CLI App

### Language

TypeScript

20.15.1

### Browser(s)

_No response_

First scenario:
1. Expand the first row and scroll to the end with either mouse wheel or by dragging the scroller. 
 => The scroll area jumps up everytime you reach the bottom.

Second scenario:
1. In app.component.ts, line 25, change the array length from 10 to 100.
2. Expand the first row and scroll down. 
=> Around when row with id 11 comes into view, when scrolling further the view will jump several rows down (for me it where 5) when usually it only scrolls around 1 row when using the mouse wheel.

### Expected behavior

I would expect that the scrolling always works without jumping back up or moving too fast ahead, possibly skipping rows that cannot be seen in this case.

### LINKED ISSUE #16564: Component: Scroll behaiviour is broken in Scroller
### Describe the bug

The Scroller is broken on the official docs of https://primeng.org/scroller

https://github.com/user-attachments/assets/35ba60e1-766e-4d23-98a8-3b10b670c2f7

### Environment

The official primeng homepages environment

### Reproducer

_No response_

### Angular version

-

### PrimeNG version

17.18.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

-

### Browser(s)

Firefox 131, Chrome & Edge 129

1. Go to https://primeng.org/scroller
2. Try to scroll to the end of the scroller a bit faster
3. The scroller will constantly jump a bit back

### Expected behavior

The scroller will just scroll till the end without jumping up again

mark7p:
Resolves #16438, resolves #16564

**Problem**: Missing optional chaining that causes `undefined` read on scroll update.

**Solution**: Fixed.

### DIFF
--- a/src/app/components/scroller/scroller.ts
+++ b/src/app/components/scroller/scroller.ts
@@ -882,7 +882,7 @@ export class Scroller implements OnInit, AfterContentInit, AfterViewChecked, OnD
                 setProp('height', numItems, this._itemSize[0]);
                 setProp('width', this._columns?.length || this._items[1]?.length, this._itemSize[1]);
             } else {
-                this.horizontal ? setProp('width', this._columns.length || this._items.length, this._itemSize) : setProp('height', numItems, this._itemSize);
+                this.horizontal ? setProp('width', this._columns?.length || this._items.length, this._itemSize) : setProp('height', numItems, this._itemSize);
             }
         }
     }


mertsincan:

Issue #16755, opened 2024-11-15
Dropdown: does not respect emitEvent false


andrii-dymov:
### Describe the bug

Dropdown component emits value change event for it's corresponding control even so there is `emitEvent: false` parameter when I call setValue. It happens only if control has no initial value or it's value is `null`. Not sure if other components have the same issue.
It looks like conditions in `allowModelChange` method should be updated to fix this.
<img width="1220" alt="image" src="https://github.com/user-attachments/assets/08ebeafc-587f-4ade-ba81-f2645f7ba6a1">

### Environment

stackblitz

### Reproducer

https://stackblitz.com/edit/stackblitz-starters-7jmis4?file=src%2Fmain.ts

### Angular version

18.1.0

### PrimeNG version

18.0.0-beta.4

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.20.3

PR #16816, opened 2024-11-22
Fix:16755 - dropdown emit event

mateuszstudentzero:
This PR fixes the issue where the Dropdown component did not respect the emitEvent=false parameter when setting its value via writeValue. This occurred when the control had no initial value or its value was null.

Changes
Updated the writeValue method to include the emitEvent parameter.

#16755


mertsincan:
Thanks a lot for your PR!

Issue #16805, opened 2024-11-21
Component: Autocomplete with updateOn blur


SpeederX:
### Describe the bug

When you setup the autocomplete with the updateOn: "blur" as value change event of the FormControl, the input will be updated only after the first selection happens, but the value will always have the last selected value by the user.
Due to how the event of blur is handled in the <input>, this doesn't make it consistent with the logic.

The stackblitz is the one on the example page with virtual scroll, so I forked that one.

I add an additional comment as we found this bug during an integration of the component into our existing webapp. If you're ok with it I can provide already some insight on how we handled it and I think might work out to solve this kind of behavior. Are open contribution doable via fork + pr? ( new here )

### Environment

Windows, Chrome

### Reproducer

https://stackblitz.com/edit/fysj2c?file=src%2Fapp%2Fautocomplete-virtual-scroll-demo.html

### Angular version

17.3.8-11

### PrimeNG version

17.18.11-2

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.11.0

### Browser(s)

Chrome 131

### Steps to reproduce the behavior

1. Click on the input of the autocomplete
2. Select one option
3. Suggestions will close and once again, click on the input of the autocomplete
4. Select one option

### Expected behavior

When a user Select the option, the autocomplete should set the value to the form accordingly

PR #16828, opened 2024-11-24
Fix autocomplete updateOn blur on select (#16805)

SpeederX:
## Related issue
Issue reported here #16805
## Changes rationale
To handle `updateOn` setup with `blur` value, the following changes have been made:
1. Adapting the `onInputBlur` conditionally calling `onModelTouched` only if `forceSelection` is `false`.
2. Adding `onModelTouched` call in the `onSelect` ( unconditionally ), which provides the `FormControl` / `FormGroup` the value selected by the user, without considering the text input the user is searching ( unless forceSelection is false ).

## Additional notes

This way was chosen because [NG_VALUE_ACCESSOR](https://angular.dev/api/forms/ControlValueAccessor#registerOnTouched) updates the ancestor calling `registerOnTouched` which by angular docs:

> When implementing registerOnTouched in your own value accessor, save the given function so your class calls it when the control should be considered blurred or "touched"

If `forceSelection` is set to `false` and `updateOn` is set as `blur`, upon  both user input text / search and when user selects an option, the `FormControl` and `FormGroup` will be updated.

### DIFF
--- a/src/app/components/autocomplete/autocomplete.ts
+++ b/src/app/components/autocomplete/autocomplete.ts
@@ -1209,7 +1209,11 @@ export class AutoComplete implements AfterViewChecked, AfterContentInit, OnDestr
         this.dirty = false;
         this.focused = false;
         this.focusedOptionIndex.set(-1);
-        this.onModelTouched();
+        /** triggered only if user can input freely text
+         * Later on it must set touched also onSelect */
+        if(!this.forceSelection){
+            this.onModelTouched();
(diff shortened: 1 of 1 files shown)


SpeederX:
Does the link to the issue must be defined during the PR opening or can be done afterwards? I see others have the issue linked

mertsincan:
Thanks a lot for your contribution! Once you open it, you can link it to issue.

PR #16889, opened 2024-11-29
fix(#16630): ensure aria-describedby is dynamically set and removed for tooltip

### LINKED ISSUE #16630: Tooltip: aria-describedBy not applied to target
### Describe the bug

The tooltip documentation states that the tooltip's id should be used for the aria-describedby attribute of the target element. While the role and id are correctly assigned to the tooltip itself, the aria-describedby attribute is not automatically applied to the target element as expected.

### Environment

PrimeNG Version 17.18.11

### Reproducer

_No response_

### Angular version

18.2.0

### PrimeNG version

17.18.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

20

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

dermartinschmitz:
fixes #16630 

This fix improves accessibility by maintaining proper ARIA relationships for tooltips.

- Added `setAriaDescribedBy` to dynamically set `aria-describedby` attribute on the target element when the tooltip is created.
- Added `removeAriaDescribedBy` to remove the `aria-describedby` attribute when the tooltip is removed.

### DIFF
--- a/src/app/components/tooltip/tooltip.ts
+++ b/src/app/components/tooltip/tooltip.ts
@@ -206,6 +206,18 @@ export class Tooltip implements AfterViewInit, OnDestroy {
             });
         }
     }
+    
+    setAriaDescribedBy() {
+        const tooltipId = this.getOption('id');
+        if (tooltipId && this.active) {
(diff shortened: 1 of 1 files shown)


mertsincan:

Issue #16709, opened 2024-11-07
Component: p-paginator


Zackysh:
### Describe the bug

Cannot append paginator dropDown to body, I've tried both `appendTo` and `dropdownAppendTo` set to `'body'` with no luck.
This make the component unusable within cropped containers (e.g, overflow hidden).

On top of that, every drop-down in the library seems to always drop "down" even if there's no space left in the parent container or even the viewport.

### Environment

PrimeNG 18 + Angular 18, tried chrome.

### Angular version

18

### PrimeNG version

18-beta3

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.13.1

### Steps to reproduce the behavior

Add paginator to the bottom of any container with overflow hidden, add `appendTo 'body'`, the paginator dropdown is cropped.

### Expected behavior

The dropdown should not crop. p-select with `appendTo 'body'` does not crop.

Zackysh:
I would love to know if p-select and every other drop-down in PrimeNG 18+ will reposition on top if there's no space below at some point. I guess so, but when?

qwadrox:
@Zackysh Hi! It works according to my tests. There were two API dropdownAppendTo and appendTo and appendTo was not in use. I removed dropdownAppendTo to align with the other API, so appendTo can be used later. 
And I have opened a separate issue for what you mentioned related to overlays: https://github.com/primefaces/primeng/issues/16898

PR #16899, opened 2024-12-01
fix: 🐛 Remove duplicated appendTo in paginator

qwadrox:
#16709


mertsincan:
Hi @qwadrox, Thanks a lot for your contribution! could you please update your PR, then I'll review it asap. Thanks!

qwadrox:
@mertsincan I did it.
Renaming will be necessary for many places because "paginator dropdown" phrases are used instead of "paginator select". Still, I guess it would be better to refactor these when you enable `strictTemplates` in the future. For now, I did only where I edited the code.

Issue #16833, opened 2024-11-25
Component: Mutliselect


stargur:
### Describe the bug

Hi, so basicly i am using multiselect in 'chip' mode and i need the (onClear) event to be called when i unselect option from the 'x'  on the chip, but its only called when an option is unselected directly from the dropdown.

### Environment

project i am working on

### Angular version

17.2.3

### PrimeNG version

17.12.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.13.0

PR #16942, opened 2024-12-05
#16833 component  mutliselect

PaterMajkel:
Fix for the issue [#16833 ](https://github.com/primefaces/primeng/issues/16833). Should be determined if the change is needed as removing chips already triggers on change event.


mertsincan:

PR #16965, opened 2024-12-06
#16607 fixed bug in inputnumber for dead keys

### LINKED ISSUE #16607: InputNumber: Cannot delete character with backspace after inserting Up Arrowhead ^
### Describe the bug

When accidentially entering an Up Arrowhead `^` into an InputNumber, it is no longer possible to delete it using `Backspace`. I first have to navigate the cursor with `ArrowLeft` to the position before the `^` until I can delete again using `Backspace`. Also, in this case, pressing `Backspace` once will delete the entrie value, not digit by digit. 

### Environment

PrimeNG: 17.18.11

### Reproducer

https://stackblitz.com/edit/zlhnsq?file=package.json

### Angular version

18.2.3

### PrimeNG version

17.18.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.19.1

### Browser(s)

Firefox 131.0.3, Chrome 129.0.6668.101

1. Open Stackblitz Reproducer
2. Press Up Arrowhead `^` on your keyboard and `Space`
3. Now try to delete it by pressing `Backspace`
4. Move the cursor one position to the left using `ArrowLeft` and now press `Backspace`. Now the whole value is deleted. 

### Expected behavior

InputNumber should not allow inserting Up Arrowhead `^`.  
Pressing `Backspace` should delete any character in an InputNumber.    
Also, pressing `Backspace` once should not delete the entire value, but digit by digit.

JeremyHefti:
fixes #16607

### DIFF
--- a/src/app/components/inputnumber/inputnumber.ts
+++ b/src/app/components/inputnumber/inputnumber.ts
@@ -835,7 +835,7 @@ export class InputNumber implements OnInit, AfterContentInit, OnChanges, Control
         }
 
         this.lastValue = (event.target as HTMLInputElement).value;
-        if ((event as KeyboardEvent).shiftKey || (event as KeyboardEvent).altKey) {
+        if ((event as KeyboardEvent).shiftKey || (event as KeyboardEvent).altKey || event.key === "Dead") {
             this.isSpecialChar = true;
             return;
         }


mertsincan:

PR #16992, opened 2024-12-10
feat(inputnumber): add anyDecimalKey option

Joery-M:
We ran into the issue where the user can't use their desired decimal key on their keyboard due to the `inputNumber`'s locale being set differently than the users' browser locale.

This PR adds an attribute, `anyDecimalKey`, to `inputNumber` to allow any decimal key to be used for adding a decimal mark, regardless of locale.


mertsincan:
Thanks a lot for your contribution! Please create an issue and link this PR to it.

Issue #16910, opened 2024-12-02
CascadeSelect panel align issue


umyanl:
Hi,

I am using the CascadeSelect component and it shows at the bottom of the page. 

When I click the control, the first drop panel shows at the top of the control, which is expected behavior.
Then when I click one of the choice to show the sub-panel dropdown, it aligns top with the first dropdown panel, When the second panel is long, it cause the scrollbar to shown on the right. Is there anyway that the all the sub-panels also shown above the control, i.e. align bottom with the first panel.

Not sure it is possible by css, since it is need to be switched on only when the first panel shown at the top of the current control.
See my fork https://stackblitz.com/edit/r3ccsc for the issue. Click US it will show a long dropdown list.

Thanks,

Lixin

PR #17027, opened 2024-12-12
fix CascadeSelect panel align issue (Issue #16910)

umyanl:
issue #16910


cetincakiroglu:
Hi,

We're planning to rework all the overlays for PrimeNG v20 to solve this kind of issues completely rather than just one component. Closing the PR.

Issue #17012, opened 2024-12-11
ThemeProvider loadCommonTheme always injects undefined nonce


Quimey-CityGIS:
### Describe the bug

Loading a theme is blocked by CSP headers because the provided nonce value is hardcoded as undefined when the ThemeProvider loads the commonTheme values:

Link to primeng source code:
https://github.com/primefaces/primeng/blob/cb7c271bed2f507e30e89af3d35c8b2836acd6dc/packages/primeng/src/config/themeprovider.ts#L65

### Environment

Angaulr application served with strict CSP headers

### Reproducer

https://stackblitz.com/edit/stackblitz-starters-jwbthjxa?file=src%2Fapp.component.html

### Angular version

18.2.9

### PrimeNG version

18

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.18.0

### Expected behavior

The ThemeProvider should use the actual CSP_NONCE provided by Angular when executing loadCommonTheme.

mertsincan:
Fixed in https://github.com/primefaces/primeng/commit/cb8449b236f8065ce84c81c0a5fc80836efe83bb

ivan-price-acted:
@mertsincan  would it be possible to have this commit applied to the 18.x branch / release as well ?

Issue #17048, opened 2024-12-13
[CSP] providePrimeNG csp nonce does not solve all csp issues when "theme" property is defined


tom-vism:
### Describe the bug

Not all CSP issues are fixed when providing the CSP nonce into the PrimeNG config when the "theme" property is an empty object or "preset" is null.
While debugging, the first nonce value is passed undefined.

### Environment

strict csp policy, no unsafe-inline, we are using nonces.

### Reproducer

https://stackblitz.com/edit/azxstcmm?file=src%2Fmain.ts

### Angular version

^18.2.11

### PrimeNG version

^18.0.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

v20.18.1

### Browser(s)

Chrome 131.0.6778.140

### Steps to reproduce the behavior

![image](https://github.com/user-attachments/assets/71eee800-49c8-491f-be75-185d2823dacb)

![image](https://github.com/user-attachments/assets/e1405e95-cefa-4a52-99af-74d5d2557313)

![image](https://github.com/user-attachments/assets/14935376-39a4-4d5a-af83-ae1d18da4123)|

![image](https://github.com/user-attachments/assets/92c13c13-2c00-4728-9b4b-283ebef24ded)

### Expected behavior

Nonce attribute is set correctly and no csp issues are thrown in the console when "theme" is defined in the primeng config

SavGRY:
Yep, this is also happening to me with the rc `19.0.0-rc.1` with Angular 19.

Although it's not a stable release, it's the same problem.

Jonathazn:
Wondering if I have the same issue, but in my case, the nonces are empty strings like so:
`<style type="text/css" nonce="" data-primeng-style-id="global-style">...</style>`

Nonce gets set in providePrimeNG like this:
`csp: {nonce: 'angularNonce'}`

Nonce is correctly set in the ngcspnonce attribute of app-root (server is nginx), but all other nonce values are empty strings.
Locally the nonces are set to the same default value the ngcspnonce has, but in production they are empty.

Angular version
`19.0.1`

PrimeNG version
`19.0.0-rc.1`

tom-vism:
> Wondering if I have the same issue, but in my case, the nonces are empty strings like so: `<style type="text/css" nonce="" data-primeng-style-id="global-style">...</style>`
> 
> Nonce gets set in providePrimeNG like this: `csp: {nonce: 'angularNonce'}`
> 
[...]

PR #17061, opened 2024-12-16
Fix CSP nonce undefined when loading common styles

Qui-mey:
Fixes: #17012 & #17048


mertsincan:
Thanks a lot for your contribution! But, fixed in https://github.com/primefaces/primeng/commit/cb8449b236f8065ce84c81c0a5fc80836efe83bb

PR #17079, opened 2024-12-17
fix: handle 'variant' property in buttonClass

### LINKED ISSUE #17078: Component: Button "variant" attribute not honored in v18
### Describe the bug

On the demo page for v18 the Text and Raised Text buttons don't use the variant attribute properly. [text]="true" still works but variant="text" does not. Variant also is not working for Outlined buttons (variant="outlined").

I've tried the variant in v19-rc1 as well and it is not working in that version either.

### Environment

V18 Demo site using Brave.

### Reproducer

https://primeng.org/button#text

### Angular version

18

### PrimeNG version

18

### Build / Runtime

Angular CLI App

### Language

TypeScript

Unknown

### Browser(s)

Brave

1. Visit https://primeng.org/button#text
2. See buttons with backgrounds.

### Expected behavior

1. Expected buttons not to have backgrounds.

navedqb:
Fixes #17078


mertsincan:

PR #17092, opened 2024-12-17
fix: useStyle when id undefined

### LINKED ISSUE #17091: useStyle: matching document.getElementById(id) when id is undefined could break css injection
### Describe the bug

In useStyle.ts there is this code: 
```ts
styleRef = this.document.querySelector(`style[data-primeng-style-id="${name}"]`) || this.document.getElementById(id) || this.document.createElement('style');
```
when we have not already injected a specific css (for example inputotp-variables) `id` is `undefined` and `this.document.getElementById(id)` could match an element that has `id="undefined"` for example `<span id="undefined">content</span>` causing a style issue for missing css.
It is very uncommon (and ugly) to create an id with value "undefined" but this isn't forbidden by html rules.

### Environment

Angular v18 webapp

### Reproducer

https://stackblitz.com/edit/github-xypllzpu

### Angular version

18.2.2

### PrimeNG version

18.0.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

22.11.0

### Browser(s)

_No response_

1. Just run the stackblitz repo

### Expected behavior

`id="undefined"` must not break style injection

luca-peruzzo:
Fixes #17091

### DIFF
--- a/packages/primeng/src/usestyle/usestyle.ts
+++ b/packages/primeng/src/usestyle/usestyle.ts
@@ -16,7 +16,7 @@ export class UseStyle {
         const { immediate = true, manual = false, name = `style_${++_id}`, id = undefined, media = undefined, nonce = undefined, first = false, props = {} } = options;
 
         if (!this.document) return;
-        styleRef = this.document.querySelector(`style[data-primeng-style-id="${name}"]`) || this.document.getElementById(id) || this.document.createElement('style');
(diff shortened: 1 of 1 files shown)


mertsincan:
Sorry, but I didn't understand this fix. 
```
document.getElementById(undefined) -> output: null 
````
So, jumps to the next check

mertsincan:
Ahh ok. I checked your attached stackblitz now. Thanks a lot for your contribution!

Issue #17132, opened 2024-12-19
Component: Tree


AlexHeidinger:
### Describe the bug

When using Tree with ng-template, the a-tag href link does not work. Either in my development environment nor on the documentation page of primeNG v18, 19

### Environment

Windows 11, using primeNG 19 with NX and Angular 19

### Angular version

19.0.0

### PrimeNG version

19.0.0

### Build / Runtime

TypeScript

### Language

TypeScript

### Node version (for AoT issues node --version)

22.12.0

### Browser(s)

Chrome

### Steps to reproduce the behavior

Just use the recommendes Code from the Template Section of the Tree Component Doc-page (https://primeng.org/tree#template)

### Expected behavior

Clicking the link, a new tab opens with the defined link the data defined in the treeNodeArray

PR #17153, opened 2024-12-19
#17132 Fix link would not work in Tree 

qwadrox:
#17132 
Move prevent default closer to the event's origin (checkbox). 
Was added here: 
https://github.com/qwadrox/primeng/commit/0f615c4bb2879213404115d34ab50d5001eae8d7


mertsincan:

Issue #17130, opened 2024-12-19
Component: P-dropdown has issue while removing items.Scroll is not happening Properly


acharugulla:
### Describe the bug

Primeng dropdown scroll is buggy, when user removes value in editable dropdown. It doesn't scroll to correct value

### Environment

Angular 

### Angular version

17

### PrimeNG version

17

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20

### Browser(s)

Chrome,Safari , Edge

### Steps to reproduce the behavior

Assume you have editable prime ng drop down which has values from 1 to 500. User enters a value in editable dropdown as 300. The dropdown opens and scrolls to 300. But when user erases value (assume 0), then value will be 30, dropdown should scroll to 30 
and its not happening

I tried implementing logic like this  which is working
https://stackblitz.com/edit/ejxbx3fy?file=src%2Fapp%2Fdropdown-editable-demo.ts,src%2Fapp%2Fdropdown-editable-demo.html
but when i have appendTo Body. it doesn't work

This needs to be fixed

### Expected behavior

It should scroll to 30 when user removes 0 in that

acharugulla:
I tried implementing logic like this  which is working
https://stackblitz.com/edit/ejxbx3fy?file=src%2Fapp%2Fdropdown-editable-demo.ts,src%2Fapp%2Fdropdown-editable-demo.html


but when i have appendTo Body. it doesn't work 

This needs to be fixed

qwadrox:
@acharugulla 
Hi. I tried your example. I type 40000 and then delete a 0 so it becomes 4000 it scrolls there then I remove another 0, and it goes to 400. 

https://github.com/user-attachments/assets/2315548e-20a3-48d8-beb9-67b0b3a45c6c

acharugulla:
Hi @qwadrox 
It is working because i have written the code for scroll

I am calling scroll function each time , based on some conditions. It should work automatically. My org uses 1000 of p-dropdown
I cant call scroll methods in each p dropdown 

You this issue needs to be fixed

acharugulla:
I have provided you a solution 
You can try with basic example without my code

acharugulla:
@qwadrox can you please make this feature available in primeng 17 version
^17.18.8 or below 

So, that this feature gets added to our code  without bumping any versions of primeng

qwadrox:
> So, that this feature gets added to our code without bumping any versions of primeng 

[...]

PR #17168, opened 2024-12-21
#17130 Dropdown: Fix search

qwadrox:
#17130
The goal was to ensure that it is possible to search backwards
https://github.com/user-attachments/assets/7ffb481d-e698-475c-a834-07965d3a5787


mertsincan:

PR #17175, opened 2024-12-22
Fix: InputNumber(#17169) InputNumber marked dirty on blur

### LINKED ISSUE #17169: On focus in p-inputNumber it marks form dirty even value is unchanged
### Describe the bug

I'm using p-inputNumber field and there is a validation that 'save' button should enable when form is dirty. Issue is when I click in p-InputNumber field and tabout or focus out it marks form as dirty even value is unchanged.

image

### Pull Request Link

_No response_

- [ ] Lack of time
- [X] Unsure how to implement the fix/feature
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/github-pteddaso?file=src%2Fapp%2Fapp.component.html

### Environment

Angular 15, Vs code

### Angular version

15.0.0

### PrimeNG version

v18 (LTS Only)

### Node version

_No response_

### Browser(s)

chrome, edge

1. click on name input and click on "Save" it's disable because form is not dirty. It became dirty only when value changed.
2. click on age than click "Save" button, it's enable now because form marked as dirty even value is not changed. this is the issue of primeNg.

### Expected behavior

form should not mark as dirty until value gets changed. It should only be dirty when value get changed

mark7p:
Closes #17169

**Problem:** Input is marked dirty on blur. This is caused by `writeValue()` writing `this.value` as `string`.

**Solution:** Parsed value first before overwriting.

### DIFF
--- a/packages/primeng/src/inputnumber/inputnumber.ts
+++ b/packages/primeng/src/inputnumber/inputnumber.ts
@@ -1508,7 +1508,7 @@ export class InputNumber extends BaseComponent implements OnInit, AfterContentIn
     }
 
     writeValue(value: any): void {
-        this.value = value;
+        this.value = value ? this.parseValue(value.toString()) : value;
         this.cd.markForCheck();
     }
 


mertsincan:
Thank you for your contribution! Currently, we’re not accepting PRs for v17 and v18. Could you please create a PR for the master branch instead?

mark7p:
@mertsincan i just created new PR #17213 for master. thanks!

Issue #17181, opened 2024-12-23
v18 / v19: Documentation for tag using property that is deprecated since v15


sandrotonon:
### Describe the bug

The documentation for PrimeNG **18.0.2 and 19.0.1** shows to use the icon property of p-tag to add an icon. However, this property has been deprecated since 15.4.2.

However, since the deprecation was removed from this property in PrimeNG version 17.18.14 (see #16220) i would also expect this to be removed in 18.0.X and 19.0.X.

![Documentation](https://github.com/user-attachments/assets/6fadce88-50f0-419e-8a79-41595de8cc58)
![IntelliJ](https://github.com/user-attachments/assets/db81e66e-dee7-465e-865c-8ff3d3dc510a)

v19: https://github.com/primefaces/primeng/blob/011989e31448f4b91e17d9a62e31dcfa727f9cbd/packages/primeng/src/tag/tag.ts#L63
v18: https://github.com/primefaces/primeng/blob/aaef4d94aabcbdbc58e0d523a52f23ae05660810/packages/primeng/src/tag/tag.ts#L63

### Pull Request Link

v19: https://github.com/primefaces/primeng/pull/17182

### Reproducer

https://stackblitz.com/edit/github-kdxhm6nf?file=src%2Fapp%2Fapp.component.html

### Environment

Irrelevant

### Angular version

18.2.13

### PrimeNG version

v18 (LTS Only)

### Node version

20.14.0

### Browser(s)

All

### Steps to reproduce the behavior

No Response

### Expected behavior

No Response

mehmetcetin01140:
Hi, the icon property is not deprecated, the documentation has been updated. Thank you for your feedback.

Issue #16220, opened 2024-08-13
Documentation: Component: Tag using property that is deprecated since v15


jpgoelz:
### Describe the bug

The documentation for PrimeNG 17,18,8 shows to use the `icon` property of `p-tag` to add an icon. However, this property has been deprecated since 15.4.2.

The deprecation warning tells us how to replace this, but I think, the current documentation should reflect the correct way to use things. v15.4.2 came out over a year ago in May.

![image](https://github.com/user-attachments/assets/14438f48-994a-4ebb-9f29-38b34ade1fd8)

![image](https://github.com/user-attachments/assets/626cb90d-5e48-4bdd-81b7-b6a93d9b71db)

### Environment

irrelevant

### Angular version

irrelevant

### PrimeNG version

17.18.8

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

irrelevant

mehmetcetin01140:
Hi, the icon property is not deprecated, the documentation has been updated. Thank you for your feedback.

PR #17182, opened 2024-12-23
Fix: Tag v19 (#17181): Remove deprecation warning from icon-property of TagComponent

sandrotonon:
### Defect Fixes
Removes the deprecation warning from the icon property in TagComponent. Since it seems not to be deprecated (see #16220)

This is the fix for v19.


mertsincan:
Thanks a lot!

Issue #17220, opened 2024-12-26
Select: Editable Dropdown search not working as expected.


akshayaqburst:
### Describe the bug

Select: Editable Dropdown search not working as expected.

In the video attached, when I search for 225, it highlights 225000 instead of 225. In case of a perfect match, I expect it to highlight that value, else it can search with other criterias.`

https://github.com/user-attachments/assets/778e6ed1-5c2c-47ac-9bb9-74d1aba93c3c

` The attached video is from primereact but this issue exist in primeng as well.

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/eimkg68s?file=src%2Fapp%2Fselect-editable-demo.ts

### Environment

System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 4.10 GB / 7.47 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.21.3 - ~/.nvm/versions/node/v14.21.3/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v14.21.3/bin/yarn
    npm: 6.14.18 - ~/.nvm/versions/node/v14.21.3/bin/npm
  npmPackages:
    react: 18.3.1 => 18.3.1

### Angular version

17.x.x

### PrimeNG version

v19

### Node version

18.x.x

### Steps to reproduce the behavior

Go to https://stackblitz.com/edit/eimkg68s?file=src%2Fapp%2Fselect-editable-demo.ts
Search for 225.
'225000' is highlighted instead of 225.

### Expected behavior

Search should check for an exact match and else it should check for startsWith.

PR #17221, opened 2024-12-26
fix: #17220, Select: Editable Dropdown search not working as expected.

akshayaqburst:
fixes https://github.com/primefaces/primeng/issues/17220


cetincakiroglu:
Hi,

Good catch! Thanks a lot for the effort and support!

PR #17236, opened 2024-12-27
docs(dynamicdialog): update documentation to include inputValues support

nesgarbo:
Updated the DynamicDialog documentation to include information about the inputValues property. Clarified its usage alongside the existing data property, explaining their differences and how they can be used together. Added in code example both properties in the DynamicDialog implementation.


mertsincan:

PR #17251, opened 2024-12-30
docs: fix `rowexpansion` template name in showcase code

lezram:
Adjusting showcase code to use `rowexpansion` instead of `expandedrow`, according to the documentation.

> [...] To use this feature, add a template named **rowexpansion** and use [...]

### Defect Fixes
Not really a "fix" as one word change in the showcase.


mertsincan:

PR #17269, opened 2025-01-01
fix(tablist): tabindex now correctly set from unwrapped computed signal

jase88:
fixes https://github.com/primefaces/primeng/issues/17270


mertsincan:

PR #17289, opened 2025-01-03
fix: #17288 MultiSelect chip sizes

### LINKED ISSUE #17288: MultiSelect: display=chip sizes
### Describe the bug

A MultiSelect with Chips in the size small or large increases its height when a value is selected, compared to when it's empty.

### Pull Request Link

#17289

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/stackblitz-starters-3vqphm7e

### Environment

Windows 10
Firefox 128.3.1esr

### Angular version

19.0.5

### PrimeNG version

v19

### Node version

_No response_

### Browser(s)

_No response_

1. Add a MultiSelect with `display="chip"` and `size="small"` or `size="large"`
2. Select a value
3. Observe an increase in the components height

### Expected behavior

MultiSelect should be displayed with a consistent height, irregartles of its selection state.

honboubao:
### Defect Fixes
Fixes #17288

### DIFF
--- a/packages/primeng/src/multiselect/style/multiselectstyle.ts
+++ b/packages/primeng/src/multiselect/style/multiselectstyle.ts
@@ -207,6 +207,15 @@ p-multiselect.ng-invalid.ng-dirty .p-multiselect-label.p-placeholder {
     padding-inline: ${dt('multiselect.sm.padding.x')};
 }
 
+.p-multiselect-sm .p-multiselect-label .p-chip {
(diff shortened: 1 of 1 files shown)


cagataycivici:
We'll resolve it later once PrimeNG moves to the shared theme module. Currently this combination is not supported. Also we'll do it for AutoComplete in Multiple mode, which is the same.

https://github.com/primefaces/primeuix/tree/main/packages/themes

19.1.0 should fix it, will verify then.

PR #17343, opened 2025-01-10
fix: #16404 InputOtp: Unclear behavior when editing the filled component

### LINKED ISSUE #16404: InputOtp: Unclear behavior when editing the filled component
### Discussed in https://github.com/orgs/primefaces/discussions/2713

<div type='discussions-op-text'>

<sup>Originally posted by **rishanaznin00** September 19, 2024</sup>
We need to press the delete or back button if we want to change a value in the OTP field. Though the input text is highlighted and gives the user the impression that it will be updated if they press any key, that's not what happens.
</div>

### Environment

-

### Reproducer

_No response_

### Angular version

18

### PrimeNG version

17

### Build / Runtime

Angular CLI App

### Language

TypeScript

-

### Browser(s)

_No response_

_No response_

### Expected behavior

When a field is highlighted/focused, the field should be editable and pass the focus to the next field after edit. 
Or the behavior could be something similar: https://ui.shadcn.com/docs/components/input-otp
Once the OTP is filled, the small inputs are not focusable. Only as whole can be edited at the next attempt.

gianluigitrontini:
This PR removes a line that previously prevented default input behavior when the input reached its maxLength. The restriction allowed only the "Backspace" button to function, requiring users to delete existing values before entering new ones. With this change, users can overwrite the value in each input slot directly, improving usability.

The change is very small and basically removes a line in the onKeyDown method. However if the change is too harsh, would love to learn more.

### Defect Fixes
Allow editing inputotp when maxLength is reached, as stated in Issue #16404

### DIFF
--- a/packages/primeng/src/inputotp/inputotp.ts
+++ b/packages/primeng/src/inputotp/inputotp.ts
@@ -378,7 +378,7 @@ export class InputOtp extends BaseComponent implements AfterContentInit {
                 break;
 
             default:
-                if ((this.integerOnly && !((event.code.startsWith('Digit') || event.code.startsWith('Numpad')) && Number(event.key) >= 0 && Number(event.key) <= 9)) || (this.tokens.join('').length >= this.length && event.code !== 'Delete')) {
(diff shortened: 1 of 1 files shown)


mertsincan:
Could you please resolve conflicts? Then we'll review it. Thanks a lot!

Issue #17109, opened 2024-12-18
Component: p-inputNumber


Xample:
### Describe the bug

When I set a value to a FormControl bound on a p-inputNumber within the same tick where the FormControl has been changed, the p-inputNumber does not display the change. The displayed value is normally the [formattedValue bound on the internal "input" tag. ](https://github.com/primefaces/primeng/blob/master/packages/primeng/src/inputnumber/inputnumber.ts#L60C1-L61C1) for some reason, the [markForCheck ](https://github.com/primefaces/primeng/blob/master/packages/primeng/src/inputnumber/inputnumber.ts#L1512)has no effect

A silly (bad) workaround is to patch the value within a setTimeout() i.e.

```typescript
                    setTimeout(
                        () => form.patchValue(adaptedInvestmentAmounts),
                        0
                    )
```

### Environment

develop, check the reproducer link

### Reproducer

https://stackblitz.com/edit/github-8l9ppt9w?file=src%2Fapp%2Fapp.component.ts

### Angular version

18.0.1

### PrimeNG version

18.0.0-rc.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

v18.20.4

### Browser(s)

Chrome

### Steps to reproduce the behavior

1. open https://stackblitz.com/edit/github-8l9ppt9w?file=src%2Fapp%2Fapp.component.ts
2. set any value in the p-inputNumber component, the value will be set to 42 as expected
3. try setting another value, while it should still be set to 42, it won't. The entered number will be displayed instead
4. Try doing the same on the <input> tag, you will get the expected behaviour i.e. unable to set anything but 42.

### Expected behavior

The component should always be in sync with the formControl's value i.e. is should never display a different value than the current formControl's one.

PR #17347, opened 2025-01-10
fix(#17109): sync p-inputnumber with formcontrol patchvalue

ANANTH4004:
This PR addresses [issue #17109](https://github.com/primefaces/primeng/issues/17109), which reports a bug with p-inputNumber not syncing correctly with FormControl when using the patchValue method.

Root Cause
The p-inputNumber component was not updating its internal value when the patchValue method was used on the associated FormControl. This issue caused the UI to remain out of sync with the form control.

Fix
Updated the p-inputNumber component to listen to FormControl value changes and properly update its state whenever patchValue is called.
Testing
Manually tested the fix with a reactive form setup.
Verified that calling patchValue correctly updates the value displayed by p-inputNumber.
Closes [#17109](https://github.com/primefaces/primeng/issues/17109).


mertsincan:
Could you please resolve conflicts in PR? Then, we'll review it. Thanks a lot!

ANANTH4004:
Conflicts are resolved Please review

Issue #17309, opened 2025-01-07
Icon avatars are not displayed correctly


mertaksoy:
### Describe the bug

[Icon avatars](https://primeng.org/avatar#icon) are not displayed correctly

<img width="1067" alt="image" src="https://github.com/user-attachments/assets/50e6ca60-e12f-461d-94cb-3de9d62e5fb8" />

### Pull Request Link

https://github.com/primefaces/primeng/pull/17308

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://wr6mm3b5-jyi8ky8y.stackblitz.io

### Environment

angular19

### Angular version

19

### PrimeNG version

v19

### Node version

22

### Browser(s)

Chrome

### Steps to reproduce the behavior

Just go to the demo https://primeng.org/avatar#icon

### Expected behavior

Icons in an avatar is displayed centered

PR #17348, opened 2025-01-10
fix(#17309): set width and height for xl icon only avatars

Bogatejc:
This pull request fixes [#17309](https://github.com/primefaces/primeng/issues/17309), which reports a bug where xlarge icon avatars are not centered within the component.

## Root Cause
The `.p-avatar-xl .p-avatar-icon` class is missing the `width` and `height` properties, unlike the styles for large icon avatars.

## Fix
This PR implements the following solution:
```
.p-avatar-xl .p-avatar-icon {
    font-size: ${dt('avatar.xl.font.size')};
    width: ${dt('avatar.lx.icon.size')};
    height: ${dt('avatar.lx.icon.size')};
}
```

## Alternative Solution
An alternative approach would be adding `display: contents` to `.p-avatar-icon`. This would eliminate the need to define `width` and `height` for each size group. However, the implemented solution was chosen for its consistency with the existing styling patterns.

Closes [#17309](https://github.com/primefaces/primeng/issues/17309).

### DIFF
--- a/packages/primeng/src/avatar/style/avatarstyle.ts
+++ b/packages/primeng/src/avatar/style/avatarstyle.ts
@@ -57,6 +57,8 @@ const theme = ({ dt }) => `
 
 .p-avatar-xl .p-avatar-icon {
     font-size: ${dt('avatar.xl.font.size')};
+    width: ${dt('avatar.lx.icon.size')};
+    height: ${dt('avatar.lx.icon.size')};
 }
 
 .p-avatar-group {


cagataycivici:
Merged manually, thanks.

DaSchTour:
@cagataycivici could you at least cherry pick the commits? For me it looks really strange that PRs and commits from the community are pushed to the repo under a different author.

Issue #17314, opened 2025-01-07
p-inputnumber doesn't initialize its value correctly


realpascuu:
### Describe the bug

Input number doesn't initialize its value correctly.

![image](https://github.com/user-attachments/assets/ced5aa70-58cf-4a89-a627-60a1c324ed24)

### Reason for not contributing a PR

- [X] Lack of time
- [X] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/github-bnbvqvg7?file=src%2Fapp%2Fapp.component.ts

### Environment

SO: Ubuntu 24.02.1
Locale: es-ES

### Angular version

19.0.5

### PrimeNG version

v19

### Node version

22.12.0

### Browser(s)

Chrome 131

### Steps to reproduce the behavior

1. Initialize component with a decimal value.
2. The input displays the value without decimal.

Example: if you initializes with 23.3, the input displays 233

### Expected behavior

- Display the correct decimal value.
- Possibility of write decimal values with minFractionDigits = 0.

imaksp:
Looks like some locale specific bug.

UrosBgd:
I can also confirm this bug appears for 'sr' locale, but it's working fine for default one. I suspect it doesn't work for locales using (,) as decimal separator instead of (.), which is actually the most of the Europe.

Steps to reproduce

- Open the first example from InputNumber docs (https://primeng.org/inputnumber) in Stackblitz.
- Import 'sr' locale in `main.ts` file `import '@angular/common/locales/global/sr';`
- Provide 'sr' locale in `main.ts` file `{ provide: LOCALE_ID, useValue: 'sr' }`
- Update third control html, set `locale="sr"`
- Update value3 to `10.1`
- Expected value `10.10`, observed value `101,00`

I can confirm this worked in v17, since I noticed it because of the failing unit test after migrating to v19.

I would also like to emphasize the importance of this issue, since this is the core feature in any financial related software.

StefaniOSApps:
same .. v19

davidda:
PR #17370 created

wertzui:
Is there any ETA when we can see this fix being released?
It is a real blocker for us.

davidda:
It will be fixed with 19.0.6 but I don't know when this will come out. In the meantime, we are using the patch-package library with this patch:
[...]

PR #17370, opened 2025-01-13
fix: #17314 -  inputnumber model binding issue

davidda:
writeValue() must not parse the model value as it is expected to be a number already. Otherwise, all kinds of locale related issues may arise.


mertsincan:

davidda:
@mertsincan I've created a follow-up PR because I found a regression during testing: #17400

Issue #16806, opened 2024-11-21
Add lazy load support to tab component


mehmetcetin01140:

NiBurhe:
Is this feature already available in v18?

Because documentations says

`lazy | InputSignalWithTransform<boolean, unknown> | false | When enabled, hidden tabs are not rendered at all. Defaults to false that hides tabs with css.`

Adrii77:
I made a [PR to add this](https://github.com/primefaces/primeng/pull/17388), but no feedback from admin team since

rostislavcz:
There is already primeng20, in your documentation is that by default lazy is false, but in fact, lazy is always on, because you totally ignore that... I mean how difficult can be add one more condition to tabpanel component to render content when lazy is false?

@Adrii77 literally made the changes to work both lazy and non lazy and you just cant merge it? I know this is opensource project, but cmon... someone do the job and after 5 months its not in production....

cagataycivici:
Duplicate #17935

Issue #17351, opened 2025-01-10
p-tabs : lazy input doesn't seem to work


Adrii77:
### Describe the bug

I don't really know if this is a bug or a misunderstanding on my part.

Before Prime v18, lazy tab were built with the help of ng-template :

```
<p-tabView>
    <p-tabPanel header="Header 2">
        <ng-template pTemplate="content"> 
             <my-complex-component />  <!-- ngOnInit from this component will only be called when tabPanel will be active -->
         </ng-template>
    </p-tabPanel>
<p-tabView>
```

Since v18, with the new tabs component, this doesn't seem to work anymore, even with the use of lazy input.

```
<p-tabs scrollable lazy value="0">
    <p-tablist>
        <p-tab value="0">tab 0</p-tab>
        <p-tab value="1">tab 1</p-tab>
    </p-tablist>
    <p-tabpanels>
         <p-tabpanel value="0">Tralala</p-tabpanel>
         <p-tabpanel value="1">
             <my-complex-component />  <!-- ngOnInit from this component is called, even if this tab isn't active... -->
         </p-tabpanel>
    </p-tabpanels>
<p-tabs>
```

I tried to encapsulate `<my-complex-component />` between ng-template, not working too.

### Reason for not contributing a PR

- [X] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/gwaxjvse?file=src%2Fapp%2Ftabs-basic-demo.html

### Environment

 "@angular/animations": "^18.2.8",
    "@angular/cdk": "^18.2.9",
    "@angular/common": "^18.2.8",
    "@angular/compiler": "^18.2.8",
    "@angular/core": "^18.2.8",
    "@angular/forms": "^18.2.8",
    "@angular/platform-browser": "^18.2.8",
    "@angular/platform-browser-dynamic": "^18.2.8",
    "@angular/router": "^18.2.8",
    "@angular/service-worker": "^18.2.8",
    "@microsoft/applicationinsights-angularplugin-js": "^15.3.4",
    "@microsoft/applicationinsights-web": "^3.3.4",
    "@microsoft/signalr": "^8.0.7",
    "@primeng/themes": "^18.0.1",
    "angular-auth-oidc-client": "^18.0.2",
    "ngx-infinite-scroll": "^18.0.0",
    "primeflex": "^3.3.1",
    "primeicons": "^7.0.0",
[...]

PR #17388, opened 2025-01-14
feat(tabs): add lazy loading for tabpanel content

Adrii77:
### Defect Fixes
https://github.com/primefaces/primeng/issues/16806
https://github.com/primefaces/primeng/issues/17351

This PR add lazy loading for tabpanel's projected content.

Inactive tabs are hidden`([hidden]="!active())"`.
`instantiated` signal is used to render content the first time the tab is active and prevent re-initalization on reselection. Maybe a `linkedSignal` could be used, but I write the code the be compatible with NG 18.
It could be really cool if this PR could be merged and available in a patch version on 18, since apps that migrate to Prime 18 end up with complex components instantiated on inactive tabs... (this is my case)


ArnaudFelixParis (approved):
LGTM

Adrii77:
Hi @mertsincan ,

Thank you for your answer and for your time :) 

My opinion : 

1) Provide an `ng-template` is already a self decision : you want a tab with lazy loading ? just project your content inside an `ng-template`. You want another tab without lazy loading ? don't add `ng-template`.
2) Smaller and simpler API for components, as we don't need to add an extra input for this feature.
3) Less code and complexity inside `Tabs` code than implementing this case via an input.
4) Easier migration : with the old `TabView` component, lazy loading was working with `ng-template`. After migration to the new `Tabs` component, devs has just to replace `pTemplate="content"` by `#content` from `ng-template` to keeping lazy loading.
5) I guess your proposal refers to a way to create this feature without providing an `ng-template`. If so, I don't think this is possible, because the projected content depends on the parent component's life cycle : content will be initialized before being projected (whereas with an `ng-template`, this wouldn't be the case as they're [rendered and initiliazed only when we instruct it to do](https://angular.dev/api/core/ng-template))
6) Today, applications that migrate to Prime 18 find themselves with a feature regression and a drop in performance, since lazy loading no longer exists.  In my case, I have a page with 5 tabs  which two of them contains a complex stateful component (with http calls). Even if one of those tab isn't active, components are initialized and backend calls are triggered.
7) Same API as [Angular material tab](https://material.angular.dev/components/tabs/overview#lazy-loading)

With this PR, lazy loading is preserved and prevents this regression (working with my test, hope you'll test too !). I think in any case it would be wise and pragmatic to have lazy loading this way to prevent this feature regression and why not refactor later if you want to find a way to do it via inputs.

What do you think ? :)

Adrii77:
@mertsincan 

More explanation if tabpanel had a `lazy` input.

Having an `ng-template` with an `ng-content` inside `TabPanel` code without projected content inside `ng-template`, such as : 

```
@Component({
....,
template: `
    <p-tabpanel lazy>
       <!-- no ng-template here.... -->
        <p>Hey !</p>
        <complex-component />
        <p>poney</p>
    </p-tabpanel>
`
})
export class RandomComponent {
....
}

@Component({
....,
template: `
  @if (initialized()) {
    <ng-template>
        <ng-content />
    </ng-template>
  }
`
})
export class TabPanel {
....
}
```

_It's already too late : `RandomComponent` has already initialized content before beeing projected._

Encapsulte content provided from `RandomComponent` inside an `ng-template`, content will be [rendered and initiliazed only when we instruct it to do](https://angular.dev/api/core/ng-template), that's why content is not initialized until `ngTemplateOutlet` does it.

If there is a solution to that, this would, I think, be an unnecessary complexity.

Adrii77:
@mertsincan Up (sorry) :/

eran10:
guys, can this be merged? 
thanks in advance

Adrii77:
+1... I'm stuck on my migration to Angular 18 / PrimeNG 18 because of this regression.

Core121:
Also waiting for this fix

ahmedbettaiebdeveloper:
Same here ,i'm eagerly waiting for this change to be merged.The lazy loading regression on Angular 18 is causing significant issues, especially for pages with complex, stateful components.Please consider merging it as soon as possible. 
Thanks!

LoaderB0T:
After looking closer at your solution I definitely agree with your proposal (not that that means anything to anybody ^^)

The only thing that might cause problems for some is that content once rendered stays in the DOM. We often have quite large tabpanels with many elements and too many DOM elements might slow things down. Could you imagine a lazy mode that also removes the content from the DOM? Preferably configurable

LoaderB0T:
> @LoaderB0T thank you for your reply !
> 
> I've actually replicated the behavior of the old TabView component in Prime 17

If that's the case, it's fine for me ! :)

MatviiKharlamenko:
Also waiting for this fix

MaSch0212:
I tested this fix, as we are also facing the issues. I noticed one issue:
We are loading some data in the constructor of the component from a server. When I have this component directly (without ng-template) in the tab-panel, the server call is done twice. I guess the content is initialized twice because `<ng-content></ng-content>` is inside an `<ng-template><ng-template>`. I would suggest changing this to `@if` instead.
If I put my component inside an `ng-template`, everything works fine.

Adrii77:
Hello @MaSch0212,

Ty for your feedback ! :)

I try to reproduce your issue, but it works well on my side. Even without `ng-template`, my component's constructor is called just one time 🤔. 

Onetheless, I also tried replacing it with `@if / @else`, and it works as well while avoiding potential re-instantiation caused by `ng-template`, as you experienced.

I pushed it, keep me informed ! :)

MaSch0212:
@Adrii77 thanks for your quick response! I will try the updated version first thing tomorrow and will post feedback.

MaSch0212:
Hey, I tried the updated version today. Unfortunately it did not solve my issue.
To get a bit more into detail what my setup is: I am trying to recreate my component on every navigation. So I want to destroy my component if the tab is not active. I did it like this:
```
      <p-tabpanel #dataTypes [value]="designerRoutes.dataTypes">
        @if (dataTypes.active()) {
          <swpe-data-data-types></swpe-data-data-types>
        }
      </p-tabpanel>
```
This worked in PrimeNG 19.0.10 but with the change from this PR the constructor is called twice for some reason. I checked all the signals that are involved but they seem to update just once:
grafik

It also only happens the first time a tab is loaded. All subsequent loads calls the constructor just once.

As I said when wrapping the content in a `ng-template` it works fine. Not sure what is happening here.

Adrii77:
@MaSch0212 

I just added a new `cache` input that, when set to false, re-renders the lazy loaded content each time a tab becomes active.

For your code , you can try this : 

 ```
<p-tabpanel [cache]="false" [value]="designerRoutes.dataTypes">
       <ng-template>
          <swpe-data-data-types></swpe-data-data-types>
        </ng-template>
      </p-tabpanel> 
```

Keep me informed :)

**Comment by cetincakiroglu** [comment]
Hi @Adrii77,

Thanks a lot for the contribution, could you please make it work as below:

- Tab (parent) component should have `lazy` input. If set the true, all the tabs should load lazy.
- TabPanel component should have `lazy` input instead of `cache`. If set the true, only the related tab should load lazy.

After this change, we can merge it.

**Comment by Adrii77** [comment]
Hello @cetincakiroglu ,

Thank you for your answer.

I have made the desired changes.

- `Tabs` and `TabPanel` have `lazy` input
- By default, all tab's content are **initialized** AND **rendered**  (hidden with CSS).
- If `lazy` is enabled, tab's content are only **rendered**  when the tab become actives (and stay in the DOM if become inactive). Note that content will be initliazed (component(s) inside projected content will call `ngOnInit`) even if the tab has not been activated yet.
- If `lazy` is enabled AND tab's content is encapsulated inside **<ng-template #content></ng-template>** : content will be **initialized** AND **rendered**  when the tab become actives (and stay in the DOM if become inactive).

Capture d’écran 2025-09-01 à 20 18 42

**Review by cetincakiroglu — APPROVED** [review]

**Comment by devshack** [comment]
> * `Tabs` and `TabPanel` have `lazy` input
> 
> * By default, all tab's content are **initialized** AND **rendered**  (hidden with CSS).
> 
> * If `lazy` is enabled, tab's content are only **rendered**  when the tab become actives (and stay in the DOM if become inactive). Note that content will be initliazed (component(s) inside projected content will call `ngOnInit`) even if the tab has not been activated yet.
> 
> * If `lazy` is enabled AND tab's content is encapsulated inside **<ng-template #content>** : content will be **initialized** AND **rendered**  when the tab become actives (and stay in the DOM if become inactive).

With the current version of Prime, this doesn't seem to work as intended? Or it doesn't reflect the "old" way of PrimeNG 17 (where we could achieve this by using cache="false" per tab)?

E.g. i have have 2 tabs - each of them has a dedicated component that fetches some http data (in ngOnInit).

With lazy & <ng-template #content> enabled, I expect to see an http call everytime I switch between the panels. But it's not happening.

I was only able to "fix" this by "hardcoding" it to the active Tab value and an @if block:

```
...
          <p-tabpanel [value]="3">
            @if (activeTab() === 3) {
              <tab-reservation [customerId]="customerId()" />
            }
          </p-tabpanel>
```

Adrii77:
Hello @devshack 

Indeed, the current implementation doesn’t allow the behavior you’re looking for, because the tab content is never destroyed once it has been rendered. What’s missing is an input to manage caching of the tab content after it has been rendered.

In my initial implementation, I suggested restoring the previous behavior of the Tabs component, but I don’t think that’s what the core team wanted, judging by the latest message from @cetincakiroglu. I implemented lazy so that the tab content is rendered only when the tab becomes active, and not destroyed once it becomes inactive.

Even now, I’m still not sure what they expect regarding cache management. A bit more clarification from them would be very welcome—I’d be happy to submit a pull request in that case.

PR #17400, opened 2025-01-15
fix: issue from previous PR #17370

davidda:
Upon testing I encountered a regression for my merged PR #17370 for issue #17314.

Setting the model value to null or undefined produced incorrect behaviour (value displayed as 0 or NaN).


mertsincan:
Thanks a lot for your update!

Issue #17384, opened 2025-01-14
When p-select is editable and a default value already exists, selectOption displays object when the filter selects other options


LiJun0107:
### Describe the bug

When p-select is editable and a default value already exists, selectOption displays object when the filter selects other options,

### Reason for not contributing a PR

- [ ] Lack of time
- [X] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/5fi9n9nf?file=src%2Fapp%2Fselect-filter-demo.html

### Environment

primeng

### Angular version

19.0.1

### PrimeNG version

v19

### Node version

19.0.0

### Steps to reproduce the behavior

1. <p-select
    [options]="countries"
    [(ngModel)]="selectOption"
    optionLabel="name"
    [filter]="true"
    filterBy="name"
    [editable]="true"
    [showClear]="true"
    placeholder="Select a Country"
    class="w-full md:w-56"
  > 
Set editable and filter
2. Click the drop-down box and enter the following options in the filter
3.  selectOption displays objects abnormally

### Expected behavior

selectOption can display the data correctly

PR #17433, opened 2025-01-19
fix(p-select): selected option when editable and filter options are e…

Bock92:
Fixes: https://github.com/primefaces/primeng/issues/17384

Changes made to fix the Issue:

Prevent resetting the selectedOption when filterValue is filled or partially typed.
Cleared the filterValue (set to null) when the user select an option from the select list.


Bock92:
Changes made to fix the Issue:
- Prevent resetting the `selectedOption` when `filterValue` is filled or partially typed.
- Cleared the `filterValue`  (set to null) when the user select an option from the select list.

PR #17474, opened 2025-01-23
fix: pass filter from listbox and allow applying external filter as w…

### LINKED ISSUE #17523: Picklist filtering does not work
Filtering in picklist is not connected to listbox component.

DaSchTour:
Fixing 3 issues that are currently blocking implementation at one of my projects

1. Filter event is currently not working because it is not connected to the listbox
2. Filtering can not be done externally, for example through searching in a backend service
3. It is not possible to add pagination to the bottom of the list box

resolves #17523


cagataycivici:
We'll resolve with a different approach, thank you.

PR #17497, opened 2025-01-25
fix: make build scripts cross platform

davidda:
Some of the PrimeNG git repo's (p)npm scripts cannot be executed on Windows due to the way environment variables are used in the scripts.

The solution is to use the "cross-env" tool where required. Tested under Ubuntu and Windows.


mertsincan:

PR #17508, opened 2025-01-27
Fix 17016 Close dynamic dialog on escape also when autoZindex is set to false

### LINKED ISSUE #17016: DynamicDialog: closeOnEscape doesn't work without autoZIndex
### Describe the bug

When setting autoZIndex to false, closeOnEscape has no effect.

### Environment

Browser

### Reproducer

https://stackblitz.com/edit/github-xfkekgrl-nd2zj1hu?file=src%2Fapp%2Fapp.component.html

### Angular version

18.x.x

### PrimeNG version

18.0.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

22.10.0

### Browser(s)

Chrome

1. Click Button in provided reproducer -> The Dialog opens
2. Press ESC

### Expected behavior

The Dialog closes, but nothing happens

shaman-apprentice:
fixes #17016 

### How I tested it

I used the exampledoc.ts and added / toggled `autoZIndex: false` [here](https://github.com/primefaces/primeng/blob/master/apps/showcase/doc/dynamicdialog/exampledoc.ts#L124) to verify existing of bug and fix of bug.

### Implementation design

I used the `ZIndexUtils` as  a nice layer tracker. I exposed the generation and removale, so I have access to it, without having to apply `zIndex` to an element. That way, still only the most top layer gets closed on escape, no matter if zIndex was applied to the an element. I think my change is none-destructive.   

Please let me know, what you think.

### DIFF
--- a/packages/primeng/src/dynamicdialog/dynamicdialog-config.ts
+++ b/packages/primeng/src/dynamicdialog/dynamicdialog-config.ts
@@ -61,7 +61,7 @@ export class DynamicDialogConfig<DataType = any, InputValuesType extends Record<
      */
     baseZIndex?: number;
     /**
-     * Whether to automatically manage layering.
+     * Whether to re-enforce layering through applying zIndex.
      * @group Props
      */
     autoZIndex?: boolean = false;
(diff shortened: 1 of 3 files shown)


mertsincan:

PR #17534, opened 2025-01-30
fix(picklist): allow disabled options in picklist

### LINKED ISSUE #17533: Picklist does not support disabled options
Picklist currently does not support disabled options although listbox already has this functionality

DaSchTour:
resolves #17533


mertsincan:

Issue #17507, opened 2025-01-27
Version 19 Accordion Icon positioning


Tguthrie500:
I recently upgraded my project to version 19 of Primeng, and with it upgraded to the new Syntax for the p-accordion from AccordionTab to AccordionPanel. However, with this change I have noticed the position of the drop down icon has shifted from left to right in the app. The AccordionTab component had a parameter to determine the position of the dropdown, but I am noticed that there is not one for AccordionPanel. Is there an option to change the position in version 19?

bychaka:
Hi @cagataycivici , it warks hot really as you described. 
According to the source code of accordion icon position is always at the end of <p-accordion-header>
[https://github.com/mtrefzer/primeng/blob/bd011b8a8b9b5d490d7046f6b75cec88ed9e2994/packages/primeng/src/accordion/accordion.ts#L127](https://github.com/mtrefzer/primeng/blob/bd011b8a8b9b5d490d7046f6b75cec88ed9e2994/packages/primeng/src/accordion/accordion.ts#L127)
```
template: `
        <ng-content /> // first content then icon 
        @if (toggleicon) {
            <ng-template *ngTemplateOutlet="toggleicon; context: { active: active() }"></ng-template>
        } @else {
            <ng-container *ngIf="active()">
                <span *ngIf="pcAccordion.collapseIcon" [class]="pcAccordion.collapseIcon" [ngClass]="pcAccordion.iconClass" [attr.aria-hidden]="true"></span>
                <ChevronDownIcon *ngIf="!pcAccordion.collapseIcon" [ngClass]="pcAccordion.iconClass" [attr.aria-hidden]="true" />
            </ng-container>
            <ng-container *ngIf="!active()">
                <span *ngIf="pcAccordion.expandIcon" [class]="pcAccordion.expandIcon" [ngClass]="pcAccordion.iconClass" [attr.aria-hidden]="true"></span>
                <ChevronUpIcon *ngIf="!pcAccordion.expandIcon" [ngClass]="pcAccordion.iconClass" [attr.aria-hidden]="true" />
            </ng-container>
        }
    `,
```

you can ensure in my example - https://stackblitz.com/edit/u1j5sp4c-9vx9meqk?file=src%2Fapp%2Faccordion-template-demo.html

sure, it is possible to reorder elements by using css order, but why did you cancel this icon positioning feature? Clases in PrimeNG  sych as `p-accordionheader-toggle-icon icon-start` are still working. 
we, devs, can find this info only in git source files.

PR #17546, opened 2025-01-31
fix: #17507 accordion icon position

mtrefzer:
### New Feature
[#17507](https://github.com/primefaces/primeng/issues/17507)  accordion icon position

Allows to pass `iconPos` values `start` or `end` in `Accordion` component like in the deprecated `AccordionTab`.
The default value is `end`.


cagataycivici:
The new accordion uses content projection so you have full control over the presentation, so there is no need props like iconPos.
`
<p-accordion-header><IconHere /> Header I <OrIconHere/> </p-accordion-header>`

PR #17548, opened 2025-01-31
fix: xlsx remove unused vulnerable dependency

mtrefzer:
### Defect Fixes
The dependency "xlsx" is vulnerable and not used in the project, therefor it should be removed.


mertsincan:

Issue #17613, opened 2025-02-07
bug found at p-datepicker


VicTMPDeV:
### Describe the bug

Good morning everyone.

I think I have found a bug in the p-datepicker component:

I am working with an inline calendar with selectionMode="range" but applying different colors to the CSS classes .p-datepicker-day-selected and .p-datepicker-day-selected-range.

To reproduce the issue, you need to select a fairly wide range where the start and end day repeat numerically in intermediate months. Then, the following happens:

For example, if I select the date range from January 1st to March 28th, the bug is that the only days that should have the .p-datepicker-day-selected class applied are January 1st and March 28th, while the rest should have the .p-datepicker-day-selected-range class. However, this is not the case.

What actually happens (and I am almost certain it is a bug because it also occurs in the PrimeNG documentation example for datepicker range, but it is not noticeable unless you inspect it with devtools since the same color is applied) is that the days February 1st, March 1st, January 28th, and February 28th also get the .p-datepicker-day-selected class instead of .p-datepicker-day-selected-range, which should be the correct class.

### Reason for not contributing a PR

- [x] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/github-yjtcmdcv-jfzul4im?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts

### Environment

Angular 19.0.5
Primeng 19.0.5

### Angular version

Angular 19.0.5

### PrimeNG version

v19

### Steps to reproduce the behavior

![Image](https://github.com/user-attachments/assets/cbe82864-0c3d-4755-95be-78902037d309)

![Image](https://github.com/user-attachments/assets/f17a8978-cb2d-48aa-8d59-bf10bbcb5f5d)

![Image](https://github.com/user-attachments/assets/3835daa8-3512-4b93-8548-a5d78111fc3d)

### Expected behavior

Only apply the correct css class to the initial-final day of the range. For that it should be controlled the day-month-year combination, not only the day

Joebeurg:
Same issue here.
Angular 18.2.11
PrimeNG 18.0.2

PR #17632, opened 2025-02-10
fix(DatePicker): range selection incorrectly styles dates in range

Joebeurg:
The DatePicker component using range type, when selecting a range spanning multiple months, the component styles the first and last selected days for each month.

Fixes [#17613](https://github.com/primefaces/primeng/issues/17613)


piotrignat-vt:
Any chances this one will be reviewed anytime soon?

Issue #17323, opened 2025-01-07
PickList onTargetSelect and onSourceSelect not wroking with mouse selection


mastersys-giovanni:
### Describe the bug

PickList onTargetSelect and onSourceSelect not wroking with mouse selection

### Reason for not contributing a PR

- [X] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/github-at6aer6m?file=src%2Fapp%2Fapp.component.ts

### Environment

windows 12

### Angular version

19

### PrimeNG version

v19

### Steps to reproduce the behavior

1 - select item in target
2 - not dispare onTargetSelect

### Expected behavior

1 - select item in target
2 - dispare onTargetSelect

neorclynn:
It's not working on v18 as well.

mariachiaravolt:
What is the status of the bug? onTargetSelect and onSourceSelect events do not work.

yklepikovska-iqmessenger:
When will this bugfix be merged?

jansowinski:
Is there hope that this bugfix will be merged anytime soon?

PR #17638, opened 2025-02-11
Feature/picklist selection

infe42:
Fixes https://github.com/primefaces/primeng/issues/17323

This PR remove a lot of code from the PickList component that tries to deal with selection and focus - which is already done by the ListBox component itself. Also, some of this code was actually never used (like `onItemClick`).

After this PR the PickList component also emits events for selection changes again, which fixes the bug mentioned above.


BeterPeter:
It looks to me like @infe42 did a great job. When can we expect this fix to make it to production?

jokarmus:
What is the status of the issue?

infe42:
I've updated the PR to resolve merge conflicts and to comply with conventional commit rules.

PR #17640, opened 2025-02-11
fix: remove scrollHandler on unbind

### LINKED ISSUE #17782: p-menu in popup mode does not close on scroll in some cases.
### Describe the bug

Szenario:
you have 2 scrollable lists with items inside.
there is one menu with [popup]="true"
clicking on an item will trigger menu.show() and it will show some menu entries.

Steps that lead to the faulty behaviour:

- click on an item of the first scrollable list
- menu appears
- scroll on the first list
- menu disappears
- click on an item of the second scrollable list
- menu appears
- scroll on the second list
- !!!!!!!!!!! menu does not disappear !!!!!!!!!!
- scroll on the first list
- ????? menu disapperas ?????

The reason is that the scrollHandler still has the reference to the list item of the first list.
This is bad for two reasons:

It holds a reference to a potentially already deleted element (the first target)
menu will never disappear on scrolling of the second list anymore because bindScrollListener wants to bind to the scroll parents of the first element (which is on the first list)
The solution is to remove the scrollHandler member of Menu on unbindScrollListener
--> no more potential leaking of deleted elements
--> works now with differenet targets from different scrollareas passed to show();

### Pull Request Link

https://github.com/primefaces/primeng/pull/17640

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/github-wjmgbgjh?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.css,src%2Fapp%2Fapp.component.ts

### Environment

Windows

### Angular version

Angular 18

### PrimeNG version

v18 (LTS Only)

### Node version

_No response_

### Browser(s)

_No response_

- click on an item of the first scrollable list
- menu appears
- scroll on the first list
- menu disappears
- click on an item of the second scrollable list
- menu appears
- scroll on the second list
- !!!!!!!!!!! menu does not disappear !!!!!!!!!!
- scroll on the first list
- ????? menu disapperas ?????

### Expected behavior

Scroll should always hide the menu in popup mode

demike:
### Defect Fixes
fixes #17782 
Fixes a bug when using one menu for multiple targets in popup mode:
p-menu in popup mode does not close on scroll in some cases.

Szenario: 
you have 2 scrollable lists with items inside.
there is one menu with `[popup]="true"`
clicking on an item will trigger menu.show() and it will show some menu entries.

Steps that lead to the faulty behaviour:
- click on an item of the first scrollable list 
 - menu appears
 - scroll on the first list
 - menu disappears
 - click on an item of the second scrollable list
 - menu appears
 - scroll on the second list
 - !!!!!!!!!!! menu does not disappear !!!!!!!!!!
 - scroll on the first list
 - ????? menu disapperas ?????

The reason is that the scrollHandler still has the reference to the list item of the first list.
This is bad for two reasons:
- It holds a reference to a potentially already deleted element (the first target)
- menu will never disappear on scrolling of the second list anymore because bindScrollListener wants to bind to the scroll parents of the first element (which is on the first list)

The solution is to remove the `scrollHandler` member of `Menu` on unbindScrollListener
**--> no more potential leaking of deleted elements
--> works now with differenet targets from different scrollareas passed to show();**

### DIFF
--- a/packages/primeng/src/menu/menu.ts
+++ b/packages/primeng/src/menu/menu.ts
@@ -803,6 +803,7 @@ export class Menu extends BaseComponent implements AfterContentInit, OnDestroy {
     unbindScrollListener() {
         if (this.scrollHandler) {
             this.scrollHandler.unbindScrollListener();
+            this.scrollHandler = null;
         }
     }
 


mertsincan:
Could you please create an issue?

demike:
@mertsincan add an issue with reproduction

Issue #17569, opened 2025-02-04
Table: totalRecords has no effect


qwenger:
### Describe the bug

In a paginated table scenario where row data are first loaded on page change, the actual total number of entries might be different from the length of the `value` property. This seems to be what `totalRecords` is for.

However, it doesn't seem to have any effect: pagination is only activated if the length of `value` is greater than the `rows` attribute, regardless of `totalRecords`. (The documentation mentions that it "*defaults* to length of value when not defined", which implies that it could be set to something else.)

### Reason for not contributing a PR

- [x] Lack of time
- [x] Unsure how to implement the fix/feature
- [x] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/github-rbpxrszq?file=src%2Fapp%2Fapp.component.html

### Environment

primeng with angular in node in alpine in docker in arch in wsl2 in win10. (it doesn't matter but you asked.)

### Angular version

19.1.3

### PrimeNG version

v19

### Steps to reproduce the behavior

Create a paginated p-table where `data.length <= rows < totalRecords`.
Behavior: pagination is disabled (only page 1 is active).

### Expected behavior

Can move to page 2 and up to page number `(totalRecords - 1)//rows + 1`.

macx:
I have the same issue, using p-table with the pagination is useless currently to handle big data. Can we please force a hot fix for the next version ASAP?

ZeroSkill830:
Is there really still such a bug open? It's impossible to use server-side pagination.

macx:
Could we please add this to Version 20.0.0? This is not just a cosmetic error.

cagataycivici:
Let's review for v20 final and backport to v18, v19 if a change is required.

PR #17654, opened 2025-02-12
fix(table): total records logic

Bock92:
Fix: https://github.com/primefaces/primeng/issues/17569

### Changes in this fix:
- The `totalRecords` input property was not considered when calculating the total number of elements, regardless of whether the table was filtered or not.

### Implementation Details:
- The `_totalRecords` property now correctly represents the original value of `@Input() totalRecords`.
- When `p-table` calculates the total number of records, it will first check if `@Input() totalRecords` is a valid number.
- If `totalRecords` is not valid/set, the table will default to using `value.length` as the total record count.


Bock92:
@cetincakiroglu  do you know if this PR is still under review?

The related issue still appears to be unresolved, so if possible, it would be great to close this PR soon. Let me know if there's anything I can do to help move it forward.

CC: @cagataycivici

macx:
Tables are not really usable using an API. Guys, this is critical to be fixed.

CC: @cagataycivici

mertsincan:

PR #17661, opened 2025-02-12
fix(table): fix filterMatchModeOptions not being set in setConfig

CoolfulSoftware:
Fixes an issue where setting the filterMatchModeOptions via the PrimeNG config would not work because the value was never set internally.


CoolfulSoftware:
@cetincakiroglu The #17664 issue is already linked. Do I need to do anything else?

CoolfulSoftware:
Fixes #17664

Issue #17659, opened 2025-02-12
File-Upload: Improve and localize formatting of maxFileSize for error messages


gregorwinkler:
The file upload component shows an error if a selected file is larger than the `maxFileSize`.
Prior to PrimeNG 18, it was formatted like this: "Invalid file size, maximum upload size is 5 MB".
From PrimeNG 18 onwards, it is formatted as: "Invalid file size, maximum upload size is 5.000 MB"

While having three decimal places may be useful in special cases, I think for normal use cases it is more confusing than helping - at most one decimal place may make sense, for example 5.5 MB. The `formatSize()` function should adapt the decimal points depending on the necessity of the concrete number.
Another issue with that is that a dot is always used for indicating the decimal places which is the default in English. But in other languages such as German a comma should be used instead and a dot is generally understood as a separator for thousands (5.000 MB -> 5000 MB -> 5 GB, while 5,000 MB would be interpreted as 5 MB).

So, I'd propose changes for the fileUpload component as followed:
```js
formatSize(bytes: number) {
    const k = 1024;
    const sizes = this.getTranslation(TranslationKeys.FILE_SIZE_TYPES);

    if (bytes === 0) {
        return `0 ${sizes[0]}`;
    }

    const i = Math.floor(Math.log(bytes) / Math.log(k));
    const formattedSize = (bytes / Math.pow(k, i)).toLocaleString(locale, {
        minimumFractionDigits: 0,
        maximumFractionDigits: 3,
    });

    return `${formattedSize} ${sizes[i]}`;
}
```

`toLocaleString` provides the functionality described above in a simple way.
`locale` could be just an `@Input()` variable with the default being `en-US`, so that the formatting for languages can be defined based on the UIs language.
Further `minimumFractionDigits` and `maximumFractionDigits` could be input variables as well but are really not that important to be changed manually and could cause confusion as their purpose is super specific.

gregorwinkler:
Potential solution implemented in https://github.com/primefaces/primeng/pull/17662

gregorwinkler:
@cetincakiroglu Is there any chance this might be added soon? Thanks in advance!

PR #17662, opened 2025-02-12
feat(fileupload): add locale and improve formatSize functionality

gregorwinkler:
Implements: https://github.com/primefaces/primeng/issues/17659

`formatSize()` shows with that decimal points only if they are needed and only as many as needed.
Further, a locale can be defined so that the decimal is properly formatted depending on the language.


gregorwinkler:
This is still relevant and the corresponding issue has several upvotes (https://github.com/primefaces/primeng/issues/17659).

PR #17679, opened 2025-02-14
Fixes error when badge applied as directive do not work with custom css tokens from dt

majkers:
### Defect Fixes
When badge is applied to element with directive and You use dt to add custom css tokens, they are not working because attribute from super method is not added to root element

### Feature Requests
Now badge applied as directive is working the same way as applied as component


SoyDiego (approved):

majkers:
@cetincakiroglu why You didn't merge this one?

Issue #17683, opened 2025-02-14
Tree: No design token for selected color of node toggle button


infe42:
### Describe the bug

It's not possible to use a different color for the toggle button of the selected node, because there is no design token and the color is explicitly set to "inherit". There are design tokens for several other states (including selectedHoverColor), so there should also be a design token for selectedColor.

### Pull Request Link

https://github.com/primefaces/primeng/pull/17684

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://primeng.org/tree#multiple

### Environment

-

### Angular version

19.1

### PrimeNG version

v19

### Steps to reproduce the behavior

- open the example on https://primeng.org/tree#multiple
- select a node and inspect the DOM
- color is set to "inherit" and not using a CSS variable

### Expected behavior

There should be a design token for the selected color of the node toggle button.

cagataycivici:
There are no plans to add this token now, you may add it as a custom token instead with the extend feature of theming.

PR #17684, opened 2025-02-14
fix(Tree): Add design token "tree.node.toggle.button.selected.color"

infe42:
Fixes https://github.com/primefaces/primeng/issues/17683


cetincakiroglu:
Hi,

We'll switch to unified styles and tokens for all Prime UI Libraries in v19.1.0. With 19.1.0 update, all tokens will be removed from the PrimeNG and will be taken from unified token library (primeuix). Because of that, we don't accept PR's that contains preset updates. Even though we accept it, tokens you've added will be removed in 19.1.0 during the unification process. Thanks for the effort and support, sadly I have to close this one.

Issue #17635, opened 2025-02-10
What happend with calender touchUI


jjs98:
I updated from v17 to v19 and suddenly my datepicker touchUI is not working anymore.

In the docu its not explicit shown with an example, but it is still in the api docu.

before:
![Image](https://github.com/user-attachments/assets/5e58314d-a8d2-4f67-b077-7bdffe7cbfab)

after:
![Image](https://github.com/user-attachments/assets/1859e2f6-62ba-43ec-b7fc-23a54a710a69)

My usage inside of a p-dialog:
````html
<p-floatLabel class="w-64">
  <p-datepicker
    id="readingDate"
    [(ngModel)]="readingDate"
    [touchUI]="true"
    [showIcon]="true"
    [showOnFocus]="false"
    [maxDate]="currentDate"
    dateFormat="dd.mm.yy"
  />
  <label for="readingDate" class="font-semibold w-6rem">{{
    translations.reading_readingDate()
  }}</label>
</p-floatLabel>
```

davesellersBG:
I've also noticed this. Looks like the `p-datepicker-touch-ui` class is not getting applied.

cagataycivici:
Unfortunately this feature has been removed after migration to new theme engine. Instead we're considering to make this a built-in feature where in mobile, DatePicker detects screen and opens in optimized mode.

PR #17689, opened 2025-02-15
fix: datepicker touchui not displayed correctly #17635

jjs98:
#17635

## What was the issue?

The problem was that the position absolute was not correct. I fixed it by using the p-datepicker-touchui class and only adding it when its actually an touchui datepicker. Otherwise the position: absolute will be used.

Also I added the touchUI back to the showcase. I added a little extra info for usage inside of p-dialog. The [appendto]="'body'" is needed because the overlay-mask is added with zIndex=1103 on the root of body. If the datepicker is not on root of body as well it will be behind the overlay-mask.

Maybe some day I have the time to find a solution that this component is smart enough to know where it should append itself.

PS. I hope the formatting is okay. I ran 'pnpm format'

### DIFF
--- a/apps/showcase/doc/datepicker/touchuidoc.ts
+++ b/apps/showcase/doc/datepicker/touchuidoc.ts
@@ -7,6 +7,7 @@ import { Component } from '@angular/core';
     template: `
         <app-docsectiontext>
             <p>When <i>touchUI</i> is enabled, overlay is displayed as optimized for touch devices.</p>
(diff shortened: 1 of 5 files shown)


imaksp:
Hi, this removal could be intentional , as its already been removed from Vue version: https://github.com/primefaces/primevue/issues/5553

mountpoint:
@cetincakiroglu the issue with centering is still in primeng 20.1.1. I use Angular 20.2.3

I left comment before https://github.com/primefaces/primeng/issues/17842#issuecomment-2714175532

And also there is no example of usage on website https://primeng.org/datepicker, but propery touchUI is present in API section https://primeng.org/datepicker#api.datepicker.props.touchUI

PR #17733, opened 2025-02-20
fix: only check for maximized when setting p-dialog-maximized

vetemi:
This is a regression bug as it worked in <v18 versions. This solves the following problems:
- This allows to toggle the maximization from the outside without using the default icon  
- Adding "maximizable: true" to the config, adds the default icon
- Not adding "maximizable: true" makes it impossible to toggle maximization
- There is no reason to check if maximizable. Checking if maximized should be the only check

### DIFF
--- a/packages/primeng/src/dynamicdialog/dynamicdialog.ts
+++ b/packages/primeng/src/dynamicdialog/dynamicdialog.ts
@@ -43,7 +43,7 @@ const hideAnimation = animation([animate('{{transition}}', style({ transform: '{
             <div
                 *ngIf="visible"
                 #container
-                [ngClass]="{ 'p-dialog p-component': true, 'p-dialog-maximized': maximizable && maximized }"
+                [ngClass]="{ 'p-dialog p-component': true, 'p-dialog-maximized': maximized }"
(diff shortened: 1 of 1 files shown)


mertsincan:
This is by design in Dialog and DynamicDialog components. Please set maximizable prop.

vetemi:
Please let me exaplain and reconsider again:

**1. Design flaw**
This is a design flaw. As I already stated, setting "maximiziable" prop will add a header with a maximize icon, which I don't want to have.

"maximizable" also doesn't work as the "closable" prop. For example I can set "closable: false" and still call DynamicDialogComponent.close() which closes the dialog. However, if I set "maximizable: false" and call DynamicDialogComponent.maximize(), it will not maximize.

This is because of the check I proposed to delete. 

**2. Marginal impact**
Removing the check doesn't harm anything within DynamicDialogComponent but prevents to maximize from outside.

**3. Regression**
Most important, this is a regression. It worked in previous versions. I created my logic around the functionality and now removing this, has impact on my work. Again, this change has no impact for your component but now I have to work around.

I hope this helps to understand my concerns. Thank you for your great work.

Issue #17732, opened 2025-02-20
Multiselect should use all options for selectedOptions not just visible


JacobSiegle:
### Describe the bug

https://github.com/primefaces/primeng/blob/master/packages/primeng/src/multiselect/multiselect.ts#L1339

In the following code it uses -
`const visibleOptions = this.visibleOptions();`

But really it should use -
`getAllVisibleAndNonVisibleOptions`

selectedOptions can get messed up due to this when used with a filter because visibleOptions + filter takes items out of the list but selectedOptions  needs to know about all the items.
https://github.com/primefaces/primeng/blob/master/packages/primeng/src/multiselect/multiselect.ts#L1271

Items are selected - 
![Image](https://github.com/user-attachments/assets/5a44bb9c-71fe-4273-ad90-28218603e5b6)

Filter and the selected items are now empty
![Image](https://github.com/user-attachments/assets/bc3a7bc9-7dee-4517-9c4d-33bc5427de9b)

Unfilter the items still are selected.

### Reason for not contributing a PR

- [x] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [x] Other

### Other Reason

I would be willing to make a PR, I am logging this so I remember when I get time.

NOTE: The stackbliz doesn't reproduce the issue all the time it tends to be race condition-ish from what i've been able to debug.

### Reproducer

https://stackblitz.com/edit/github-cxuzxajo

### Environment

Win10

### Angular version

19

### PrimeNG version

v19

### Node version

18

### Steps to reproduce the behavior

Open dropdown
Filter
Selected items can vanish sometimes

### Expected behavior

Selected items always have the full list in them

PR #17734, opened 2025-02-20
fix: selectedOptions not taking into account all options sometimes

JacobSiegle:
### Defect Fixes
https://github.com/primefaces/primeng/issues/17732


mertsincan:
Thanks a lot!

PR #17741, opened 2025-02-21
Fix ConfirmDialog breakpoints

alezhu:
**Bug:** The `ConfirmDialog` component receives the `breakpoints` input property and generates corresponding CSS `@media` classes for it. However, it assigns names in the format `.p-dialog[pn_id_<id>]`, where `<id>` is the identifier of the `ConfirmDialog` component itself, not the ID of the internal `p-dialog` component. As a result, the CSS classes do not work as intended. Meanwhile, the `p-dialog` component is capable of accepting the `breakpoints` property directly and correctly generates the appropriate class names.  

**Fix:** Pass the `breakpoints` property through to the internal `p-dialog` component.

**PS:** Also define separate type for dialog position 

#17753


mertsincan:
Thanks a lot for your contribution! Could you please create an issue and link it to PR?

alezhu:
> Thanks a lot for your contribution! Could you please create an issue and link it to PR?

OK. 
#17753

alezhu:
I'm sorry, but I believe there might be some misunderstanding. The issue was closed, however, my pull request has not been merged yet, and the necessary changes have not been made in the code. As a result, the bug is still present.

Issue #17540, opened 2025-01-30
Binding [dt] on Select component has no effect


WillooWisp:
```
<p-select
    appendTo="body"
    [options]="cars"
    [(ngModel)]="selectedCar"
    (onChange)="onCarChange($event.value)"
    optionLabel="label"
    [dt]="noSelectFocus()"
/>

public noSelectFocus = signal({
    root: {
        background: 'green',
        filledBackground: 'green',
        filledHoverBackground: 'green',
        filledFocusBackground: 'green',
        borderColor: 'transparent',
        hoverBorderColor: 'green',
        focusBorderColor: 'green',
        color: 'green',
        shadow: 'none',
        paddingX: 0,
        paddingY: 0
    },
    option: {
        focusBackground: 'green',
        selectedBackground: 'green',
        selectedFocusBackground: 'green',
        color: 'green',
        focusColor: 'green',
        selectedColor: 'green',
        selectedFocusColor: 'green',
    },
});
```

WillooWisp:
sm and lg properties also missing from root in SelectDesignTokens.

PR #17742, opened 2025-02-21
fix: correctly apply custom design token (badge, dropdown, select, tabmenu)

rfreydi:
issue: https://github.com/primefaces/primeng/issues/17540


mertsincan:
Thanks a lot!

Issue #17752, opened 2025-02-23
Accordion p-accordion toggle arrows seem reversed


esthersoftwaredev:
### Describe the bug

As I was using the accordion, placing it in a component in my v19 app, I took a look at the toggle arrows: 

![Image](https://github.com/user-attachments/assets/859feffc-c472-4a4f-b11d-35aad4adcd92)

when collapsed they are all pointing up and the element is `chevronupicon` is 

when I toggle the arrow of course does change to the opposite and becomes a `chevrondownicon`

![Image](https://github.com/user-attachments/assets/5d6bb561-afe7-4815-9e48-cc57fa239460)

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [x] Other

### Other Reason

I am not sure yet if this is intentional on not, and I prefer verifying before taking any further actions

### Reproducer

https://stackblitz.com/run?file=components%2Flayout%2Fapp.component.html

### Environment

Windows
Chrome
Angular v19.1.0
PrimeNG v19.0.6
Node v20.14.0

### Angular version

19.1.0

### PrimeNG version

v19

### Node version

20.14.0

### Browser(s)

Chrome

### Steps to reproduce the behavior

1. Go to [https://primeng.org/accordion#accessibility](https://primeng.org/accordion#accessibility)
2. Toggle panels of the accordion
3. Observe the toggle arrows

### Expected behavior

In my other apps where I am using accordions and just most accordions and similar actions have the arrows reversed and I am used to associating arrow-up with collapse & arrow down with - expand...

One of my apps with Angular Material UI:

![Image](https://github.com/user-attachments/assets/f9a907f6-1ed5-4533-ba4f-55f5d9b2e259)

So in my latest app with PrimeNG, I am rotating them like this:

![Image](https://github.com/user-attachments/assets/e2989cff-92a1-4062-9ccc-55b4bbcbc5ba)

It just feels more intuitive this way, to me at least.

Please let me know if it is ok to work on this and I can prepare a PR later on.

almenke:
I would agree.

esthersoftwaredev:
Thank you @cetincakiroglu for the replay, this is very interesting, I am genuinely curious @cagataycivici when you have a moment - could you let us know what lead to this design decision, I would love to know the backstory :)

PR #17764, opened 2025-02-25
fix: #17752, Accordion p-accordion toggle arrows seem reversed

akshayaqburst:
fix: #17752, Accordion p-accordion toggle arrows seem reversed


Wendelstein7:
I feel like this may more be a case of preference than one of mistake. Though I think I agree with your proposed changes.

akshayaqburst:
> I feel like this may more be a case of preference than one of mistake. Though I think I agree with your proposed changes.

PrimeNg handles it correctly as shown here - https://primevue.org/accordion/#dynamic.

So, I guess this is a bug and needs to be fixed.

cetincakiroglu:
Hi,

It's not a bug, it's a choice of design. Thanks for the effort and support, closing the issue.

PR #17765, opened 2025-02-25
Use PrimeNG's interface in examples, instead of manually creating a duplicate (bad practice in documentation)

Wendelstein7:
The current paginator documentation, shows the practice of manually defining a type interface in every example where the paginator is used, for the `onPageChange($event)` event emitter. In every example, the type of `$event` is manually defined like so:

```ts
interface PageEvent {
    first: number;
    rows: number;
    page: number;
    pageCount: number;
}

...

onPageChange(event: PageEvent) {
  this.first = event.first;
  this.rows = event.rows;
}
```

Since PrimeNG internally uses its own interface definition for this event, manually defining it in the users' project code could lead to errors in future updates where PrimeNG changes its internal interface for the `onPageChange($event)` event emitter. Having users use their own defined type would undo the benefits of typed javascript, as it would obscure potential issues or functionality.

In fact, the current duplicate definition of the interface you see above, is already unpaired/unsynchronised with PrimeNG's internal interface, which defines all properties as nullable. This already introduces possibilities for forgotten undefined-checks in the users code, as this is hidden by the use of the manually defined interface. This is a good example of why the current example could be considered a bad practice.

Therefore, I propose to remove the manual definition of the type `PageEvent` in the examples, and instead refer to the PrimeNG, which is available as `PaginatorState` in `primeng/paginator`.

Users could simply acquire this interface trough an import; and it would always be up to date with PrimeNG's functionality:
```ts
import { PaginatorState } from 'primeng/paginator';

...

onPageChange(event: PaginatorState) {
  this.first = event.first;
  this.rows = event.rows;
}
```

I hope this helps! With kind regards :)

### DIFF
--- a/apps/showcase/doc/paginator/basicdoc.ts
+++ b/apps/showcase/doc/paginator/basicdoc.ts
@@ -1,12 +1,6 @@
 import { Code } from '@/domain/code';
 import { Component } from '@angular/core';
-
-interface PageEvent {
-    first: number;
-    rows: number;
-    page: number;
-    pageCount: number;
-}
+import { PaginatorState } from 'primeng/paginator';
 
 @Component({
     selector: 'basic-doc',
@@ -29,7 +23,7 @@ export class BasicDoc {
 
     rows: number = 10;
 
-    onPageChange(event: PageEvent) {
+    onPageChange(event: PaginatorState) {
         this.first = event.first;
         this.rows = event.rows;
     }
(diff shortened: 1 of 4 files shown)


Wendelstein7:
Ofcourse, now that we're using the advantages of a typed language, we immediately notice how we need to adjust a little bit of code so the code can expect all the outcomes PrimeNG would emit :)

Issue #15655, opened 2024-05-22
Calendar: rselectionMode range + showTime doesn't work together


Emiliorth:
### Describe the bug

p-calendar with selectionMode="range" and [showTime]="true" gives error, when clicking arrows that changes hours
![image](https://github.com/primefaces/primeng/assets/46076240/f31acc47-27f5-47b7-a2a1-2f0770f0d748)

### Environment

Windows 10 Angular 17, PrimeNG 17.17.0

### Angular version

17.3.0

### PrimeNG version

17.17.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

21.6.0

### Browser(s)

Chrome

### Steps to reproduce the behavior
Click on arrow that changes hours
`  <p-calendar [formControlName]="controlName"
                        [inputId]="finalInputId"
                        appendTo="body"
                        [dateFormat]="'dd-mm-yy'"
                        [showIcon]="true"
                        [showClear]="true"
                        [showButtonBar]="true"
                        [showTime]="true"
                        [minDate]="min"
                        [maxDate]="max"
                        selectionMode="range"
                        [required]="requiredAutoForValidation">
                <ng-template pTemplate="triggericon">
                    <app-google-icon [icon]="GoogleIcon.CALENDAR_TODAY"></app-google-icon>
                </ng-template>
            </p-calendar>`

matteobertozzi:
Same problem with 18.0.0-rc.2 and 17.18.12
```
ERROR TypeError: value.toDateString is not a function
    at _Calendar.toggleAMPMIfNotMinDate (primeng-calendar.mjs:3895:43)
    at _Calendar.incrementHour (primeng-calendar.mjs:3889:10)
    at _Calendar.repeat (primeng-calendar.mjs:3930:35)
    at _Calendar.onTimePickerElementMouseDown (primeng-calendar.mjs:3905:12)
    at Calendar_div_3_div_5_Template_p_button_mousedown_2_listener (primeng-calendar.mjs:1576:36)
    at executeListenerWithErrorHandling (core.mjs:25797:12)
    at wrapListenerIn_markDirtyAndPreventDefault (core.mjs:25829:18)
    at HTMLElement.<anonymous> (platform-browser.mjs:745:112)
    at _ZoneDelegate.invokeTask (zone.js:402:33)
    at core.mjs:6717:49
```

```
<p-calendar [(ngModel)]="rangeDates" selectionMode="range"
[...]

PR #17781, opened 2025-02-27
fix datepicker increment hour - using range + showtime

stajne:
### Defect Fixes
<ins>https://github.com/primefaces/primeng/issues/15655</ins> 
though for datepicker only first. if accepted, it should be applied to calendar also


packages/primeng/src/datepicker/datepicker.ts
```diff
@@ -2869,7 +2869,7 @@ export class DatePicker extends BaseComponent implements OnInit, AfterContentIni
     }
 
     toggleAMPMIfNotMinDate(newPM: boolean) {
-        let value = this.value;
+        let value = Array.isArray(this.value) ? this.value[0] : this.value;
```
FilipKovac, line 2872: Please use `this.values[this.values.length - 1]` to have same functionality as before. This fix would change time only for the first value resp. the start of the interval in range.

stajne: Not sure if i understood you. Firstly this.values does not exist. As i read this function this valueDateString is compared only with minDate which can only be the start of the interval. If you meant this.value, if length is 2 (which is when the bug was happening) we would use this.value[1] which would compare end of range with minDate. Did I miss something?

FilipKovac: Ok, `this.values` was my copy/paste error from git history, it should be just `this.value`. `this.minDate` is not start of the interval but the possible minimal value from inputs. You have to still check the last value in `this.value` if it is Array because that is the only value you can update. Otherwise it makes no sense to check first value when you can only update time of the last value.

stajne: Yes, i maybe miss spoke about minDate, sorry.  I just tested what you said and this is what confuses me. If i select **start date** in range and then I change time, variable **value** will be null because this.value[1] is null (and this.value.length-1 will always be 1 for range since length is 2) and that null will but compared to minDate and start date value will never be checked.

FilipKovac: I would say that the idea of the time picker for multiselect and rangepicker is somewhere about that you can edit time only for last added value. For the range you pick the start date and then edit time (this.value.length === 1), then you add end date and edit the time. So when editing start date, the value should have only one element inside. The `this.pm` is there only to say, what to show (AM/PM) for edited time. It does not make sense to compare startDate in range.

I am not sure, I did not debug the picker, I just run into the same issue you did.

stajne: this.value.length is never 1 for range because there is second value which is null (I wasn't completely sure how JS will interpret it) so i debugged it and length is two even when just one value is added. Also one more thing, why i did what how i did it. When you enter first value this is compared with minDate (with my code), if you select a date before that it will also become first date so increment hour will use that for this function. I saw no point in comparing second value since second (end)  date has to be after first date, and since first (start) date is equal or after minDate this would be after. The only possible usage for second (end) date I just realized is when it is the same as the first date, so i came up with a new solution that would work as you expected it by using : 
this.value.slice().reverse().find(v => v != null) 

i pushed now

FilipKovac: Ok, I see and I agree with you. One more thing, you can write reverse find function in more optimized way without creating new array. 
`arr.reduceRight((acc, cur) => cur !== null ? cur : acc, null);` where `arr = [123431231, null]`

stajne: done

stajne: is this resolved?

mehmetcetin01140:
Fixed in PrimeNG v22.

PR #17798, opened 2025-03-02
fix(dialog): Ensure mask visibility state is consistent with dialog visibility

### LINKED ISSUE #17301: Dialog not visible after quickly hiding and showing again
### Describe the bug

When I quickly hide and then show a `Dialog`, then it is not visible. It ends up in an inconsistent state:
- `dialog._visible` is `true` which is correct
- `dialog.maskVisible` is ❌ `false` (should be `true` instead)

### Reproducer

https://stackblitz.com/edit/snxqxzny

Faulty state:

image

Note that the dialog is supposed to be visible but it is not.

### Angular version

19

### PrimeNG version

v19

- Lack of time
- Unsure how to implement the fix/feature

AlexFHer:
fix #17301

Animations if they are quickly changed between they lost sync because they are not done programatically.

### DIFF
--- a/packages/primeng/src/dialog/dialog.ts
+++ b/packages/primeng/src/dialog/dialog.ts
@@ -1060,6 +1060,10 @@ export class Dialog extends BaseComponent implements OnInit, AfterContentInit, O
                 this.onContainerDestroy();
                 this.onHide.emit({});
                 this.cd.markForCheck();
+
+                if (this.maskVisible !== this.visible) {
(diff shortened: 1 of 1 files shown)


SoyDiego:
Thanks for your contribution. 
Please replace the word `ref` to `fix #17301" and your PR will be linked with the issue and when PrimeNG Team approves, eveything will be closed together.

AlexFHer:
Done

PR #17815, opened 2025-03-05
fix: p-datepicker highlight selected year range

fxprunayre:
Using the date picker with a range and view mode year or month does not highlight selected range.
It was only highlighting the first date.

image


cloggy45:
Any joy with this?, it would be great to have.

maksymprokopenko:
It's so strange this PR not merged yet as it's more like a fix for a bug in component.

PR #17829, opened 2025-03-06
fix(tabs): "lazy" property doesn't do anything

### LINKED ISSUE #16806: Add lazy load support to tab component

LoaderB0T:
Fixes #16806 
There is already a PR that resolved that issue, however I kept the original [lazy] flag functionality so you can toggle it on and off (false by default as described in the docs).

### DIFF
--- a/apps/showcase/doc/tabs/lazydoc.ts
+++ b/apps/showcase/doc/tabs/lazydoc.ts
@@ -1,0 +1,134 @@
+import { Code } from '@/domain/code';
+import { Component } from '@angular/core';
+
+@Component({
+    selector: 'lazy-doc',
+    standalone: false,
+    template: `
+        <app-docsectiontext>
+            <p>Enabling <i>lazy</i> property of a Tab removes the content from the DOM until the tab is activated.</p>
+        </app-docsectiontext>
+        <div class="card">
+            <p-tabs value="0" [lazy]="true">
+                <p-tablist>
+                    <p-tab value="0">Header I</p-tab>
+                    <p-tab value="1">Header II</p-tab>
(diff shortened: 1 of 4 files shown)


Adrii77:
Hello @LoaderB0T 

I already tried to implements your solution before sending my PR. I still retested what you did and it doesn't work. See my video with your code : 

https://github.com/user-attachments/assets/91dc3c81-f6b8-4293-9682-ae856a01083a

`ngOnInit` should not appear until you click on the tab and that's not the case with your solution.

Also, I believe it would have been more respectful and efficient to discuss your approach directly in my PR rather than creating a new one. Let’s work together to resolve this issue.

Thank you.

LoaderB0T:
@Adrii77 
Fair enough, I thought it to be easier to have two solutions to compare, no disrespect intended. Just options.

Regarding OnInit being called, that is correct for me the elements being in the DOM or not was the only thing that mattered and that I checked.

Would you agree keeping the flag is good for compatibility with older versions of primeng? If so feel free to include it in your PR and I'll close this one

Adrii77:
@LoaderB0T 
Thanks for your reply and sorry for thinking that. I hope you understand me.

Even if elements are not present in the DOM, components are still instantiated, and in my opinion, this is really problematic for stateful components, especially those that trigger API calls, for example.

In fact, I tried many things to get it to work with the lazy flag, but it turned out to be more complicated than expected. I explain why here: https://github.com/primefaces/primeng/pull/17388#issuecomment-2592858363

LoaderB0T:
Makes sense to me, my mistake I should have taken a closer look to your PR. I will continue commenting on your PR to it easier to follow for others :)

Issue #17794, opened 2025-03-01
p-dialog: dismissableMask only works if modal is set


mathisvester:
### Describe the bug

If I want to enable https://primeng.org/dialog#api.dialog.props.dismissableMask in Dialog i have to also enable https://primeng.org/dialog#api.dialog.props.modal.

I think this is the case because of `pointer-events: none` which is set if `modal` isn't enabled as seen here https://github.com/primefaces/primeng/blob/fd48576f863fa5bdf5d46afbd14959b13ad80475/packages/primeng/src/dialog/dialog.ts#L62C17-L62C58.

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [x] Other

### Other Reason

Needs to be discussed with the core team before applying any change.

### Reproducer

https://stackblitz.com/edit/wohghtbd?file=src%2Fapp%2Fdialog-basic-demo.html

### Environment

-

### Angular version

19.0.1

### PrimeNG version

v19

### Node version

18.20.3

### Steps to reproduce the behavior

1. In StackBlitz click on `Show` button
2. Try to close by clicking outside the dialog

### Expected behavior

Enabling `dismissableMask` alone has no effect on the expected behavior. The user has to also enable `modal`.

If this is intended I would suggest an extended description in the docs of the component property: https://primeng.org/dialog#api.dialog.props.dismissableMask.

floreauwe:
same this for me

ApplY3D:
https://github.com/primefaces/primeng/pull/17834

cagataycivici:
dismissableMask only applies a modal dialog so this seems to be expected. Doc is also not ambiguous for me;

`Specifies if clicking the modal background should hide the dialog.`

PR #17834, opened 2025-03-07
fix: bind mask click listener if dismissableMask:true but modal:false

ApplY3D:
https://github.com/primefaces/primeng/issues/17794

### DIFF
--- a/packages/primeng/src/dynamicdialog/dynamicdialog.ts
+++ b/packages/primeng/src/dynamicdialog/dynamicdialog.ts
@@ -434,7 +434,7 @@ export class DynamicDialogComponent extends BaseComponent implements AfterViewIn
     }
 
     enableModality() {
-        if (this.ddconfig.closable && this.ddconfig.dismissableMask) {
+        if (this.ddconfig.dismissableMask) {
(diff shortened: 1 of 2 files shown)


packages/primeng/src/dynamicdialog/dynamicdialog.ts
```diff
@@ -438,7 +434,7 @@ export class DynamicDialogComponent extends BaseComponent implements AfterViewIn
     }
 
     enableModality() {
-        if (this.ddconfig.dismissableMask) {
+        if (this.ddconfig.closable && this.ddconfig.dismissableMask) {
```
ApplY3D, line 437: I've added a `closable` check like it is done in `dialog`.

ApplY3D: It looks like we should not check for `closable` either in `dynamicdialog` or `dialog`, as it "Adds a close icon to the header to hide the dialog"

cagataycivici:
dismissableMask only applies a modal dialog so this seems to be expected. Doc is also not ambiguous for me;

`Specifies if clicking the modal background should hide the dialog.`

PR #17958, opened 2025-03-24
fix: picklist documentation

khayo:
I corrected the picklist documentation, adding the missing pTemplate attribute and specifying the object name representing the template.


khayo:
fix #17959

PR #18041, opened 2025-04-04
fix: fixed #18040 - cascadeselect: clicking a few times causes the co…

### LINKED ISSUE #18040: CascadeSelect: clicking a few times causes the component to freeze

mehmetcetin01140:
fixes #18040


packages/primeng/src/overlay/overlay.ts
```diff
@@ -494,21 +494,25 @@ export class Overlay extends BaseComponent implements AfterContentInit, OnDestro
 
         switch (event.toState) {
             case 'visible':
-                this.show(container, true);
+                if (this.visible) {
+                    this.show(container, true);
+                }
                 this.bindListeners();
```
rosenthalj, line 500: should this.bindListeners be inside your newly added  "if block"?

PR #18044, opened 2025-04-07
fix(autocomplete): support full option object in selectedItemTemplate with optionValue and multiple

### LINKED ISSUE #17332: autocomple multiple : can't display labels
### Describe the bug

When using **autocomplete** with **multiple** = true and objects like **[{label: 'toto', value: {}}]**,
**the selected items cannot display their labels properly** : 

```html
<li
                    #token
                    *ngFor="let option of modelValue(); let i = index"
                    [ngClass]="{ 'p-autocomplete-chip-item': true, 'p-focus': focusedMultipleOptionIndex() === i }"
                    [attr.id]="id + '_multiple_option_' + i"
                    role="option"
                    [attr.aria-label]="getOptionLabel(option)"
                    [attr.aria-setsize]="modelValue().length"
                    [attr.aria-posinset]="i + 1"
                    [attr.aria-selected]="true"
                >
```
**let option of modelValue()** is wrong : it's not an option (label+value) but only a value.
This should be renamed as **value**, and a function similar to **inputValue()** should be used to fetch the selected options among the options list

BR

### Pull Request Link

_No response_

- [X] Lack of time
- [X] Unsure how to implement the fix/feature
- [X] Difficulty understanding the codebase
- [ ] Other

### Other Reason

_No response_

### Reproducer

x

### Environment

prime 18.0.2

### Angular version

18.0.2

### PrimeNG version

v18 (LTS Only)

### Node version

_No response_

### Browser(s)

_No response_

use an autocomplete with complexe objects

### Expected behavior

labels should display properly

gmaggiodev:
When using `optionValue` with `multiple=true`, the selected items rendered in `selectedItemTemplate` receive only the primitive value (e.g., an ID), making it impossible to access the full object context (e.g., to render a label or additional fields).

This PR introduces a complete internal solution that ensures the model (`modelValue`) always contains full objects — even if the bound `ngModel` uses primitive values — by mapping them from the available suggestions during `writeValue()` and after suggestion updates.

This avoids relying on potentially filtered or dynamic `suggestions` arrays at render time, ensuring stable rendering and correct template behavior.

### ✨ Key Changes

- `writeValue()` sets `modelValue` to the full option(s) when `optionValue` is used, using a `resolvePendingValue()` mechanism.
- `resolvePendingValue()` ensures that selected values are resolved to objects once suggestions are available (supports async/lazy).
- Modified `onOptionSelect()` and `updateModel()` to maintain consistency of full objects in `modelValue`.
- Keeps full compatibility with `optionValue`, `optionLabel`, and `multiple=true`.
- Added new documentation samples explaining both cases (with and without optionValue)

image

image

### ✅ Benefits

- Full context is always available for templates, without needing to search in `suggestions` dynamically.
- Works even when suggestions are loaded lazily or filtered.
- No breaking changes.

### 🛠 Example Use Case

```html
<p-autoComplete
  [(ngModel)]="selectedIds"
  [suggestions]="options"
  [optionValue]="'id'"
  [optionLabel]="'name'"
  [multiple]="true"
>
  <ng-template let-option pTemplate="selecteditem">
    {{ option.name }}
  </ng-template>
</p-autoComplete>
```
Here, selectedIds is an array of IDs, but option in the template will still be the full object ({ id, name }).

### Fixes:
Closes #17332


gmasci-ca:
Hi, any update on this?
This issue is quite critical for us — if it doesn't get resolved, we'll unfortunately be forced to consider switching to a different library.

its-dibo:
nice

filiptrivan:
Working with reactive forms?

gmaggio-ca:
> Working with reactive forms?

Yes. It doesn't depend on forms type. It depends on how the value is managed internally.

gmaggio-ca:
Trying to collaborate, BUT:

- Posted this PR to solve #17332 3 months ago.
- Tried to resolve conflicts one month ago. 

Is there anybody out there? @cagataycivici

Issue #16827, opened 2024-11-22
`Angular 19` Support


mertsincan:
Details; https://github.com/primefaces/primeng/pull/16783

AStoker:
If it helps, here's a list of files that needed to be updated in my testing. The Docs still have a lot of work though. https://github.com/AStoker/primeng/commit/a1bad87497c480a7b62abd392a5a9540d52c9a05

NateRadebaugh:
Can this be back-ported to a 17.x release once this is ready?

mertsincan:
Today, we had a detailed meeting about Angular 19 and decided to support it with PrimeNG 19. We are very close to the PrimeNG 18.Final. After releasing it, we plan to release PrimeNG 19.rc.1 within 1 to 2 weeks. Thanks a lot for your understanding!

mertsincan:
> Can this be back-ported to a 17.x release once this is ready?

Unfortunately, we do not plan to port it to version 17. Primeng versions will be released following the Angular versions. Otherwise, there will be many breakages in frameworks and libraries that are under such active development. I think this is the most logical thing.

igordanilovski:
So as I understand v17 & v18 (which still is in beta) will not have support for Angular 19? First release to support Angular 19 will be 19.x?

David-Ben-Mesecke:
@mertsincan I agree with all of what @aberonni said, however, as opposed to him, if I do not at least see a drastic attitude change from your team, I am switching back to free and well supported angular material, this is completely unacceptable for my workflow. When angular updates, I update, I can wait a maximum of a week to update to a new major version upgrade but not even an day longer. I am pretty sure many of your paying customers feel the same. That said, this gives me a much needed kick in the butt to finally abstract my use of frontend UI components using facades. Yet @mertsincan please, pretty please, with sugar on top, please don't make me go back to angular material, right now, I have better things to do. Also, from what I can tell, the fix does not demand much work from you, so please do it, and do it right now.

David-Ben-Mesecke:
> Today, we had a detailed meeting about Angular 19 and decided to support it with PrimeNG 19. We are very close to the PrimeNG 18.Final. After releasing it, we plan to release PrimeNG 19.rc.1 within 1 to 2 weeks. Thanks a lot for your understanding!

[...]

PR #18058, opened 2025-04-08
feat: Angular 19 support for PrimeNG 17

jacobg213:
# Feature Request - Angular 19 support for PrimeNG 17

This relates to issue #16827 and [this discussion](https://github.com/orgs/primefaces/discussions/3065).

## ⁉ Motivation

We have a big project that's using PrimeNG 17. We'd like to upgrade to Angular 19 but the migration to PrimeNG 19 is really blocking us. I tried to figure out how to do it but without a guide the amount of research and work required is too much for us to justify at the moment. It seemed that adding Angular support to v17 could be the easier approach.

This seems to be an issue many others also struggle with and that makes this update even more important in my opinion.

## 🛠 Changes 

- Upgraded all dependencies required to support Angular 19
- Run Angular's upgrade command - `ng update @angular/core@19 @angular/cli@19`
  - All `standalone` flags were changed to match v19's new default
  - Angular's core dependencies were updated
- Disabled the new "prefer standalone" ESLint rule - when running `fix` with that rule all `standalone: false` flags get removed which obviously breaks stuff
- Fixed errors introduced by the TypeScript version upgrade
- Fixed a few tests
- Run formatting
- Update required versions in `peerDependencies`

## 📝 Notes

I can see that the tests are failing on `v17`. I tried to fix up as much as I could but there are still a few failures coming from components I haven't used before. I'd very much appreciate if you have any directions or time to help me fix those up.

## ❔ Questions

- If this gets approved could you quickly release it as `v17.19.0`?
- Would you like me to revert any removals of the `standalone` flag set to `true` (I guess for backwards compatibility)?


jacobg213:
@cagataycivici Upgrading from PrimeNG 17 to 19 is an insanely huge task. And there's no good guide on what on earth to do. Are you basically saying that support for PrimeNG 17 is now dropped?

jacobg213:
If anyone else is looking for a PrimeNG 17 that supports Angular 19 I published a forked version on NPM https://www.npmjs.com/package/@jacobg213/primeng-17

chibisuke:
To be honest, this is a big nono to me and makes me reconsider using Prime alltogether. The changes from prime 17 to prime 18 in the themeing are way to extensive to migrate this properly within reasonable time. When we're touching this we might aswell switch frameworks.

jacobg213:
@chibisuke It's the same in my team - we'll be migrating to something else. The migration to v18+ is just too much without a proper guide/changelog.

olgeni:
I'll just throw Claude Code Max Ultra Pro Whatever at the task, as I am too old to deal with this kind of stuff anymore 😃

xantao:
primeng 17 is the latest "classic" primeng, it should at least have the dependencies corrected for latest angular versions. It works fine. Me and my team wont switch to primeng 18+ due to the absolute changes it would take on our projects.

Issue #18055, opened 2025-04-08
pTemplate directive documented yet not referenced in example


todaszy:
### Describe the bug

In https://primeng.org/selectbutton#template (v19):

"Template
For custom content support define a ng-template with pTemplate where the local ng-template variable refers to an option in the options collection."

It mentions "pTemplate" yet it is never used in an example:
```
<p-selectbutton [options]="justifyOptions" [(ngModel)]="value" optionLabel="justify">
    <ng-template #item let-item>
        <i [class]="item.icon"></i>
    </ng-template>
</p-selectbutton>
```

Also, what this is even doing? After upgrading from v17 to v19 this stopped working:
`<ng-template pTemplate let-item>`
I had to do to fix it:
`<ng-template #item let-item>`
No idea why... I hoped that docs will explain this but now I'm confused.

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [x] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/jvpq5g71?file=src%2Fapp%2Fselect-button-template-demo.html

### Environment

documentation

### Angular version

19

### PrimeNG version

v19

### Node version

22

### Browser(s)

chrome

### Steps to reproduce the behavior

Documentation mentioned doesn't make sense.

### Expected behavior

Documentation mentioned explains "pTemplate" and shows appropriate example.

Nicoss54:
@todaszy, p-template is a structural directive, if you want to use it, you have to write

<ng-content *pTemplate="'item'; let item">

if you take a look on the code you can see that the rendering is done like that

```typescript
 @ContentChildren(PrimeTemplate) templates: QueryList<PrimeTemplate> | undefined;

    ngAfterContentInit() {
        (this.templates as QueryList<PrimeTemplate>).forEach((item) => {
            switch (item.getType()) {
                case 'item':
                    this._itemTemplate = item.template;
                    break;
            }
        });
    }
```

PrimeTemplate take in input a name, and the getTye return the name that'why you have to write 

```html
<ng-content *pTemplate="'item'; let item">

...
</ng-content>

I will make a PR to show an exemple in the documentation

PR #18087, opened 2025-04-15
feat(selectbutton): enhance template customization options

Nicoss54:
fixes: #18055

Added support for using the `pTemplate` directive for custom option rendering in the SelectButton component. Updated docs and examples to demonstrate the usage of this feature, providing more flexible and structured customization.


cagataycivici:
We'll update the docs regarding this for all, not just this particular component, thank you for the heads up.

PR #18113, opened 2025-04-18
Add 'normal' as a size option

SinS3i:
Context
When wrapping primeNG components, an undefined value cannot be passed through. Allowing for a no-op 'normal' value to be passed through does not impact functionality, but does allow for the normal state to be hit.

### Defect Fixes
Creating an issue after PR to link the two, will edit PR description once created.
https://github.com/primefaces/primeng/issues/18117


cagataycivici:
normal is the default so not necessary, it basically has no use on css side. Also your PR does not cover all components that have size prop, thank you.

PR #18119, opened 2025-04-19
fix(input-number): avoid redundant calculation

### LINKED ISSUE #18118: [Input Number] - write value redundant calculation
In the writeValue hook, a redundant calculation is done and could be optimized

Nicoss54:
fixes: #18118

### DIFF
--- a/packages/primeng/src/inputnumber/inputnumber.ts
+++ b/packages/primeng/src/inputnumber/inputnumber.ts
@@ -40,6 +40,7 @@ export const INPUTNUMBER_VALUE_ACCESSOR: any = {
     useExisting: forwardRef(() => InputNumber),
     multi: true
 };
+
 /**
  * InputNumber is an input component to provide numerical input.
  * @group Components
@@ -402,6 +403,7 @@ export class InputNumber extends BaseComponent implements OnInit, AfterContentIn
      * @group Props
      */
     @Input({ transform: booleanAttribute }) autofocus: boolean | undefined;
+
     /**
      * When present, it specifies that the element should be disabled.
      * @group Props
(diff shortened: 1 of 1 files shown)


packages/primeng/src/inputnumber/inputnumber.ts
```diff
@@ -1516,7 +1521,7 @@ export class InputNumber extends BaseComponent implements OnInit, AfterContentIn
     writeValue(value: any): void {
         this.value = value ? Number(value) : value;
         if (this.input) {
-            this.input.nativeElement.value = value ? Number(value) : value;
+            this.input.nativeElement.value = String(this.value);
         }
```
rosenthalj, line 1525: Is your change correct?
Because of line 1517,  did you mean:
`this.input.nativeElement.value = this.value`

Nicoss54: Hi :), thanks for review. Yes, this.value is cast a number. this.input.nativeElement.number is a type of string.

It's not raise an error with the terniary, due to the type any of value.

rosenthalj: Your explanation makes sense  but I'm still a little confused:
If the method writeValue is passed the argument value of  null
then this.value = null
and String(this.value) = String(null) = "null" and not null?  If this is ok or if my analysis is incorrect then your change is ok with me.

Nicoss54: My bad you're right, it seems that it will be more correct to have this like that

```typescript
this.input.nativeElement.value = this.value ? String(value) : null
```

mehmetcetin01140:
Fixed in PrimeNG v22

Issue #17791, opened 2025-02-28
Scrollable tabs wrongly styled


majkers:
### Describe the bug

When we use scrollable tabs styles are added to :

```
.p-tablist-tab-list {
    position: relative;
    display: flex;
    background: ${dt('tabs.tablist.background')};
    border-style: solid;
    border-color: ${dt('tabs.tablist.border.color')};
    border-width: ${dt('tabs.tablist.border.width')};
}
```

but when we scroll it, this div is not visible and all the styles are gone... IMHO one of parents of this div ie. the one with classes:

`p-tablist-content p-tablist-viewport`

should have tokens too.

Have a look at demo:

I set var --p-tabs-tablist-background to red. And initially it is ok

![Image](https://github.com/user-attachments/assets/d6d493a0-7b72-4ae0-b980-44638f491495)

but after scrolling to right red background is gone:

![Image](https://github.com/user-attachments/assets/e938c254-9003-4173-bc92-61f379277d4b)

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://primeng.org/tabs

### Environment

Primeng demo page

### Angular version

19.1.7

### PrimeNG version

v19

### Node version

n/a

### Browser(s)

all

### Steps to reproduce the behavior

1. Go to https://primeng.org/tabs
2. Go to scrollable section
3. Change var --p-tabs-tablist-background to red eg
4. Scroll tabs to right

### Expected behavior

Background should stay red

cagataycivici:
Provide a reproducer please.

majkers:
Repriducer is on demo page. I have attached screenshots

minhperry:
![Image](https://github.com/user-attachments/assets/50849ccf-de73-463d-828c-129559a29e56)

I think this is also somewhat similar to the issue: Scrolling background does not span the full width. Component is used in root.

StackBlitz https://stackblitz.com/~/github.com/minhperry/skyblock-simulator, branch preview.

GitHub repo at the exact point: https://github.com/minhperry/skyblock-simulator/tree/0b06772387bdcfe51701434db2f21b2397d084d6

PR #18121, opened 2025-04-20
fix(tabs): scrollable background does not span the whole content

minhperry:
With references to the linked issue:
`display: flex` only currently make the background color span the viewable width. Changing this to `display: inline-flex` will make it span the whole content strip.

`min-width: 100%` cover the cases where there is no scroll buttons, then it will span the whole width of its parent.

### Defect Fixes
Fixes https://github.com/primefaces/primeng/issues/17791


cetincakiroglu:
Hi, 

This issue has been fixed in the next branch during PrimeUIX Migration and the fix will be available by PrimeNG v20. Closing the PR.

PR #18166, opened 2025-04-25
feat: add missing closeAriaLabel option to ConfirmDialog component

### LINKED ISSUE #18163: Accessibility Issue in p-confirmdialog
Hi Team,
After updating to PrimeNG v19, I noticed a key issue with the p-confirmdialog component:

**Issue: Accessibility Issue with the Close Button**
The close button in p-confirmdialog does not have an aria-label or accessible text by default. While the documentation mentions the closeAriaLabel property, setting it does not apply the aria-label to the close button.

**Example:**

<p-confirmdialog closeAriaLabel="{{ 'FORM.CLOSE' | translate }}"></p-confirmdialog>

**Expected Behavior:**

The aria-label should appear on the close button, ensuring accessibility compliance (e.g., WCAG standards).

**Actual Behavior:**

The aria-label is missing from the close button, resulting in accessibility flags during audits.

I created an issue before regarding this; the ticket is closed but it seems the issue is still there. link follows
[https://github.com/primefaces/primeng/issues/17404](https://github.com/primefaces/primeng/issues/17404)

### Pull Request Link

_No response_

- [x] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/5fqkywvc?file=src%2Fapp%2Fconfirm-dialog-basic-demo.html

### Environment

Angular v19
Primeng v 19

### Angular version

19.2.8

### PrimeNG version

v19

### Node version

_No response_

### Browser(s)

Chrome latest

See the above example, inspect the close button

### Expected behavior

There should be an aria-label attribute in the button

prasanthnv:
### Defect Fixes
Fixes #18163

The `aria-labelledby` attribute in the ConfirmDialog component had a broken reference when the target ID did not exist.  
This PR fixes the issue by ensuring the ID is properly set and referenced, improving accessibility compliance.

### Notes
- Only a small fix (non-breaking).
- Tested locally on Chrome Accessibility tools.

### DIFF
--- a/packages/primeng/src/confirmdialog/confirmdialog.ts
+++ b/packages/primeng/src/confirmdialog/confirmdialog.ts
@@ -57,6 +57,7 @@ const hideAnimation = animation([animate('{{transition}}', style({ transform: '{
             [style]="style"
             [dismissableMask]="dismissableMask"
             [draggable]="draggable"
+            [closeAriaLabel]="option('closeAriaLabel')"
         >
             @if (headlessTemplate || _headlessTemplate) {
                 <ng-template #headless>


StefanTieber:
Thanks. This seems logical.
Could you also take a look at the propagation of the `closeButtonProps` one can set in `ConfirmationService.confirm()`? I think they have a similar issue, and I think it would be fitting to fix it in one PR.
You can see what I mean here: https://stackblitz.com/edit/github-gbpwkjgt?file=src%2Fapp%2Fapp.component.ts

pnellikkaparambil:
Hi @StefanTieber 

Thank you for the suggestion!

I have incorporated the changes to include closeButtonProps. I also noticed that it was missing in the p-dialog attributes, so I have added it there as well.

For reference, here is the closeButtonProps usage in the Dialog component:
[dialog.ts](https://github.com/primefaces/primeng/blob/b92c04e84b908e6ba82f0fe70cdbf52d01ceecf2/packages/primeng/src/dialog/dialog.ts#L342).

Thanks again!

> Thanks. This seems logical. Could you also take a look at the propagation of the `closeButtonProps` one can set in `ConfirmationService.confirm()`? I think they have a similar issue, and I think it would be fitting to fix it in one PR. You can see what I mean here: https://stackblitz.com/edit/github-gbpwkjgt?file=src%2Fapp%2Fapp.component.ts

Issue #18103, opened 2025-04-17
InputNumber error in undefined value (19.1.0)


mastersys-giovanni:
### Describe the bug

When InputNumber start with undefined value, cause error

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [x] Other

### Reproducer

https://stackblitz.com/edit/buhuq2vx?file=src%2Fapp%2Finput-text-basic-demo.ts

### Environment

stackblitz

### Angular version

19

### PrimeNG version

v19

### Browser(s)

Chrome

### Steps to reproduce the behavior

1 - Run example
2 - InputNumber show undefined in editor

### Expected behavior

1 - Run example
2 - InputNumber show empty

mastersys-giovanni:
Correction:

formatValue(value: any) {
        if (value != null **&& value != undefined**) {
            if (value === '-') {
                // Minus sign
                return value;
            }

            if (this.format) {
                let formatter = new Intl.NumberFormat(this.locale, this.getOptions());
                let formattedValue = formatter.format(value);

                if (this.prefix && value != this.prefix) {
                    formattedValue = this.prefix + formattedValue;
                }

                if (this.suffix && value != this.suffix) {
                    formattedValue = formattedValue + this.suffix;
                }

                return formattedValue;
            }

            return value.toString();
        }

        return '';
    }

imaksp:
Hi @mertsincan this could be due to this PR: https://github.com/primefaces/primeng/pull/17347
it works correctly in previous version. I hope that v20 adds some kind of testing to prevent such issues in future.

mastersys-giovanni:
@mertsincan , please put on 19.1.1

mastersys-giovanni:
[https://github.com/primefaces/primeng/pull/18167](url)

PR #18167, opened 2025-04-25
fix(inputnumber): resolve issue eith undefined value (#181003)

mastersys-giovanni:
inputnumber with undefined value, cause wrong formatValue


mastersys-giovanni:
[https://github.com/primefaces/primeng/issues/18103](url)

PR #18197, opened 2025-04-29
fix(InputMask): don't trigger unexpected events

### LINKED ISSUE #11566: InputMask triggers value change event on blur
**I'm submitting a ...**
```
```

**Plunkr Case (Bug Reports)**

https://stackblitz.com/edit/github-6hugjj?file=src/app/app.component.html
https://github-6hugjj.stackblitz.io

The console prints all value changes. Just type some letters and then blur the control. You can see that the last value gets printed two times.

**Current behavior**
When blurring the input mask a change event gets fired although the value doesn't change.

**Expected behavior**
The change event gets only fired when the value actually changes.

Use the stackblitz example above.

Or create a input-mask control via form builder and subscribe to the valueChanges observable on the form group.

It should work properly.

Windows, Rider, yarn, Angular 13, PrimeNg 13.3.3

* **Angular version:** 13.X

* **PrimeNG version:** 13.3.3

* **Browser:** At least Chrome and Edge
 
* **Language:** Typescript 4.6.3

* **Node (for AoT issues):** `node --version` =  14.18.3

DerStimmler:
Fix:
- don't trigger change event on blur
- only update model and trigger change event if value actually changes

fixes #11566

### DIFF
--- a/packages/primeng/src/inputmask/inputmask.ts
+++ b/packages/primeng/src/inputmask/inputmask.ts
@@ -582,13 +582,6 @@ export class InputMask extends BaseComponent implements OnInit, AfterContentInit
         }
         this.updateFilledState();
         this.onBlur.emit(e);
-
-        if (this.inputViewChild?.nativeElement.value != this.focusText || this.inputViewChild?.nativeElement.value != this.value) {
-            this.updateModel(e);
-            let event = this.document.createEvent('HTMLEvents');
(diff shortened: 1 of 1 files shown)


DerStimmler:
This issue is still present in v20.
I updated this pr to resolve merge conflicts.

DerStimmler:
Resolved merge conflicts.

PR #18298, opened 2025-05-09
fix(picklist): Exclude `undefined` item from `selectedItems` push in picklist

Will-Smith-007:
Fixes primefaces#18198


Will-Smith-007:
forgot that the list could have `undefined` as well

PR #18344, opened 2025-05-15
fix: select should use padding y and not padding x

Chocobozzz:
Fix a small typo in chip theme


Chocobozzz:
Oup's sorry, I didn't see it. Thank you for your work!

PR #18375, opened 2025-05-21
fix: check all filter rules to determine if column has a filter

afermon:
the hasFilter context variable on the ColumnFilter icon template only checks the first rule of the array. Other rules might still apply. This check for any non blank rule when more than one.


afermon:
fixes #18376

afermon:
Closes: 18376

afermon:
Closes: #18376

PR #18387, opened 2025-05-22
fix(table): ensure filter timeout is cleared on destroy

makepanic:
This makes sure that the scheduled timeout is always cleared on destroying the Table component.


makepanic:
We found this issues due to a table being used in tabs and when switching tabs quickly while the table is starting a filter action.
Our table uses resize-able columns and stores in localstorage which means after the timeout, it tries to compute the column widths of detached DOM. This in turn creates some bogus widths and persists it for future use.

PR #18390, opened 2025-05-23
Add type to toast severity, also fixed typo.

icenitipol:
### Defect Fixes
When submitting a PR, please also <ins>**create an issue**</ins> documenting the error and [manually link to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#manually-linking-a-pull-request-or-branch-to-an-issue-using-the-issue-sidebar).

### Feature Requests


SoyDiego (approved):

icenitipol:
Hi, after quite some time, I noticed that there is a build error in previous attempt of pull requests.
I have merged latest code from master, then fix some build errors caused by static severity of ToastMessageOptions  in showcases.

Here is more information on impact of adding ToastSeverity as a fixed type to ToastMessageOptions:

Previously, severity in ToastMessageOptions was typed as a loose string (or untyped). This PR introduces a strict union type:
```typescript
export type ToastSeverity = 'success' | 'info' | 'warn' | 'error' | 'secondary' | 'contrast' | 'custom';
```

### Breaking change for existing projects:
```typescript
// these all worked before, now type errors
this.messageService.add({ severity: 'warning', ... });  // common typo/alias for 'warn'
this.messageService.add({ severity: 'primary', ... });  // copied from Button severity
this.messageService.add({ severity: 'critical', ... }); // custom app-defined value
```

### Headless / custom severity use case

Projects using headless templates often set an arbitrary severity as a data carrier — the value is read in their own template logic, not by PrimeNG. These break too unless `custom` covers their case, which it only does if they happened to use that exact string.

Workaround for projects that need arbitrary values

They would need to cast: severity: `'myvalue' as ToastSeverity` which defeats the type safety purpose.

### Suggestion

```typescript
// Consider widening the type slightly to allow escape using `(string & {})`:
export type ToastSeverity = 'success' | 'info' | 'warn' | 'error' | 'secondary' | 'contrast' | 'custom' | (string & {});
```

This keeps autocomplete for known values while still accepting arbitrary strings without a cast. It's a common pattern used in React's CSSProperties and similar APIs. Runtime behavior is identical — only the DX changes.

Issue #1234, opened 2016-11-08
Change the firstDay automatically


CHOUAIAR:
I know that using jQuery we can do it but using Angular 2 is there a way to change the firstDay dynamically after the rendering of the schedule?

cagataycivici:
Please use forum for questions;

http://forum.primefaces.org/viewforum.php?f=35&sid=276c5ae7570d9b3751325405c6fc5510

It can be done by editing the firstDay propert of the bound locale object.

PR #18394, opened 2025-05-23
feat(dynamicDialog): add position-based transform options for DynamicDialog

trzhnnkv:
This change introduces dynamic transform animations based on the dialog
position (top, bottom, left, right, center) for DynamicDialog. It
aligns behavior with p-dialog, which already supports transformOptions
per position.

Previously, DynamicDialog only used a fixed scale transform regardless
of position, resulting in inconsistent UX.

Relates to [issue]( https://github.com/orgs/primefaces/discussions/4030)


trzhnnkv:
Relates to [#1234](https://github.com/orgs/primefaces/discussions/4030#discussion-8363798)

Issue #18388, opened 2025-05-22
Picklist target not update


LucasAlumni:
### Describe the bug

In the picklist component, when I add an item to the “target” list, the first element is added correctly. However, when I add another item to the “target” list, targetProduct contains only one element, even though the UI shows two elements in the “target” list.

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/xammse4a?file=src%2Fapp%2Fpicklist-basic-demo.html,src%2Fapp%2Fpicklist-basic-demo.ts

### Environment

- MacOS (15.5)
- Safari + Chrome

### Angular version

19

### PrimeNG version

v19

### Steps to reproduce the behavior

1 . add source element to target list
2 . add new source element to target list

### Expected behavior

targetProduct must have two elements.

gualbertof:
Confirmed, Same here

hkjeffchan:
Only works on "Ctrl" multi select. If add one by one, only the last one in the target data list.

sathyaanarayanan:
@mehmetcetin01140, wondering if this fix could be merged soon. It's breaking production

hkjeffchan:
I compiled and replace the files in node_modules/primeng/fesm2022 directly

You will need to remove the txt extension as github does not allow mjs and map extension.

[primeng-picklist.mjs.txt](https://github.com/user-attachments/files/20517703/primeng-picklist.mjs.txt)
[primeng-picklist.mjs.map.txt](https://github.com/user-attachments/files/20517702/primeng-picklist.mjs.map.txt)

sathyaanarayanan:
The weird workaround I implemented is as follows: 
Creating a new instance of the array every time the value is updated. 
 
**HTML file:** 

```
    <p-pickList 
        [source]="deselectedColumns" [target]="selectedColumns"
        (onMoveToTarget)="onMoveToTarget($event)"
        (onMoveToSource)="onMoveToSource($event)" 
        (onMoveAllToTarget)="onMoveAllToTarget($event)" 
        (onMoveAllToSource)="onMoveAllToSoaurce($event)"
>
```

**.ts file:**

```

  onMoveAllToTarget(event: any): void {
    this.selectedColumns = [...this.selectedColumns]
    this.deselectedColumns = [...this.deselectedColumns]
  }

  onMoveToTarget(event: any): void {
[...]

PR #18414, opened 2025-05-27
fix: fix picklist component for not updating source and target values

hkjeffchan:
moved source and target from input() to model() signal

Fix https://github.com/primefaces/primeng/issues/18388


cetincakiroglu:
fixed in https://github.com/primefaces/primeng/commit/591b02e5aff5f2011b7ff6b939b8e7ba77865cf1

PR #18432, opened 2025-06-02
feat(select): Add tooltip support for options

Will-Smith-007:
I've created a feature request about this: https://github.com/orgs/primefaces/discussions/4062


Will-Smith-007:
This PR is still relevant and was updated

PR #18437, opened 2025-06-02
fix(accordion,stepper): replace transformToBoolean with booleanAttribute for input properties

### LINKED ISSUE #18436: Accordion | Input "multiple" not transformed correctly
### Describe the bug

Given the following
```html
<p-accordion multiple [value]="activePanels()">
```
would internally set `multiple` to `false`

Because
```
export const transformToBoolean = (value: any): boolean => {
    return !!value;
};
```

just checks for truthiness.
But usually you would expect it this way
```
export function booleanAttribute(value: unknown): boolean {
  return typeof value === 'boolean' ? value : value != null && value !== 'false';
}
```
https://github.com/angular/angular/blob/a70f03a9b40a4f1f0e554541f236270025b41501/packages/core/src/util/coercion.ts#L21

### Pull Request Link

https://github.com/primefaces/primeng/pull/18437

### Reproducer

### Environment

MacOS

### Angular version

19.2.12

### PrimeNG version

v19

### Node version

_No response_

### Browser(s)

Chrome

Use `<p-accordion multiple>` and see that `multiple` is `false`

### Expected behavior

Use `<p-accordion multiple>` and `multiple` should be `true`

jase88:
fixes https://github.com/primefaces/primeng/issues/18436


jase88:
ready for review, still an issue

PR #18447, opened 2025-06-04
fix(#18448): hardcoded css variable in domhandler breaks overlay styles

mariovyord:
In `packages/primeng/src/dom/domhandler.ts` in the static method `relativePosition` there is a styling that is using hardcoded CSS variables. 

This is creating a bug where if you use different prefix than '--p', then `--p-anchor-gutter` is undefined and the styles are not applied.

Fixes Issue: https://github.com/primefaces/primeng/issues/18448


RBantutova (approved):

mariovyord:
Already fixed.

Issue #18380, opened 2025-05-22
primeng v18/19 - Menu: Keyboard navigation broken when using custom template


davidla06:
### Describe the bug

Hello,

when using a custom template to display menu items for a p-menu, selecting an element using the keyboard does not work anymore (the command is not triggered). Only mouse selection works.

same as it was in v17: https://github.com/primefaces/primeng/issues/15794

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/7xdwbhaw-r5fkif4g?file=src%2Fapp%2Fmenu-template-demo.ts

### Environment

*

### Angular version

18.2.13

### PrimeNG version

v18 (LTS Only)

### Steps to reproduce the behavior

1. navigate the menu with the keyboard up and down arrows
2. try to select an item pressing enter

### Expected behavior

the command of the item should be triggered

davidla06:
same happen for p-menubar

volkorros:
also they have added event.stopPropagation(); in tab component, so you can't manually configure keyboard behavior.
For example there is focused tab. I need on arrow down go to item below, and I can't.

maydrori:
They already fixed it in previous version here: https://github.com/primefaces/primeng/pull/15815/files,
but it seems that this fix was dropped in newer versions. 
I created a pr for them https://github.com/primefaces/primeng/pull/18469.

Anyway, a workaround for now is to add `[attr.data-pc-section]="'action'"` to the `<a>` tag in your item template

PR #18469, opened 2025-06-08
fix(menu, menubar): fix key enter on menu with custom template for item

maydrori:
fixes: [#18380](https://github.com/primefaces/primeng/issues/18380)


cetincakiroglu:
Hi,

Thanks for the support, we've reviewed the PR and it seems the issue happens in every menu component. Issue is fixed by the core team in https://github.com/primefaces/primeng/commit/af5844faed72b6596282e79ed64016439dad3646

Issue #16469, opened 2024-10-01
Tree/TreeSelect Selection Propagation not working correctly with filter applied


randes73:
### Describe the bug

Selection propagation in the treeview (and treeselect) only occur with filtered nodes and affects child selections based on parent selections and parent selection state (full or partially selected) based on child selections.  

### Environment

Angular, NodeJS & VS Code for development, C# .NET Core backendend

### Angular version

18.2.2

### PrimeNG version

17.18.9

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.16

### Browser(s)

Chrome,Edge (both latest version)

### Steps to reproduce the behavior

Associated Stackblitz: https://stackblitz.com/edit/gnvdu7-7zrhrz?file=package.json

Both scenarios below are using a treeview or treeSelect component, filtering enabled, and selection propagation enabled.  

Using the referenced stackblitz above where the node structure follows:
* Documents
  * Work
    * Expenses.doc
    * Resume.doc
  * Home
    * Invoices.doc
    
Scenario 1 - Parent does not select all children
1. Make sure all nodes are not selected
2. Apply a filter so that a Parent node is shown and at least one child filtered but not all ("Exp")
3. Select Parent node (Work)
4. Remove filter
5. Observe child selections where only filtered nodes (Expenses.doc) were affected.  Parent nodes is fully selected but not all children are selected (Resume.doc).

Scenario 2 - Parent not reflecting child node selections
1. Make sure all nodes are not selected
2. Apply a filter so that a Parent node is shown and at least one child filtered but not all
3. Select child node (Expenses.doc).  Observe the Parent node being fully selected.
5. Remove filter
6. Observe parent node still showing fully selected but not all child nodes selected (Resume.doc)

Other information: The following issue looks to be very similar to what I have reported: https://github.com/primefaces/primeng/issues/15776

### Expected behavior

Our experience prior to 16.9.6 with selection propagation was that the selections (or deselections) would occur regardless of filter and the parent properly reflects child selections regardless of filter.

randes73:
Updated associated stackblitz from 17.18.9 to 17.18.11.  Verified issue still exists with 17.18.11.

randes73:
[...]

PR #18472, opened 2025-06-09
fix tree-select propagation bug with filter (#16469)

Navaja970927:
https://github.com/primefaces/primeng/issues/16469


Navaja970927:
https://github.com/primefaces/primeng/issues/16469

PR #18493, opened 2025-06-17
feat(frozenColumn): Performances issues

### LINKED ISSUE #18481: pFrozenColumn not recalculating when table layout changes and terrible performances when resizing window
### Describe the bug

I found an issue (https://github.com/primefaces/primeng/issues/15922) talking about the performances problems.

It is not fixed in latest v19.1.3.
Performances are awful when you have frozen columns directive in a table with a lot a columns and rows.
All our columns need to have the pFrozenColumn directive and we play with the frozen attribute to activate the frozen status : a window resize takes more than 30 seconds to render.

After analyzing, even if the recalculateColumns do nothing at all (empty function) the rendering takes a lot of time.

It seems the reason is the spamming of event listeners on window:resize (one for each cell in the table so row*columns). Removing that resolves the performances problems.

To add to that problem, the window:resize isn't even a good solution because the table can be resized through other means and frozen column width is not recalculated then. Examples :

adding columns might reduce the frozen column size
having resizable columns makes frozen columns react weirdly
allow resize of the container of the table through slider might change column size without recalculating the left or right value of the frozen columns.

### Pull Request Link

https://github.com/primefaces/primeng/pull/18493

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/9yw4wtgu

### Environment

Microsoft Edge

### Angular version

19.1.4

### PrimeNG version

v19

### Node version

_No response_

### Browser(s)

Microsoft Edge

In the stackblitz either :
- remove columns, then add columns.
- activate resizableColumns and resize the last column to see that the previous frozen column is not moved
- change the window size to see the table resizing taking a lot of time

### Expected behavior

Listening to window:resize seems a bad idea when you have 2000 cells doing it.
The best way would be to react to resize events from the parent table.

Loic-fr:
Optimize performances when window resizing.
Recalculate when display of table changes

Closes #18481

### Defect Fixes
When submitting a PR, please also <ins>**create an issue**</ins> documenting the error and [manually link to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#manually-linking-a-pull-request-or-branch-to-an-issue-using-the-issue-sidebar).

### Feature Requests


Loic-fr:
Fix still needed because performances seems even worse on v20.

PR #18521, opened 2025-06-24
feat(floatlabel): Global config of default FloatLabel variant

djayard:
Added support for defining the default FloatLabel variant.
I followed the example of the input components.

Issue: https://github.com/primefaces/primeng/issues/18523


djayard:
Commits rebased against master.

Issue #18534, opened 2025-06-26
Autocomplete when multiple=true and typeahead=false ignores unique=true


lblu12555:
### Describe the bug

As you can see in the Autocomplete examples in both v18 and v19, in the Multiple section, the "with typeahead" unique works but the "without typeahead" unique doesn't work.  The "Without Typeahead" example allows you to enter multiple of the same entry.

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/ey7saxmy?file=src%2Fapp%2Fautocomplete-multiple-demo.html

### Environment

Windows environment, Chrome browser

### Angular version

v19

### PrimeNG version

v19

### Steps to reproduce the behavior

1.  Go to https://primeng.org/autocomplete
2.  Go to the Multiple section
3.   In the "Without Typeahead" example, enter "test" multiple times
4. Observe that non-unique entries are allowed

According to the API, unique=true is the default.  But even going into Stackblitz of the example and explicitly setting [unique]="true", it still allows the same text entered multiple times.

### Expected behavior

When unique=true, it should not allow the same string to be added if it has already been added.

PR #18535, opened 2025-06-27
fix(autocomplete): avoid adding the same value when no typeahead

Nicoss54:
fixes: #18534


alaa0ezzat:
when will merge the fix? @Nicoss54

PR #18550, opened 2025-07-03
improve severity type for toast

imaksp:
Improve `severity` field type for Toast/MessageService.


ggroupas:
Maybe a better approach would be to add a type like 

```typescript 
type ToastSeverity = 'success' | 'info' | 'error' | 'warn';
```

as this can be reused

PR #18585, opened 2025-07-09
fix: allow defaultDate to be null in DatePicker component

FlavienR:
- The @Input() defaultDate property was declared as Date but the PrimeNG docs state its default value is null.

- Passing null currently causes a type error, which contradicts the documented behavior.

- This PR corrects the mismatch by changing the getter/setter signature to accept Date | null.

(I'm creating an issue right away)


FlavienR:
Issue created: https://github.com/primefaces/primeng/issues/18586

FlavienR:
Fixes #18586

PR #18593, opened 2025-07-13
fix(autocomplete): onClear event not emitted when input loses focus 

### LINKED ISSUE #18592: Autocomplete doesn´t emit onClear with forceSelection set on
### Describe the bug

When you use forceSelection=true on p-autocomplete and type something which is not a suggestions. The input gets removed which is ok. But onClear was not emitted, which is not ok. When using backspace to clear the input, the onClear gets emitted just fine. Also wenn you click on an item in a suggestion the modelValue gets updated to null at first and miliseconds later to the suggested value. This is a timing issues due to having onInputChange on onChange and onOptionSelect on click event handler.

### Pull Request Link

#18593 

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/github-gcc7sdrq?file=src%2Fapp%2Fapp.component.html

### Environment

Angular 20.1 with primeng v20

### Angular version

20.1.0

### PrimeNG version

v20

### Node version

_No response_

### Browser(s)

Chrome

1. p-autocomplete with forceSelection=true
2. type in something 
3. click away so to blur

### Expected behavior

onClear gets emitted

eduardheller:
fixes #18592

### DIFF
--- a/packages/primeng/src/autocomplete/autocomplete.ts
+++ b/packages/primeng/src/autocomplete/autocomplete.ts
@@ -1134,23 +1134,7 @@ export class AutoComplete extends BaseInput implements AfterViewChecked, AfterCo
     }
 
     onInputChange(event) {
-        if (this.forceSelection) {
-            let valid = false;
-
-            if (this.visibleOptions()) {
-                const matchedValue = this.visibleOptions().find((option) => this.isOptionMatched(option, this.inputEL.nativeElement.value || ''));
-
-                if (matchedValue !== undefined) {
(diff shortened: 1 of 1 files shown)


wsacha:
Is there a way to prioritize merging this with the main branch? @cetincakiroglu

cetincakiroglu:
Hi,

Could you please resolve the conflicts so we can merge safely?

eduardheller:
@cetincakiroglu I resolved the conflict

Issue #18619, opened 2025-07-16
Primeng 20 showcase still shows rc.1


junan-trustarc:
### Describe the bug

<img width="557" height="319" alt="Image" src="https://github.com/user-attachments/assets/82977124-8c6a-40df-9a4c-087c241c409f" />

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://primeng.org/installation

### Environment

https://primeng.org/installation

### Angular version

20

### PrimeNG version

v20

### Steps to reproduce the behavior

Go https://primeng.org/installation

### Expected behavior

should be without rc.1

PR #18623, opened 2025-07-17
fix(downloaddoc): update package installation commands to remove version specifier

jase88:
fixes https://github.com/primefaces/primeng/issues/18619

### DIFF
--- a/apps/showcase/doc/installation/downloaddoc.ts
+++ b/apps/showcase/doc/installation/downloaddoc.ts
@@ -14,12 +14,12 @@ import { Component } from '@angular/core';
 export class DownloadDoc {
     code: Code = {
         command: `# Using npm
-npm install primeng@20.0.0-rc.1 @primeuix/themes
+npm install primeng @primeuix/themes
 
 # Using yarn
-yarn add primeng@20.0.0-rc.1 @primeuix/themes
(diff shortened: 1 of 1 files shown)


rosenthalj:
@jase88
With this change,  how would a developer determine which version of PrimeNG the demo is running?  A while back the dropdown included the version number.

Angular_Table_Component

jase88:
> @jase88
> With this change,  how would a developer determine which version of PrimeNG the demo is running?  A while back the dropdown included the version number.
> 
> Angular_Table_Component
> 

Would not be possible. But imho would be fine to use the latest version on the latest docs. 

What are you suggesting? Alternatively the corresponding package.json could be used to print the corresponding version.

rosenthalj:
> > @jase88
> > With this change,  how would a developer determine which version of PrimeNG the demo is running?  A while back the dropdown included the version number.
> > Angular_Table_Component
> 
> Would not be possible. But imho would be fine to use the latest version on the latest docs.
> 
> What are you suggesting? Alternatively the corresponding package.json could be used to print the corresponding version.

@jase88
I don't have an alternative solution.  Since I'm always using the demo as a reference,  it would be very useful to know the **exact** version of the demo.

mertsincan:
Good catch! Thanks a lot!

PR #18631, opened 2025-07-18
fix(InputNumber): fix decimal value on spin

### LINKED ISSUE #18630: InputNumber: wrong internal value for decimal numbers
### Describe the bug

Defining a InputNumber with minFractionDigits={1} and step={0.1} when pressing arrows button (+0.1) internal values are randomly incorrect:
value 0.1
value 0.2
value 0.30000000000000004
value 0.4
value 0.5
value 0.6
value 0.7
value 0.7999999999999999
value 0.9

### Pull Request Link

_No response_

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

.

### Environment

.

### Angular version

20.0.3

### PrimeNG version

v20

### Node version

20

### Browser(s)

_No response_

1. Set minFractionDigits=1 and step=0.1 
2. The internal values are incorrect 

### Expected behavior

The expected behavioor is that the e.value increases by 0.1

navedqb:
Fixes #18630

### DIFF
--- a/packages/primeng/src/inputnumber/inputnumber.ts
+++ b/packages/primeng/src/inputnumber/inputnumber.ts
@@ -674,10 +674,14 @@ export class InputNumber extends BaseInput implements OnInit, AfterContentInit, 
         this.spin(event, dir);
     }
 
+    addWithPrecision(base: number, increment: number, precision: number = 10): number {
+        return Math.round((base + increment) * precision) / precision;
+    }    
+
     spin(event: Event, dir: number) {
         let step = (this.step() ?? 1) * dir;
         let currentValue = this.parseValue(this.input?.nativeElement.value) || 0;
-        let newValue = this.validateValue((currentValue as number) + step);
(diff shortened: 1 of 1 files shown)


packages/primeng/src/inputnumber/inputnumber.ts
```diff
@@ -674,10 +674,14 @@ export class InputNumber extends BaseInput implements OnInit, AfterContentInit,
         this.spin(event, dir);
     }
 
+    addWithPrecision(base: number, increment: number, precision: number = 10): number {
+        return Math.round((base + increment) * precision) / precision;
```
rosenthalj, line 678: Since there is no way to change the precision,  this will result in an incorrect value for the following case:

base = 1.99
increment = 1

expected  result:  2.99
result after changes: 3

rosenthalj:
Dealing with floating point numbers is always tricky.  Should the component using the PrimeNG InputNumber add the rounding logic if necessary?

PR #18677, opened 2025-07-23
fix: Column dragend listener, fixes #18668

### LINKED ISSUE #18668: Table: column reorder indicator does not go away the column is dropped on invalid location
### Describe the bug

In a table with column reorder enabled, if you drag a column header so the indicators are visible and then drop it in a different (invalid) location, the drag stop and the column does not change location but the indicators remain visible. the only way to remove them is to drag another column.

Image

 

### Pull Request Link

https://github.com/primefaces/primeng/pull/18677

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/r9mphodt

### Environment

MacOS 15.5

### Angular version

20.1.0

### PrimeNG version

v20

### Node version

22.16.0

### Browser(s)

Safari 18.5, Chrome 138

1. Drag a column header to another position so the drag indicator arrows are visible (do not drop it)
2. Drop the column header in an invalid location, outside the table.
3. Drop indicator remain visible even though it is no longer in drag mode.

### Expected behavior

Drop indicators should also go away if the column header is dropped somewhere else and the column does not change its position.

afermon:
### Defect Fixes
In a table with column reorder enabled, if you drag a column header so the indicators are visible and then drop it in a different (invalid) location, the drag stop and the column does not change location but the indicators remain visible. the only way to remove them is to drag another column.

image

Closes: #18668

### DIFF
--- a/packages/primeng/src/table/table.ts
+++ b/packages/primeng/src/table/table.ts
@@ -2661,6 +2661,17 @@ export class Table<RowData = any> extends BaseComponent implements OnInit, After
         event.dataTransfer.setData('text', 'b'); // For firefox
     }
 
+    onColumnDragEnd(event: any) {
+        event.preventDefault();
+        if (this.draggedColumn) {
+            (<ElementRef>this.reorderIndicatorUpViewChild).nativeElement.style.display = 'none';
+            (<ElementRef>this.reorderIndicatorDownViewChild).nativeElement.style.display = 'none';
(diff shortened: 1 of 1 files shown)


afermon:
fixes https://github.com/primefaces/primeng/issues/18668

fkwq34:
Can we re-open this pull request? The issue still exists.

PR #18681, opened 2025-07-23
fix: Dialog: Escape listener closes everything always https://github.com/primefaces/primeng/issues/16955 

johanneszelger:
by using same logic in p-dialog that is also applied in p-dynamicdialog (z-index based check if dialog should be closed, e.g. when calendar is open inside dialog)

### Defect Fixes
https://github.com/primefaces/primeng/issues/16955


johanneszelger:
I cant manage to link the issue as "Development" is not clickable on my sidebar,..,

Issue #18618, opened 2025-07-16
ConfirmDialog: ignored acceptButtonProps


RadoslawTaborski:
### Describe the bug

In ConfirmDialog acceptButtonProps.label is ignored.

<img width="1919" height="869" alt="Image" src="https://github.com/user-attachments/assets/e7d7619c-8901-4508-95e6-30e708b8e57b" />

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://primeng.org/confirmdialog

### Environment

Windows 11

### Angular version

20.0.6

### PrimeNG version

v20

### Node version

22.12.0

### Steps to reproduce the behavior

1. Open https://primeng.org/confirmdialog
2. Go to 'Basic' section
3. Click 'Save' button

### Expected behavior

'Accept' button should have 'Save' label instead of 'Yes'

LukasSinger:
This is the case with `rejectButtonProps` as well. It doesn't show up in the example screenshot because the Delete button was pressed. That one has a standalone `rejectLabel` property, which does correctly set the property, and it remains on the `ConfirmationService` for all subsequent popups.

It makes sense that the standalone properties (`acceptLabel` and `rejectLabel`) wouldn't have counterparts in `acceptButtonProps` and `rejectButtonProps` respectively, because otherwise it would be ambiguous which property would take priority if both were defined. This is probably just a documentation issue.

RadoslawTaborski:
Perhaps it's just a documentation issue, although I've used it this way before (including in v19.1.3) and it worked. It stopped working in v20.

Now I see there's also another problem here.

I pressed the `Save` button, but before that I also pressed the `Delete` button. Dialog for `Save` button inherited the buttons from the `Delete` dialog:

https://github.com/user-attachments/assets/9416f501-2722-475d-bc81-3cdbfcefb5ab

LukasSinger:
True, you can observe the old behavior here still: https://v19.primeng.org/confirmdialog
This should be added to https://primeng.org/migration/v20#removals if the behavior isn't changed back.

gianluca-moro:
> Perhaps it's just a documentation issue, although I've used it this way before (including in v19.1.3) and it worked. It stopped working in v20.
> 
> Now I see there's also another problem here.
> 
[...]

Issue #18602, opened 2025-07-15
reusing p-confirmationdialog


honboubao:
### Describe the bug

Reusing a single p-confirmationdialog instance by calling confirmationService.confirm() with differing parameters does not work as expected.

Settings from previous calls are persistent if they are not explicitely passed in the configuration of later calls.

This happens because the properties of the config are copied over to the confirmationdialog instance: https://github.com/primefaces/primeng/blob/d5bb9892b92d8c26758c2f409bf797380747256c/packages/primeng/src/confirmdialog/confirmdialog.ts#L436
...from where they continue to have an effect on subsequent displays of the dialog.

### Reason for not contributing a PR

- [x] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/github-ubbpryky-wszbkqgr?file=src%2Fapp%2Fapp.component.html

### Environment

Windows 10
Firefox 128.3.1esr

### Angular version

20.0.6

### PrimeNG version

v20

### Steps to reproduce the behavior

1. call this.confirmationService.confirm({ message: 'test' });
2. dialog is displayed as expected with two buttons
3. call this.confirmationService.confirm({ message: 'test', rejectVisible: false });
4. dialog is displayed as expected with a single accept button
5. call this.confirmationService.confirm({ message: 'test' });
6. dialog is expected to show two buttons, but it only shows a single accept button

### Expected behavior

Calling this.confirmationService.confirm({ message: 'test' }); with a default instance of <p-confirmdialog></p-confirmdialog> should always show a dialog with accept and reject buttons. Settings from previous calls should not affect this call.

davisb10:
+1

We have noticed the same issues in our applications where the confirm dialog is used pretty heavily but different options each time.

PR #18684, opened 2025-07-24
Fix confirmation dialog reset options

AtiposEiji:
Fixed an issue in the ConfirmationDialog component in versions 19 and 20 where configuration options (such as message, icon, and button labels) were retained across multiple dialog openings even when not explicitly specified.

Previously, if you opened a dialog with certain settings, those settings would persist and be applied to subsequent dialogs, causing unexpected behavior.

With this fix, the options are reset to their default values every time the dialog is opened, ensuring each instance starts with a clean and consistent state.

This fix was implemented for version 19. However, the same issue also exists in version 20. I’m not sure about the workflow or process to address this for both versions simultaneously.

https://github.com/primefaces/primeng/issues/18618
https://github.com/primefaces/primeng/issues/18602


cetincakiroglu:
Hi,

Could you please resolve the conflicts so we can merge it.

AtiposEiji:
@cetincakiroglu Resolved :)

PR #18706, opened 2025-07-29
Fix p-multiselect to only use p-variant-filled when appropriate

johnhwright:
It is currently always applying the p-variant-filled class if _any_ value is provided in the variant.

You can see from the references - https://github.com/search?q=repo%3Aprimefaces%2Fprimeng%20p-variant-filled&type=code) it should be comparing to `filled` to match all the other components.


Chanachai-Pansailom:
Yes, I am also stuck with this problem.

Issue #18687, opened 2025-07-24
Table: p-columnFilter icon changes before filter is applied


YahorBabrouski:
### Describe the bug

The filter icon indicates that the filter has a value inside it, not that the filter has been applied and affects which rows are displayed.

### Reason for not contributing a PR

- [x] Lack of time
- [ ] Unsure how to implement the fix/feature
- [x] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/upicdyv5

### Environment

Win 11

### Angular version

20.0.4

### PrimeNG version

v20

### Node version

20.19.1

### Steps to reproduce the behavior

1. Open 'Name' or 'Country' filter
1. Start writing the filter value. For example, 'foobar'
1. Close the filter by clicking outside of it or pressing the 'Escape' key
1. The filter is not applied, but the icon is 'filter-fill'

### Expected behavior

The filter icon should not change until the filter is applied.

rra-topdev:
I made a PR to fix this #18743

rra-topdev:
The problem mentioned in this issue is still ongoing.

mehmetcetin01140:
Fixed in PrimeNG v22.

PR #18743, opened 2025-08-12
[#18687] fix: use filled filter icon when filter is applied

rra-topdev:
### Defect Fixes
[#18687](https://github.com/primefaces/primeng/issues/18687)

image


rosenthalj (changes requested):
Your PR code changes appear to fix the issue.

I am proposing changes to your PR for the following reasons:

1.  Remove redundant logic
2.  Remove duplicate code
3.  Centralize code for future enhancements (reduce the likelihood of future bugs when the code is updated)

Add the following method

```
	setHasFilter(newValue: boolean): void {
	    let fieldFilter = this.dt.filters[<string>this.field];
	    if (fieldFilter && newValue) {
	        if (Array.isArray(fieldFilter)){
				this.filterApplied = !this.dt.isFilterBlank((<FilterMetadata[]>fieldFilter)[0].value);	
			}
	        else  {
				this.filterApplied = !this.dt.isFilterBlank(fieldFilter.value);
			}
	    } else {
	    	this.filterApplied = false;
		}
	}
```

and replace  hasFilter()

with:

```
    get hasFilter(): boolean {
		return this.filterApplied;
    }
```

Note:  The two changes above move the logic from the old hasFilter() method to the new setHasFilter() method

Replace line 6117 with:

```
this.setHasFilter(false);
```
 
Replace line 6123 with:

```
this.setHasFilter(true);
```

Replace lines 6267, 6276-6279, 6285 with

```
this.colFilter.setHasFilter(true);
```

Note:  The logic for the last three changes have been simplified since the new setHasFilter() method will handle the additional logic that you just added for those lines

rosenthalj (commented):
The new changes look good.  I downloaded your changes and ran both the basic and advanced table column filter demos.  The changes appear to work, but ...

The advance filter demo's "clear filter" button  clears the table's column  filters but doesn't reset  each of the individual table column filter indicators.  I confirmed that both your initial PR change and your current PR change broke this functionality.   This is either a  problem induced by your changes or a  bug introduced to the demo code.  This need to be investigated and fixed.

The first video shows the official PrimeNG website's  advance table column filter demo.  The column filter indicator is broken (Issue #18687), but the demo's "clear filter" works.

https://github.com/user-attachments/assets/6d9e3dee-a766-4706-aa42-6e0221ef0171

The second video is running the PrimeNG advance table column filter demo with your changes .  The column filter indicator is now fixed (Issue #18687), but the demo's "clear filter" is broken.


https://github.com/user-attachments/assets/933d2c39-7873-4d83-90a5-70259e7fe7f7


Note:   In the current version of PrimeNg, The "county" column filter is broken in the  basic table column demo (and works in the advanced table filter demo).  Your  changes did not cause this issue.

rosenthalj:
Untested fix:

Update table's clearFilterValues() method:  For each column setHasFilter(false) or filterApplied = false.

rra-topdev:
> Untested fix:
> 
> Update table's clearFilterValues() method: For each column setHasFilter(false) or filterApplied = false.

This is not working because in clearFilterValues() this.columns and this._columns are both undefined.


But i added a commit which addresses this issue.
https://github.com/primefaces/primeng/pull/18743/commits/909f8795f10cc3114d48506f97794545ae580b68

rosenthalj (commented):
I have tested the following alternative change:

```
	get hasFilter(): boolean {
		if (!this.filterApplied) {
			return false;
		}
		// Because Table's clearFilterValues method may have been called (which clears all filters, but doesn't update filterApplied),  must call setHasFilter to make sure that filterApplied is up to date
		this.setHasFilter(true);
	    return this.filterApplied;
	}
```

rra-topdev:
@rosenthalj Thank you! added your change

rosenthalj (approved):
I would recommend adding the comment (or similar comment) to hasFilter or someone may just remove it without understanding the consequences

`// Because Table's clearFilterValues method may have been called (which clears all filters, but doesn't update filterApplied),  must call setHasFilter to make sure that filterApplied is up to date`

rosenthalj (approved):
:thumbsup:

rosenthalj:
@rra-topdev 

You may want to check out this issue:  **Calling Table.filter() causes the column filter menu to misbehave and not show value or condition #18764**.  I believe the solution to this issue is independent of your PR, but ...

The Issue #18764 is an implementation of column filtering triggered by  an external button action that calls a method which programmatically applies a column filter.  In the reproducer supplied for Issue #18764 ,  the current "hasFilter" works  and the column filter indicator properly shows that a filter has been applied when the not-table button is pressed (non-column filter action).  Unfortunately,  I'm pretty sure (untested) your new PR will break this use case.  I think you need to add additional calls to setHasFilter for this use case.

rra-topdev:
@rosenthalj  I will test this use case.

You're right. At the moment my fix breaks this use case.

rra-topdev:
@rosenthalj 

I've changed some logic and now it works in the demo application and also with the filter() method.

[record.webm](https://github.com/user-attachments/assets/6842aeb4-bb11-45f2-9dc0-b1b12ec6ec3f)

rosenthalj (changes requested):
Theses change appear to work.  Also these changes would  work for my application,  but I do not programmatically filter columns.

but ...

I don't think its a good idea to change the FilterMetadata interface.  

1. Your changes are forcing the developer who programmatically filter columns to also change/manage filter indicators.  I don't think this is a good use case and I think it will cause inconsistencies.  When programmatically setting/clearing a column filter,  PrimeNG should  be able to figure this out and change the column filter indicator.
2. filterApplied is optional in the FilterMetadata interface.  If it is not properly set,  the table column filter indicator will be wrong.  As a result, functionally the newly added field to the interface is required.  After your update,  developers who have code that programmatically changed column filters, will have no indication/direction that they  have to update their code.  This change to some developers will be equivalent of a "Breaking Code Change"

rra-topdev:
@rosenthalj 
I understand your concerns and have therefore reverted the changes.
I have added a minimal [commit](https://github.com/primefaces/primeng/pull/18743/commits/472e38f5950a854466b363107c9a049eadcde23d) that achieves the desired behavior without forcing a code change.

rosenthalj (commented):
@rra-topdev 

line 5701 how can this work?
I thought fieldConstraints was an array of  FilterMetadata

On line 2150  you added applyFilter which I think is a neat idea.  But it never gets cleared.  I believe that some of the column filter code updates the filter object and doesn't replace it, as a result the applyFilter  (added on line 2150) will become a barnacle and may cause errors with hasFilter.

rra-topdev:
@rosenthalj line 5701 works because when you call the filter() method or use the Basic table example _fieldConstraints_ will return an _FilterMetadata_ object and not an array. The type is missing in line 5652.
image

Please use the console output to confirm my statement @rosenthalj.

rra-topdev:
@rosenthalj I added a new [commit](https://github.com/primefaces/primeng/pull/18743/commits/80d17de12569b3804f086664c2a5fbfe3f7ec35f) which removes the property.

rosenthalj (changes requested):
@rra-topdev 
I took another  look at the code and you are correct about the filter method (note:   I did  check the console with your changes). 

The additional changes for edge cases are getting to be complicated and "hidden"  to anyone who looks at the code without some comments.

Looking at your code I think this will work and be a little cleaner.

Replace get hasFilter() with the following (untested):

```
    get hasFilter(): boolean {
        // Special case, this can only happen when legacy "filter(value: any, field: string, matchMode: string) " is programmatically called
        if ((<any>this.fieldConstraints)?.applyFilter) {
            (<any>this.fieldConstraints).applyFilter = null;
            setHasFilter(true);
        }
        if (!this.filterApplied) {
            return false;
        }
        // Because Table's clearFilterValues method may have been called (which clears all filters, but doesn't update filterApplied), must call setHasFilter to make sure that filterApplied is up to date.
        this.setHasFilter(true);
        return this.filterApplied;
    }
```

rra-topdev:
@rosenthalj replace setFilter method?

rosenthalj:
> @rosenthalj replace setFilter method?

@rra-topdev 

My bad.

I just updated my previous response.  Update "**get hasFilter()**"

rra-topdev:
@rosenthalj 

What's the reason to set applyFilter to null instead setting it to undefined or remove the property?

rosenthalj:
> @rosenthalj
> 
> What's the reason to set applyFilter to null instead setting it to undefined or remove the property?

@rra-topdev 
Using null is my personal preference.  Using undefined or removing the property will also work. I actually don't know what's approach.

rra-topdev:
@rosenthalj Could you please re-check your suggestions from [comment](https://github.com/primefaces/primeng/pull/18743#pullrequestreview-3109864424) and how i implemented them? In the showcase it is working, but i'm not sure whether it's correct.

rosenthalj:
> @rosenthalj Could you please re-check your suggestions from [comment](https://github.com/primefaces/primeng/pull/18743#pullrequestreview-3109864424) and how i implemented them? In the showcase it is working, but i'm not sure whether it's correct.

@rra-topdev 
What is your concern?

Most importantly, I assume you have verified that your changes work for the following:

- PrimeNG basic column filter demo
- PrimeNG advanced column filter demo
- Programmatic use case (Issue #18764)

I agree that "get hasFilter()" is  "not the nicest" code,  but it  would be "nice" if it didn't handle the two "special" cases.  Special processing always causes problems,  when you don't want to redesign code. 

This is your PR and if the three cases above are satisfied you should feel free to submit any solution you feel is best to PrimeNG for their approval. I try to monitor all changes to Table,  because it's key functionality to my application and in the past submitted/approved PR have unexpectedly broken Table functionality that is critical to my application.

rra-topdev:
@rosenthalj 
You [said](https://github.com/primefaces/primeng/pull/18743#pullrequestreview-3109864424) to replace the [hasFilter()](https://github.com/primefaces/primeng/blob/2c3288714a5f866ec19b3c8da81e7dbb3038496f/packages/primeng/src/table/table.ts#L2294) method which belongs to the Table and replace it with **get hasFilter()** which tries to access **this.filterApplied** property which does not exists on the table. That's what i'm concerned about.

In my project the solution seems not to work completely because of this.

rosenthalj:
> @rosenthalj You [said](https://github.com/primefaces/primeng/pull/18743#pullrequestreview-3109864424) to replace the [hasFilter()](https://github.com/primefaces/primeng/blob/2c3288714a5f866ec19b3c8da81e7dbb3038496f/packages/primeng/src/table/table.ts#L2294) method which belongs to the Table and replace it with **get hasFilter()** which tries to access **this.filterApplied** property which does not exists on the table. That's what i'm concerned about.
> 
> In my project the solution seems not to work completely because of this.

I'm a still confused and I''m sure I was the one who started all the confusion.  I probably made a  reference to  "get hasFilter()" incorrectly.  

**Regardless** 

I like your last commit  [#82fbd3211a1c27ead68f39d7052fbbd7ed68cfab](https://github.com/primefaces/primeng/pull/18743/commits/82fbd3211a1c27ead68f39d7052fbbd7ed68cfab).  I like the current state of the PR code.  As long as your current changes  have been tested,  I think your PR is good to go.  Also,  your comments in the code will help ensure that future changes to Table and Column Filter will less likely break this functionality.

rra-topdev:
@rosenthalj 
I will review the code as soon as possible and let you know when i verified the correct functionality in my personal project.

rosenthalj (commented):
@rra-topdev 

I like your comments.

Is comment on line 6045 correct if so is lines 2967-2969 needed?
(Note:   I have never worked with table state)

rra-topdev:
@rosenthalj 

I [updated](https://github.com/primefaces/primeng/pull/18743/commits/425fcd4fec25e2191e5a5328e87d2859637e94ad) the comment in line 6045.

Yes this is needed. In my application the table state is restored. This is quite similar to calling filter() programmatically. The difference is:
- **filter():** When called **this.fieldConstraints** in **get hasFilter()** will be **FilterMetadata**, never **FilterMetadata[]**
- r**estoreState()**: When called **this.fieldConstraints** in **get hasFilter()** can be **FilterMetadata**, or **FilterMetadata[]** - depending on what your saved table state was.

I will re-test everything and after that i will re-request reviewing the code changes from you.

rosenthalj:
> @rosenthalj
> 
> I [updated](https://github.com/primefaces/primeng/pull/18743/commits/425fcd4fec25e2191e5a5328e87d2859637e94ad) the comment in line 6045.
> 
> Yes this is needed. In my application the table state is restored. This is quite similar to calling filter() programmatically. The difference is:
> 
> * **filter():** When called **this.fieldConstraints** in **get hasFilter()** will be **FilterMetadata**, never **FilterMetadata[]**
> * r**estoreState()**: When called **this.fieldConstraints** in **get hasFilter()** can be **FilterMetadata**, or **FilterMetadata[]** - depending on what your saved table state was.
> 
> I will re-test everything and after that i will re-request reviewing the code changes from you.

@rra-topdev 
I probably wrote my previous question poorly.

I understand your newest changes and I believe your state logic is correct (unverified). 

I think the old comment in hasFilter:

**// Special case, this can only happen when legacy "filter(value: any, field: string, matchMode: string)" is programmatically called or when the table state is restored with filter value (type object)**

is now incorrect because of lines 2967-2969 and needs to be updated.    There are now **two** special cases for lines 6044-6047

rra-topdev:
@rosenthalj I updated the comments

rra-topdev:
I tested various test cases and from my side the pr is ready now. Thank you @rosenthalj

rosenthalj (commented):
👍

mehmetcetin01140:
Fixed in PrimeNG v22

Issue #18791, opened 2025-08-21
The minus icon in indeterminate checkbox is not visible in v20.0.1


rokamon1997:
### Describe the bug

After upgrading to v20.0.1 the minus in the indeterminate checkbox is not visible anymore which can also be seen on the component demo page:

https://primeng.org/checkbox

I've also checked and it was displaying fine in v20.0.0. So the issue appears from v20.0.1 specifically.

### Reason for not contributing a PR

- [x] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://primeng.org/checkbox

### Environment

Node - v20.19.4
Angular - v20.2.1
Primeng - v20.0.1

### Angular version

20.2.1

### PrimeNG version

v20

### Node version

20.19.4

### Browser(s)

Chrome 139.0.7258.128

### Steps to reproduce the behavior

Add a checkbox with indeterminate:

https://primeng.org/checkbox

### Expected behavior

It should show the `minus` icon correctly.

The-Real-1:
Also facing the same problem on the same version.

Edit : The problem is also visible in the Treetable with the select mode set to checkbox

cetincakiroglu:
fixed in https://github.com/primefaces/primeng/pull/18794/files

The-Real-1:
> fixed in https://github.com/primefaces/primeng/pull/18794/files

🤩🤩🤩

PR #18794, opened 2025-08-21
fix/restore_minus_icon_on_indeterminate_checkbox

Elyes3:
### What is the purpose of this PR?

Fixes the issue where an indeterminate button in the Checkbox component did not display the minus icon as expected.

This addresses the existing issue #18791.

### Changes made:

- Imported the missing minus icon into the project.
- Updated the Checkbox component to render the minus icon for indeterminate state.
- Ensured proper CSS classes are applied so the icon displays correctly.
- Verified the change in the showcase app to confirm visual correctness.

### Notes

- Only affects the Checkbox component.
- No breaking changes; fully backward-compatible.


Elyes3 (commented):
Minor changes. Added minus icon.

Issue #18802, opened 2025-08-25
Select: Select field becomes invalid/dirty before releasing selected option


shinkhouse:
### Describe the bug

When using reactive forms with a required select field, clicking on an option briefly flashes the invalid state before going to valid state. I also found that the select field will be invalid while holding down the selected option before releasing it. This issue occurs in both v19 and v20. 

### Reason for not contributing a PR

- [x] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/hzcumnrt?file=src%2Fapp%2Fselect-reactive-forms-demo.ts,src%2Fapp%2Fselect-reactive-forms-demo.html,src%2Fmain.ts

### Environment
```
---------------------------------------
@angular-devkit/architect       0.2002.0
@angular-devkit/build-angular   20.2.0
@angular-devkit/core            20.2.0
@angular-devkit/schematics      20.2.0
@angular/cdk                    20.2.0
@angular/cli                    20.2.0
@schematics/angular             20.2.0
rxjs                            7.8.2
typescript                      5.8.3
zone.js                         0.15.1
```

### Angular version

20.0.0

### PrimeNG version

v20

### Node version

20.19.1

### Browser(s)

Chrome 139.0.7258.139 

### Steps to reproduce the behavior

See video:

https://github.com/user-attachments/assets/8cc269ed-1cde-4bb6-accd-354d0c57188d

### Expected behavior

The field should not become invalid because the value hasn't been selected yet and it should also not be dirty.

cetincakiroglu:
fixed in https://github.com/primefaces/primeng/pull/18817

PR #18817, opened 2025-08-29
fix/remove invalid behaviour before releasing selected option

Elyes3:
## What is the purpose of this PR?

Fixes the issue where the **Select** component briefly displays a validation error when a required field is interacted with but the selected option is not yet released. The invalid state also flashes when the selection is released.

This addresses the existing issue [#18802]
---

## Changes made

- **Updated the `onInputBlur` method:**  
  Only marks the control as "touched" if the overlay is not visible and model touch prevention is not active.
- **Prevents premature validation error display:**  
  Ensures errors do not show while the dropdown is open or the user is still interacting.
- **Verified the change in the showcase app:**  
  Confirmed the error no longer flashes unexpectedly.

---

## Notes

- Only affects the **Select** component.
- No breaking changes; fully backward-compatible.

### DIFF
--- a/packages/primeng/src/select/select.ts
+++ b/packages/primeng/src/select/select.ts
@@ -1383,7 +1383,7 @@ export class Select extends BaseInput implements OnInit, AfterViewInit, AfterCon
         this.focused = false;
         this.onBlur.emit(event);
 
-        if (!this.preventModelTouched) {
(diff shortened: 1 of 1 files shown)


shinkhouse:
@cetincakiroglu Is it possible to release this for PrimeNG v19 too?

PR #18830, opened 2025-09-02
fix: primefaces#18821, Select: EmptyFilterMessage does not work on Select with virtualscroll

akshayaqburst:
fix: primefaces#18821, Select: EmptyFilterMessage does not work on Select with virtualscroll

https://stackblitz.com/edit/mayx8jwd?file=src%2Fapp%2Fselect-virtualscroll-demo.html


akshayaqburst:
@cetincakiroglu Please review and merge, thanks.

PR #18837, opened 2025-09-03
fix(Menu): fixes sanitizeHtml is undefined

### LINKED ISSUE #18836: Menu - sanitizeHtml is undefined because inject() is called outside the injection context
### Describe the bug

When escape is set to false on a MenuItem, santizeHtml is undefined and this results in a loop of exceptions.

### Pull Request Link

_No response_

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/github-uknitcug?file=src%2Fapp%2Fapp.component.ts

### Environment

I used the primeng showcase to reproduce this behaviour with version 20.1.0

### Angular version

20

### PrimeNG version

v20

### Node version

_No response_

### Browser(s)

Chrome

just open the example provided from Stackblitz

### Expected behavior

sanitizeHtml works as intended and does not crash the application

maxschroers:
when item.escape === false sanitizeHtml is undefined resulting in exceptions

Closes #18836

this probably occurs because a function outside a component is not avaliable in a components template. Moreover this wouldn't have worked anyways, even if it would have been avaliable for the template, because inside the function inject() is called outside the injection context of angular.

### DIFF
--- a/packages/primeng/src/menu/menu.ts
+++ b/packages/primeng/src/menu/menu.ts
@@ -61,21 +61,10 @@ export class SafeHtmlPipe implements PipeTransform {
     }
 }
 
-export function sanitizeHtml(value: string): SafeHtml | string {
-    const platformId = inject(PLATFORM_ID);
-    const sanitizer = inject(DomSanitizer);
-
-    if (!value || !isPlatformBrowser(platformId)) {
-        return value;
-    }
-
-    return sanitizer.bypassSecurityTrustHtml(value);
-}
-
 @Component({
     selector: '[pMenuItemContent]',
     standalone: true,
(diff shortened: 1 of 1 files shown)


sandrotonon:
The menu component is basically unusable due to this error, if you want to escape the content of a MenuItem.
I had to refactor the menu to be template driven. Now it works as intended. But this may not be applicable in other scenarios.

maxschroers:
> The menu component is basically unusable due to this error, if you want to escape the content of a MenuItem. I had to refactor the menu to be template driven. Now it works as intended. But this may not be applicable in other scenarios.

yeah, thats why i reverted the changes in this pullrequest and this should fix it now

PR #18854, opened 2025-09-08
[#18853] return true if the filter matchmode notEquals is undefined, null or empty string

### LINKED ISSUE #18853: Filter matchmode notEquals is broken
### Describe the bug

MatchMode **notEquals** should return **true** if the filter is undefined, null or empty string. Currently it returns **false**.

### Pull Request Link

[PR](https://github.com/primefaces/primeng/pull/18854)

### Reproducer

https://github.com/primefaces/primeng/blob/master/packages/primeng/src/api/filterservice.ts#L100

### Environment

PrimeNG Table

### Angular version

20.2.4

### PrimeNG version

v20

### Node version

24.5.0

### Browser(s)

Chrome 140

use the clearFilterValues() method from primeng table.

### Expected behavior

MatchMode **notEquals** should return **true** if the filter is undefined, null or empty string.

rra-topdev:
### Defect Fixes
#18853


rosenthalj:
~~This change would make  matchmode notEquals inconsistent with the other matchmodes.~~

~~If this PR is accepted,  this would be a breaking change.~~

~~Maybe the solution to your problem is to add a new boolean parameter that would change the behavior of all matchmodes?~~

rra-topdev:
@rosenthalj 
I dont understand why my change would make notEquals inconsistent?

I maintain that my change would restore the consistency of the filter in conjunction with the others because consistency is currently lacking.

I will explain it to you:

In the [filterservice.ts](https://github.com/primefaces/primeng/blob/4636c91ec1998a116b7f36af6dcf8eea7cdd8868/packages/primeng/src/api/filterservice.ts) every matchmode returns **true** when the filter is _empty_ (except **notEquals**).
For Example:
`
dateIs: (value: any, filter: any): boolean => {
if (filter === undefined || filter === null) {
    return true;
}
`

`
dateIsNot: (value: any, filter: any): boolean => {
if (filter === undefined || filter === null) {
    return true;
}
`

`
between: (value: any, filter: any[]): boolean => {
if (filter == null || filter[0] == null || filter[1] == null) {
    return true;
}
`

`
notContains: (value: any, filter: any, filterLocale?: any): boolean => {
if (filter === undefined || filter === null || (typeof filter === 'string' && filter.trim() === '')) {
    return true;
}
`



**notEquals** returns false when the filter is "empty" which makes no sense. When restoring the table with an _empty_ value for **notEquals**, the table is empty because the matchMode returns **false** on every element in the table.

rosenthalj:
> @rosenthalj I dont understand why my change would make notEquals inconsistent?
> 
> I maintain that my change would restore the consistency of the filter in conjunction with the others because consistency is currently lacking.
> 
> I will explain it to you:
> 
> In the [filterservice.ts](https://github.com/primefaces/primeng/blob/4636c91ec1998a116b7f36af6dcf8eea7cdd8868/packages/primeng/src/api/filterservice.ts) every matchmode returns **true** when the filter is _empty_ (except **notEquals**). For Example: `dateIs: (value: any, filter: any): boolean => { if (filter === undefined || filter === null) { return true; }`
> 
> `dateIsNot: (value: any, filter: any): boolean => { if (filter === undefined || filter === null) { return true; }`
> 
> `between: (value: any, filter: any[]): boolean => { if (filter == null || filter[0] == null || filter[1] == null) { return true; }`
> 
> `notContains: (value: any, filter: any, filterLocale?: any): boolean => { if (filter === undefined || filter === null || (typeof filter === 'string' && filter.trim() === '')) { return true; }`
> 
> **notEquals** returns false when the filter is "empty" which makes no sense. When restoring the table with an _empty_ value for **notEquals**, the table is empty because the matchMode returns **false** on every element in the table.

@rra-topdev 
My comment about inconsistency is incorrect.  Your  PR will make notEquals consistent with the other  matchmodes.

PR #18373, opened 2025-05-21
fix(18369): fix autocomplete form submit when pressing enter

gianluca-moro:
Fixes #18369 

We skip `event.preventDefault()` when the overlay is not visible. This ensures that the form submit works as intended. Note that before, `this.onArrowDownKey(event)` would have returned immediately anyway since it has an early return for `if (!this.overlayVisible)`, thus we can get rid of it.

PR #18896, opened 2025-09-18
fix: autocomplete form submit when pressing enter (take 2)

perceptron8:
Since #18373, [Enter] has triggered a form submit even if [multiple]="true". `event.preventDefault()` should be called not only when the overlay is visible, but also when multiple selections are allowed.


perceptron8:
Fixes #18897

Issue #18953, opened 2025-09-29
Picklist: move all buttons don't update input source or target references


dylanbehetre:
Hello,

I saw an issue in my project about source and target reference set in input of picklist component.

Change about source and target were missing. I saw that copy was used in moveAllRight and moveAllLeft function.

I expect that using push method and not a copy will fix this issue.

cetincakiroglu:
Hi,

Please follow the issue format and provide a StackBlitz reproducer and link the PR with the issue correctly. We can't identify "the issue in your project".

aleesaan:
Hi @cetincakiroglu I'm also having this issue, I created a stackblitz to reproduce it so you can see it:
https://stackblitz.com/edit/mcdmkgew?file=src%2Fapp%2Fpicklist-basic-demo.ts

I simply copied the [basic](https://primeng.org/picklist#basic) picklist from the docs, and added the `targetProducts` names in the template. You can see that when clicking the select all button, they're not updated.
In fact, even by using the single select button, the list is only updated the first time, any subsequent updates are ignored. So this issue is probably related to https://github.com/primefaces/primeng/issues/18997 and https://github.com/primefaces/primeng/issues/18969.

If it can help, this issue was introduced in v20.2.0 (when drag and drop was fixed 😅), it wasn't there till v20.1.2.

dylanbehetre:
Hi @cetincakiroglu,

I did a [PR](https://github.com/primefaces/primeng/pull/18954) to fix this problem one month ago. The problem was obvious for me : `update on input references aren't did in input reference due to copy` so I doesn't want to waste time that I doesn't have for that. Sorry if it wasn't but just ask for it if needed.  Thanks @aleesaan for your example.

We don't have enough right to `link the PR with the issue correctly`, we can just set a title and a message.

dylanbehetre:
#18954

PR #18954, opened 2025-09-29
fix: fix moveAllRight and moveAllLeft picklist buttons

dylanbehetre:
Fix moveAllRight and moveAllLeft picklist buttons to continue to update parent source and target references (no copy which break the parent vision).

https://github.com/primefaces/primeng/issues/18953

### DIFF
--- a/packages/primeng/src/picklist/picklist.spec.ts
+++ b/packages/primeng/src/picklist/picklist.spec.ts
@@ -748,14 +748,20 @@ describe('PickList', () => {
         }));
 
         it('should allow drag&drop after moveAllRight', fakeAsync(() => {
+            // Verify initial state and keep tracking source and target form original references as a parent component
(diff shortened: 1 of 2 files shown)


jrcunningham21:
Also having this issue, would love for this to be fixed

dylanbehetre:
Hello @cetincakiroglu ! Is something missing about this PR to be merged ?
If community fix about primeng issues aren't merged, community will stop to fix issues.

Xriuk:
Just a suggestion: instead of copying all the elements inside the for loops, for each element, you could just copy them all at once after the loop like this:
```ts
if(picklist.target)
    picklist.target = [...picklist.target];
```

dylanbehetre:
I close it because this PR isn't considered by primeng team and this bug should be fixed by another PR arrived later and merged

PR #18512, opened 2025-06-20
fix(drawer): set default transition options when full screen sets to false

AlexFHer:
fix #17504

PR #18964, opened 2025-10-01
Fixed an issue that the drawer always opens from the left after fulls…

mabakl:
…creen.

Adds on https://github.com/primefaces/primeng/pull/18512 to fix an issue that the drawer defaults to opening from the left after fullscreen.

Issue: https://github.com/primefaces/primeng/issues/18965


danielvieira1:
I would have done it this way to avoid introducing another class property:

https://github.com/primefaces/primeng/pull/18512#issuecomment-3384597863

PR #18971, opened 2025-10-02
fix: #12595 - Two onSort emits were triggered, when both sortField and sortOrder changed at once

### LINKED ISSUE #12595: p-table emits multiple lazyLoading events when we have dynamic [sortField]
### Describe the bug

We have component template that looks like this:
`<p-table 
      [scrollable]="true"
      scrollHeight="flex"
      responsiveLayout="scroll"
      [columns]="header"
      [value]="items" 
      [paginator]="true"
      [rows]="config.pageSize"
      [pageLinks]="config.pageLinkSize"
      [rowsPerPageOptions]="[10, 50, 100, 200]"
      [totalRecords]="config.totalCount"
      [lazy]="true" 
      (onLazyLoad)="onLazyLoading($event)"
      [sortMode]="'single'"
      [sortField]="settings.sortField"
      [sortOrder]="settings.sortOrder"
      (onRowSelect)="onRowSelected($event)">
.
.
.
</p-table>`

Almost all the columns that I have are sortable and filterable. When I initially load the table the sortField and the sortOrder are null.
The first problem that I have is every time when I click the sort icon from the table header the lazy loading function is called 2 times therefore we have 2 backend calls. 
Another problem is also when I have some sort applied and save the table state and than when I come back to this page restore the state I also have 2 lazy loading calls.
Is there a way to restore the table state without extra calls. Is there a way to tell the table that there is a pre selected sort column and the table to mark this column without triggering onLazyLoading call ?
Ideally would be whenever I do
 `table.sortField = settings.sortField;
  table.sortOrder = settings.sortOrder`
  the table to mark them as selected or highlighted.

### Environment

Windows 11 Pro
Processor	AMD Ryzen 7 5700U with Radeon Graphics            1.80 GHz
Installed RAM	16,0 GB (14,8 GB usable)
System type	64-bit operating system, x64-based processor

### Reproducer

_No response_

### Angular version

14.2.10

### PrimeNG version

14

### Build / Runtime

Angular CLI App

### Language

TypeScript

16.16.0

### Browser(s)

Chrome

_No response_

### Expected behavior

_No response_

EstelonAgarwaen:
Fixes: https://github.com/primefaces/primeng/issues/12595

When using `p-table` with `lazy=true` and setting the `sortOrder` and `sortField`-Inputs, the `onSort`-Emitter emitted two events. 

This PR fixes that issue, so that no matter which constellation of changes these inputs receive, there will be one emit on the `onSort`-emitter.

### DIFF
--- a/packages/primeng/src/table/table.spec.ts
+++ b/packages/primeng/src/table/table.spec.ts
@@ -120,6 +120,37 @@ describe('Table', () => {
     @Component({
         standalone: false,
         template: `
+            <p-table [value]="products" [sortField]="sortField" [sortOrder]="sortOrder" [lazy]="true">
+                <ng-template #header>
+                    <tr>
+                        <th pSortableColumn="name">Name <p-sortIcon field="name"></p-sortIcon></th>
+                        <th pSortableColumn="price">Price <p-sortIcon field="price"></p-sortIcon></th>
+                        <th pSortableColumn="category">Category <p-sortIcon field="category"></p-sortIcon></th>
+                    </tr>
+                </ng-template>
+                <ng-template #body let-product>
+                    <tr>
+                        <td>{{ product.name }}</td>
+                        <td>{{ product.price | currency }}</td>
+                        <td>{{ product.category }}</td>
+                    </tr>
+                </ng-template>
+            </p-table>
+        `
+    })
+    class TestSingleSortingTableComponent {
+        sortField = 'name';
+        sortOrder = -1;
+        products = [
+            { id: '1001', name: 'Gaming Laptop', price: 1299.99, category: 'Electronics' },
+            { id: '1002', name: 'Wireless Mouse', price: 29.99, category: 'Accessories' },
+            { id: '1003', name: 'Mechanical Keyboard', price: 149.99, category: 'Accessories' }
+        ];
+    }
+
+    @Component({
+        standalone: false,
+        template: `
             <p-table [value]="products" [globalFilterFields]="['name', 'category']">
                 <ng-template #header>
                     <tr>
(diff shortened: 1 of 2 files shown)


packages/primeng/src/table/table.ts
```diff
@@ -1382,18 +1382,15 @@ export class Table<RowData = any> extends BaseComponent<TablePassThrough> implem
             }
         }
 
-        if (simpleChange.sortField) {
+        // if both the sort field and the sort direction change at once,
+        // extra care is needed to prevent the sort and lazy load events from being emitted twice
+        if(simpleChange.sortField) {
```
Maran23: Minor: Indentation is off

packages/primeng/src/table/table.ts
```diff
@@ -1382,18 +1382,15 @@ export class Table<RowData = any> extends BaseComponent<TablePassThrough> implem
...
-            //avoid triggering lazy load prior to lazy initialization at onInit
-            if (!this.lazy || this.initialized) {
-                if (this.sortMode === 'single') {
-                    this.sortSingle();
-                }
-            }
         }
-
-        if (simpleChange.groupRowsBy) {
+        if(simpleChange.sortOrder) {
```
Maran23: Indentation is off

packages/primeng/src/table/table.ts
```diff
@@ -1382,18 +1382,15 @@ export class Table<RowData = any> extends BaseComponent<TablePassThrough> implem
...
+            this._sortOrder = simpleChange.sortOrder.currentValue;
+        }
+        if(simpleChange.sortField || simpleChange.sortOrder) {
```
Maran23: Indentation is off

packages/primeng/src/table/table.spec.ts
```diff
@@ -361,6 +392,78 @@ describe('Table', () => {
...
+
+            testComponent.sortOrder = 1;
+            testFixture.detectChanges();
+
+            expect(tableInstance.sortField).toBe('name');
+            expect(tableInstance.sortOrder).toBe(1);
+            expect(emitterSpy).toHaveBeenCalledTimes(1);
+        });
+
+        it('should only emit one sort event when onyl field changes', () => {
```
Maran23: Minor: Typo

Maran23 (changes requested):
Fix makes sense. Have some minor comments.

Nice to see some more tests for the sorting!
I can confirm that one test fails before and passes after. The other tests pass before and after, as expected.

Maran23 (approved):
Looks good now!

Maran23 (approved):
Still good. Hope there will be an official review at one point. Thanks for changing it to the new testing approach.

mehmetcetin01140:
Fixed in PrimeNG v22

Issue #18536, opened 2025-06-27
Unresponsive browser if pressing ESC-key in combination with lots of popovers


gioid:
### Describe the bug

In our applications we sometimes have quite large pages with complex component structures and many (PrimeNg) component instances. Each instance can display a custom popover or tooltip.

If we use popovers, we face performance issues with ESC-key handling in general. So e.g. if wanting to close the popover or we just press ESC-key by mistake, it can take several seconds until the popover disappears and the page returns responsive.

Maybe is something similar to what happened with the issue https://github.com/primefaces/primeng/issues/16729

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [x] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/github-meubaa3m?file=src%2Fapp%2Fapp.component.ts

### Environment

Reproducible with current PrimeNG versions (19.x and older).

### Angular version

19.2

### PrimeNG version

v19

### Steps to reproduce the behavior

1. Open the stackblitz on a new tab
2. hover one of the number on the page to show the popover
3. press ESC-key

-> the tab freezes for a few seconds, depending on the amount of elements on the table

### Expected behavior

The popover closes immediately and browser must not get unresponsive in any case.

gioid:
@cagataycivici @mehmetcetin01140 @cetincakiroglu
Hi! Just checking in to see if there are any updates on this Issue. In particular on the PR that I've created to fix it.
Please let me know if it won’t be merged because it targets an older version (v19), so I can adjust or close it accordingly. Thanks!

PR #18990, opened 2025-10-07
Fix: Popover(18536) Unresponsive browser if pressing ESC-key in combination with lots of popovers

gioid:
Fix for [#18536](https://github.com/primefaces/primeng/issues/18536)

I modified the behavior so that the document:keydown event listener is only active when the Popover is visible. This change significantly reduces the number of active @HostListener, improving overall performance and preventing browser freezes.

### DIFF
--- a/packages/primeng/src/popover/popover.ts
+++ b/packages/primeng/src/popover/popover.ts
@@ -496,6 +496,7 @@ export class Popover extends BaseComponent implements AfterContentInit, OnDestro
         this.unbindDocumentClickListener();
         this.unbindDocumentResizeListener();
         this.unbindScrollListener();
+        this.unbindDocumentEscapeListener();
     }
 
     ngOnDestroy() {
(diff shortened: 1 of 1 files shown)


rosenthalj (commented):
I quickly looked at your changes.

Do you need to add

`this.unbindDocumentEscapeListener();`

to the onContainerDestroy method?

gioid:
> I quickly looked at your changes.
> 
> Do you need to add
> 
> `this.unbindDocumentEscapeListener();`
> 
> to the onContainerDestroy method?

You're right, I'll do it. Thanks for the feedback

rosenthalj (commented):
👍

gioid:
@mehmetcetin01140 I don't know if I can ask to you. Someone can take a look to this PR and merge it? The issue that this PR addresses is very important on our opinion.

gioid:
@cagataycivici @mehmetcetin01140 @cetincakiroglu
Hi! Just checking in to see if there are any updates on this PR.
Please let me know if it won’t be merged because it targets an older version (v19), so I can adjust or close it accordingly. Thanks!

gioid:
We think that the PR is still relevant.
We can also port it to newer version if approved.

Issue #16483, opened 2024-10-03
p-password toggleMask


vikore:
### Describe the bug

Keyboard navigation with tab/arrow key is not working on toggeMask icon eyeicon.
Please find the below video -

https://github.com/user-attachments/assets/2cce25bc-df2b-408b-a73a-54759f8b200e

### Environment

Development

### Angular version

17.3.6

### PrimeNG version

17.16.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

18.16.1

### Browser(s)

Chrome, Firfox, safari, IE-Edge

### Steps to reproduce the behavior

1. Navigate to password input having toggleMask property using keyboard tab key
2. Press tab key
3. Observe the behavior it will skip toggleMask icon i.e. eye icon and will jump to next input/element.

### Expected behavior

Pressing tab key from password input should navigate to toggleMask i.e. eye icon when toggleMask is added to password input

fgeierst:
This is a critical feature for keyboard users and users of other assistive technologies.

fgeierst:
https://github.com/primefaces/primeng/pull/19039 — there is a PR, can this be merged?

fgeierst:
Another note for this button is that the target size (not necessarily the visible eye icon) must be at least 24x24px, compare https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html
<img width="513" height="107" alt="Image" src="https://github.com/user-attachments/assets/ed74c224-8b8a-4c54-a8de-1dbd45c948c5" />

PR #19039, opened 2025-10-28
fix: #16483 - add accessibility features to password toggle icons

Zerak630:
Add missing attributes :
- `role="button"` : help screen readers to understand icon as a button
- `tabindex="0"`: Add icon to keyboard navigation
- `(keypress)=""` : activate "Enter" instead of click-only
- `[aria-label]`: needed to identify button usage. Default to "Hide password / Show password"

Hope it helps !


fgeierst:
Nice you are working on this, thank you!! Would it make sense to wrap the the svg into a <button> element rather to work with aria attributes. That could be more robust in the future.

PR #19076, opened 2025-11-10
fix(#19075): autocomplete optionLabel for selected item with string values correctly displayed

### LINKED ISSUE #19075: [AutoComplete] optionLabel not applied to selected item if suggestions are strings
### Describe the bug

For AutoComplete components, we can use `optionLabel` to display a custom label in the suggestions and selected item.   
This works fine when the suggestions are objects but it does not work properly when the suggestions are just strings. In that case, the selected item does not display the label provided by `optionLabel` but just the value itself, while with objects, the selected item displays the label generated by `optionLabel`.

Image

<br/><br/>

Edit:   
After some initial investigation of the source code, I saw in https://github.com/primefaces/primeng/blob/993680d9202bce66570663d2c3c3eeda6860e56d/packages/primeng/src/autocomplete/autocomplete.ts#L860-L875 and https://github.com/primefaces/primeng/blob/993680d9202bce66570663d2c3c3eeda6860e56d/packages/primeng/src/autocomplete/autocomplete.ts#L921-L923 that the `optionLabel` is only applied to the selected item if the value is an object or if it's a string and the `optionValue` property is set.    
But this does not really make sense to me since why would you set `optionValue` if the value is a string??    
I think the logic in the `inputValue` signal computation is flawed. In my opinion, `optionLabel` should always be applied, no matter the type of the value. 

So, apparently there is a workaround to fix this issue: Add an `optionValue` function that just returns the item itself, aka:
```javascript
optionValue(item: any): any {
  return item;
}
```

### Pull Request Link

https://github.com/primefaces/primeng/pull/19076

### Reproducer

https://stackblitz.com/edit/dz8njim1-jg7bfq1g?file=src%2Fapp%2Fautocomplete-objects-demo.html

### Environment

PrimeNG 20.3.0, Angular 20, Firefox

### Angular version

20.0.4

### PrimeNG version

v20

### Browser(s)

Firefox

1. Go to https://stackblitz.com/edit/dz8njim1-jg7bfq1g?file=src%2Fapp%2Fautocomplete-objects-demo.html
2. The top AutoComplete uses an array of objects for the suggestions, while the bottom AutoComplete uses an array of strings
3. The optionLabel function is the same for both and just returns `test` 
4. When we select and item in both AutoCompletes, the top one still displays `test` (which is intended behaviour) while the bottom one displays the actual value instead of `test`

### Expected behavior

The selected item in AutoCopmlete component should always display the label generated by `optionLabel` no matter if the suggestions are objects or primitive types like strings.

gianluca-moro:
Fixes https://github.com/primefaces/primeng/issues/19075

The current implementation of `get optionValueSelected()` does not make sense since why would you set `optionValue` if the value is a string? The `optionValue` property only really makes sense for object values.    
Therefore, we change the logical AND to a OR to solve the issue. Now, `optionLabel` is also applied correctly to the selected item if the value is a string.

### DIFF
--- a/packages/primeng/src/autocomplete/autocomplete.ts
+++ b/packages/primeng/src/autocomplete/autocomplete.ts
@@ -919,7 +919,7 @@ export class AutoComplete extends BaseInput<AutoCompletePassThrough> {
     }
 
     get optionValueSelected() {
-        return typeof this.modelValue() === 'string' && this.optionValue;
+        return typeof this.modelValue() === 'string' || this.optionValue;
     }
 
     chipItemClass(index) {


gianluca-moro:
Can this get reviewed, please? We are currently facing issues because of this and I don't want to have to implement the 1000th workaround becuase PrimeNG is a buggy mess.

Issue #19181, opened 2025-12-05
Duplicate text in Angular Animate On Scroll Directive in llms.txt


pmroz1:
Duplicate text in Angular Animate On Scroll Directive in llms.txt

PR #19182, opened 2025-12-05
fix: correct duplicate text in Angular Animate On Scroll Directive in llms.txt

pmroz1:
On Scroll Directive description text duplicate  llms.txt instructions https://github.com/primefaces/primeng/issues/19181


cetincakiroglu:
llms.txt file is generated dynamically. Your solution will removed the next time we build the package.

PR #19184, opened 2025-12-08
fix(autocomplete): #19090 - fixes initial input not displayed

### LINKED ISSUE #19090: Autocomplete initial input value displayed to the user doesn't use the optionLabel provided.
### Describe the bug

There is a problem using the autocomplete component and the properties `optionLabel` and `optionValue`. If the compoment has a preselected value, check the example for `selectedCountry` this one is displayed in the input without checking in the provided suggestions and usign the `optionLabel` provided.
In the demo provided you can see that for the pre selected value `CA` should display in the input `Canada` but instead if display `CA`

### Pull Request Link

_No response_

- [x] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/xudvtudh?file=src%2Fapp%2Fautocomplete-objects-demo.ts

### Environment

Quite normal env, Mac, Chrome, I don't think this is relevant for this issue.

### Angular version

20

### PrimeNG version

v20

### Node version

22.19.0

### Browser(s)

Chrome but I think this isn't browser dependant

There is no need for steps, this happens on first load of the autocomplete component, it doesn't use the `optionLabel` and display the value.

### Expected behavior

When `selectedCountry` is initialized with a value it should display the suggestion label if avaialble respeting the `optionLabel` property. This means if `selectedCountry` is pre filled with `CA` and the `optionLabel` is provided and there is an entry in the suggestions that match this criteria the input should display `Canada` and not `CA`.

eduardheller:
Fixes #19090

### DIFF
--- a/packages/primeng/src/autocomplete/autocomplete.ts
+++ b/packages/primeng/src/autocomplete/autocomplete.ts
@@ -951,7 +951,8 @@ export class AutoComplete extends BaseInput<AutoCompletePassThrough> {
     }
 
     get optionValueSelected() {
-        return typeof this.modelValue() === 'string' && !!this.optionValue;
+        // When modelValue is a primitive and optionValue is defined, we consider that optionValue is used
+        return typeof this.modelValue() !== 'object' && !!this.optionValue;
     }
 
     chipItemClass(index) {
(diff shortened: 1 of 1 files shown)


eduardheller:
As you see I removed updateInputLabel from the commits. You get the reactive state already from the inputValue itself due to computed and reacting to model change and suggestions.

PR #19194, opened 2025-12-11
Add role "progressbar" to progressbar

bahuma20:
Re-adding the a11y role "progressbar" to the progressbar component, that was (accidentally?) removed in the migration to primeuix in commit d204244b8e0f46865240a40dacad4de8a415ccdf


SoyDiego (approved):

cetincakiroglu:
Hi,

This PR tries to bind progressbar property and since it doesn't exist, it'll throw the error 

```
✘ [ERROR] NG8002: Can't bind to 'role' since it isn't a known property of 'p-progressBar'. [plugin angular-compiler]
```

PR #19225, opened 2025-12-17
Handle maximum height for spacer for table

jjasloot:
As described in the issue: if the spacer gets a pixel height greater than 33 or 17 million pixels the browser ignores the height style. This results in the scrollbar height just matching the loaded data, not the virtual data. 

This PR aims to mitigate this issue by making sure the height never becomes too big. However, it will result in a moving scrollbar height when the table would originally have been too big. #19226


jjasloot:
Yes it is still relevant

Issue #19227, opened 2025-12-17
v21 DynamicDialog Memory Leak


rosenthalj:
### Describe the bug

Dynamic Dialog has a memory leak.

The video listed below demonstrates the issue with Dynamic Dialog memory leak.
Video Notes:

1. At the start of the video, I hard refresh the page. This clears the page's memory
2. I go to the Dynamic Dialog Demo Page
3. I open and close the Dynamic Dialog multiple times
4. I navigate from the Dynamic Dialog page to the "about page" because "about page" is a very simple page
5. I use the Chrome Development Tool's Memory Profiler to take a snapshot of the browsers JavaScript VM memory (profiling type = detached elements)
6. Chrome has multiple detached Dynamic Dialog elements

https://github.com/user-attachments/assets/233494c7-e739-4243-a9bb-b41c52b8e70f

Examining the source code,  it appears that the memory leak is caused because the DynamicDialog component is incorrectly overriding  the ngOnDestroy method.

<img width="1910" height="860" alt="Image" src="https://github.com/user-attachments/assets/3d854d2f-f47e-42e8-90a8-252b9952fda6" />

### Pull Request Link

https://github.com/primefaces/primeng/pull/19228

### Reproducer

https://primeng.org/dynamicdialog

### Environment

n/a

### Angular version

21+

### PrimeNG version

v21

### Node version

n/a

### Browser(s)

all browsers

### Steps to reproduce the behavior

Repeat the steps in the video above

### Expected behavior

no memory leak

PR #19220, opened 2025-12-16
fix #19215, fixes v21 regression Accordion memory leak

rosenthalj:
Fix for issue #19215

This pull request fixes the PrimeNG BaseComponent's handling of listeners that had callbacks that were "bind(this)".  I believe the bind was causing some of the v21 memory leaks.  

The following video shows the Pull Request changes fixes the Memory Leak of the PrimeNG Accordion Demo. 

https://github.com/user-attachments/assets/c8f2b291-c8aa-4172-8322-06d0fdf0f6e3

~~**Note:  I retested Table and it still has problems.**~~

I retested again and it appears that this Pull Request fixes the Table problems.  The video listed below shows the Memory Leak of PrimeNG Table Demo is fixed.

https://github.com/user-attachments/assets/80b933f5-57c6-4a98-82b4-56e8938c7a70

As an additional test,  I verified that the memory leak is fixed for the PrimeNG Demo  for lots of components that I use in my own application.  The video listed below shows the Memory Leak of PrimeNG multiple demos without a memory leak.

https://github.com/user-attachments/assets/045df9b5-3b9a-4da7-8b4d-2ebfd2b3817d

Issue #19215, opened 2025-12-16
v21 Accordion Regression Memory Leak


rosenthalj:
### Describe the bug

PrimeNG v21 Accordion functionality has a regression memory leak. 
This regression was fixed in v19 (Closed Issue #18296) and was also fixed in v20(Closed Issue #18546).

The video listed below demonstrates the issue with v21 Accordion.
Video Notes:

1. At the start of the video, I hard refresh the page. This clears the page's memory
2. In the Video I navigate away from the Accordion Demo Page 3 times
-- I navigate from the Accordion page to the "about page" because 'about page' is a very simple page
3. I use the Chrome Development Tool's Memory Profiler to take a snapshot of the browsers JavaScript VM memory (profiling type = detached elements)
4. Chrome has detected 3 detached Accordion elements

https://github.com/user-attachments/assets/e456967f-bae9-4bab-838f-e07892979ac9

I "arbitrarily" tested (following similar steps) a few PrimeNG Components that I use in my application:
 1. MegaMenu, MultiSelect, Select, SelectButton did not appear to have a memory leak
 2. **Table appears to also have a memory leak**

### Pull Request Link

https://github.com/primefaces/primeng/pull/19220

### Reproducer

https://primeng.org/accordion

### Environment

N/A

### Angular version

21+

### PrimeNG version

v21

### Node version

n/a

### Browser(s)

all browsers

### Steps to reproduce the behavior

See closed Issue #18546 "Steps" on how to reproduce the bug

### Expected behavior

No memory leak.

bvx5009:
any update on this? are they going to use your PR @rosenthalj  or use another solution?

rosenthalj:
> any update on this? are they going to use your PR [@rosenthalj](https://github.com/rosenthalj) or use another solution?

@bvx5009 this bug should have been closed when issue #19250 was close

The fix is in the current version of PrimNG v21
Note: the bug was fixed with PR v21 DynamicDialog Memory Leak #19227 (which include the change in PR #19220)

I'm closing this issue

PR #19250, opened 2026-01-06
fix: memory leak in basecomponent

### LINKED ISSUE #19249: Memory Leak in BaseComponent
### Describe the bug

3 months ago, a previously fixed memory leak in the BaseComponent of all primeng controls was reintroduced.

This leads to raising memory & degraded performance over time, especially in large applications.

### Pull Request Link

https://github.com/primefaces/primeng/pull/19250

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://primeng.org (can be reproduced everywhere, follow repro stept)

### Environment

The (at the time of writing) latest main branch has the issue and the released v21 version aswell.
The issue was reintroduced here: https://github.com/primefaces/primeng/commit/9b5c3c72cdfe0cf0fd0d32bcc78f5042af07b6bd#diff-315219ce5e29cd6185c317544d0a096d9f46f310d322405400a1fdd72f0737f8

### Angular version

21.0.5

### PrimeNG version

v21

### Node version

_No response_

### Browser(s)

_No response_

The issue can be easily reproduced on the primeng docs page.

My repro steps:

1. serve the app locally (for unobfuscaded code)
2. open the docs at http://localhost:4200/colorpicker
3. reload the page for clean memory
4. open the dev tools
5. navigate to the memory in dev tools and force a garbage collection
6. create a heap snapshot
7. in the primeng docs navigate to the datepicker section, wait until it has loaded & then navigate back to the color picker
8. in the devtools force anther garbace collection
9. take another heap snapshot
10. compare snapshot one with two (objects allocated between snapshot 1 and snapshot 2)
11. Search for "date"

The result will show 33 retained instances of the datepicker component, including their DOM nodes.

Image

### Expected behavior

With my fix, no component instances are retained anymore:

Image

LoaderB0T:
Fixes #19249

### DIFF
--- a/packages/primeng/src/basecomponent/basecomponent.ts
+++ b/packages/primeng/src/basecomponent/basecomponent.ts
@@ -42,6 +42,8 @@ export class BaseComponent<PT = any> implements Lifecycle {
 
     private _themeScopedListener: () => void;
 
+    private _themeChangeListeners: (() => void)[] = [];
+
     /******************** Inputs ********************/
 
     /**
@@ -396,13 +398,19 @@ export class BaseComponent<PT = any> implements Lifecycle {
 
     private _themeChangeListener(callback = () => {}) {
         Base.clearLoadedStyleNames();
-        ThemeService.on('theme:change', callback.bind(this));
+        const listener = callback.bind(this);
+        ThemeService.on('theme:change', listener);
+        this._themeChangeListeners.push(listener);
     }
 
     private _removeThemeListeners() {
         ThemeService.off('theme:change', this._loadCoreStyles);
         ThemeService.off('theme:change', this._load);
         ThemeService.off('theme:change', this._themeScopedListener);
+        this._themeChangeListeners.forEach((listener) => {
+            ThemeService.off('theme:change', listener);
+        });
+        this._themeChangeListeners = [];
     }
 
     /********** Passthrough **********/


rosenthalj:
I believe PR #19220 will fix this issue

LoaderB0T:
@rosenthalj 
I tested my fix of course and yes it fixes the issue. The only event listener that is leaking is the one passed by callback. And since this method can be called many times, I just store all listeners in the array and clean all of them up when the rest is cleaned up.

The other listeners are already cleaned up correctly, as they are not some random callback coming from the outside, but are well-known. (_loadCoreStyles, _load & _themeScopedListener can be used to unregister the listener w/o any issues)

So the only real change required is the callback-based listener in the _themeChangeListener. This solves the leak.

Edit: Feel free to check out my branch and test for yourself :) I am ready to eat my words if it is still leaking.

rosenthalj:
> @rosenthalj I tested my fix of course and yes it fixes the issue. The only event listener that is leaking is the one passed by callback. And since this method can be called many times, I just store all listeners in the array and clean all of them up when the rest is cleaned up.
> 
> The other listeners are already cleaned up correctly, as they are not some random callback coming from the outside, but are well-known. (_loadCoreStyles, _load & _themeScopedListener can be used to unregister the listener w/o any issues)
> 
> So the only real change required is the callback-based listener in the _themeChangeListener. This solves the leak.
> 
> Edit: Feel free to check out my branch and test for yourself :) I am ready to eat my words if it is still leaking.

@LoaderB0T
I believe the memory leak is caused by  the following original code:

`       ThemeService.on('theme:change', callback.bind(this));`

As a result,  the original code is failing to remove the listeners which is causing the memory leak:

```
        ThemeService.off('theme:change', this._loadCoreStyles);
        ThemeService.off('theme:change', this._load);
        ThemeService.off('theme:change', this._themeScopedListener);
```

In your PR,  you store all listeners (including _loadCoreStyle,  _load, and _themeScopedListener) and then remove all the listeners.  That is why I believe your changes fixed the memory leak for your test cases.

I still stand by my code review comments.  I believe your PR can be improved.


Hopefully PrimeNG (@mertsincan, @mehmetcetin01140, @cetincakiroglu, @cagataycivici) will choose/create a solution that fixes the critical basecomponent v21 memory leak

I was surprised the following memory leak issues where not resolved in PrimeNG v21.0.3;

#19249 and #19215. - same issue
#19227

LoaderB0T:
@rosenthalj 
I retract everything I’ve said and claim the exact opposite.
You are right, I did not see that the three internal callbacks are using the same private method. I thought it was a fourth way to add a listener. With my solution, listeners would stack up when the design tokens change. With the map, they don't. I vote in favor of your solution.
Sorry for my confusion & thanks for being persistent!

rosenthalj:
> @rosenthalj I retract everything I’ve said and claim the exact opposite. You are right, I did not see that the three internal callbacks are using the same private method. I thought it was a fourth way to add a listener. With my solution, listeners would stack up when the design tokens change. With the map, they don't. I vote in favor of your solution. Sorry for my confusion & thanks for being persistent!

@LoaderBOT
The only thing I care about is fixing this memory leak (and hopefully fixing the memory leak correctly).
Thank-you for your persistence!

PR #19257, opened 2026-01-08
fix(dialog): use textContent instead of innerHTML for dynamic styles

moraisacr:
### What’s changed
Replaced `innerHTML` with `textContent` when injecting dynamic CSS into the `<style>` element used for responsive dialog breakpoints.

### Why
Avoids using the `innerHTML` sink when no HTML parsing is required, improving compatibility with CSP / Trusted Types environments.

### Tests
- `pnpm --filter primeng test:unit -- --include="**/dialog/*.spec.ts"`

### DIFF
--- a/packages/primeng/src/dialog/dialog.ts
+++ b/packages/primeng/src/dialog/dialog.ts
@@ -807,29 +807,29 @@ export class Dialog extends BaseComponent<DialogPassThrough> implements OnInit, 
         }
     }
 
-    createStyle() {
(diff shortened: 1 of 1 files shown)


moraisacr:
Hi team! This is a small internal fix replacing `innerHTML` with `textContent` for dynamic dialog styles to improve CSP / Trusted Types compatibility.  
Dialog unit tests were executed locally and passed. Thanks!

Issue #19196, opened 2025-12-11
DynamicDialog does not emit onClose on pressing ESC or close icon


qwadrox:
### Describe the bug

The dynamic dialog onClose event does not emit when the dialog is closed with ESC or the built-in closed icon.

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://primeng.org/dynamicdialog#example

### Environment

As the showcase app

### Angular version

21

### PrimeNG version

v21

### Steps to reproduce the behavior

https://primeng.org/dynamicdialog#example

1. open the example
2. hit the button - Select a product
3. Press ESC or the close icon
4. The toast message does not pop up 

### Expected behavior

The dynamic dialog onClose event should emit when the dialog is closed with ESC or with the built-in closed icon. The toast message should appear on ESC.

ES-Six:
Also got this since PrimeNG v20.x, still present in v21.x but not documented on the PrimeNG documentation or migration guide.

quentinFonck:
@cetincakiroglu can we also have this fix for PrimeNG v20.x ? we just upgraded to 20 and came across this issue.

PR #19209, opened 2025-12-15
fix: #19196 - Ensure onClose fires on close icon click and ESC key press

johnson-jnr:
Fixes #19196 

In `DynamicDialog`, visibleChange event handler is added, in which we call close on the dialogRef.

The fix ensures that `onClose` fires and the toast message shows up.

PR #19260, opened 2026-01-08
fix: #19196 - Ensure onClose fires on close icon click and ESC key press (v20)

danielvieira1:
Can we please create a hotfix for the v20 version regarding this fix [PR](https://github.com/primefaces/primeng/pull/19209)

Solves https://github.com/primefaces/primeng/issues/19196

In DynamicDialog, visibleChange event handler is added, in which we call close on the dialogRef.
The fix ensures that onClose fires and the toast message shows up.


gianluca-moro:
Any update on this? We are facing the same issue

danielvieira1:
I forked the project and we are now using our own version of primeNG.
they take ages to simply merge a fix for a regression they made and create a hotfix for it.

Issue #18010, opened 2025-03-31
PanelMenu: Different item is focussed when item is pressed


vdbongard:
### Describe the bug

If a PanelMenu item is actively being selected, another item in the PanelMenu group is focused as well.

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://primeng.org/panelmenu#template

### Environment

Chrome 134, Windows 11

### Angular version

19.2.0

### PrimeNG version

v19

### Steps to reproduce the behavior

1. Go to the "Template" section of the PanelMenu docs (https://primeng.org/panelmenu#template)
2. Open the "Mail" group
3. Actively select the second entry of "Inbox" (hold the mouse down while clicking the item)
4. The first item "Compose" also gets highlighted, while pressing the second item "Inbox" 

![Image](https://github.com/user-attachments/assets/15355290-96ca-4d87-8fd0-6b8e45ebf5b7)

### Expected behavior

The first item should not get highlighted while pressing the second item.

vdbongard:
The bug still exists. The error has changed slightly. When actively selecting the second entry of "Inbox" (step 3), the last item "Trash" will get highlighted. (instead of the first item "Compose" in step 4)

<img width="612" height="463" alt="Image" src="https://github.com/user-attachments/assets/e5585daa-9b5c-4734-80cc-ee2d52e73d4e" />

bae-sh:
Same issue as #13940, but the error still continues.
When I run PanelMenu in dev mode, it works well, but in prod mode, this issue happens.
@cagataycivici Can I work on this issue?

bae-sh:
@cagataycivici I've fixed the bug and opened PR #19296. Could you please review it?

cetincakiroglu:
Hi,

This issue already fixed in here and the fix will available with upcoming release. https://github.com/primefaces/primeng/commit/2f431599c6a27d7cc382449efe8bbc2caa3d2b6f#diff-96a09a58c7c12744d56b6a66388fd0da4643399cf3739accc097bf025f616495R159

vdbongard:
I have just tested this. The bug still exists like it is described in the issue.

Environment
Chrome 134, Windows 11

PrimeNG version
v21

I created a new issue as I can't reopen this.

https://github.com/primefaces/primeng/issues/19526

PR #19296, opened 2026-01-18
fix: #18010 - Different item is focussed when item is pressed

bae-sh:
Fixed Issue #18010

## Root Cause
The issue was caused by an incorrect `data-pc-name` property value. In production mode, due to code minification, the component name gets mangled, causing isElementInPanel method to fail in finding the panel and return null. This leads to incorrect focusedItem assignment in the onFocus event handler (defaulting to this.findLastItem).

### Dev mode
Screenshot 2026-01-18 at 6 39 28 PM

### Prod mode
Screenshot 2026-01-18 at 6 40 43 PM

## Solution
Used `this['componentName']` value from basecomponent's `$name` to ensure consistent component identification across build modes.

## Alternative Approach
Another potential solution would be to disable script optimization in `apps/showcase/angular.json`, though this should be evaluated based on the performance trade-offs.

```json
...
"configurations": {
    "production": {
        "optimization": {
            "scripts": false,
            "styles": true,
            "fonts": true
        }
    }
}
...
```

### DIFF
--- a/packages/primeng/src/panelmenu/panelmenu.ts
+++ b/packages/primeng/src/panelmenu/panelmenu.ts
@@ -948,6 +948,7 @@ export class PanelMenuList extends BaseComponent {
     hostDirectives: [Bind]
 })
 export class PanelMenu extends BaseComponent<PanelMenuPassThrough> {
+    componentName = 'PanelMenu';
     /**
      * An array of menuitems.
      * @group Props


cetincakiroglu:
Hi,

This issue already fixed in here and the fix will available with upcoming release. https://github.com/primefaces/primeng/commit/2f431599c6a27d7cc382449efe8bbc2caa3d2b6f#diff-96a09a58c7c12744d56b6a66388fd0da4643399cf3739accc097bf025f616495R159

PR #19303, opened 2026-01-21
fix(listbox): aria-multiselectable always true

mrkshx:
In Listbox, aria-multiselectable was always set to true but it should be depending on "multiple"

The fix uses "multiple" to set value of aria-multiselectable attribute value.


mrkshx:
fixes #19304

PR #19335, opened 2026-01-27
fix #19326 - pagination rows and first not overwritten by undefined (Table)

### LINKED ISSUE #19326: Table: corrupted table content when pagination stored to localStorage and content changed by third party
### Describe the bug

Take a p-table with pagination on true if the element count is larger than x and the state saved to the localStorage 

Now user A sorts the table when elements < x 

Then user B does something that causes elements > x 

When user A reloads the page the content of the p-table is corrupted because the default value of [rows] and [first] of the table is overwritten by undefined from the localStorage. 

### Pull Request Link

https://github.com/primefaces/primeng/pull/19335

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/github-xmjt3sdt-o6u7cj7d?file=src%2Fapp%2Fapp.component.html

### Environment

Windows 10 

angular v20.1.3 and v21.1.0

### Angular version

20.1.3

### PrimeNG version

v20

### Node version

22.14.0

### Browser(s)

Chrome Version 143.0.7499.170 (Official Build) (64-bit)

1. Sort table by pressing the column Name at least once
2. press the green button (which will simulate a content change by another user and reload the page)
3. See that the p-table displays no content 

### Expected behavior

Table content should be shown as if there was no saved state.

S1m0zK:
This way missing data in the localStorage of the user should not overwrite the default values of rows and first from the table component.. Closes #19326

### DIFF
--- a/packages/primeng/src/table/table.ts
+++ b/packages/primeng/src/table/table.ts
@@ -2971,12 +2971,12 @@ export class Table<RowData = any> extends BaseComponent<TablePassThrough> implem
             let state: TableState = JSON.parse(stateString, reviver);
 
             if (this.paginator) {
-                if (this.first) {
+                if (this.first !== undefined && state.first) {
                     this.first = state.first;
                     this.firstChange.emit(this.first);
                 }
 
-                if (this.rows) {
+                if (this.rows !== undefined && state.rows) {
                     this.rows = state.rows;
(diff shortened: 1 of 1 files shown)


johanneszelger:
any update on this? an issue was created a while ago

PR #19341, opened 2026-01-29
Fix: filterFields input is not applied in MultiSelect filtering

Indragith20:
**Issue**

filterFields input in MultiSelect was defined in the public API but not used anywhere in the filtering logic.
As a result, providing filterFields had no effect and filtering always relied on the default field (filterBy/ optionLabel / implicit string matching), causing incorrect filtered results.

This leads to incorrect filtering behavior and makes the filterFields API ineffective.

This PR will fix the above issue.

### DIFF
--- a/packages/primeng/src/multiselect/multiselect.ts
+++ b/packages/primeng/src/multiselect/multiselect.ts
@@ -1448,7 +1448,7 @@ export class MultiSelect extends BaseEditableHolder<MultiSelectPassThrough> {
     }
 
     searchFields() {
-        return (this.filterBy || this.optionLabel || 'label').split(',');
+        return this.filterBy?.split(',') || this.filterFields || [this.optionLabel];
     }
 
     findNearestSelectedOptionIndex(index, firstCheckUp = false) {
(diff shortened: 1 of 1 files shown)


Indragith20:
Fix for the issue [#19342](https://github.com/primefaces/primeng/issues/19342)

packages/primeng/src/multiselect/multiselect.ts
```diff
@@ -1448,7 +1448,7 @@ export class MultiSelect extends BaseEditableHolder<MultiSelectPassThrough> {
     }
 
     searchFields() {
-        return (this.filterBy || this.optionLabel || 'label').split(',');
+        return this.filterBy?.split(',') || this.filterFields || [this.optionLabel];
     }
```
rosenthalj: I only took a quick look and I have not verified your change actually works, but ...

Do you need to add **'label'** as the final "default"?  This would make you change more consistent with the previous logic.

Indragith20: Added the 'label' as final default value

Issue #19374, opened 2026-02-04
Tree passthrough options for nodeToggleButton only apply one level deep


RGamberini-Harris:
### Describe the bug

When styling the `nodeToggleButton` using passthrough options like so:
```html
<p-tree [pt]="{nodeToggleButton: {class: 'red'}}" [value]="files" />
```
The styling will only be applied to the first level of `p-tree-node` see truncated example below
```html
<p-treenode>
  <!-- omitted -->
  <button class="p-ripple p-tree-node-toggle-button red"></button>
  <p-treenode>
    <button class="p-ripple p-tree-node-toggle-button"></button>
  </p-treenode>
</p-treenode>
```
Tested on `primeng@21.1.1`
Image: 

<img width="1073" height="476" alt="Image" src="https://github.com/user-attachments/assets/af198f7f-f006-4a17-8f2b-5cc11fa96b1c" />

### Pull Request Link

https://github.com/primefaces/primeng/pull/19375

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/~/github.com/RGamberini-Harris/primeng-issue-template

### Environment

Stackblitz package.json should show everything but it's Angular 21 / PrimeNG 21

### Angular version

21

### PrimeNG version

v21

### Steps to reproduce the behavior

Try to style nested tree-node-toggle-button using passthrough

### Expected behavior

The nested toggle buttons should have the same styling as the top level ones

RGamberini-Harris:
Edited my post to add the pull request resolving this issue

ArcherTrent:
Same with nodeLabel, only applied to top nodes.

cetincakiroglu:
fixed in v22

PR #19375, opened 2026-02-04
Tree-node: Add missing parentheses for PT signal

RGamberini-Harris:
### Defect Fix
Resolves https://github.com/primefaces/primeng/issues/19374 an issue where styling was not passed through to children tree-nodes as the PT parameters were being passed as a raw signal rather than being called.


Georg632:
When I look at the `tree.ts` file I am finding two more occurrences of `[pt]="pt"`, line `821 `and `843`.

PR #19423, opened 2026-02-23
fix(select): blank panel with virtualScroll + lazy in dialog (

### LINKED ISSUE #19401: p-select with virtualScroll + lazy inside p-dialog renders blank dropdown until user interaction (v21)
### Describe the bug

p-select with [virtualScroll]="true" and [lazy]="true" inside a p-dialog with appendTo="body" renders a blank dropdown panel when the application uses provideZoneChangeDetection(). This appeared after Angular 21 upgrade since applications now use zoneless change detection by default,

The issue is Iitems are loaded and available in memory, but the **dropdown panel opens completely empty**. Clicking anywhere else on the page triggers change detection and the items suddenly appear.

This issue is only happening when zone detection is enabled

Workarounds:

Force scroller reinitialization on panel show:

`<p-select #selectRef (onShow)="onPanelShow()" ... ></p-select>`

```
@ViewChild('selectRef') selectRef!: Select;

private readonly ngZone = inject(NgZone);

onPanelShow(): void {
  this.ngZone.runOutsideAngular(() => {
    setTimeout(() => {
      this.ngZone.run(() => {
        const scroller = this.selectRef?.scroller;
        if (scroller) {
          if (!scroller.initialized) {
            scroller.viewInit();
          } else {
            scroller.init();
          }
        }
      });
    });
  });
}

```

### Pull Request Link

_No response_

- [x] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/a8jsmgda?file=src%2Fmain.ts

### Environment

Environment

OS: Windows
Browser: Chrome (all browsers affected)

### Angular version

21.0.0

### PrimeNG version

v21

### Node version

22.x.x

### Browser(s)

_No response_

**To reproduce on your own:**
1. Bootstrap the app with provideZoneChangeDetection() 
2. Place a p-select with [virtualScroll]="true" [lazy]="true" appendTo="body" inside a p-dialog
3. Open the dialog
4. Click the dropdown
5. Dropdown panel opens blank — no items visible
6. Click anywhere else on the page → items suddenly appear

Or go to the following reproducer:  https://stackblitz.com/edit/a8jsmgda?file=src%2Fmain.ts
Click open dialog, the dropdown will show but blank. Clicking anywhere on the screen or on the keyboard will force items to rended.

### Expected behavior

Dropdown items should render immediately when the panel opens, regardless if provideZoneChangeDetection ir enabled or not

khaledbaghdadii:
Fixes #19401

## Problem

When `p-select` is configured with `[virtualScroll]="true"` and `[lazy]="true"` inside a `p-dialog`
with `appendTo="body"`, the dropdown panel opens blank. Items only appear after an additional user
interaction (click, keypress).

## Root Cause

`p-motion`'s `afterRenderEffect` uses `await nextFrame()` (double `requestAnimationFrame`), which
causes execution to escape `NgZone`. The `onAfterEnter` animation callback fires **outside NgZone**,
so the scroller's entire initialization chain (`init → setTimeout → calculateOptions → onLazyLoad`)
runs outside the zone. Angular never detects the subsequent item updates, leaving the panel blank.

## Fix

Add an `onOverlayAfterEnter` handler that re-enters NgZone via `zone.run()` before triggering
`cd.detectChanges()`. This ensures all downstream async operations (setTimeout, lazy load events,
data fetches) are intercepted by zone and trigger change detection when they resolve.

```typescript
onOverlayAfterEnter(event: any) {
    if (this.virtualScroll) {
        this.zone.run(() => {
            this.cd.detectChanges();
        });
    }
}

### DIFF
--- a/packages/primeng/src/select/select.spec.ts
+++ b/packages/primeng/src/select/select.spec.ts
@@ -4,6 +4,7 @@ import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { FormBuilder, FormControl, FormGroup, FormsModule, ReactiveFormsModule, Validators } from '@angular/forms';
 import { By } from '@angular/platform-browser';
 
+import { Dialog } from 'primeng/dialog';
 import { BehaviorSubject, timer } from 'rxjs';
 import { map, take } from 'rxjs/operators';
 import { Select } from './select';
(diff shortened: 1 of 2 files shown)


domvanrob:
I've got the same issue with the overlay in my TreeSelect component. Running this manually after the `onShow` output fixes it.

Issue #17558, opened 2025-02-03
Accordion BUG: TextArea doesn't work properly when it is placed inside an accordion


htdong:
### Describe the bug

When a textarea is placed inside an accordion, cursor move up and move down will not work.

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/github-mzz9qowd?file=src%2Fapp%2Fapp.component.html

### Environment

Angular 19, PrimeNG 19

### Angular version

19

### PrimeNG version

v19

### Steps to reproduce the behavior

1.Create a component
2.Import a drawer
3.Import an accordion
4.Place a textarea inside the accordion
5.Issue: Can not move up and move down cursor

### Expected behavior

Cursor can be moved up and down within an accordion

PR #19434, opened 2026-02-25
fix: #17558 ignore keydown event from an input or textarea in accordion

tarkkeim:
Fixes issue https://github.com/primefaces/primeng/issues/17558


Cr3aHal0:
Fix is not sufficient as even a set of p-radiobutton won't be able to go up/down the list within an accordion. Not sure either that checking every type of element dispatching an event within the accordion may be a good idea, as people could come with their own custom component needing to handle arrow keys for accessibility purpose, for instance.

tarkkeim:
Thanks for your feedback. You're right for the case of radiobuttons, and there are probably other cases that face the same issue. I'll close this PR for now and try to find a better way to fix this.

Issue #19325, opened 2026-01-26
Block UI: DOM elements inside Block UI shouldn't be visible when `blocked` is `false`


Poolmann:
### Describe the bug

Displaying icons inside a block UI mask is no longer possible (since v21 I believe) since DOM elements inside a block UI component are visible despite the block UI not being "blocked". This behavior arises, when the block ui state is `false` when the page is rendered first.

### Reason for not contributing a PR

- [x] Lack of time
- [x] Unsure how to implement the fix/feature
- [x] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/h8suq8kd?file=src%2Fapp%2Fblockui-basic-demo.ts

### Environment

Firefox 145.0.1
Fedora 43

### Angular version

21.0.8

### PrimeNG version

v21

### Node version

22.20.0

### Steps to reproduce the behavior

1. Go to reproducer
2. See that spinner is visible
3. Block and unblock
4. Spinner is no longer visible (expected behavior)

<img width="580" height="328" alt="Image" src="https://github.com/user-attachments/assets/417dee28-5f38-4366-8651-eeb3567ebe45" />

### Expected behavior

The spinner should not be visible when the Block UI isn't visible.

MAN-Sendance:
I wish this issue would get more recognition, since it is a major bug in my opinion.

cs-StephanieDM:
I had the same problem and just added an @if as a workaround:

```
<p-blockUI [blocked]="blocked" class="justify-center items-center">
  <!-- We have to add the "if" because of https://github.com/primefaces/primeng/issues/19325 -->
  @if (blocked) {
    <p-progressSpinner animationDuration="3s"></p-progressSpinner>
  }
</p-blockUI>
```

MAN-Sendance:
I wish this would get more attention since it blocks us from upgrading to Angular v21

davidsybit:
i have the same problem

"@angular/core": "21.2.1",
"primeng": "21.1.3",

I have a similar workaround:

`
@if (loading()) {
  <p-blockUI [blocked]="loading()">
    <p-progressSpinner ariaLabel="loading" strokeWidth="5"></p-progressSpinner>
  </p-blockUI>
}
`

buecherlstefan:
Same problem here, please fix because it breaks a lot of my pages!

MAN-Sendance:
We are still forced to use Angular v20 due to this major bug.

What can we do so that this will get more attention?

[...]

PR #19439, opened 2026-02-26
fix(19325): properly set overlay as display none if blocked equals false

Cr3aHal0:
Fixes [19325](https://github.com/primefaces/primeng/issues/19325)

Disclaimer : As of today, PrimeNG implementation of BlockUI is quite different from PrimeVue & PrimeReact  one which is unified (the component is a wrapper of content and the overlay is projected within the wrapper boundaries, whereas in PrimeNG BlockUi is much likely a sibling of the target to be blocked). I did not know if this component had any major plan coming from PrimeNG team, so I opted for a quick working fix which just makes sure that if the blocked flag is false at startup then it should consider the overlay as invisible (hence the display: none, which is already called when manually hiding the overlay and then prevents the overlay content to show in the first place - which was the description of the issue)

### DIFF
--- a/packages/primeng/src/blockui/blockui.ts
+++ b/packages/primeng/src/blockui/blockui.ts
@@ -99,6 +99,7 @@ export class BlockUI extends BaseComponent<BlockUIPassThrough> {
 
     onAfterViewInit() {
         if (this._blocked) this.block();
+        else this.hideSelf();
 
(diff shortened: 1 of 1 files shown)


MAN-Sendance:
I wish this would get more attention since it blocks us from upgrading to Angular v21

PR #19454, opened 2026-03-03
add pt support for p-tableCheckbox

AnasArafeh:
the p-tableCheckbox component does not pass the pass through property to the checkbox component

so I have just added it similarly to what we have in p-tableHeaderCheckbox component


AnasArafeh:
@mehmetcetin01140 may you please review this

AnasArafeh:
@cagataycivici it is sad to see such PR and other similar PRs are ignored for weeks while the library contains a lot of basic issues which makes the library require workarounds to be able to achieve what is needed.

Issue #17008, opened 2024-12-11
v18 Component: Autocomplete optionValue overriding optionLabel


sandrotonon:
### Describe the bug

**Actual behaviour**
If an autocomplete has objects as data and you specify a optionLabel, everything works as expected. As soon as you set the optionValue, the displayed value is also the optionValue, instead of the spezified optionLabel.

For example:

This is the data the autocomplete has to work with:
```typescript
data = [
  {
    id: 1,
    someProperty: 'One',
  },
  {
    id: 2,
    someProperty: 'Two',
  },
];
``` 

This is the autocomplete config
```typescript
<p-autocomplete
  [(ngModel)]="valueTwo"
  [suggestions]="suggestions"
  [dropdown]="true"
  optionLabel="someProperty"
  optionValue="id"
  (completeMethod)="search($event)"
/>
``` 

If you select object with id "1" in the autocomplete, the form value is "1" (which is correct), but the displayed Value (or label) is also "1".

### Environment

Updating from 17.18.13 to 18.0.0

### Reproducer

https://stackblitz.com/edit/stackblitz-starters-6ub1avdm?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.html

### Angular version

18.2.13

### PrimeNG version

18.0.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

### Node version (for AoT issues node --version)

20.14.0

### Browser(s)

Firefox latest, Chrome latest

### Steps to reproduce the behavior

1. Within card "With optionValue" type "o"
2. Option "One" should appear
3. Select option "One"

### Expected behavior

If a optionLabel is provided, the autocomplete always shows the specified property (or calculated value), no matter if the optionValue is also provided or not.

Applied to the example above:

If you select object with id "1" in the autocomplete, the form value is "1", and the displayed Value (or label) is "One".

JonathanWenabi:
Hello i have the same issue on primeNG V19, do you have a date for a fix ?

maha90:
This is frustrating as we can not have the label of a suggestion object shown in the input.
Also it is not possible to use a template as the selectedItem template is ignored without multiple=true (see https://github.com/primefaces/primeng/issues/12983).
I would expect this to behave like the Select component where the optionLabel is also used for the input and not the optionValue.

GuillaumeWenabi:
Hello @mertsincan any plan to fix this issue in coming patches ? That's a blocker for us

[...]

PR #19483, opened 2026-03-19
fix(autocomplete): Patching Form Value displays OptionValue

### LINKED ISSUE #19433: v21 Autocomplete: Patching Form Value displays OptionValue
### Describe the bug

With reactive forms I have a FormControl of type string.
I want to use it with the autocomplete component, where the suggestions are an array of objects with a label and a key. The key is type string and is used in the optionValue attribute of the autocomplete component.
The label is also string and used for the optionLabel attribute.
If want to call the forms patchValue method to set the value for the component with a valid key value, the autocomplete component displays the key, not the label (produced by optionLabel).
I do realize there is a specific section in the documentation for using objects.
And if I do use the object as the type of the form control and only use the optionLabel attribute, the component works.
This is unexpected as the select and multiSelect components do work differently and in the past I had trouble using objects as form control values as there might be issues with equality when patching the form.

### Pull Request Link

_No response_

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/stackblitz-starters-2mvsjwdv?file=src%2Fapp.component.ts

### Environment

Angular v21
PrimeNG v21

### Angular version

21.1.0

### PrimeNG version

v21

### Node version

_No response_

### Browser(s)

_No response_

1. Just click the button "patch form" and you'll see "de" displayed in the autocomplete.

### Expected behavior

It should be "Germany", which is the label of the patched country.

O2sa:
Fixes https://github.com/primefaces/primeng/issues/19433 https://github.com/primefaces/primeng/issues/17008

## Scope
- inputValue logic only
- No breaking changes
- No visual or API changes
- No behavior change for the multiple mode or if the optionValue input dose not set
- No CSS or DOM structure changes

### DIFF
--- a/packages/primeng/src/autocomplete/autocomplete.ts
+++ b/packages/primeng/src/autocomplete/autocomplete.ts
@@ -895,7 +895,16 @@ export class AutoComplete extends BaseInput<AutoCompletePassThrough> {
 
     inputValue = computed(() => {
         const modelValue = this.modelValue();
-        const selectedOption = this.optionValueSelected ? (this.suggestions || []).find((option: any) => equals(option, modelValue, this.equalityKey())) : modelValue;
+        const selectedOption = this.optionValueSelected
+            ? (this.suggestions || []).find((option: any) => {
+                  if (typeof option === 'object') {
+                      const optVal = resolveFieldData(option, this.optionValue);
+                      return optVal == modelValue;
+                  }
+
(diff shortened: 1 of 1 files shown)


gianluca-moro:
Can this get reviewed, please? We are facing this issue (amongst many others, lol) since several months and it's annoying having to implement workarounds.

PR #19556, opened 2026-04-29
chore: replace docs button plain examples with contrast ones

julien-deramond:
> [!NOTE]
> The template was suggesting to create an issue first, but there were some mandatory fields (e.g. StackBlitz link) that seemed more related to a bug in the library, rather than a docs enhancement. If the issue creation is still needed, please let me know, and I'll create one :)

---

This PR replaces button plain examples in the docs with button contrast. Indeed plain examples are basically the normal examples, and some contrast examples were missing to showcase the rendering of this `severity`.

## Context

Most examples in https://primeng.org/button (and also in https://primeng.org/splitbutton) showcase the Contrast severity buttons, such as:

Screenshot 2026-04-29 at 09 18 58

However, some examples are rather showcasing a Plain variant of the Button component:

Screenshot 2026-04-29 at 09 20 27

If one looks at the Plain variant code, they would see that it's basically the Primary button with a different label.

In the sections that are showcasing the Plain variant, my suggestion in this PR is to replace the Plain variant by the Contrast variant, that is a real `severity`, and has a different rendering:

Screenshot 2026-04-29 at 09 22 59

## Modifications

Manually modified files:
- `apps/showcase/doc/button/iconsonly-doc.ts`
- `apps/showcase/doc/button/raisedtext-doc.ts`
- `apps/showcase/doc/button/text-doc.ts`

The rest was generated after running `pnpm build` at the root-level of the project.


apps/showcase/public/llms/llms-full.txt
```diff
@@ -16916,236 +16921,6 @@ IftaLabel is used to create infield top aligned labels.
 
 ---
 
-# Angular Image Component
```
julien-deramond, line 16919: I'm not sure why this section is entirely removed after running `pnpm build`. Maybe I shouldn't add the the result of `pnpm build` in the PR?

PR #19632, opened 2026-06-18
Fix FrozenColumn.updateStickyPosition misplacing cells

### LINKED ISSUE #19631: FrozenColumn.updateStickyPosition misplacing cells
### Describe the bug

`FrozenColumn.updateStickyPosition` assumes the next row is always a filter row, which is wrong. It sets a css position on the cell below the current, that usually does not cause an issue. But in my example, I use rowspan for a column, and the row below will have 1 less columns, so it sets styles on a cell that's not right below the current one, misaligning it.

### Pull Request Link

#19632

### Reproducer

https://stackblitz.com/edit/github-ewcjlidn?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts

### Environment

macOS Tahoe 26.5.1

### Angular version

21.2.1

### PrimeNG version

v21

### Node version

24.11.0

### Browser(s)

Chrome 149.0.7827.115

1. Go to https://stackblitz.com/edit/github-ewcjlidn?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts
2. Look at the second row
3. "Action 1" and "Action 2" overlap on the right edge (e.g. both get `right: <action width>px`).
4. In DevTools, inspect the second row's last frozen cell: its inline `style.right` is wrong (it should be `right: 0`, but it matches the width of the column).
5. Resize the window: the last column briefly snaps to `right: 0`, then jumps back (wrong) once the first row's frozen cell re-runs and re-copies its offset.

### Expected behavior

"Action 2" cell in the second row should have `right: 0`.

kabalage:
Fixes #19631

Edit: the issue was closed by mistake, my changes do apply to v22 as well. I can prepare a PR for the v22 branch too.

### DIFF
--- a/packages/primeng/src/table/table.ts
+++ b/packages/primeng/src/table/table.ts
@@ -3644,12 +3644,15 @@ export class FrozenColumn extends BaseComponent {
                 this.el.nativeElement.style.left = left + 'px';
             }
 
-            const filterRow = this.el.nativeElement?.parentElement?.nextElementSibling;
-            if (filterRow) {
+            const nextRow = this.el.nativeElement?.parentElement?.nextElementSibling;
+            const isHeaderCell = this.el.nativeElement?.tagName === 'TH';
+            const nextRowIsFilterRow = !!nextRow && !!DomHandler.findSingle(nextRow, '[data-pc-section="columnfilter"]');
+
+            if (isHeaderCell && nextRow && nextRowIsFilterRow) {
                 let index = DomHandler.index(this.el.nativeElement);
-                if (filterRow.children && filterRow.children[index]) {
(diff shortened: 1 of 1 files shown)


benjaminforras:
Awesome job! Thank you for looking into this. Hopefully it can be in the next PrimeNG release so it can unblock us!

krook1024:
Good fix. The root cause is clear — nextElementSibling has no guarantee of being a filter row, so blindly applying sticky positioning to it could corrupt layout for any table without a filter row, or worse, misplace cells in the first body row. Would appreciate this in production as soon as possible.

PR #19636, opened 2026-06-23
[19635] fix: Add scope="col" to th elements for td-has-header a11y compliance

### LINKED ISSUE #19635: Table and TreeTable: missing scope attribute on th elements causes td-has-header a11y violation
### Describe the bug

<td> elements in Table and TreeTable components lack proper header associations, causing the _td-has-header_ accessibility rule to fail.

Affected components:

Table (p-table)
TreeTable (p-treetable)
References:

axe rule: td-has-header
WCAG: 1.3.1 Info and Relationships (Level A)

### Pull Request Link

https://github.com/primefaces/primeng/pull/19636

### Reproducer

https://stackblitz.com/edit/github-hrd4vzvf?file=src%2Fapp%2Fapp.component.ts

### Environment

A library that wraps primeng and provide its own styling.

### Angular version

21

### PrimeNG version

v21

### Node version

_No response_

### Browser(s)

_No response_

1. Open the Table demo page (e.g., Basic Table example)
2. Run an accessibility audit using browser DevTools (Lighthouse) or axe DevTools extension
3. Observe the td-has-header violation is reported

Or manually:

1. Open browser DevTools
2. Inspect any <th> element in the table header
3. Notice the scope attribute is missing

### Expected behavior

All <th> elements in Table and TreeTable should have scope="col" attribute automatically applied, ensuring:

- Screen readers announce column headers when navigating to data cells
- The td-has-header axe accessibility rule passes
- WCAG 1.3.1 (Info and Relationships) compliance is met

fkwq34:
fixes: #19635 

Description
Adds scope="col" attribute to <th> elements in Table and TreeTable components to fix the td-has-header accessibility violation.

Changes

- Added updateThScope() method that sets scope="col" on all <th> elements
- Method is called on view init, when data loads, and when columns change
- Ensures dynamically loaded content and toggled columns maintain accessibility compliance

Why
Without the scope attribute, screen reader users cannot understand data context when navigating table cells. The scope="col" attribute creates a programmatic relationship between data cells and their column headers.

Testing

1. Run axe DevTools or Lighthouse accessibility audit on Table/TreeTable demos
2. Verify td-has-header rule passes
3. Inspect <th> elements and confirm scope="col" is present

References

- axe rule: td-has-header
- WCAG: 1.3.1 Info and Relationships (Level A)

### DIFF
--- a/packages/primeng/src/stepper/stepper.ts
+++ b/packages/primeng/src/stepper/stepper.ts
@@ -80,8 +80,7 @@ export interface StepPanelContentTemplateContext {
     changeDetection: ChangeDetectionStrategy.OnPush,
     encapsulation: ViewEncapsulation.None,
     host: {
-        '[class]': 'cx("root")',
-        '[attr.role]': '"presentation"'
+        '[class]': 'cx("root")'
     },
     providers: [StepListStyle, { provide: STEPLIST_INSTANCE, useExisting: StepList }, { provide: PARENT_INSTANCE, useExisting: StepList }],
     hostDirectives: [Bind]
@@ -145,7 +144,6 @@ export class StepperSeparator extends BaseComponent<StepperSeparatorPassThrough>
     encapsulation: ViewEncapsulation.None,
     host: {
         '[class]': 'cx("root")',
-        '[attr.role]': '"presentation"',
         '[attr.data-p-active]': 'isActive()'
     },
     providers: [StepItemStyle, { provide: STEPITEM_INSTANCE, useExisting: StepItem }, { provide: PARENT_INSTANCE, useExisting: StepItem }],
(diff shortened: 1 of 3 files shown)


rosenthalj:
I only quickly reviewed the changes you made to table.ts.  Wouldn't it be more efficient to be proactive vs reactive by adding the missing attribute when the "**th**" is created?  How does querying/setTimeout effect performance when a table has a large number of columns or a large number of rows? 

What would happen if a table (HTML and/or PrimeNG) was implemented within a PrimeNG table?  PrimeNG  (and your change) should not be modifying any embedded non-PrimeNG tables.

fkwq34:
@rosenthalj Good points. I have reworked the approach:

- Switched to a MutationObserver on the <thead> instead of running querySelectorAll on every change — it now only fires when th elements are added
- Scoped the selector to :scope > tr > th so we don't accidentally change any nested tables embed

Let me know if you see any other issues.

